PR #31, opened 2026-02-12
Design built-in file storage, plan blob removal

aeplay:
## Summary

- **New spec: blob removal** (`a_week_2026_02_09/removing_blob_feature.md`) — precise list of every code and status-quo spec edit needed to remove the blob abstraction this week, before the team builds on it
- **New spec: built-in file storage** (`b_mvp/built_in_file_storage.md`) — relational `files` + `file_parts` tables with `bytea` columns, `uuid[]` array references, content-addressed chunking, and policy inheritance via FK declarations
- **New spec: cascade integration** (`c_launch/file_storage_cascade_integration.md`) — ON DELETE CASCADE for automatic file cleanup, deferrable to later
- **New spec: mutable files** (`d_later/mutable_files_and_smart_chunking.md`) — FastCDC chunking and merge strategies for editable files
- **Expanded: on_delete_cascade** — reference-counted cascade semantics and distributed refcounting (eager soft delete, authoritative hard delete)
- Updated cross-references in 3 existing specs

## Test plan

- [ ] Verify all cross-references between specs are correct
- [ ] Review phasing makes sense (MVP → Launch → Later)
- [ ] Review blob removal spec covers all code locations


specs/todo/b_mvp/built_in_file_storage.md
```diff
@@ -0,0 +1,92 @@
...
+create table files (
+  name text,
+  mime text not null,
+  parts uuid[] references file_parts not null,
+  part_sizes integer[] not null
+);
+
+create table file_parts (
+  data bytea
+);
```
gdorsi, line 22: I think files should live on a separate storage or filesystem.

It does have different requirements compared to data (no multiplayer, ok to be slower), takes more space and should be possible for us to move them on a slower but more portable storage.

Unsure about storing them in a table, shall we look at some prior art here?

PR #42, opened 2026-02-12
feat: added placeholder wasm for build

antoniomuso:
Just some corrections to make everything work and start working immediately with jazz2.


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "scripts": {
+    "postinstall": "git update-index --skip-worktree crates/groove-wasm/pkg/groove_wasm.js crates/groove-wasm/pkg/groove_wasm.d.ts 2>/dev/null || true",
```
antoniomuso: This allows us to not track groove_wasm.js and groove_wasm.d.ts in git; in this way, we can build all packages without ts errors.

crates/groove-wasm/pkg/groove_wasm.js
```diff
@@ -0,0 +1 @@
+throw new Error("You should build the WASM module with `pnpm build` before using it.");
```
antoniomuso, line 1: This is just a placeholder.

package.json
```diff
@@ -1,9 +1,10 @@
 {
   "scripts": {
+    "postinstall": "git update-index --skip-worktree crates/groove-wasm/pkg/groove_wasm.js 2>/dev/null || true",
```
antoniomuso, line 3: This is to avoid updating the placeholder.

PR #44, opened 2026-02-12
docs: add a spec to clarify merge design

gdorsi:
Tried to play some scenarios in my head, but something here doesn't clicks out.

My main concerns are:
- if commits store full snapshots, how are we going to optimize transmission when people need only small updates on specific fields?
- how would we handle custom merge strategies if, by storing full snapshots, we lose the info on what the actual user action was?
- when does the merge of the tips happen?

So I'm adding a research task to:
- research if these problems can be handled while staying on the current design or if we should change it
- define a set of user stories to describe when tips are going to be merged


specs/todo/a_week_2026_02_09/merge_design.md
```diff
@@ -0,0 +1,279 @@
...
+   - **Disjoint changes**: combine both deltas (no conflict)
+   - **Same column, deterministic strategy**: apply strategy (LWW, counter-add, etc.)
+   - **Same column, no strategy**: conflict — surface to application
+4. Produce merge commit M with the resolved snapshot
+
+The diff operates on decoded `Value` columns — not raw bytes. The row encoding (`encoding.rs`) supports per-column decode via `decode_column()`, so diffing doesn't require decoding the entire row if only a few columns changed.
+
+### Why this works
+
+- **Optimizes for reads**: every commit is a complete snapshot, no chain-walking to reconstruct state
```
gdorsi, line 122: This is true only if there is only one tip.

If there are more tips a merge should be done to show the current state to the user, which involves walking the history to apply the merge correctly.

specs/todo/a_week_2026_02_09/merge_design.md
```diff
@@ -0,0 +1,279 @@
...
+- **Intent is recoverable**: three-way diff reveals what each side changed, even though we don't store explicit deltas
+- **Merge overhead is pay-on-conflict**: most objects won't have concurrent edits; those that do pay the diff cost only at merge time
```
gdorsi, line 124: Well, this is also true

specs/todo/a_week_2026_02_09/merge_design.md
```diff
@@ -0,0 +1,291 @@
...
+- **Deterministic**: given the same DAG, every peer computes the same merge result (assuming deterministic strategy + deterministic timestamps)
```
gdorsi, line 125: Timestamp are considered only to choose the winner on "unsolvable" conflicts on concurrent tips right?

Normally the last commit always wins

specs/todo/a_week_2026_02_09/merge_design.md
```diff
@@ -0,0 +1,291 @@
...
+- Checkpoint commits (snapshot every N-th, deltas in between) — adds complexity
+- Separate commit cadence per table or column type
+
+For MVP, target the task-board / CRUD cadence. Real-time text is a harder problem that may need RGA/Yjs-style CRDTs regardless of snapshot vs. delta choice.
+
+### Interaction with truncation
+
+Branch truncation (`set_branch_tails()`) deletes old commits. If we truncate too aggressively, the common ancestor needed for three-way merge may be gone. Options:
+
+- **Keep at least one common ancestor** per diverged pair (truncation-aware tail selection)
```
gdorsi, line 166: We can't, a user might be offline for an indefinite time and any commit might be a common ancestor

specs/todo/a_week_2026_02_09/merge_design.md
```diff
@@ -0,0 +1,291 @@
...
+
+When does the system produce a merge commit?
+
+### Status quo timing
+
+Today, merge happens **implicitly on the next local write** — `update()` uses all tips as parents. There is no merge step between receiving diverged commits via sync and the next user action. During that window, reads use LWW.
+
+### Proposed timing
+
+**Reads**: compute merged result ad-hoc whenever tips are diverged. No commit is created — the merged view is ephemeral, computed in the read path (`load_row_from_object_on_branch` or equivalent). This means:
```
gdorsi, line 222: This isn't really applicable.

If two clients merge tips concurrently they may end up in an infinite loop

specs/todo/a_week_2026_02_09/merge_design.md
```diff
@@ -0,0 +1,291 @@
...
+| Use case | Commit frequency | Snapshot overhead | Merge granularity |
+|----------|-----------------|-------------------|-------------------|
+| Task board / CRUD | On user action (click, submit) | Low — commits are infrequent | Coarse — rarely concurrent |
+| Real-time text editing | Per keystroke or debounced (50–200ms) | High — many snapshots | Fine — frequent concurrency |
+| Canvas / drawing | Per stroke or debounced | Medium–High | Medium |
+| Presence / cursors | Continuous (throttled) | High | Fine, but conflicts are trivial (LWW) |
+
+Key question: **is snapshot-per-commit affordable for high-frequency use cases?** The storage benchmarking spike (`storage_benchmarking_spike.md`) will provide numbers. If not, options include:
+
+- Debounced/batched commits (reduce frequency, coarser granularity)
```
gdorsi, line 156: If we want to rely on denouncing or throttling we should implement it on the framework side.

Like make the DB overwrite the last commit and sync/persist only after X ms.

It won't be easy and will probably penalize benchmarks

PR #59, opened 2026-02-13
fix: filter by primary and foreign keys

nrainhart:
While adding cross-table references, I ran into the following issues:
- filtering by FK was not working, as ids were being sent as TEXT wasm values instead of UUID
- filtering by PK was not working either because the `id` column was not being mapped to `_id` when translating to WASM Query JSON

Both of these issues are now fixed.

Next task: make `.include()` work


examples/todo-client-localfirst-ts/schema/migration_v1_v2_311995e9a178_7d55b70e0186.ts
nrainhart, line 1: I collapsed the migration history so that all tasks belong to a project

aeplay (approved):

PR #74, opened 2026-02-17
Fix server hang: dedicated OS thread for batched_tick

joeinnes:
## Problem

While building the moon lander multiplayer example, where each participant hits the database with a position update every 200ms, the server would hang after some time.

Root cause: `TokioScheduler` ran `batched_tick()` via `tokio::spawn`. This acquires `Arc<Mutex<RuntimeCore>>` and holds it while processing sync callbacks and flushing the WAL. Every HTTP request handler also needs this lock. Under load, a slow tick blocks a tokio worker while holding the mutex, and other request handlers block further workers waiting for it. Eventually all tokio workers are parked on the mutex and the server stops responding entirely.

## Solution

Replace `tokio::spawn` with a dedicated OS thread (`groove-tick`) that parks on `std::sync::mpsc::channel`. When `schedule_batched_tick()` is called, it sends a wake signal. The thread wakes, acquires the mutex, runs `batched_tick()`, and parks again. Tokio workers never hold the lock during tick processing, so the event loop stays responsive.

Three details:

- **Tokio context**: the thread calls `Handle::enter()` on startup, so sync callbacks can use `tokio::spawn` and timers transparently.
- **Panic resilience**: each tick is wrapped in `catch_unwind`. A bad tick logs the error and continues rather than silently killing all future ticks.
- **Lifecycle**: the thread exits when the `Sender` drops (i.e. when the runtime is dropped).

The coarse `Mutex<RuntimeCore>` still exists — individual request handlers can still briefly contend with a running tick. The improvement is that the tick itself never occupies a tokio worker: degradation is gradual (higher latency under load) rather than catastrophic (complete hang).

## Testing

`test_slow_tick_does_not_block_tokio`: uses a `current_thread` tokio runtime (single worker, worst case) with a sync callback that sleeps 500ms (simulating slow I/O). After triggering a tick via insert, asserts that `yield_now` returns within 100ms. This test fails on main (blocked for ~500ms) and passes with the dedicated thread.

## Changes

- `crates/groove-tokio/src/lib.rs` — TokioScheduler rewritten from `tokio::spawn` to dedicated thread
- `specs/status-quo/batched_tick_orchestration.md` — updated to document the new architecture

### DIFF
--- a/crates/groove-tokio/src/lib.rs
+++ b/crates/groove-tokio/src/lib.rs
@@ -2,12 +2,12 @@
 //!
 //! Provides `TokioRuntime<S>` - a thin wrapper around
 //! `RuntimeCore<S, TokioScheduler<S>, CallbackSyncSender>`
-//! that handles async scheduling via `tokio::spawn`.
+//! that handles async scheduling via a dedicated tick thread.
 //!
 //! # Architecture
 //!
 //! - `S: Storage + Send + 'static` provides synchronous storage
-//! - `TokioScheduler<S>` implements `Scheduler` using tokio::spawn for batched ticks
+//! - `TokioScheduler<S>` implements `Scheduler` using a dedicated OS thread
 //! - `CallbackSyncSender` implements `SyncSender` with a user-provided callback
 //! - `TokioRuntime<S>` wraps `Arc<Mutex<RuntimeCore<...>>>`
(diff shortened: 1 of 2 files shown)


aeplay:
Closing this as we really shouldn't need threads - the idea is that batched ticks are sequential

PR #75, opened 2026-02-17
feat: add arrays to Typescript DSL

nrainhart:
- Adds support for array columns in the Typescript DSL
- Lenses now preserve column types
- Lenses support adding and removing array columns

I'll add support for array queries (eq/contains) as a fup.


packages/jazz-tools/src/sql-gen.test.ts
```diff
@@ -212,6 +243,80 @@ describe("lensToSql", () => {
...
+    const lens = getCollectedMigration()!;
+
+    expect(lensToSql(lens, "fwd")).toBe(`ALTER TABLE projects ADD COLUMN todos TEXT DEFAULT ARRAY[];
+`);
+  });
+
+  it("supports adding array columns in lenses", () => {
+    resetCollectedState();
+    migrate("projects", {
+      todos: col.add().array({ of: "UUID", default: [] }),
```
nrainhart, line 286: The lenses API could be nicer (especially for nested array types), but want to get something working for now

packages/jazz-tools/src/dsl.ts
```diff
@@ -35,13 +44,17 @@ class ColumnBuilder {
       nullable: this._nullable,
     };
   }
+
+  get _references(): string | undefined {
+    return undefined;
+  }
```
gdorsi, line 50: Why the _ prefix?

gdorsi (approved):
LGTM

packages/jazz-tools/src/dsl.ts
nrainhart: That's how we're distinguishing these internal methods from the public-facing API for now

PR #76, opened 2026-02-17
Set up realistic benchmarks

aeplay:


gdorsi (approved):
The PR is quite big and without any description, and so hard to review.

I can only approve on "trust" but would be really helpful to have some screenshots displaying:
- the CI runs results
- how does the local results look like 

Also, are there some safeguards around reducing noise to keep the results stable?
Like, if I do run the same benchmark twice with the same code, am I going to get similar results?

PR #81, opened 2026-02-18
feat: add contains query operator

nrainhart:
## Description

FUP to https://github.com/garden-co/jazz2/pull/75.

Adds a `contains` query operator for array fields.


packages/jazz-tools/tests/browser/worker-bridge.test.ts
```diff
@@ -212,6 +219,67 @@ describe("Worker Bridge with OPFS", () => {
     expect(results[0].name).toBe("Project A");
   });
 
+  it("queries by array column equality", async () => {
```
nrainhart: This is totally not the place where I should be doing this sort of integration tests, but I wanted to test things as close to the user-facing Typescript API as possible. Maybe we should have a test suite where we actually generate the TS client?

nrainhart: Ended up adding a separate folder for running integration tests on the generated typescript client: https://github.com/garden-co/jazz2/pull/81/changes/03ecc334f0cd70e9c24f2b4a7e6847324e396d60

aeplay (approved):

PR #103, opened 2026-02-23
runtime: add multi-tab leader election and leader-routed sync

aeplay:
## Summary
- add `specs/todo/a_mvp/multi_tab_leader_election.md` with the full browser-tab leader election design, including heartbeat/failover behavior, main-thread upstream swap behavior, and zombie-lock fallback notes
- implement tab leader election in runtime (`tab-leader-election.ts`) with heartbeats, lease takeover, and peer lifecycle hooks
- extend worker protocol + bridge/runtime plumbing so follower tabs route sync writes through the elected leader and rebind upstream server peers on leader change
- add opt-in leader-election debug hooks to aid diagnosis in browser tests
- add/expand runtime and browser test coverage for leader routing, failover, WAL recovery, and cross-client server sync
- fix browser test harness stability by making global setup idempotent and aligning sync test auth with the test server’s configured modes

## Testing
- `pnpm --dir packages/jazz-tools exec vitest run src/runtime/tab-leader-election.test.ts src/runtime/worker-bridge.test.ts`
- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.browser.ts tests/browser/worker-bridge.test.ts`


gdorsi (changes requested):
This kind of feature requires a more carful testing strategy.

The unit tests are not covering the whole implementation, and are just too minimal.

Also I think we need a sort-of e2e test that checks how the election reacts to tab changes, close and reloads.

packages/jazz-tools/src/runtime/tab-leader-election.ts
```diff
@@ -0,0 +1,382 @@
...
+  removeEventListener(type: "message", listener: (event: MessageEvent) => void): void;
+  close(): void;
+}
+
+function randomTabId(): string {
+  const cryptoObj = (globalThis as { crypto?: Crypto }).crypto;
+  if (cryptoObj && typeof cryptoObj.randomUUID === "function") {
+    return cryptoObj.randomUUID();
+  }
+  return `tab-${Math.random().toString(36).slice(2, 12)}`;
```
gdorsi, line 84: Since this fallback is good-enough, shall we always return this to keep the funcition simple?

packages/jazz-tools/src/runtime/tab-leader-election.ts
```diff
@@ -0,0 +1,382 @@
...
+    this.now = options.now ?? (() => Date.now());
+    this.channelName = `jazz-leader:${options.appId}:${options.dbName}`;
+
+    this.readyPromise = new Promise<LeaderSnapshot>((resolve, reject) => {
+      this.readyResolve = resolve;
+      this.readyReject = reject;
+    });
+  }
+
+  start(): void {
```
gdorsi, line 167: How about we document how this lifecycle works with some ASCII diagram?

packages/jazz-tools/src/runtime/tab-leader-election.ts
```diff
@@ -0,0 +1,382 @@
...
+    });
+
+    this.leaseTimer = setInterval(
+      () => {
+        this.checkLease();
+      },
+      Math.max(100, Math.floor(this.heartbeatMs / 2)),
+    );
+
+    this.initialElectionTimer = setTimeout(() => {
```
gdorsi, line 187: Why a timer here?

packages/jazz-tools/src/runtime/tab-leader-election.ts
```diff
@@ -0,0 +1,382 @@
...
+
+    this.channel = new ChannelCtor(this.channelName);
+    this.channel.addEventListener("message", this.onMessage);
+
+    this.postMessage({
+      type: "leader-request",
+      requesterTabId: this.tabId,
+    });
+
+    this.leaseTimer = setInterval(
```
gdorsi: Is polling the only choice we had here?

I would prefer something more reliable

specs/todo/a_mvp/multi_tab_leader_election.md
```diff
@@ -1,7 +1,251 @@
...
+
+Important: on-demand lock per read/write is not viable for current architecture because acquiring a sync handle is async and would force async boundaries into the synchronous storage/query path.
+
+## Lifecycle Integration
+
+Page listens to:
+
+- `visibilitychange`
+- `pagehide`
+- `freeze` / `resume` where available
```
gdorsi, line 216: I think we should use a SharedWorker as central authority for tab election.

It would give us a reliable way of tracking which tabs are still open, and a central place to track which one is the current leader so it would be less likely that different tabs disagree.

Depending only on this kind-of events leaves us too much in the hands of the browser-specific implementations. 

Also couldn't find any implementation of this lifecycle in the PR.

gdorsi: Done some prior-art research and found out that PGLite implements something similar: https://github.com/electric-sql/pglite/blob/main/packages/pglite/src/worker/index.ts#L723

It is based on navigator.lock, which would give us similar properties as a central SharedWorker (a single source of truth)

aeplay: Unfortunately Android for Chrome doesn't support SharedWorker: https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker

packages/jazz-tools/src/runtime/tab-leader-election.ts
aeplay: done!

packages/jazz-tools/src/runtime/tab-leader-election.ts
aeplay: replaced with leases + deadline timeouts

packages/jazz-tools/src/runtime/tab-leader-election.ts
aeplay: replaced with leases + deadline timeouts

PR #107, opened 2026-02-23
fix: update combined_descriptor after array subquery compilation

joeinnes:
## Summary

Using `.include({ sender: true })` on message queries in the chat app resulted in the message text being garbled. The root cause: `graph.combined_descriptor` was not updated after array subqueries were compiled, so `QueryUpdate.descriptor` reflected only the base table columns. Decoding rows with this stale descriptor corrupts variable-length column data because the offset table is shorter than expected, shifting `var_data_start`.

The fix adds `graph.combined_descriptor = current_descriptor.clone()` before the output node in both compilation paths (`compile_execution_plan_with_session` and `compile_execution_plan_with_schema_context`).

Includes a Rust test (`manager_tests`) and a TypeScript test (`query-api`) that both fail on the parent commit and pass with the fix.

## Test plan

- [x] Rust test: `cargo test -p jazz-tools --lib -- array_subquery_update_descriptor`
- [x] TS test: `pnpm --filter jazz-tools exec vitest run tests/ts-dsl/query-api.test.ts`
- [x] Both tests fail on the parent commit (stale descriptor / capacity overflow panic)
- [x] Both tests pass with the fix

### DIFF
--- a/packages/jazz-tools/tests/ts-dsl/query-api.test.ts
+++ b/packages/jazz-tools/tests/ts-dsl/query-api.test.ts
@@ -58,13 +58,40 @@ describe("TS Query API", () => {
     expect(baseline[0].title).toBe("Hello world");
 
     const withInclude = await db.all(
-      app.todos
-        .where({ id: { eq: todoId } })
-        .include({ project: true }),
+      app.todos.where({ id: { eq: todoId } }).include({ project: true }),
     );
 
     expect(withInclude.length).toBe(1);
     expect(withInclude[0].title).toBe("Hello world");
+  });
+
+  it("include returns the related entity", async () => {
+    const db = track(
+      await createDb({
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/tests/ts-dsl/query-api.test.ts
```diff
@@ -38,6 +38,33 @@ describe("TS Query API", () => {
     expect(results[0].name).toBe("Project A");
   });
 
+  it("text is not corrupted when using include", async () => {
```
nrainhart, line 41: Mind adding another test that asserts include works properly? (i.e. the included entity is returned)

nrainhart (approved):
The fix looks reasonable to me!

gdorsi (approved):

PR #110, opened 2026-02-23
feat(jazz-tools): add text contains filter and fix query updates on useAll

gdorsi:
- Updated Predicate::Contains to support substring matching for text columns.
- Added tests for filtering todos by title substring and handling empty substring matches.
- Improved the overall filtering logic in FilterNode to accommodate new text matching functionality.


packages/jazz-tools/tests/browser/useAll.test.tsx
```diff
@@ -0,0 +1,584 @@
...
+
+    await waitForCondition(
+      () => getText("rows").includes("open-task") && !getText("rows").includes("done-task"),
+      5000,
+      "expected initial query to show only open task",
+    );
+
+    container
+      ?.querySelector('[data-testid="toggle-query"]')
+      ?.dispatchEvent(new MouseEvent("click", { bubbles: true }));
```
gdorsi: Switching to use the Vitest user interactions APIs.

PR #118, opened 2026-02-24
Add Real (float) column type support

joeinnes:
## Summary

- `col.float()` is permitted in the TypeScript DSL and emits `REAL` in SQL, but the Rust SQL parser rejected it with `UnsupportedType`. This came up while working on the chat app, where I wanted to use `col.float()` for stroke width on the canvas. Investigating the failure led to discovering that the WASM bridge had no Real type at all and was silently coercing floats to integers as a shortcut.
- This PR adds end-to-end support for Real (f64) columns: Rust core types, binary encoding, order-preserving index key encoding, SQL parser (accepts REAL/FLOAT/DOUBLE), schema manager, WASM and NAPI bridges, and TypeScript codegen/runtime.
- Implements IEEE 754 negative zero semantics in index queries: `-0.0` and `0.0` are stored as distinct bit patterns but treated as equal in lookups and range scans. Documents a known upstream limitation where `-0.0` may be lost crossing the WASM boundary due to serde buffering.

## Test plan

- [x] 14 new Rust tests covering SQL parsing, binary encoding round-trips, index key ordering, and negative zero query semantics
- [x] 2 updated TypeScript tests (codegen expects `{ type: "Real" }`, value-converter handles Real)
- [x] `cargo test -p jazz-tools --lib` passes (705 tests)
- [x] `cargo clippy` and `cargo fmt` clean
- [x] Relevant TypeScript tests pass (89 tests)

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -44,7 +44,7 @@ use jazz_tools::sync_manager::{
 enum NapiValue {
     Integer(i32),
     BigInt(i64),
-    Real(f64),
+    Double(f64),
     Boolean(bool),
     Text(String),
     Timestamp(u64),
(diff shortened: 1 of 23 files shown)


gdorsi (approved):
Code LGTM

Unsure about the naming, in JS decimals are named as float (parseFloat) rather than real.

But I guess real gives us a more real DB vibe (pun intended ahahah)

joeinnes:
Fair point on the naming, I ran with `real` because that was the SQL being generated by the codegen from the ts `col.float()`, but in Rust it's an `f64`, so maybe `double` would be better 😕. 

REAL/FLOAT/DOUBLE from the SQL all map onto the same `f64` anyway, so any of them could work. I think most folks will still interact with this using `col.float()`.

Happy to change to whatever we think sounds best.

PR #119, opened 2026-02-24
feat: replace in-memory OPFS-btree with pure in-memory storage in WASM

nrainhart:
## Description

While profiling, we found we're spending lots of time running crc32 checksums in the in-memory OPFS-btree storage in the main thread:

before

Decided to replace the in-memory OPFS-btree storage with a more light-weight, pure in-memory storage, which speeds up main-thread "storage" by ~4x:

after

## Discussion

Went with boxed dynamic references to storages to support runtime polymorphism for different storage implementations in the WASM runtime. I found this approach way less verbose than generating an enum wrapper over both types of storage and doing pattern matching, and checked performance is not noticeably affected. Still, I'm relatively new to Rust, so we can change this if you think there's a better way to do this.

### DIFF
--- a/crates/jazz-tools/src/storage/mod.rs
+++ b/crates/jazz-tools/src/storage/mod.rs
@@ -181,6 +181,122 @@ pub trait Storage {
     fn flush_wal(&self) {}
 }
 
+// Box<Storage> is used to allow for dynamic dispatch of the Storage trait.
+impl<T: Storage + ?Sized> Storage for Box<T> {
+    fn create_object(
+        &mut self,
+        id: ObjectId,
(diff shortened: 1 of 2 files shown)


gdorsi (approved):

gdorsi:
The flamegraph is starting to get nicer:

Screenshot 2026-02-24 at 15 44 28

There is still the issue with the subscription emitting "onfulfilled" too early
Tried with a more aggressive batching (non-viable) and it shows that we could squeeze some more perf:

Screenshot 2026-02-24 at 15 45 09

We gotta fix the "settled" subscription timing @aeplay

PR #121, opened 2026-02-24
fix: fixes sorting and makes id the default sorting field

gdorsi:
- Fixed query sorting correctness issues (including default/no-orderBy behavior and explicit id ordering semantics).
- Stabilized ordering across updates so results are deterministic instead of drifting/reordering unexpectedly.
- Added a browser integration test suite focused on subscribeAll sorting scenarios to lock in behavior and prevent regressions going forward.

Main changes:
- added `process_with_ordered_input` to LimitOffset and Output node to make them sorting aware, before queries with limit weren't considering the sorting position
- added an indexed delta in the Subscription output, to provide to clients the information of where to place changes in the row list
- introduced a `moved` vector to make it possible to signal to the above layers which tuples has been moved without sending them huge deltas


crates/jazz-tools/src/query_manager/graph_nodes/output.rs
```diff
@@ -93,6 +93,67 @@ impl OutputNode {
             .collect()
     }
 
+    /// Ordered tuples as received from upstream nodes.
+    pub fn ordered_tuples(&self) -> &[Tuple] {
+        &self.ordered_tuples
+    }
+
+    fn compute_tuple_delta(&self, old_tuples: &[Tuple], new_tuples: &[Tuple]) -> TupleDelta {
```
gdorsi, line 101: Shall we compute the index delta here?

crates/jazz-tools/src/query_manager/graph_nodes/output.rs
```diff
@@ -93,6 +93,67 @@ impl OutputNode {
...
+                delta.updated.push((old.clone(), new.clone()));
+            }
+        }
+
+        delta
+    }
+
+    /// Rebuild ordered output from a full ordered upstream input.
+    pub fn process_with_ordered_input(&mut self, ordered_tuples: &[Tuple]) -> TupleDelta {
+        let old_tuples = self.ordered_tuples.clone();
```
gdorsi: We can remove this clone by doing compute_tuple_delta earlier

crates/jazz-tools/src/query_manager/graph_nodes/output.rs
```diff
@@ -93,6 +93,67 @@ impl OutputNode {
...
+        self.ordered_tuples = ordered_tuples.to_vec();
+        self.current_tuples = self.ordered_tuples.iter().cloned().collect();
```
gdorsi, line 152: Do we need cloned here?

What's the difference between self.ordered_tuples and self.current_tuples?

crates/jazz-tools/src/query_manager/graph.rs
```diff
@@ -736,20 +731,15 @@ impl QueryGraph {
...
+        // Sort node (default: id ASC when order_by is omitted)
+        let sort_keys = sort_keys_from_order_by(&plan.order_by, &current_descriptor);
+        if !sort_keys.is_empty() {
+            let sort_tuple_desc =
+                TupleDescriptor::single_with_materialization("", current_descriptor.clone(), true);
+            let sort_node = SortNode::with_tuple_descriptor(sort_tuple_desc, sort_keys);
+            let sort_id = graph.add_node(GraphNode::Sort(sort_node));
+            graph.add_edge(sort_id, phase2_input);
+            phase2_input = sort_id;
         }
```
gdorsi, line 743: This simply removed the check `!plan.order_by.is_empty()` because we want to always sort by id

crates/jazz-tools/src/query_manager/graph.rs
```diff
@@ -1983,14 +1970,23 @@ impl QueryGraph {
...
+                        _ => None,
+                    });
 
                     if let Some(GraphNode::LimitOffset(lo_node)) = self.get_node_mut(node_id) {
-                        let delta = RowNode::process(lo_node, input_delta);
+                        let delta = if let Some(ordered) = ordered_input {
+                            lo_node.process_with_ordered_input(&ordered)
+                        } else {
+                            RowNode::process(lo_node, input_delta)
+                        };
```
gdorsi, line 1989: Shall we move input_delta here?

Also, since ordering is always computed, should we always go with `process_with_ordered_input`?

crates/jazz-tools/src/query_manager/graph.rs
```diff
@@ -2038,14 +2034,26 @@ impl QueryGraph {
...
+                        }
+                        _ => None,
+                    });
 
                     if let Some(GraphNode::Output(output_node)) = self.get_node_mut(node_id) {
-                        let delta = RowNode::process(output_node, input_delta);
+                        let delta = if let Some(ordered) = ordered_input {
+                            output_node.process_with_ordered_input(&ordered)
+                        } else {
+                            RowNode::process(output_node, input_delta)
```
gdorsi, line 2055: Same here

Also, are we computing ordering twice if we have a limit node?

crates/jazz-tools/src/query_manager/graph.rs
```diff
@@ -2269,13 +2277,27 @@ fn sort_keys_from_order_by(
...
         .filter_map(|(col, dir)| {
-            descriptor.column_index(col).map(|idx| SortKey {
-                col_index: idx,
-                direction: *dir,
-            })
+            if col == "id" || col == "_id" {
+                Some(SortKey {
+                    target: SortTarget::RowId,
+                    direction: *dir,
+                })
```
gdorsi, line 2296: Do we really need a spiecialized SortTarget for ids?

gdorsi: Yes, we need the specialized RowId target because id/_id is a virtual key (not a payload column), and we must preserve explicit ORDER BY id|_id direction/mixed-key semantics. The final tie-breaker alone is not equivalent.

examples/todo-client-localfirst-react/src/TodoList.tsx
```diff
@@ -16,7 +16,7 @@ export function TodoList() {
 
   // #region reading-reactive-hooks-react
   const db = useDb();
-  const todos = useAll(todosQuery);
+  const todos = useAll(todosQuery.orderBy("title").limit(15)) ?? [];
```
nrainhart: We should revert this change, right?

crates/jazz-wasm/src/types.rs
```diff
@@ -90,12 +90,40 @@ pub struct WasmRow {
...
+    pub id: String,
+    pub index: usize,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct WasmUpdated {
+    pub id: String,
+    #[serde(rename = "oldIndex")]
+    pub old_index: usize,
```
nrainhart: What's the `oldIndex` used for, if we're removing updated rows by id?

packages/jazz-tools/src/runtime/subscription-manager.ts
```diff
@@ -47,31 +60,45 @@ export class SubscriptionManager<T extends { id: string }> {
...
+
+      if (change.row) {
+        const item = transform(change.row);
+        this.currentResults.set(change.id, item);
+      }
+
+      const existing = this.currentResults.get(change.id);
+      if (!existing) continue;
+
+      this.insertIdAt(change.id, change.newIndex);
```
nrainhart: Are newIndexes sent in asc order? Otherwise won't we have problems?

crates/jazz-tools/src/query_manager/graph_nodes/limit_offset.rs
```diff
@@ -53,28 +53,49 @@ impl LimitOffsetNode {
...
         let mut delta = TupleDelta::new();
+        let old_ids: std::collections::HashSet<Vec<crate::object::ObjectId>> =
+            old_tuples.iter().map(|t| t.ids()).collect();
+        let new_ids: std::collections::HashSet<Vec<crate::object::ObjectId>> =
+            new_tuples.iter().map(|t| t.ids()).collect();
 
         // Find removed tuples (in old but not in new)
         for old in old_tuples {
-            if !new_tuples.iter().any(|t| t == old) {
+            if !new_ids.contains(&old.ids()) {
```
nrainhart, line 72: Have you done any performance testing to see how fast this is with large arrays?

nrainhart (commented):
I did a partial review, will finish it later

crates/jazz-wasm/src/types.rs
gdorsi: Good point.

Going to remove it

crates/jazz-tools/src/query_manager/graph_nodes/limit_offset.rs
gdorsi: Compared main vs this branch on loading 5k elements, can't see any difference in the load.

This branch:
Screenshot 2026-02-25 at 21 23 17

Main:
Screenshot 2026-02-25 at 21 16 11

Update a row

Branch
Screenshot 2026-02-25 at 21 23 58

Main
Screenshot 2026-02-25 at 21 21 31


Will prepare a heavier test tomorrow

gdorsi: Updates are slower than I was expecting @aeplay

crates/jazz-tools/src/query_manager/graph_nodes/output.rs
aeplay: yes this seems like the most natural place

packages/jazz-tools/src/runtime/subscription-manager.ts
gdorsi: You are right!

Fixed by revisiting the delta data structure to simplify the apply of the changes in the right order.

https://github.com/garden-co/jazz2/pull/121/changes/157a6f0805e45b13f537e977ef1801b035bb1984

crates/jazz-tools/src/query_manager/manager.rs
```diff
@@ -735,9 +752,23 @@ impl QueryManager {
...
                     descriptor: subscription.graph.combined_descriptor.clone(),
                 });
             } else if !delta.is_empty() {
+                let post_ids: Vec<ObjectId> = subscription
+                    .graph
+                    .current_result()
+                    .iter()
+                    .map(|row| row.id)
+                    .collect();
+                let projected = project_row_delta_with_post_ids(
```
aeplay: "Projected" is overloaded as a term and is confusing here. All this does is produce an "ordered delta" (indexed is also overloaded)

aeplay (changes requested):
I feels like there is redundant order index calculation going on and like there are still a bunch of code branches for ordered/not-ordered where it might be easier to just assume everything is always ordered.

The JS tests seem a lot more comprehensive than the Rust ones, would be nice to make Rust ones reach parity, because we can run them faster.

The only real issue in my opinion is the prevalent "projection" and "indexed" naming, both of which have existing, different meanings, which is likely confusing for someone trying to understand QueryManager. Please fix that - everything else can be done in the future

gdorsi:
> I feels like there is redundant order index calculation going on and like there are still a bunch of code branches for ordered/not-ordered where it might be easier to just assume everything is always ordered.

You mean that there is always a Sort node as input?

I also had the same feeling, but the change isn't simple unless we panic if the Sort node isn't available.

My knowledge of the codebase is still limited, I've tried other solutions but all felt worse than the current one.

> The JS tests seem a lot more comprehensive than the Rust ones, would be nice to make Rust ones reach parity, because we can run them faster.

That's because they are way more useful, while writing test in Rust now feels quite useless.

The API on the JS side is more limited and unlikely to change, it also covers a way more complex system where we have main + worker communicating.

On the Rust side instead, the API doesn't feel as stable because the surface area is way more wider and we are still tweaking everything.

> The only real issue in my opinion is the prevalent "projection" and "indexed" naming, both of which have existing, different meanings, which is likely confusing for someone trying to understand QueryManager. Please fix that - everything else can be done in the future

On it!

aeplay (approved):

PR #122, opened 2026-02-24
fix: expo

antoniomuso:
This PR makes the RN sync work.


packages/jazz-tools/src/runtime/client-session.ts
```diff
@@ -104,14 +134,123 @@ function encodeBase64(bytes: Uint8Array): string {
...
+      const bytes = utf8Encode(input);
+      const digest = await cryptoObj.subtle.digest("SHA-256", bytes as unknown as BufferSource);
+      return new Uint8Array(digest);
+    } catch {
+      // Fall through to pure JS implementation.
+    }
   }
 
-  throw new Error("No SHA-256 implementation available in this runtime");
+  return sha256PureJs(input);
```
antoniomuso, line 253: We can implement this in Uniffi. I will take note.

packages/jazz-tools/src/runtime/client-session.ts
```diff
@@ -18,6 +18,17 @@ interface BufferLike {
   from(input: string | Uint8Array, encoding?: string): { toString(encoding?: string): string };
 }
 
+const SHA256_K = new Uint32Array([
```
gdorsi, line 21: What's this?

antoniomuso: This is used from the SHA256 function; we don't have it in RN. My idea is to move that function to Uniffi.

PR #124, opened 2026-02-24
feat: implement bytea column type across rust/wasm/ts

aeplay:
## Summary
- add native Bytea support to schema/value types, SQL/schema encoding, storage indexing, and row encoding/decoding
- enforce a 1 MiB per-cell limit and explicitly reject unsupported bytea ordering/range operations
- wire bytea through wasm + TypeScript runtime, DSL, codegen, and query adapter behavior
- add/adjust tests for bytea roundtrips, validation errors, and query semantics across Rust and TS

## Validation
- cargo test -p jazz-tools
- cargo test -p jazz-wasm
- pnpm -C crates/jazz-wasm build
- pnpm -C packages/jazz-tools test


nrainhart (approved):

gdorsi (approved):
LGTM

Could you add this new column type to the browser tests (db.all and db.subscribeAll)?

PR #125, opened 2026-02-24
Implement FK-declared policy inheritance end-to-end

aeplay:
## Summary
- add `inherit_policy` metadata for FK declarations across schema types, SQL parser/generator, and wire encodings
- implement declared inbound FK policy inheritance for read and write permission paths with recursion guards
- extend Rust + JS/WASM/RN bindings and jazz-tools DSL/codegen/sql generation for `INHERIT POLICY`
- add targeted tests for semantics (scalar/array edges, cycles, reactivity), schema compatibility, and TS tooling mappings

## Notes
- this moves the earlier direct-to-main change into a standard PR flow

### DIFF
--- a/packages/jazz-tools/src/permissions/index.test.ts
+++ b/packages/jazz-tools/src/permissions/index.test.ts
@@ -391,6 +391,43 @@ describe("permissions DSL", () => {
     });
   });
 
+  it("supports allowedTo.readReferencing helper", () => {
+    const compiled = definePermissions(app, ({ policy, allowedTo }) => [
+      policy.projects.allowRead.where(allowedTo.readReferencing(policy.todos, "projectId")),
+    ]);
+
+    expect(compiled.projects.select?.using).toEqual({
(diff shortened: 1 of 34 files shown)


packages/jazz-tools/src/permissions/index.test.ts
```diff
@@ -391,6 +391,43 @@ describe("permissions DSL", () => {
...
+    ]);
+
+    expect(compiled.projects.select?.using).toEqual({
+      type: "InheritsReferencing",
+      operation: "Select",
+      source_table: "todos",
+      via_column: "projectId",
+      max_depth: 3,
+    });
+  });
```
gdorsi, line 513: How about indirect recursion?

Project -> Person -> Project

gdorsi (approved):
Code-wise LGTM

I think we need some integration tests for this sync-server features that works like the browser tests.

We need this kind of guardrails so we can ensure that there is no regression when we do huge refactorings.

PR #128, opened 2026-02-24
Persist catalogue schema and lens IDs

aeplay:
## Summary
- adds a failing cloud-server integration test reproducing restart behavior where an existing client no longer gets Edge-settled results without explicit catalogue re-sync
- extends `Storage` with append-only catalogue manifest support using `schemaSeen` / `lensSeen` ops
- implements manifest op storage + manifest materialization in `MemoryStorage`, `OpfsBTreeStorage`, and `SurrealKvStorage`
- adds backend/unit tests for manifest op idempotency and manifest roundtrip

## New Storage API
- `append_catalogue_manifest_op(app_id, op)`
- `append_catalogue_manifest_ops(app_id, ops)`
- `load_catalogue_manifest(app_id)`
- `CatalogueManifestOp::{SchemaSeen, LensSeen}`

## Validation
- `cargo test -p jazz-tools catalogue_manifest -- --nocapture`
- `cargo test -p jazz-tools --features surrealkv surrealkv_catalogue_manifest_roundtrip -- --nocapture`
- `cargo test -p jazz-cloud-server --test client_multi_integration jazz_tools_existing_client_keeps_working_after_server_restart_without_catalogue_resync -- --nocapture` (fails as expected, reproducing bug)


gdorsi (approved):
High-level looks good to me.

Do you think that would be doable to do an integration test on OPFS and Memory using the browser suite?

gdorsi (approved):
High-level looks good to me.

Do you think that would be doable to do an integration test on OPFS and Memory using the browser suite?

aeplay:
> High-level looks good to me.
> 
> Do you think that would be doable to do an integration test on OPFS and Memory using the browser suite?

I was going to say that this isn't necessary because the client always receives schemas from the code but you actually ma de me realize that there's an edge case: a deployed, outdated client might receive new schemas and lenses from the sync server that make new data readable for the old client. Without catalogue persistence, it would have to re-sync the new schemas and lenses at every start.

Going to write a test for this scenario, which would cover catalogue persistence on the client.

PR #134, opened 2026-02-25
/schema endpoints

Takeno:


crates/jazz-tools/src/schema_manager/catalogue_export.rs
```diff
@@ -0,0 +1,394 @@
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct CatalogueColumnDescriptor {
+    pub name: String,
+    pub column_type: CatalogueColumnType,
+    pub nullable: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub references: Option<String>,
+}
+
+impl From<ColumnType> for CatalogueColumnType {
```
aeplay, line 75: Why couldn't we just use ColumnType instead of introducing a new very similar type?

crates/jazz-tools/src/schema_manager/catalogue_export.rs
```diff
@@ -0,0 +1,394 @@
...
+        exprs: Vec<CataloguePolicyExpr>,
+    },
+    Not {
+        expr: Box<CataloguePolicyExpr>,
+    },
+    True,
+    False,
+}
+
+impl From<PolicyExpr> for CataloguePolicyExpr {
```
aeplay, line 223: same question here

aeplay (approved):

PR #135, opened 2026-02-25
perf(opfs-btree): stop range scans at first key >= end in leaf scan

nrainhart:
## Description

Changes leaf range scan behavior to stop the leaf-chain walk as soon as the scan reaches the exclusive upper bound.

`raw_leaf_scan` now returns `None` (instead of `header.next_page_id`) when it encounters the first key >= end, signaling there is no useful continuation for the current range query.

B-tree ordering guarantees that:
- keys inside a leaf are sorted
- keys in subsequent leaves are also greater than or equal to keys in the current leaf

So once we hit key >= end, neither the rest of the current leaf nor any next leaf can contain keys in [start, end).

## Performance comparison

**Before**

2-before

[before.json.gz](https://github.com/user-attachments/files/25553210/before.json.gz)

**After**

2-after

[after.json.gz](https://github.com/user-attachments/files/25553213/after.json.gz)


crates/jazz-wasm/src/runtime.rs
```diff
@@ -27,7 +27,7 @@ fn init_tracing() {
     static INIT: Once = Once::new();
     INIT.call_once(|| {
         let config = wasm_tracing::WasmLayerConfig::new()
-            .with_max_level(tracing::Level::TRACE)
```
nrainhart: Also increased the log level in jazz-wasm to improve performance. See main thread comparison:

**Trace**

Image

**Warn**

Image

gdorsi (approved):
Sweet!

crates/opfs-btree/src/page.rs
```diff
@@ -367,6 +368,7 @@ pub(crate) fn raw_leaf_scan<'a>(
         }
         let key = &payload[key_off..key_end];
         if key >= end {
+            reached_end = true;
```
gdorsi, line 375: Could you add a comment documenting what reached_end means?

nrainhart: Do you think it's not clear by itself? It literally means you've reached the end of the range you want to scan. Maybe what's less clear is the function's returned value. I added a Rustdoc comment to document that

aeplay:
Apparently this was a regression introduced in https://github.com/garden-co/jazz2/commit/21ebdf9
thanks for catching it!

crates/opfs-btree/src/page.rs
gdorsi: I meant the context of the choice.

Was thinking that might be helpful to store it as a comment.

Non blocking, thanks for documenting the function return value!

PR #136, opened 2026-02-26
Expose TIMESTAMP in TS DSL and make Date the high-level default

aeplay:
## Summary
- expose `TIMESTAMP` in the TypeScript schema DSL (`col.timestamp()` plus migration helpers)
- map DSL `TIMESTAMP` to WASM `Timestamp`
- switch high-level generated TS types to use `Date` for timestamp columns
- make runtime timestamp conversion Date-friendly for writes and reads
- extend query translation to accept timestamp `Date`, ISO string, and numeric string values
- add tests across codegen, value conversion, row transformation, query translation, and SQL generation

## Validation
- `pnpm exec vitest run --config vitest.config.ts src/runtime/query-adapter.test.ts src/runtime/value-converter.test.ts src/runtime/row-transformer.test.ts src/codegen/codegen.test.ts src/sql-gen.test.ts`
- `pnpm exec vitest run --config vitest.config.ts src/dsl.test.ts`


docs/content/docs/react-quick-start.mdx
```diff
@@ -42,6 +42,7 @@ There are a number of column types available which cover most common use cases:
 | `col.string()`     | TEXT      | `string`          |
 | `col.boolean()`    | BOOLEAN   | `boolean`         |
 | `col.int()`        | INTEGER   | `number`          |
+| `col.timestamp()`  | TIMESTAMP | `Date`            |
```
nrainhart, line 45: Wish it wasn't too early to use [Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) 🥲

packages/jazz-tools/src/schema.ts
```diff
@@ -31,9 +31,11 @@ type TSTypeFromScalarSqlType<T extends ScalarSqlType> = T extends "TEXT"
       ? number
       : T extends "REAL"
         ? number
-        : T extends "UUID"
-          ? string
-          : never;
+        : T extends "TIMESTAMP"
+          ? number
```
nrainhart: Is this ok? Shouldn't it be Date?

packages/jazz-tools/src/runtime/query-adapter.ts
```diff
@@ -97,12 +130,15 @@ function toWasmValue(value: unknown, columnType: ColumnType): object {
...
-      return { Timestamp: value };
+      return { Timestamp: toTimestampMs(value) };
     }
     // Use Integer for all numbers - WASM will handle type coercion
     return { Integer: value };
   }
   if (typeof value === "string") {
+    if (columnType?.type === "Timestamp") {
+      return { Timestamp: toTimestampMs(value) };
+    }
```
nrainhart, line 157: The query-builder-generator does not support string params, should we either add them as an option, or remove this conversion?

nrainhart (approved):

packages/jazz-tools/src/schema.ts
aeplay: this is an input type for defaults in migrations, changed it to `Date | number`

packages/jazz-tools/src/runtime/query-adapter.ts
aeplay: I think we do a JSON.stringify conversion somewhere in between and that's why it needs to take a string

PR #148, opened 2026-02-26
feat: retry queue

antoniomuso:
We keep the client-side server payload retry queue because reconnect alone does not close the delivery gap. RuntimeCore drains its outbox and hands messages to JS before the HTTP /sync POST is confirmed; if that POST fails at that moment, the payload would be lost without JS-level retry. Reconnect restores the stream and re-attaches server state, but that is a heavier recovery path and not a direct replacement for immediate payload delivery guarantees. The retry queue gives ordered, eventual resend with exponential backoff for transient outages.


aeplay:
I think retry logic should live in RuntimeCore / http-transport so it's shared across all environments.

PR #149, opened 2026-02-26
Add JSON column type with schema validation

aeplay:
## Summary
- add `JSON` / `JSON('<schema>')` column type support across TS DSL, schema SQL parsing/serialization, and wasm bridges
- validate JSON payloads in Rust mutation paths (`insert`, `update`, `undelete`, and server-side write checks), including optional JSON Schema validation
- parse JSON columns back to structured JS values in query results while storing mutation payloads as raw JSON text
- update codegen/query DSL typing to expose JSON columns as `JsonValue` and add JSON query restrictions (`eq/ne/in/isNull`; no order-by)
- cover behavior with Rust + TS tests, including zod v4 schema source usage

## Validation
- `pnpm run format`
- `cargo fmt`
- `pnpm run lint` (passes; oxlint still reports existing repo-wide warnings)
- `cargo test -p jazz-tools json -- --nocapture`
- `cargo test -p jazz-wasm json -- --nocapture`
- `pnpm --filter jazz-tools exec vitest run src/codegen/codegen.test.ts src/runtime/value-converter.test.ts src/runtime/query-adapter.test.ts src/runtime/row-transformer.test.ts`


crates/jazz-tools/src/query_manager/encoding.rs
```diff
@@ -1025,6 +1037,12 @@ fn decode_array_element(data: &[u8], element_type: &ColumnType) -> Result<Value,
             })?;
             Ok(Value::Text(s.to_string()))
         }
+        ColumnType::Json(_) => {
```
nrainhart: Looks like there's a lot of overlap between `decode_array_element` and `decode_column`. Not necessary for this PR, but would be nice to avoid the duplicate logic

crates/jazz-napi/src/lib.rs
```diff
@@ -129,6 +129,8 @@ struct JsColumnType {
     variants: Option<Vec<String>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     columns: Option<Vec<JsColumnDescriptor>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    schema: Option<serde_json::Value>,
```
nrainhart, line 134: So... all column types can have a schema? Is this something we'd want for anything other than JSON columns?

packages/jazz-tools/src/runtime/value-converter.ts
```diff
@@ -54,6 +54,23 @@ export function toValue(value: unknown, columnType: ColumnType): WasmValue {
       }
       throw new Error("Expected Uint8Array or byte array for Bytea column type");
     }
+    case "Json": {
```
nrainhart: Nit: use `toJsonText` here

packages/jazz-tools/src/dsl.ts
```diff
@@ -33,6 +37,48 @@ function normalizeEnumVariants(variants: readonly string[]): string[] {
...
+    throw new Error("JSON schema must be an object or implement ~standard.jsonSchema.");
+  }
+  return schema;
+}
+
+function jsonColumn(): JsonBuilder<JsonValue>;
+function jsonColumn<Schema extends StandardJSONSchemaV1<unknown, unknown>>(
+  schema: Schema,
+): JsonBuilder<StandardJSONSchemaV1.InferOutput<Schema>>;
+function jsonColumn(schema: JsonSchema): JsonBuilder<JsonValue>;
```
nrainhart, line 77: JsonSchema is just Record<string, unknown>. Do we really want to be that permissive with user-provided schemas?

packages/jazz-tools/package.json
```diff
@@ -75,7 +77,8 @@
     "typescript": "^5.3.0",
     "vite-plugin-top-level-await": "^1.6.0",
     "vite-plugin-wasm": "^3.5.0",
-    "vitest": "catalog:default"
+    "vitest": "catalog:default",
+    "zod": "^4.3.6"
```
nrainhart, line 90: Now that we support standard schema, do we really want to ship zod as part of jazz?

nrainhart (approved):

packages/jazz-tools/src/runtime/value-converter.ts
aeplay: Addressed in b1c47e5: mutation JSON serialization in `value-converter` now reuses a shared `toJsonText` helper.

packages/jazz-tools/src/dsl.ts
aeplay: Addressed in b1c47e5: `JsonSchema` is now narrowed from `Record<string, unknown>` to `Exclude<JSONSchema, boolean>` (from `json-schema-to-ts`).

crates/jazz-napi/src/lib.rs
aeplay: Addressed in b1c47e5: Rust now enforces that only `Json` columns can carry `schema` metadata, and there is a unit test covering the rejection path.

crates/jazz-tools/src/query_manager/encoding.rs
aeplay: Agree on the duplication. I left this as follow-up for now since it is outside this JSON-column scope, but can take a cleanup pass in a separate PR.

packages/jazz-tools/package.json
aeplay: Good call. We are not shipping zod in runtime dependencies here; it remains a `devDependency` and is only used in tests for a standard-schema provider.

crates/jazz-tools/src/query_manager/encoding.rs
aeplay: Implemented in 03bceb1: `decode_column` and `decode_array_element` now share a common `decode_non_null_value` path (including shared text/enum UTF-8 handling), so the duplicate decode logic is removed.

PR #161, opened 2026-02-27
ci/e2e-test-expo

antoniomuso:
CI-only Android E2E suite for Expo and runs it on Maestro Cloud.


examples/todo-client-localfirst-expo/.expo/prebuild/cached-packages.json
```diff
@@ -1,4 +1,4 @@
 {
```
aeplay: should this file be checked in?

aeplay (changes requested):
Let's discuss whether to use CloudFlare or sth else for the server

PR #162, opened 2026-02-27
feat: deleteBrowserStorage API

gdorsi:
It is currently impossible to delete browser storage to start fresh, so I'm adding an API to make it possible.

### DIFF
--- a/examples/todo-client-localfirst-react/src/App.tsx
+++ b/examples/todo-client-localfirst-react/src/App.tsx
@@ -27,13 +27,15 @@ function defaultConfig(
 }
 // #endregion context-setup-react
 
-const jazzClient = await createJazzClient(defaultConfig());
+const jazzClient = createJazzClient(defaultConfig());
 
 if (location.origin.includes("localhost")) {
-  Object.defineProperty(window, "jazzClient", {
-    value: jazzClient,
-    writable: true,
-  });
+  jazzClient.then((client) =>
+    Object.defineProperty(window, "jazzClient", {
+      value: client,
+      writable: true,
+    }),
+  );
 }
 
 // #region context-setup-react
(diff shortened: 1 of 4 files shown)


examples/todo-client-localfirst-react/src/App.tsx
```diff
@@ -27,54 +27,19 @@ function defaultConfig(
...
-  }
-
-  if (!client) {
-    return <>{fallback ?? <p>Loading...</p>}</>;
-  }
+if (location.origin.includes("localhost")) {
+  Object.defineProperty(window, "jazzClient", {
+    value: jazzClient,
+    writable: true,
+  });
```
gdorsi: Exposing this to window on dev makes it easier to test stuff

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -1006,6 +1059,57 @@ export class Db {
...
+   * This only deletes `${namespace}.opfsbtree` for the current namespace and does not touch
+   * localStorage-based auth or synthetic-user state.
+   *
+   * Behavior:
+   * - Browser worker-backed Db only (throws in non-browser/non-worker runtimes)
+   * - Serializes with worker reconfigure operations
+   * - Tears down worker + clients, deletes OPFS file, respawns worker
+   * - If file deletion fails, still respawns worker and then rethrows the deletion error
+   */
+  async deleteBrowserStorage(): Promise<void> {
```
Takeno: What about `deleteLocalStorage`? For the same API for other platforms

Takeno (commented):
LGTM

Just a doubt: anything to do with the multi-tab management?

packages/jazz-tools/src/runtime/db.ts
gdorsi: Used `deleteClientStorage`

deleteLocalStorage was making Codex go nuts about the naming including `localStorage`

gdorsi:
> Just a doubt: anything to do with the multi-tab management?

Added a throw in case the current tab is not the leader.

Don't want to mess with multi-tab flow, for now we can just ask the user to close the other tabs.

Takeno (approved):

aeplay (approved):

PR #163, opened 2026-02-27
perf(opfs-tree): switch to FxHashMap and optimize page cache eviction

gdorsi:
Optimizations:
- Switch to [FxHashMap](https://github.com/rust-lang/rustc-hash)
- use top-k instead of full sorting to select the victims to deallocate.
- trigger the eviction during navigation using an allowance, which means we don't start evicting pages immediately after reaching max if we are in the middle of a navigation
- kill victims until we are under the 75% of the max page, to reduce the frequence of which the eviction occurs

Testing on 15k rows reduces the loading times from 2s to 650ms

Before
Screenshot 2026-02-27 at 13 03 53

After
Screenshot 2026-02-27 at 14 51 53

Question: is it ok to use the fast hash everywhere?

### DIFF
--- a/crates/opfs-btree/src/db.rs
+++ b/crates/opfs-btree/src/db.rs
@@ -1,4 +1,5 @@
 use rustc_hash::{FxHashMap, FxHashSet};
+use std::collections::BinaryHeap;
 
 use crate::BTreeError;
 use crate::file::SyncFile;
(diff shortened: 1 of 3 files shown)


crates/opfs-btree/src/db.rs
```diff
@@ -1,4 +1,5 @@
-use std::collections::{HashMap, HashSet};
+use rustc_hash::FxHashMap;
+use std::collections::HashSet;
```
nrainhart: Does it make sense to use FxHashSet as well?

crates/opfs-btree/src/db.rs
```diff
@@ -1101,8 +1104,12 @@ impl<F: SyncFile> OpfsBTree<F> {
             };
             self.pages.insert(current_page_id, page_raw.to_vec());
             self.touch_page(current_page_id);
-            self.evict_pages_if_needed(Some(page_id));
+            self.evict_pages_if_needed_with_allowance(
+                Some(page_id),
+                self.options.read_coalesce_pages,
+            );
         }
+        self.evict_pages_if_needed(Some(page_id));
```
nrainhart: Why do we need to call `evict_pages_if_needed` after the for loop? Aren't the `evict_pages_if_needed_with_allowance` calls enough?

nrainhart (approved):
This is great! Did you check how much of the speedup came just from switching the hash function? It's fine if you didn't, I'm just curious

crates/opfs-btree/src/db.rs
gdorsi: To evict pages after the navigation if we are over max but under the allowance we haveduring the navigation

crates/opfs-btree/src/db.rs
gdorsi: Nice, this shaved other 100ms from the initial load of 15k items!

PR #164, opened 2026-02-27
perf(query-manager): avoid spurious moved deltas on index shifts

nrainhart:
## Motivation

This PR fixes spurious moved events in query result deltas by changing move detection from absolute index comparison to relative-order comparison among shared tuples.

Previously, adding/removing a tuple could shift indices and incorrectly mark many unaffected tuples as moved. Now, tuples are only marked as moved when their order changes relative to each other.

## Validation

Tested inserting an element after loading 10k rows.

### Before

[before.json.gz](https://github.com/user-attachments/files/25604382/before.json.gz)

Main thread:
before-main-thread

Worker: 
before-worker

### After

[after.json.gz](https://github.com/user-attachments/files/25604389/after.json.gz)

Main thread:

after-main-thread

Worker:

after-worker


gdorsi (approved):
Great job!

PR #183, opened 2026-03-02
feat: localOnly query propagation for devTool extension

Takeno:
This PR adds runtime-aware query propagation controls to the inspector.

- Introduces runtime info (standalone vs extension) in the devtools context.
- Enforces full propagation in standalone mode.
- Adds an extension-only switch in Data Explorer (next to Tables) to toggle propagation between:
  - localOnly (local-only in runtime options)
  - network (full in runtime options)
- Wires table data subscriptions to use the selected propagation mode.

### DIFF
--- a/packages/inspector/src/components/data-explorer/TableDataGrid.tsx
+++ b/packages/inspector/src/components/data-explorer/TableDataGrid.tsx
@@ -7,7 +7,7 @@ import {
   useReactTable,
 } from "@tanstack/react-table";
 import { allRowsInTableQuery, type DynamicTableRow } from "jazz-tools";
-import { useAll } from "jazz-tools/react";
+import { useDb } from "jazz-tools/react";
 import { useEffect, useMemo, useState } from "react";
 import { Navigate, useParams } from "react-router";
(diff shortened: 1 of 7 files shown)


packages/inspector/src/components/data-explorer/TableDataGrid.tsx
```diff
@@ -30,18 +30,32 @@ export function TableDataGrid() {
...
+      (delta) => {
+        setRows(delta.all);
+      },
+      { propagation: queryPropagation },
+    );
+
+    return () => {
+      unsubscribe();
+    };
+  }, [db, queryBuilder, queryPropagation]);
```
gdorsi, line 57: How about adding `queryPropagation` to `useAllBase`?

We could use that internally without exposing the option to users

gdorsi (approved):

packages/inspector/src/components/data-explorer/TableDataGrid.tsx
Takeno: Since we created a new `QueryOptions` type, I will make a new PR to introduce that parameter to the react hooks

PR #196, opened 2026-03-03
Runtime scheduler hook for deferred subscribe binding

gdorsi:
I've started working on testing the UX of a "filter" input and noticed that sometimes the input becomes laggy due to the main thread building the subscription object

Screenshot 2026-03-03 at 15 54 59

Since there is no good reason for blocking the main thread I've wrapped the subscribe in the [Scheduler API](https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/postTask) to make the subscription fire without blocking the user interactions.

Screenshot 2026-03-03 at 16 01 27

Currently the priority is set to `user-visible`, we could also make it lower and go for `background`:

> [user-visible](https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API#user-visible)
Tasks that are visible to the user but not necessarily blocking user actions. This might include rendering non-essential parts of the page, such as non-essential images or animations.

> This is the default priority for scheduler.postTask() and scheduler.yield().

> [background](https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API#background)
Tasks that are not time-critical. This might include log processing or initializing third party libraries that aren't required for rendering.

For this kind of task is probably best suited to go with `background`, but it feels more risky.
With the background the task is scheduled after React non-critical rendering parts, and we probably want to the subscription start as early as possible without interrupting the user interaction.

Screenshot 2026-03-03 at 16 14 18

### DIFF
--- a/packages/jazz-tools/src/runtime/client.for-request.test.ts
+++ b/packages/jazz-tools/src/runtime/client.for-request.test.ts
@@ -30,6 +30,7 @@ function makeClient() {
   const subscribeCalls: Array<
     [string, string | undefined, string | undefined, string | undefined]
   > = [];
+  const unsubscribeCalls: number[] = [];
   const subscribeCallbacks: Array<Function> = [];
 
   const runtime: Runtime = {
@@ -66,7 +67,9 @@ function makeClient() {
       subscribeCallbacks.push(onUpdate);
       return 1;
     },
-    unsubscribe: () => {},
+    unsubscribe: (handle: number) => {
+      unsubscribeCalls.push(handle);
+    },
     insertDurable: async () => "00000000-0000-0000-0000-000000000001",
     updateDurable: async () => {},
     deleteDurable: async () => {},
@@ -97,8 +100,13 @@ function makeClient() {
     client: new JazzClientCtor(runtime, context, "edge"),
     queryCalls,
     subscribeCalls,
+    unsubscribeCalls,
     subscribeCallbacks,
   };
+}
+
+async function flushMicrotasks(): Promise<void> {
+  await Promise.resolve();
 }
 
 function makeClientWithContext(context: AppContext): JazzClient {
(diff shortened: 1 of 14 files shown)


packages/jazz-tools/src/runtime/client.for-request.test.ts
```diff
@@ -369,9 +383,219 @@ describe("JazzClient.forRequest", () => {
...
+  });
+
+  it("unsubscribe unknown handle is a no-op", () => {
+    const { client } = makeClient();
+    expect(() => client.unsubscribe(123_456)).not.toThrow();
+  });
+
+  it("subscribe bind failure cleans up subscription state", async () => {
+    const runtime: Runtime = {
+      insert: () => "00000000-0000-0000-0000-000000000001",
```
nrainhart: Nit: i'd extract this into a function and just pass the subscribe function, to reduce visual noise across tests

packages/jazz-tools/src/runtime/client.ts
```diff
@@ -704,8 +779,28 @@ export class JazzClient {
    * @param subscriptionId ID returned from subscribe()
    */
   unsubscribe(subscriptionId: number): void {
-    this.runtime.unsubscribe(subscriptionId);
+    const pending = this.pendingSubscriptions.get(subscriptionId);
+    if (pending) {
+      pending.canceled = true;
```
nrainhart: Do we really need this field? The only place we're checking it is in `bindPendingSubscription`, and it's reading the map and checking the canceled field synchronously

packages/jazz-tools/src/runtime/client.ts
```diff
@@ -387,6 +416,10 @@ export class JazzClient {
   private streamController: SyncStreamController;
   private serverClientId: string = generateClientId();
   private subscriptions = new Map<number, SubscriptionCallback>();
+  private pendingSubscriptions = new Map<number, PendingSubscriptionState>();
+  private boundSubscriptions = new Map<number, number>();
+  private nextProvisionalSubscriptionId = 1;
```
nrainhart: Mind adding a small comment for these new fields?

nrainhart (approved):
The approach is kind of hacky, but I think the improvement is worth it, and agree it doesn't make sense to change the runtime.subscribe api right now

PR #200, opened 2026-03-03
perf: encode main thread<>worker sync messages using postcard

nrainhart:
## Description

Replaces the existing JSON serialization for sync messages sent between the main thread and the worker thread with a faster binary encoding using [postcard](https://docs.rs/postcard/latest/postcard/).

## Discussion

None of the faster-than-JSON Rust<>Rust serialization libraries I found support schema evolution, so I decided to keep binary encoding scoped to main thread<>worker (& cross-tab communication). For communication across the network we continue to use JSON.

I tested multiple Rust crates:
- https://docs.rs/postcard/latest/postcard/
- https://docs.rs/bitcode/latest/bitcode/
- https://docs.rs/rkyv/latest/rkyv/

And found `postcard` to be the fastest in WASM. Keeping benchmarks here for future reference:
- [json-encoding.json.gz](https://github.com/user-attachments/files/25723029/1-json-encoding.json.gz)
- [postcard-encoding.json.gz](https://github.com/user-attachments/files/25723034/2-postcard-encoding.json.gz)
- [bitcode-encoding.json.gz](https://github.com/user-attachments/files/25723039/3-bitcode-encoding.json.gz)
- [rkyv-encoding.json.gz](https://github.com/user-attachments/files/25723041/4-rkyv-encoding.json.gz)

The code is in https://github.com/garden-co/jazz2/tree/perf/sync-payload-binary-encoding-pocs.

## Performance

Scenario: load 10k rows

#### Before

before-main-thread

#### After

after-main-thread

#### Before

before-worker

#### After

after-worker

### DIFF
--- a/crates/jazz-tools/src/query_manager/types/value.rs
+++ b/crates/jazz-tools/src/query_manager/types/value.rs
@@ -1,4 +1,4 @@
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::object::ObjectId;
 
@@ -10,8 +10,7 @@ use super::*;
 /// `PartialEq`/`Eq` are implemented manually because `f64` does not implement
 /// `Eq`. We use bitwise comparison (`f64::to_bits`) so that NaN == NaN and
 /// -0.0 != 0.0, which is the correct semantics for storage identity.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(tag = "type", content = "value")]
+#[derive(Debug, Clone)]
 pub enum Value {
     Integer(i32),
     BigInt(i64),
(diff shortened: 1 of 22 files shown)


packages/jazz-tools/src/worker/jazz-worker.ts
```diff
@@ -106,7 +106,21 @@ function enqueueSyncMessageForMain(payload: string): void {
...
+
+function collectPayloadTransferables(payloads: (Uint8Array | string)[]): Transferable[] {
+  const transferables = [];
+  for (const payload of payloads) {
+    if (payload instanceof Uint8Array) {
+      transferables.push(payload.buffer);
+    }
+  }
+  return transferables;
 }
```
gdorsi, line 124: Why are we doing this?

gdorsi (approved):

packages/jazz-tools/src/worker/jazz-worker.ts
nrainhart: Payloads could be either strings or Uint8Arrays, and we can only transfer the latter. Catalogue sync updates are not binary-encoded, but we still send them to the main thread

crates/jazz-tools/src/query_manager/types/value.rs
```diff
@@ -30,6 +29,137 @@ pub enum Value {
...
+    Text(String),
+    Timestamp(u64),
+    Uuid(ObjectId),
+    Bytea(Vec<u8>),
+    Array(Vec<ValueHuman>),
+    Row(Vec<ValueHuman>),
+    Null,
+}
+
+/// Use externally-tagged enum for binary serialization (postcard does not support internally-tagged enums).
```
nrainhart, line 49: Postcard messages are not self-describing, and therefore incompatible with internally tagged enums (https://docs.rs/postcard/latest/postcard/struct.Deserializer.html)

PR #206, opened 2026-03-04
feat(sql): parse EXISTS in policy expressions

joeinnes:
`parse_policy_primary` did not handle `EXISTS`, causing it to fall through to the column-name path and fail with `Expected: policy operator after column, got LParen`.

## Changes

- Add `Token::Exists`, `Token::From`, `Token::Where` to the lexer
- Handle `Token::Exists` in `parse_policy_primary`, producing `PolicyExpr::Exists { table, condition }`
- Two tests: parse and round-trip


joeinnes:
Replacing with a clean branch on top of main

PR #209, opened 2026-03-04
feat: in-memory storage driver

Takeno:
**Summary**
This PR introduces a first-class in-memory runtime mode for Jazz clients and backend contexts, intended for apps that must always reflect remote state and do not want local-first persistence (for example data explorers/admin consoles/ephemeral dashboards).

It can also be used as a fallback if OPFS is unavailable, like incognito mode in Firefox.

**Key changes**
- Added explicit runtime driver modes in TS:
  - `driver: { type: "persistent", dbName? }` (default)
  - `driver: { type: "memory" }`
  
- Frontend/runtime:
  - `createDb` now routes by `driver.type`
  - `memory` mode skips browser worker/OPFS/leader-election entirely
  - `memory` mode requires `serverUrl`
  - default durability becomes `edge` in memory mode
  
- Backend:
  - `createJazzContext` now uses `driver` config:
    - persistent => existing SurrealKV path
    - memory => new in-memory native runtime path
    
- Native (`jazz-napi`):
  - Added `NapiRuntime.inMemory(...)`
  - Runtime now supports both persistent and memory storage backends

**Open question**

Is `driver` the best naming/option to define this behavior?


specs/status-quo/storage.md
```diff
@@ -97,6 +101,14 @@ Single process, no worker needed. SurrealKvStorage backed by regular files.
 
 NAPI bindings exposing RuntimeCore to Node.js via TokioRuntime.
 
+Current runtime modes:
+
+- `new NapiRuntime(..., dataPath, ...)` for persistent SurrealKV storage.
+- `NapiRuntime.inMemory(...)` for non-persistent in-memory storage.
+
+Internally, the runtime core now uses `Box<dyn Storage + Send>` so both backends are selected
```
nrainhart: Nit: don't think this is too relevant for the spec?
```suggestion
```

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -480,7 +491,8 @@ export class Db {
         localAuthMode: this.config.localAuthMode,
         localAuthToken: this.config.localAuthToken,
         adminSecret: this.config.adminSecret,
-        tier: this.worker ? undefined : "worker",
+        tier: this.worker || useMemoryDriver ? undefined : "worker",
```
nrainhart: Shouldn't we fail if both this.worker and useMemoryDriver are set?

nrainhart: ⚠️ Also: in https://github.com/garden-co/jazz2/pull/200/changes#diff-57ba7760fcc05f780046cd3c10e57548fbc7d9fdc7fc5bc4419c305bb6b600d7R273 I switched to using binary encoding when the tier is "client" (to make main thread<>worker communication faster). I'm pretty sure this will break your PR if you rebase on top of main.

We need to propagate the worker/in-memory mode to the runtime so we can decide when to use binary encoding vs json. Happy to do this myself on top of your PR

nrainhart (approved):
LGTM! Left an important note about a likely merge conflict, happy to work on it

crates/jazz-wasm/src/runtime.rs
```diff
@@ -248,12 +248,14 @@ impl Scheduler for WasmScheduler {
 /// The callback is set lazily via `on_sync_message_to_send()`.
 pub struct JsSyncSender {
     callback: RefCell<Option<Function>>,
+    use_binary_encoding: bool,
 }
 
 impl JsSyncSender {
-    fn new() -> Self {
+    fn new(use_binary_encoding: bool) -> Self {
```
nrainhart, line 337: I like this change!

crates/jazz-wasm/src/runtime.rs
```diff
@@ -326,13 +327,16 @@ impl WasmRuntime {
...
+    /// * `use_binary_encoding` - Optional outgoing sync payload encoding mode.
+    ///   `Some(true)` emits postcard bytes (`Uint8Array`), otherwise JSON strings.
     #[wasm_bindgen(constructor)]
     pub fn new(
         schema_json: &str,
         app_id: &str,
         env: &str,
         user_branch: &str,
         tier: Option<String>,
+        use_binary_encoding: Option<bool>,
```
nrainhart, line 421: I think we can just make this `bool`

crates/jazz-wasm/src/runtime.rs
```diff
@@ -263,15 +265,14 @@ impl JsSyncSender {
...
+    fn send_sync_message(&self, message: OutboxEntry) {
         if let Some(ref callback) = *self.callback.borrow() {
             let is_catalogue = message.payload.is_catalogue();
             let (destination_kind, destination_id) = match message.destination {
                 Destination::Server(server_id) => ("server", server_id.0.to_string()),
                 Destination::Client(client_id) => ("client", client_id.0.to_string()),
             };
-            let use_binary_encoding = sender_tier == "client" || destination_kind == "client";
-            if use_binary_encoding {
+            if self.use_binary_encoding {
```
nrainhart: ⚠️ We should keep the `destination_kind` check. For that case it always makes sense to send binary payloads

PR #219, opened 2026-03-05
feat: make use-all work with concurrent rendering

gdorsi:
Fixes use-all to enable concurrent rendering on React

Enables nice patterns on loading with suspense, and [ViewTransition stuff](https://react.dev/reference/react/ViewTransition#animating-enter-exit-with-activity)

https://github.com/user-attachments/assets/482f6fa1-1e96-4f6c-a7ec-e3feeae4a0aa

### DIFF
--- a/docs/content/docs/reading-data.mdx
+++ b/docs/content/docs/reading-data.mdx
@@ -60,6 +60,16 @@ Use framework bindings immediately after subscription concepts:
   </Tab>
 </Tabs>
 
+### React Suspense and Transitions
+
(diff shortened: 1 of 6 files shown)


aeplay (changes requested):
Implementation looks good to me - but can you please show the usage using a dedicated docs example app, not the main one (would like to keep that one clean for people who don't know about suspense/transitions yet)

You can then show snippets from the docs example app in the docs for reading data with react

aeplay (approved):

PR #223, opened 2026-03-05
Fix cold offset query sync scope

gdorsi:
## Summary
- sync the ordered prefix required to replay paginated queries locally instead of syncing only the current server-side window
- track the pagination node in the query graph so downstream sync scope can include `offset + limit` rows when needed
- add a regression test for a cold remote subscription with a non-zero offset

## Repro
- with a large dataset, a direct cold load at `offset=500` returned no rows
- navigating from page 1 to page 10 in the same session worked because earlier rows had already been cached locally

## Testing
- cargo test -p jazz-tools e2e_client_ -- --nocapture
- cargo test -p jazz-tools offset_limited_subscription_shifts_window_when_deleting_row_before_window -- --nocapture


aeplay:
superseded by #226

PR #235, opened 2026-03-06
feat(example): add moon-lander-react example with walkthrough

joeinnes:
## Summary

- Adds `examples/moon-lander-react` — a real-time multiplayer moon lander game built with Jazz, React, and Phaser, with a full browser test suite (102 tests across 8 files)
- Adds a Marp walkthrough (`jazz-moon-lander.md`) with a shared `theme-jazz.css` (theme name: `jazz`) covering the key Jazz APIs used
- Applies the same `jazz` Marp theme to the existing `examples/wequencer` walkthrough
- Fixes `useAll` for uncollected deposits to use the `"edge"` tier so `undefined` correctly drives settled detection

## Test plan

- [x] All 102 browser tests pass (`pnpm exec vitest run --config vitest.config.browser.ts`)
- [x] Marp walkthrough renders correctly with `--theme-set walkthrough/theme-jazz.css --html`


aeplay (changes requested):
1) for the walkthroughs, can we use snippet ranges like we do in the doc examples so the code doesn't get stale?

2) can we actually not do write batching? How does it run at 60FPS writes?

joeinnes:
## Stress test results — unthrottled vs 200ms interval

Tested on commit fca92f89 (no throttle) vs f8b9a3fb (200ms interval).

| Test | NO_THROTTLE (5 runs) | THROTTLED — 200ms interval (10 runs) |
|---|---|---|
| deposit collected by A disappears for B | 5/5 FAIL | 0/10 FAIL |
| shared deposit appears in receiver's inventory | 5/5 FAIL | 3/10 FAIL |
| All others | 0/5 FAIL | 0/10 FAIL |

**Root cause:** removing the 200ms flush interval eliminated rate-limiting on `updateState()`. The game runs at 60fps; with position changing every frame, the refactor produces up to 60 player-row writes/second vs 5/second before. This floods the Jazz server and causes stream connect timeouts for isolated BrowserContext clients.

**"shared deposit" is pre-existing flaky** (3/10 failures even on the throttled version) — the refactor turns it from flaky into deterministic.

joeinnes:
extensive manual testing, successful soak tests, successful repeated runs of the browser test suite, seems the rebase to use fjall as backend has worked

gdorsi (approved):
To the moon!

PR #250, opened 2026-03-08
Add $canRead/$canEdit/$canDelete permission introspection columns

aeplay:
## Summary
- add `$canRead` / `$canEdit` / `$canDelete` as explicit permission introspection magic columns that stay out of `SELECT *`
- evaluate them through the shared policy-context machinery used by real mutations, with reactive dependency tracking, joined-query support, and graph memory accounting for magic-column dependencies
- extend the generated TypeScript DSL so magic columns work end to end in `select(...)`, `where(...)`, `orderBy(...)`, and projected result types
- reserve the `$` column prefix for system magic fields so future magic columns cannot conflict with user-defined schema columns or introduced migration columns
- document magic columns in `reading-data`, add TS end-to-end coverage for all three columns, and add schema/lens validation tests for the reserved namespace

## Testing
- pnpm lint
- pnpm --filter jazz-tools exec vitest run src/codegen/codegen.test.ts src/runtime/query-adapter.test.ts src/runtime/row-transformer.test.ts tests/ts-dsl/generated-app.test.ts tests/ts-dsl/query-api.test.ts src/dsl.test.ts src/sql-gen.test.ts
- cargo test -p jazz-tools magic_columns_ --lib
- cargo test -p jazz-tools join_subscription_can_filter_and_project_magic_columns --lib
- cargo test -p jazz-tools estimate_memory_size_counts_magic_column_dependencies --lib
- cargo test -p jazz-tools reserved_magic_column --lib


examples/docs/todo-client-localfirst-ts/src/docs-snippets.ts
```diff
@@ -70,6 +70,22 @@ export async function readTodoTitlesWithSelectedProject(db: Db) {
...
+  );
+}
+
+export async function readEditableTodos(db: Db) {
+  return db.all(app.todos.where({ _canEdit: true }).select("title", "_canEdit"));
+}
+
+export async function readDeletableTodos(db: Db) {
+  return db.all(app.todos.where({ _canDelete: true }).select("title", "_canDelete"));
+}
```
gdorsi, line 86: API-wise, we've learned from Jazz1 that having `_` as prefix for "framework" gives too much the "internal you should not use this" kind of vibe.

How about switching to `$`:

```ts
export async function readTodoPermissionIntrospection(db: Db) {
  return db.all(
    app.todos.select("title", "$canRead", "$canEdit", "$canDelete").orderBy("title", "asc"),
  );
}

export async function readEditableTodos(db: Db) {
  return db.all(app.todos.where({ $canEdit: true }).select("title", "$canEdit"));
}

export async function readDeletableTodos(db: Db) {
  return db.all(app.todos.where({ $canDelete: true }).select("title", "$canDelete"));
}
```

examples/docs/todo-client-localfirst-ts/src/docs-snippets.ts
```diff
@@ -70,6 +70,22 @@ export async function readTodoTitlesWithSelectedProject(db: Db) {
 }
 // #endregion reading-select-ts
 
+// #region reading-magic-columns-ts
+export async function readTodoPermissionIntrospection(db: Db) {
+  return db.all(
+    app.todos.select("title", "_canRead", "_canEdit", "_canDelete").orderBy("title", "asc"),
+  );
```
gdorsi, line 77: What's the usefulness of `canRead`?

If I can't read, I can't select the row in first place, am I right?

examples/docs/todo-client-localfirst-ts/src/docs-snippets.ts
```diff
@@ -70,6 +70,22 @@ export async function readTodoTitlesWithSelectedProject(db: Db) {
...
+}
+
+export async function readEditableTodos(db: Db) {
+  return db.all(app.todos.where({ _canEdit: true }).select("title", "_canEdit"));
```
gdorsi: How does this resolves when we have multi-table joins?

Do we look for edit permissions in all the tables?

gdorsi (commented):
From Codex:

New magic-column tracking is not included in graph memory accounting.
- QueryGraph::estimate_memory_size updates for many node classes, but does not include magic_column_tables in [crates/jazz-tools/src/query_manager/graph.rs (line 2518)](app://-/index.html?hostId=local#).
- Impact: memory estimates can be underreported for queries with many magic-column dependencies; low direct risk, but it weakens sizing/guarding behavior.
- Suggestion: include magic_column_tables in the estimation pass for consistency.

gdorsi:
Another point, if we want to take the road of magic fields, we should block people from using them in their columns.

This way we can always add new magic fields without caring too much about possible data-conflicts.

examples/docs/todo-client-localfirst-ts/src/docs-snippets.ts
aeplay: You can in theory create data locally that you won't be allowed to read

examples/docs/todo-client-localfirst-ts/src/docs-snippets.ts
aeplay: selects before the join would be per-table, during the join magic columns like normal ones should get projected into their precise table.col format

PR #262, opened 2026-03-09
codegen: fix relation include overrides for foreign-key columns

Elfo404:
# Summary

Fix generated include result types so relation includes override same-named foreign-key scalar fields only when that relation is actually included.

# Problem

Generated types were incorrect:
```ts
const rows = useAll(app.items.include({ owner: true }));
rows?.[0].owner.id; // error: owner was still typed as string
```

# Root Cause

The generated type combined base fields and included relations in a way that dropped relation overrides for overlapping keys like owner, and it still evaluated I["owner"] even when no include key was present.

# Fix

Generate included relations separately, guard them with key presence, and then replace overlapping base keys only for relations that were actually included.

### DIFF
--- a/packages/jazz-tools/src/codegen/type-generator.ts
+++ b/packages/jazz-tools/src/codegen/type-generator.ts
@@ -204,22 +204,20 @@ function generateRelationsTypes(relations: Map<string, Relation[]>): string[] {
  * Generate WithIncludes types for type-safe include results.
  *
  * Example output:
- *   type TodoIncludedRelations<I extends TodoInclude = {}> =
+ *   type TodoIncludedRelations<I extends TodoInclude = {}, B extends object = Todo> =
  *     ("project" extends keyof I
  *       ? (NonNullable<I["project"]> extends infer RelationInclude
- *           ? RelationInclude extends true
+ *           ? [I] extends [{ [P in "project"]-?: unknown }]
  *             ? { project?: Project }
- *             : RelationInclude extends ProjectQueryBuilder<infer QueryInclude extends ProjectInclude>
- *               ? { project?: ProjectWithIncludes<QueryInclude> }
- *               : RelationInclude extends ProjectInclude
- *                 ? { project?: ProjectWithIncludes<RelationInclude> }
- *                 : {}
+ *             : ("project" extends keyof B
+ *                 ? { [P in keyof Pick<B, "project">]: Pick<B, "project">[P] | Project }
+ *                 : { project?: Project })
  *           : {})
  *       : {}) &
  *     {};
  *
  *   export type TodoWithIncludes<I extends TodoInclude = {}> =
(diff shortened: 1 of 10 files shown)


Elfo404:
@codex review

packages/jazz-tools/src/codegen/type-generator.ts
```diff
@@ -204,17 +204,20 @@ function generateRelationsTypes(relations: Map<string, Relation[]>): string[] {
...
+ *             ? { project?: Project }
+ *             : ("project" extends keyof B
+ *                 ? { [P in keyof Pick<B, "project">]: Pick<B, "project">[P] | Project }
+ *                 : { project?: Project })
+ *           : {})
+ *       : {}) &
+ *     {};
+ *
+ *   export type TodoWithIncludes<I extends TodoInclude = {}> =
+ *     Omit<Todo, Extract<keyof I, keyof TodoRelations>> & TodoIncludedRelations<I, Todo>;
```
nrainhart, line 220: Why is Extract needed? Don't `*Include` and `*Relations` contain the same keys?
```suggestion
 *   export type TodoWithIncludes<I extends TodoInclude = {}> =
 *     Omit<Todo, keyof I> & TodoIncludedRelations<I, Todo>;
```

packages/jazz-tools/tests/codegen/typecheck-generated.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    const diagnostics = await compileTypecheckFixtureDir("nested-reverse-array");
+
+    expect(diagnostics, formatDiagnostics(diagnostics)).toHaveLength(0);
+  });
+
+  it("reports compile errors for unsafe access on maybe-included overlapping relations", async () => {
+    const diagnostics = await compileTypecheckFixtureDir(
+      "include-collisions",
+      "negative-typecheck.ts",
+    );
```
nrainhart, line 118: ⚠️ Do we really need to run the typechecker programatically? Can't we just generate the files and rely on our project's typechecking?

packages/jazz-tools/tests/codegen/fixtures/include-collisions/negative-typecheck.ts
```diff
@@ -0,0 +1,9 @@
+import { app, type TodoInclude } from "./app.ts";
```
nrainhart, line 1: Where is this file coming from? I don't see it as part of this PR 😕

nrainhart (commented):
I don't love how verbose the generated app.ts file is getting, but this is definitely an improvement over what we have. I'd go ahead and merge it, and then we can make the generated types more concise.

Not approving yet because I'm not convinced about the testing approach

Elfo404:
closing this in favor of https://github.com/garden-co/jazz2/pull/291

PR #263, opened 2026-03-09
feat: make query optional in `useAll`

nrainhart:
## Description

React's Rules of hooks make it inconvenient to run queries when some of their inputs aren't yet available (either because of a missing user input, or a previous query not having completed). 

```typescript
const db = useDb();
const session = useSession();
const userId = session?.user_id ?? null;
// Fails because `userId` can be null
const users = useAll(app.users.where({ user_id: { eq: userId } }));
```

This PR makes the query provided to `useAll` optional, to enable the following pattern:

```typescript
const users = useAll(userId ? app.users.where({ user_id: { eq: userId } }) : undefined);
```

When a query is not provided, `useAll` will return `undefined` (and `useAllSuspense` will suspend).

## Discussion

I took a look at how TanStack Query solves this issue. I found their model for [disabling/pausing queries](https://tanstack.com/query/v5/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken) excessively complex (as it introduces a new "paused" state to the query lifecycle which I don't think we need), and it also doesn't solve typing properly (see https://tkdodo.eu/blog/react-query-and-type-script#type-safety-with-the-enabled-option). Our solution is closer to their newer approach of avoiding to provide a query with [skipToken](https://tanstack.com/query/v5/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken).

### DIFF
--- a/packages/jazz-tools/src/react-core/use-all.ts
+++ b/packages/jazz-tools/src/react-core/use-all.ts
@@ -3,17 +3,28 @@ import { type Usable, use } from "react";
 import type { QueryBuilder } from "../runtime/db.js";
 import { useJazzClient } from "./provider.js";
 
-function useAllBase<T extends { id: string }>(
-  query: QueryBuilder<T>,
-  suspense: boolean,
-): T[] | undefined {
-  const { manager } = useJazzClient();
+type UseAllOptions = {
+  suspense?: boolean;
+};
 
-  const key = manager.makeQueryKey(query);
-  const entry = manager.getCacheEntry<T>(key);
-  const dispatch = React.useReducer((_, action) => action, entry.state)[1];
+const SUSPEND_FOREVER: Promise<never> = new Promise(() => {});
+
+function useAllBase<T extends { id: string }>(
+  query?: QueryBuilder<T>,
+  options?: UseAllOptions,
+): T[] | undefined {
+  const { suspense = false } = options ?? {};
+  const { manager } = useJazzClient();
+  const entry = React.useMemo(() => {
+    if (!query) return null;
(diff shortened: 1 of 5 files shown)


packages/jazz-tools/src/react-core/use-all.ts
```diff
@@ -46,10 +64,10 @@ function useAllBase<T extends { id: string }>(
...
 
-export function useAll<T extends { id: string }>(query: QueryBuilder<T>): T[] | undefined {
-  return useAllBase(query, false);
+export function useAll<T extends { id: string }>(query?: QueryBuilder<T>): T[] | undefined {
+  return useAllBase(query, { suspense: false });
 }
 
-export function useAllSuspense<T extends { id: string }>(query: QueryBuilder<T>): T[] {
-  return useAllBase(query, true) as T[];
+export function useAllSuspense<T extends { id: string }>(query?: QueryBuilder<T>): T[] {
```
gdorsi, line 71: Is it right to make the query optional on `useAllSuspense`?

Last time I worked on this stuff, I've found some issues on making the input optional.

gdorsi (approved):

packages/jazz-tools/src/react-core/use-all.ts
nrainhart: Do you remember what type of issue you ran into? From my tests this is working correctly, but I'm not 100% sure if having a separate promise for suspending in this case could cause any issues (I remember the React folks recommend keeping the promise identity constant)

aeplay (approved):

PR #265, opened 2026-03-09
Add opfs-btree crash-window regression test

aeplay:
## Summary
- add a crash-injection SyncFile test helper for opfs-btree
- reproduce the checkpoint window where data pages are durable before the superblock swap
- assert reopen can hit the same stale total_pages / out-of-bounds page id failure seen after reload mid-write

## Testing
- cargo test -p opfs-btree
- cargo clippy --workspace -- -D warnings
- cargo fmt --all


aeplay:
superseded/fixed by #266

PR #270, opened 2026-03-10
spec: schema directory layout and checkpoint-based migrations

joeinnes:
## Summary

- Adds `specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md` specifying two planned changes to the build tooling:
  - Move auto-generated versioned schema and migration files into `versions/` and `migrations/` subdirectories, keeping the schema root clean
  - Replace auto-versioning-on-every-build with explicit `--checkpoint` flag, so iterative edits during development don't accumulate permanent version history

## Test plan

- [ ] Spec reviewed and agreed before implementation begins

### DIFF
--- a/specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md
+++ b/specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md
@@ -26,7 +26,9 @@ schema/
 └── migrations/
     ├── migration_v1_v2_101ec784deec_ba2b07e7513e.ts
     ├── migration_v1_v2_fwd_101ec784deec_ba2b07e7513e.sql
-    └── migration_v1_v2_bwd_101ec784deec_ba2b07e7513e.sql
+    ├── migration_v1_v2_bwd_101ec784deec_ba2b07e7513e.sql
+    ├── migration_wip_fwd.sql   ← auto-regenerated each build, deleted at checkpoint
+    └── migration_wip_bwd.sql   ← auto-regenerated each build, deleted at checkpoint
 ```
 
 ### Files to change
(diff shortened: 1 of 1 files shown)


specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md
```diff
@@ -0,0 +1,110 @@
...
+### Current behaviour
+
+`jazz-tools build` creates a new `versions/schema_v{N}_{hash}.sql` and corresponding migration files on every build where the schema hash has changed. This means each iterative schema edit during development produces a new permanent version + migration pair, which is rarely what the developer wants.
+
+### Target behaviour
+
+**`jazz-tools build` (default)**
+
+- Regenerates `current.sql` and `app.ts`.
+- Does **not** create any new file in `versions/`.
```
nrainhart, line 54: I thought we'd get the file on first build, and then update it/re-create it on every subsequent build

specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md
```diff
@@ -0,0 +1,110 @@
...
+
+    ```
+    Schema has changed since last checkpoint.
+    Run `jazz-tools build --checkpoint` to save this version and generate
+    migrations so old clients can migrate to the current schema.
+    ```
+
+  If `versions/` is empty (no checkpoint exists yet) the wip migration and message are suppressed.
+
+**`jazz-tools build --checkpoint`**
```
nrainhart, line 67: I'd try to use the "migration" term, to make this more consistent with other DBs. Either `build --migrate` or just `migrate` is fine by me

specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md
nrainhart: Discussed it offline. My rationale was making it more explicit that you have not-yet-checkpointed schema changes, but that's also obvious when you see there's a diff in current.ts/sql

specs/todo/a_mvp/schema_directory_layout_and_checkpoints.md
joeinnes: I kind of specifically wanted to avoid `migrate`, because as a verb, it implies to me that it actively *does* something, whereas actually the operation doesn't touch your data at all just generates the files that describe the data. That said, maybe I'm overthinking it, I'm not passionately against `migrate`, and happy to adjust if we think that `migrate` is better/easier to understand?

aeplay:
Closing because we'll do something else 🤷‍♂️

PR #274, opened 2026-03-10
spec: ordered-index top-k query path

gdorsi:
## Minimal Mental Model

This spec is explainable in one sentence:

"Read the rows in the order the query wants, probe anything that can be answered exactly by index, materialize only what survives, and stop as soon as the window boundary is known."

I've implemented a POC for this spec, which lives here: https://github.com/garden-co/jazz2/pull/271/

Results are quite good.


nrainhart (approved):
Everything sounds pretty reasonable to me

antoniomuso (approved):
LGTM

PR #277, opened 2026-03-10
QueryOptions for frameworks bindings

Takeno:
**Summary**

This PR adds `QueryOptions` support to framework-level live query APIs so UI subscriptions can use the same delivery controls as `db.subscribeAll(...)`.

React and React Native hooks now accept `QueryOptions`, and Svelte `QuerySubscription` now accepts the full `QueryOptions` shape instead of only `{ tier }`. Internally, subscription cache keys now include canonicalized query options so default-equivalent option objects reuse the same entry while meaningful option changes produce a new subscription.

**Changes**

- Added `options?: QueryOptions` to:
  - `useAll(...)`
  - `useAllSuspense(...)`
  - `new QuerySubscription(query, options?)`
- Forwarded full `QueryOptions` from framework subscriptions to `db.subscribeAll(...)`
- Updated framework docs/examples to show query options only in option-specific sections, especially durability/read-option usage
- Documented `propagation` alongside existing read options in the shared reading-data docs

### DIFF
--- a/packages/jazz-tools/src/react-core/use-all.ts
+++ b/packages/jazz-tools/src/react-core/use-all.ts
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { type Usable, use } from "react";
-import type { QueryBuilder } from "../runtime/db.js";
+import type { QueryBuilder, QueryOptions } from "../runtime/db.js";
 import { useJazzClient } from "./provider.js";
 
 type UseAllOptions = {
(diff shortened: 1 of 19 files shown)


packages/jazz-tools/src/react-core/use-all.ts
```diff
@@ -11,15 +11,16 @@ const SUSPEND_FOREVER: Promise<never> = new Promise(() => {});
 
 function useAllBase<T extends { id: string }>(
   query?: QueryBuilder<T>,
+  queryOptions?: QueryOptions,
   options?: UseAllOptions,
 ): T[] | undefined {
   const { suspense = false } = options ?? {};
   const { manager } = useJazzClient();
   const entry = React.useMemo(() => {
```
gdorsi, line 19: Unrelated, but I don't think that a useMemo here would give us any gain.

Both query and queryOptions refs are going to be new on each render.

packages/jazz-tools/src/subscriptions-orchestrator.ts
```diff
@@ -341,3 +341,31 @@ export class SubscriptionsOrchestrator {
     this.queryDefinitions.delete(entry.key);
   }
 }
+
+function normalizeQueryOptions(options?: QueryOptions): QueryOptions | undefined {
```
gdorsi: Which problems does this function solve?

gdorsi (approved):

packages/jazz-tools/src/subscriptions-orchestrator.ts
Takeno: Initially, it was to have a string more readable. But it has no sense, and it will cause additional changes when we add new QueryOptions in the future.
Removed in favor of a simpler `JSON.stringify`

PR #283, opened 2026-03-11
Add indexed query path for ordered top-k scans

gdorsi:
Implementation of the spec https://github.com/garden-co/jazz2/blob/main/specs/todo/a_mvp/ordered_index_topk_query_path.md


gdorsi:
On the benchmarks, I think that the difference is not big enough to indicate a real regression.

Feeded the result to Codex, the also deemed the flagged regressions as noise.

gdorsi:
Split the NAPI Fjall integration coverage into #309 so this PR stays focused on the ordered top-k query path changes.

PR #287, opened 2026-03-11
fix(sync_manager): merge ephemeral payload claims into server session

joeinnes:
## Summary

Commit 35e32128 prevented payload sessions from overriding the server-established client session, which was correct for `user_id` (identity must be authoritative from auth), but it also discarded ephemeral per-subscription claims such as `join_code` supplied for invite flows — breaking the private chat invite flow.

When both sessions carry the same `user_id`, this fix merges their claims: payload claims serve as the base, client session claims extend and override them. This means:

- `user_id` cannot be spoofed (explicit mismatch check rejects and warns)
- Auth-established claims (`auth_mode`, `local_mode`, etc.) cannot be overwritten by the payload — client session wins on collision
- Ephemeral per-subscription claims like `join_code` pass through to policy evaluation

## Security notes

- Reviewed for `__jazz_outer_row` injection: not a risk — outer-row context is resolved via expression binding (`bind_outer_row_refs`), not via session claims; by the time the inner EXISTS evaluates, all outer-row refs are substituted with literals
- Write-path permission checks (`ObjectUpdated`, `ObjectTruncated`) use `client.session` directly and are unaffected by this change
- The `(None, payload_session)` branch (unauthenticated clients) is pre-existing behaviour, unchanged here

## Test plan

- [ ] Existing sync_manager tests pass
- [ ] Invite flow in chat-react example works end-to-end (Alice creates private chat, Bob joins via invite link)


joeinnes:
Superseded by #198 which already landed on main — origin/main now passes the raw payload session directly to PendingQuerySubscription, so join_code claims reach policy evaluation correctly.

PR #293, opened 2026-03-11
Inspector - Data Explorer mutations

Takeno:
Data mutation + data filtering

Untitled2


packages/inspector/src/components/data-explorer/TableDataGrid.tsx
```diff
@@ -46,43 +54,43 @@ export function TableDataGrid() {
...
+        builder = builder.where({ [filter.column]: filter.value });
+      } else {
+        builder = builder.where({
+          [filter.column]: {
+            [filter.operator]: filter.value,
+          },
+        });
+      }
+    }
+    return builder.orderBy(sortColumn, sortDirection).limit(queryLimit).offset(queryOffset);
```
gdorsi, line 85: Shall we add a + 1 to `queryLimit`?

Otherwise `hasNextPage` will be always false

Takeno: Nice catch! It was already [in place](https://github.com/garden-co/jazz2/pull/293/changes#diff-233cec59e3d11f377b2ce8ca30b88ffd3050952e3f3040fe8e72dfb7282c7eb4L60), but I removed it in some refactoring and forgot to add it again.

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
```diff
@@ -0,0 +1,366 @@
...
+export function TableFilterBuilder({
+  schemaColumns,
+  clauses,
+  onClausesChange,
+}: TableFilterBuilderProps) {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const dialogRef = useRef<HTMLDialogElement | null>(null);
+  const canUseModalApi =
+    typeof HTMLDialogElement !== "undefined" &&
+    typeof HTMLDialogElement.prototype.showModal === "function";
```
gdorsi: Should this be on module level?

Maybe in a shared constant or getter

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
```diff
@@ -0,0 +1,366 @@
...
+  useEffect(() => {
+    if (!canUseModalApi) return;
+    const dialog = dialogRef.current;
+    if (!dialog) return;
+    if (isModalOpen && !dialog.open) {
+      dialog.showModal();
+    } else if (!isModalOpen && dialog.open) {
+      dialog.close();
+    }
+  }, [isModalOpen, canUseModalApi]);
```
gdorsi: nit: do we need this effect?

This looks like stuff that could be moved in the event handlers

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
gdorsi: Well actually we could just use the API without checking the support.

Coverage is good enough.

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
```diff
@@ -0,0 +1,366 @@
...
+      </button>
+      {clauses.length > 0 ? (
+        <div className={styles.summaryBox}>
+          <code className={styles.summary}>{activeFiltersSummary}</code>
+        </div>
+      ) : null}
+      <dialog
+        ref={dialogRef}
+        className={styles.modal}
+        {...(!canUseModalApi ? { open: isModalOpen } : {})}
```
gdorsi: We should either control open or call showModal, not both.

From a quick research on my phone, open doesn't work really well with React so I would suggest going with showModal

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
```diff
@@ -0,0 +1,366 @@
...
+        onClose={() => setIsModalOpen(false)}
+        aria-label="Filter rows"
+      >
+        <form
+          className={styles.form}
+          onSubmit={(event) => {
```
gdorsi: nit: this handler looks quite bit, how about moving this on the function body?

packages/jazz-tools/src/dev-tools/dev-tools.ts
```diff
@@ -312,6 +313,86 @@ function hookRegistration(
           return;
         }
 
+        const resolveCommandClient = async (): Promise<JazzClient> => {
```
gdorsi, line 316: This stuff is starting to look quite complicated.

How about considering adding something like TRPC, oRPC or Comlink  to abstract the communication?

From the phone now, so can't evaluate them properly, but I think the custom code we are going to remove would probably pay off the library cost.

gdorsi (approved):

packages/inspector/src/components/data-explorer/RowMutationSidebar.tsx
```diff
@@ -0,0 +1,286 @@
...
+    }
+    const nextFields: Record<string, FieldState> = {};
+    for (const column of schemaColumns) {
+      nextFields[column.name] = getInitialFieldState(rowValues[column.name], mode, column);
+    }
+    setFields(nextFields);
+    setErrors({});
+    setSaveError(null);
+    setIsSaving(false);
+  }, [rowValues, mode, schemaColumns]);
```
gdorsi: What's this effect for?

packages/inspector/src/components/data-explorer/RowMutationSidebar.tsx
```diff
@@ -0,0 +1,286 @@
...
+export function RowMutationSidebar({
+  mode,
+  tableName,
+  schemaColumns,
+  targetRowId,
+  rowValues,
+  onCancel,
+  onSave,
+}: RowMutationSidebarProps) {
+  const [fields, setFields] = useState<Record<string, FieldState>>({});
```
gdorsi: Have you evaluated a form library for this?

Complexity-wise I think it would help.

packages/inspector/src/components/data-explorer/TableDataGrid.tsx
aeplay: Anselm's Codex: Confirmed. The branch keeps queryLimit at pageSize + 1 again, and TableDataGrid.test.tsx still asserts the 11-row limit so we don't regress this in the next refactor.

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
aeplay: Anselm's Codex: Done. I removed the controlled open fallback and kept a single imperative dialog path.

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
aeplay: Anselm's Codex: Agreed. That sync effect is gone now; the dialog is opened and closed from the event handlers directly.

packages/inspector/src/components/data-explorer/RowMutationSidebar.tsx
aeplay: Anselm's Codex: I didn't pull a form library into this PR, but I did take the smaller refactor path here: centralized the field state helpers and removed the reset effect. If these editors grow further, I agree react-hook-form is worth revisiting in a follow-up.

packages/inspector/src/components/data-explorer/RowMutationSidebar.tsx
aeplay: Anselm's Codex: It isn't anymore; I removed it. The sidebar now initializes once on mount, and TableDataGrid keys the sidebar by mode:rowId so switching rows still resets cleanly without wiping drafts on live updates.

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
aeplay: Anselm's Codex: Done. The submit, column, operator, value, and remove handlers now live in the function body instead of inline JSX.

packages/inspector/src/components/data-explorer/TableFilterBuilder.tsx
aeplay: Anselm's Codex: Done. The filter dialog now uses showModal()/close() directly, and the shared inspector test setup provides the tiny dialog polyfill the unit tests need.

aeplay: Anselm's Codex: I ended up removing that support flag entirely, so there isn't anything left to hoist to module scope here.

packages/jazz-tools/src/dev-tools/dev-tools.ts
aeplay: Anselm's Codex: I left Comlink/tRPC/oRPC out of this PR to keep the scope tight, but I agree the bridge is a good candidate for a follow-up transport abstraction once this lands.

PR #294, opened 2026-03-11
feat: skip FK validation on unchanged columns during partial updates

antoniomuso:
This PR solves the error in this PR: https://github.com/garden-co/jazz2/pull/259

> The sync system is query-driven. Only objects that match an active query's result set get synced to the client. Here's what the DataPage of the stress test subscribes to:
> ```js
> // Query 1: 50 todos
> app.todos.where({ done: true }).orderBy(...).limit(50)
> 
> // Query 2: 20 projects  
> app.projects.limit(20)
> ```
> With 15,000 todos and 2,000 projects, there is a high probability that the 50 todos retrieved by our query won't have their corresponding projects in memory.
> Consequently, marking a todo as 'done' will trigger a UuidForeignKeyViolation error. Because the update process validates all columns - including associations - the operation fails if the corresponding project entity has not been loaded into memory.
> ```
> MemoryStorage after sync completes:
> 
>   projects index:  [p1, p2, ... p20]          ← 20 out of 15K
>   todos:           [t1, t2, ... t50]          ← each refs some project
>   
>   t1.project = p4217  → not in index → 💥    (even after waiting forever)
>   t2.project = p891   → not in index → 💥
>   t3.project = p12    → in index → ✓         (lucky)
> ```
> 
> This fix skips FK validation on unchanged columns during partial updates.

### UPDATE

This PR now removes checks on FKs completely.


crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -849,13 +922,29 @@ impl QueryManager {
...
+                changed,
+            )?;
+        } else {
+            self.validate_foreign_keys_for_values(
+                storage,
+                &table_name,
+                &descriptor,
+                values,
+                &self.current_branch(),
+            )?;
```
gdorsi: Can't we always go with `validate_foreign_keys_for_changed_columns`?

gdorsi (approved):

crates/jazz-tools/src/query_manager/writes.rs
antoniomuso: Yeah, I update it!

PR #306, opened 2026-03-12
Add failing nested inherits Wasm regression test

aeplay:
## Summary
- add a focused local Wasm runtime regression for a nested permissions graph
- compile the schema through `definePermissions(...)`
- seed `teams -> resource_access_edges -> resources -> brandings` locally and query as `alice`

## Repro
```bash
pnpm --dir packages/jazz-tools exec vitest run src/runtime/local-nested-inherits.integration.test.ts
```

## Current failure
The new test fails on `main` with:

```text
Error: query timed out: {"table":"resource_access_edges","array_subqueries":[],"relation_ir":{"TableScan":{"table":"resource_access_edges"}}}
```

So with this nested grant schema present, even the direct `resource_access_edges` read hangs instead of resolving the inherited access path.

### DIFF
--- a/packages/jazz-tools/src/runtime/local-nested-inherits.integration.test.ts
+++ b/packages/jazz-tools/src/runtime/local-nested-inherits.integration.test.ts
@@ -1,0 +1,253 @@
+import { createHmac } from "node:crypto";
+import { describe, expect, it } from "vitest";
+import type { WasmSchema } from "../drivers/types.js";
+import { definePermissions } from "../permissions/index.js";
+import { JazzClient, type Row } from "./client.js";
+import { translateQuery } from "./query-adapter.js";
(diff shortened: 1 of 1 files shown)


aeplay:
Re-tested this locally. The original repro was invalid: the isolated Wasm path passes once the direct runtime query is called without an explicit settled tier. My earlier version was timing out because I forced  on , not because the nested permissions graph was failing. Closing this PR rather than leaving a bad regression around.

aeplay:
Closing because the current repro is invalid after fixing the direct Wasm query usage.

aeplay:
Re-tested this locally. The original repro was invalid: the isolated Wasm path passes once the direct runtime query is called without an explicit settled tier. My earlier version was timing out because I forced `tier: "worker"` on `runtime.query(...)`, not because the nested permissions graph was failing.

Closing this PR rather than leaving a bad regression around.

PR #309, opened 2026-03-12
test(napi): add Fjall integration coverage

gdorsi:
Includes:
- durable `db.all` integration tests for the NAPI Fjall runtime
- durable `db.subscribeAll` integration tests for the NAPI Fjall runtime
- shared Fjall test helpers and the supporting runtime test utility updates

### DIFF
--- a/packages/jazz-tools/src/runtime/napi.fjall.db.all.integration.test.ts
+++ b/packages/jazz-tools/src/runtime/napi.fjall.db.all.integration.test.ts
@@ -1,0 +1,131 @@
+import { describe, expect, it } from "vitest";
+import {
+  allTodosQuery,
+  createNapiFjallTestEnv,
+  readRowDone,
+  readRowTitle,
+  todoValues,
+} from "./napi.fjall.test-helpers.js";
+
+const env = createNapiFjallTestEnv();
+
+describe("db.all NAPI Fjall integration", () => {
+  it("queries durable rows from a Fjall-backed runtime", async () => {
+    const store = await env.createPersistentStore("query-durable");
+    const { client } = await env.openPersistentClient(store);
+
+    const first = await client.createDurable("todos", todoValues("Task A", false), {
+      tier: "worker",
+    });
+    const second = await client.createDurable("todos", todoValues("Task B", true), {
+      tier: "worker",
+    });
+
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/runtime/napi.fjall.test-helpers.ts
```diff
@@ -0,0 +1,229 @@
...
+  dataPath: string;
+  cleanup(): Promise<void>;
+};
+
+type PersistentClientHandle = {
+  client: JazzClient;
+  shutdown(): Promise<void>;
+};
+
+export const TEST_SCHEMA: WasmSchema = {
```
nrainhart, line 28: What do you think about adding an actual ts schema and relying on codegen for these tests? It'd certainly cut down the boilerplate we need for creating queries and parsing their results

packages/jazz-tools/src/runtime/napi.fjall.db.all.integration.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    const rows = await env.waitForRows(client, allTodosQuery, (entries) => entries.length === 2);
+    const titles = rows.map(readRowTitle).sort();
+
+    expect(titles).toEqual(["Task A", "Task B"]);
+    expect(rows.find((row) => row.id === first.id)).toBeDefined();
+    expect(rows.find((row) => row.id === second.id)).toBeDefined();
+    expect(readRowDone(rows.find((row) => row.id === first.id)!)).toBe(false);
+    expect(readRowDone(rows.find((row) => row.id === second.id)!)).toBe(true);
+  }, 20_000);
```
nrainhart: Wow, we really need such a long timeout?

nrainhart (approved):
This is great!! Left only two non-blocking question

packages/jazz-tools/src/runtime/napi.fjall.test-helpers.ts
gdorsi: Gong to give it a try in a separate PR

packages/jazz-tools/src/runtime/napi.fjall.db.all.integration.test.ts
gdorsi: Removed

PR #311, opened 2026-03-12
spec: add COUNT aggregation intent

gdorsi:
This PR adds the COUNT aggregation MVP intent spec.

It covers:
- the user-facing `count()` query shape and semantics
- the execution and sync model for reactive `COUNT(*)`
- the planned IR, graph, and runtime shaping work for the first implementation slice

[Preview](https://github.com/garden-co/jazz2/blob/codex/count-aggregation-intent/specs/todo/a_mvp/count_aggregation_intent.md)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -25,6 +25,7 @@
 
 - [**client-state-cleanup**](todo/ideas/1_mvp/client-state-cleanup.md) — Garbage collection of server-side state (sync cursors, query subscriptions, session records) for permanently disconnected clients.
 - [**complex-merge-strategies**](todo/ideas/1_mvp/complex-merge-strategies.md) — Per-column/per-table merge strategies beyond LWW (counters, sets, rich text, custom logic).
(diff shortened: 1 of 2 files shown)


specs/todo/a_mvp/count_aggregation_intent.md
```diff
@@ -0,0 +1,392 @@
...
+
+  let baseQuery = app.todos.orderBy("id", "desc");
+  if (showDoneOnly) {
+    baseQuery = baseQuery.where({ done: true });
+  }
+
+  const pageQuery = baseQuery.limit(PAGE_SIZE).offset(page * PAGE_SIZE);
+  const countQuery = baseQuery.count();
+
+  const todos = useAll(pageQuery) ?? [];
```
gdorsi: I think that would be ok for now to just fit `count` in the existing hooks/APIs.

I think at some point we will have to revisit hooks, but for now it should be ok to have a single generic API.

PR #320, opened 2026-03-12
Optimize opfs-btree CRC32 for wasm

gdorsi:
### Summary

- Replaced the `crc32fast` dependency (which falls back to `crc-rs`'s software implementation on WASM) with a hand-rolled slicing-by-16 CRC32 tuned for WASM page checksumming.
- Halves the lookup table footprint from 32 KB (slicing-by-32) to 16 KB (slicing-by-16) for better L1 cache behavior on WASM engines.

### Why not crc32fast?

`crc32fast` is fast on native targets — it uses SSE4.2 CRC32C instructions on x86 and NEON on ARM. On WASM, none of that exists. It falls back to a generic software path (via `crc-rs`) that:

1. **Defaults to bytewise** (`Table<1>`) unless you explicitly opt into `Table<16>`. The bytewise path is ~6x slower than slicing-by-16 per crc-rs's own benchmarks.
2. **Carries generic abstraction overhead.** crc-rs supports CRC-8/16/32/64/82 with arbitrary polynomials, reflection, init/xorout. Each lookup goes through generic code that handles `refin`/`refout` and runtime-configurable polynomial — none of which we need for IEEE CRC32.
3. **Inhibits inlining across crate boundaries.** Without LTO (typical in WASM builds), each `digest.update()` crosses a function boundary. Our `crc32::update` is `#[inline]` within the same crate as `page_checksum`.
4. **Adds WASM binary size.** Monomorphized generic code for `Crc<u32, Table<16>>` plus the crate's supporting types inflate the `.wasm` binary. Our implementation is ~200 lines total.

### Why slicing-by-16 instead of slicing-by-32?

Both process exactly 1 table lookup per input byte — the algorithmic throughput is identical. The difference is cache pressure:

| Approach | Table size | L1 fit (32 KB typical) |
|---|---|---|
| Slicing-by-32 | 32 KB | Saturates L1, evicts page data |
| Slicing-by-16 | 16 KB | Fits with ~16 KB headroom |

For our hot path (checksumming 16 KB B-tree pages), slicing-by-32's tables barely fit in L1 and compete with the page buffer and stack for cache lines. Slicing-by-16 leaves room for other working data, giving consistently higher hit rates after the first few chunks warm the tables.

WASM lacks carryless multiply (PCLMULQDQ) and hardware CRC instructions, so table-based slicing is the ceiling for software CRC32. Slicing-by-16 is the sweet spot: half the memory of slicing-by-32 with no throughput loss per byte.

### What this doesn't change

- On-disk checksum format (still IEEE CRC32).
- Non-WASM builds use `update_state_slow` (bytewise) — native targets can rely on hardware CRC or a dedicated native crate if needed later.
- Public API: `crc32::hash(&[u8]) -> u32` and `crc32::update(u32, &[u8]) -> u32`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -660,15 +660,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crc32fast"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
-dependencies = [
- "cfg-if",
-]
-
-[[package]]
 name = "criterion"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2422,7 +2413,6 @@ name = "opfs-btree"
 version = "0.1.0"
 dependencies = [
  "bf-tree",
- "crc32fast",
  "criterion",
  "fjall",
  "js-sys",
--- a/crates/opfs-btree/Cargo.toml
+++ b/crates/opfs-btree/Cargo.toml
@@ -13,7 +13,6 @@ crate-type = ["cdylib", "rlib"]
 compare-native = ["dep:bf-tree", "dep:rocksdb", "dep:fjall"]
 
 [dependencies]
-crc32fast = "1"
 rustc-hash = "2"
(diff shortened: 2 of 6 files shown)


aeplay:
closing in favor of #326

PR #325, opened 2026-03-13
Fix permissions.ts app import loading during schema build

tobiaslins:
## Summary
- load  through 's CJS loader while keeping  and migrations on the existing ESM path
- preserve generated  compatibility for , , and bare  imports from 
- add CLI regression tests covering all three import spellings

## Testing
- pnpm exec vitest run src/cli.test.ts --config vitest.config.ts
- reproduced the failure in  with published 
- verified the patched local wrapper succeeds against a temp workspace built from the real  schema

### DIFF
--- a/packages/jazz-tools/src/cli.test.ts
+++ b/packages/jazz-tools/src/cli.test.ts
@@ -76,10 +76,10 @@ table("todos", {
 `;
 }
 
-function permissionsSchema(): string {
+function permissionsSchema(appImportPath: string = "./app.js"): string {
   return `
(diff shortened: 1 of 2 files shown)


joeinnes:
As discussed on Discord, adding a test that bypasses the Vitest module loading pipeline.

To ensure red test against main:

```
git checkout main
git cherry-pick ee64de2b
npx tsc --build --force
npx vitest run src/cli.test.ts -t "loads permissions.ts that imports ./app via bin entry point"
```

Thank you @tobiaslins!

PR #326, opened 2026-03-13
Replace crc32fast with xxHash32 for opfs-btree page checksums

gdorsi:
## Summary

- Replace `crc32fast` (CRC32) with `xxhash-rust` (xxh3_64) for B-tree page and superblock checksums.
- Introduces a thin `checksum` module wrapping xxh3 with a streaming `Hasher` for the page checksum's skip-4-bytes pattern.
- The xxh3_64 output is truncated to `u32` to fit the existing 4-byte checksum field — no page format change.

## WASM checksum benchmark (Chromium, 1M iterations per cell)

Throughput in MB/s, measured via `pnpm --filter @jazz/opfs-btree run bench:wasm:hash -- --count 1000000`:

| Data size | crc32 | xxh32 | rapidhash | **xxh3_64** |
|-----------|------:|------:|----------:|------------:|
| 4 KB      | 2,117 | 2,987 |     5,878 |  **16,672** |
| **16 KB** (page size) | 2,144 | 3,049 | 5,676 | **17,165** |
| 64 KB     | 1,945 | 2,788 |     5,744 |  **15,695** |

At the 16 KB page size used by opfs-btree, xxh3 is **3x faster than rapidhash**, **5.6x faster than xxh32**, and **8x faster than crc32**.

## Why xxh3 dominates on WASM

All three algorithms fall back to pure software on WASM (no hardware CRC, no SIMD by default). Their performance separates along two axes: **memory access pattern** and **instruction-level parallelism**.

|  | CRC32 (software) | xxHash32 | xxh3_64 |
|---|---|---|---|
| **Static tables** | 16 KB (16 × 256 × 4 B) | 0 B | 0 B |
| **Per-byte work** | 1 random table lookup + XOR | ~1 arith op (add/mul/rot) | ~1 arith op (add/mul/rot) |
| **Accumulator lanes** | 1 (serial) | 4 (32-bit) | 8 (64-bit) |
| **ILP** | Each byte depends on the previous | 4-way parallel | 8-way parallel, wider multiply |
| **Cache pressure** | Every byte touches a random cache line in the table | None | None |

CRC32's table lookups are the worst possible pattern on WASM engines — data-dependent loads from a 16 KB table that compete for L1 with the page data itself. On native x86/ARM, `crc32fast` bypasses this entirely with single-cycle hardware CRC instructions (SSE4.2 / NEON), but those don't exist in WASM.

xxh32 eliminates the tables but is limited to 4 × 32-bit accumulator lanes. xxh3 goes further: 8 × 64-bit lanes with wider multiplies, giving the WASM engine's pipeline much more independent work per loop iteration. The result is near-linear scaling with data size — hitting ~17 GB/s at 16 KB, close to memory bandwidth.

### What about WASM SIMD?

We benchmarked with `-C target-feature=+simd128` enabled. Surprisingly, explicit SIMD made xxh3 **slower** (7.4 GB/s vs 16.8 GB/s at 16 KB). The V8 JIT appears to do a better job auto-optimizing the scalar xxh3 loop than executing the explicit `core::arch::wasm32` SIMD intrinsics from xxhash-rust. SIMD is left off.

Before:
Screenshot 2026-03-13 at 17 37 01

After:
Screenshot 2026-03-13 at 17 38 58

### DIFF
--- a/crates/opfs-btree/Cargo.toml
+++ b/crates/opfs-btree/Cargo.toml
@@ -17,7 +17,7 @@ rustc-hash = "2"
 thiserror = "2"
 tracing = "0.1"
 serde = { version = "1", features = ["derive"] }
-xxhash-rust = { version = "0.8", features = ["xxh32"] }
+xxhash-rust = { version = "0.8", features = ["xxh3"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 bf-tree = { version = "0.4.8", optional = true }
--- a/crates/opfs-btree/Cargo.toml
+++ b/crates/opfs-btree/Cargo.toml
@@ -13,11 +13,11 @@ crate-type = ["cdylib", "rlib"]
 compare-native = ["dep:bf-tree", "dep:rocksdb", "dep:fjall"]
 
 [dependencies]
-crc32fast = "1"
 rustc-hash = "2"
 thiserror = "2"
 tracing = "0.1"
 serde = { version = "1", features = ["derive"] }
+xxhash-rust = { version = "0.8", features = ["xxh32"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 bf-tree = { version = "0.4.8", optional = true }
--- a/.changeset/pr-326-opfs-checksum.md
+++ b/.changeset/pr-326-opfs-checksum.md
@@ -1,0 +1,8 @@
+---
+"jazz-tools": patch
+"jazz-wasm": patch
+---
+
+Use xxHash-based checksums for `opfs-btree` pages and superblocks to reduce checksum overhead in persistent browser storage.
+
+Existing OPFS stores created by older builds are not checksum-compatible with this change and will need to be recreated after upgrading.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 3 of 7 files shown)


antoniomuso (approved):
LGTM! 🚀

crates/opfs-btree/Cargo.toml
```diff
@@ -13,11 +13,11 @@ crate-type = ["cdylib", "rlib"]
 compare-native = ["dep:bf-tree", "dep:rocksdb", "dep:fjall"]
 
 [dependencies]
-crc32fast = "1"
 rustc-hash = "2"
 thiserror = "2"
 tracing = "0.1"
 serde = { version = "1", features = ["derive"] }
+xxhash-rust = { version = "0.8", features = ["xxh32"] }
```
nrainhart: Looks like there's also simd support for wasm: https://github.com/DoumanAsh/xxhash-rust#hw-acceleration

In theory it should work on browsers, wanna try it out to see if it's even faster?

nrainhart (approved):
This is great! Thanks for looking into this

crates/opfs-btree/Cargo.toml
gdorsi: Been playing with WASM SIMD when tuning CRC32, it become actually slower but I thought that was me.

So I made Claude follow your suggestion and:

```
The surprising finding: SIMD doesn't help here, and may even hurt.

xxh32 (a scalar algorithm) is roughly the same with or without the SIMD flag, as expected.
xxh3 with SIMD is actually slower than without SIMD at 1KB–16KB — the sizes most relevant to Jazz's page checksums. At 16KB (your page size), scalar xxh3 hits 9,456 MB/s vs 7,378 MB/s with SIMD.
Possible explanations:

wasm-opt interference — enabling +simd128 changes LLVM's codegen globally. wasm-opt may produce worse optimized code when SIMD instructions are mixed in.
V8 WASM SIMD JIT overhead — browser VMs may not yet optimize WASM SIMD as aggressively as their scalar paths.
xxhash-rust's WASM SIMD path — the explicit core::arch::wasm32 intrinsics may have more overhead than V8's auto-vectorization of the scalar fallback.
Recommendation for now: Don't enable +simd128 globally. The scalar xxh3 already gives a nice bump over xxh32 at your page size (9,456 vs 7,063 MB/s = 34% faster) without SIMD. If you want to switch the checksum from xxh32 to xxh3, that's worth doing on its own merits — but leave SIMD off.
```

This aligns with the profiling I've done manually, so I've switched to xxh3.

nrainhart: Huh, that's interesting. Sucks that we can't get rely on SIMD support in browsers though

PR #330, opened 2026-03-13
fix: preserve undefined refs in forward array refs, omit in reverse refs

nrainhart:
Added test coverage for resolving missing references across tables.

While doing this I uncovered that we were returning no refs (both for forward and reverse includes) when at least one of them was missing.

This PR establishes how we'll handle missing FKs by default:
- return undefined for scalar FKs
- return array with undefined entries for forward FKs
- skip missing FKs in reverse relations


packages/jazz-tools/tests/ts-dsl/query-api.test.ts
```diff
@@ -112,6 +120,76 @@ describe("TS Query API", () => {
...
+    await db.delete(app.projects, project.id);
+
+    const result = await db.one(
+      app.todos.where({ id: { eq: todo.id } }).include({ project: true }),
+    );
+    assert(result, "Result is not defined");
+    expect(result.project).toBeUndefined();
+  });
+
+  it("include returns 'undefined' for missing array referenced entities", async () => {
```
nrainhart: I'm not 100% sure this is the behavior we want, but since we'll be adding a `requireIncludes` option to hide rows where references are missing, I think we should return undefined refs for array FKs

gdorsi: This means that the return type would be `Array<V | undefined>` which is meh.

We don't have better choices on missing scalar referenced entities, but here I think we can safely hide the missing value.

gdorsi (approved):

packages/jazz-tools/tests/ts-dsl/query-api.test.ts
nrainhart: Yeah, agree. I also went on a side quest to learn how drizzle handles this (https://orm.drizzle.team/docs/relations-v2) and they do the same thing. I'll change this

nrainhart: Done: [`c60db99` (this PR)](https://github.com/garden-co/jazz2/pull/330/commits/c60db9982874f8a4a826cb10cffc1ac1a7f1b42a)

PR #340, opened 2026-03-16
feat: JWKS rotation support for self-hosted server

joeinnes:
## Summary

The self-hosted server (jazz-tools) fetched JWKS once at startup and never refetched. If your identity provider rotated keys, JWTs signed with the new key would be rejected until the server was restarted. The cloud server had rotation support but lacked DoS protection and stale-if-error resilience.

This PR addresses all three:

### JWKS rotation (jazz-tools, new)

Ports the cloud server's cache + retry pattern to the self-hosted server:

- **TTL-based cache** — JWKS cached for 5 minutes (configurable via `JAZZ_JWKS_CACHE_TTL_SECS`)
- **On-demand refresh** — unknown kid or signature mismatch triggers one forced refetch, then retries
- **Error classification** — `Retryable` (may succeed after refresh) vs `Fatal` (malformed JWT, will never succeed)
- **Startup warmth preserved** — cache warmed at startup, misconfigured URLs still fail fast

### Forced-refresh cooldown (both servers)

Prevents unauthenticated callers from using fabricated key IDs to trigger unbounded outbound JWKS fetches:

- 10-second cooldown between forced refreshes
- Per-app tracking on cloud server (App A's rotation doesn't block App B)
- Per-cache-instance on self-hosted (one app per server)

### Stale-if-error fallback (both servers)

When the JWKS endpoint goes down after the cache TTL expires, the server continues validating against the stale cached keyset rather than returning 401 for every request:

- Falls back to stale keyset on any fetch failure (network error, 500, empty keyset)
- Capped by `max_stale` (default 5 min past TTL, configurable via `JAZZ_JWKS_MAX_STALE_SECS`) — revoked keys can't be valid indefinitely
- If no cached keyset exists (first fetch at startup), the error still propagates

### Cleanup

- Removed vestigial `AuthConfig.jwks_set` field — production path always uses `JwksCache`
- Removed `validate_jwt_identity` and `validate_jwt` (superseded by `validate_jwt_with_cache`)

## Files changed

| File | What |
|------|------|
| `jazz-tools/src/middleware/auth.rs` | `JwksCache` (configurable TTL, cooldown, stale-if-error, max_stale), `JwtVerificationError`, async `extract_session` |
| `jazz-tools/src/commands/server.rs` | `JwksCache` on `ServerState`, env var support |
| `jazz-tools/src/routes.rs` | Updated call sites to use cache |
| `jazz-tools/src/main.rs` | Removed `jwks_set` from `AuthConfig` construction |
| `jazz-cloud-server/src/server.rs` | Per-app cooldown on `CachedJwks`, stale-if-error with max_stale, env var support |
| `jazz-tools/tests/test_server.rs` | Programmable JWKS mock with hit counting + custom TTL/max_stale |
| `jazz-tools/tests/auth_test.rs` | 7 new integration tests |
| `jazz-cloud-server/tests/auth_jwks.rs` | 3 new integration tests |

## Environment variables

| Variable | Default | Description |
|----------|---------|-------------|
| `JAZZ_JWKS_CACHE_TTL_SECS` | 300 | JWKS cache TTL in seconds |
| `JAZZ_JWKS_MAX_STALE_SECS` | 300 | Max time (past TTL) to serve stale keyset on fetch failure |

## Test plan

- [x] Unknown kid triggers JWKS refresh and succeeds (jazz-tools)
- [x] Bad signature stays 401 after refresh (jazz-tools)
- [x] Cached JWKS serves consecutive requests without refetch (jazz-tools)
- [x] Rapid fabricated kids trigger at most one refresh within cooldown (both servers)
- [x] Stale JWKS served when endpoint goes down after TTL expiry (both servers)
- [x] Stale JWKS refused after max_stale expires (both servers)
- [x] All 21 unit tests pass
- [x] All 15 jazz-tools integration tests pass (5 pre-existing failures on `main` unrelated)
- [x] All 9 cloud server integration tests pass
- [x] Full workspace: 982 tests passed, 0 failed
- [x] `cargo fmt --check` and `cargo clippy -D warnings` clean


crates/jazz-tools/src/middleware/auth.rs
```diff
@@ -381,6 +493,98 @@ pub fn validate_jwt_identity(token: &str, config: &AuthConfig) -> Result<Verifie
...
+    })?;
+
+    match verify_jwt_signature_with_jwks(token, &refreshed_jwks) {
+        Ok(verified) => Ok(verified),
+        Err(JwtVerificationError::Retryable(e) | JwtVerificationError::Fatal(e)) => {
+            warn!(error = %e, "JWT validation failed after JWKS refresh");
+            Err(JwtError::Invalid(e))
+        }
+    }
+}
```
Takeno, line 630: Should we add a kind of protection from this `force` refresh? An unauthenticated caller can send a stream of bogus JWTs and force one outbound JWKS fetch per request, which turns /sync and /events into a simple DoS/amplification path against both the server and the IdP

crates/jazz-tools/src/middleware/auth.rs
```diff
@@ -139,6 +145,104 @@ impl std::fmt::Display for JwtError {
...
+
+        let jwks = fetch_jwks(&self.http_client, &self.endpoint).await?;
+
+        *self.cached.write().await = Some(CachedJwksEntry {
+            endpoint: self.endpoint.clone(),
+            fetched_at_us: now_timestamp_us(),
+            set: jwks.clone(),
+        });
+
+        Ok(jwks)
```
Takeno: If the request fails, we reject the request instead of using the old cached value. If that’s intentional to match cloud, it may still be worth documenting; otherwise stale-if-error would improve availability a lot.

joeinnes: Maybe we need some kind of configurable grace period where we continue using the stale JWKS for an additional X (maybe 5 by default) minutes with some kind of polling or something? This is just a straight lift-and-shift from the cloud server, but likely would be a useful addition for both?

crates/jazz-tools/src/middleware/auth.rs
joeinnes: Good idea.

PR #342, opened 2026-03-16
.never() permission

Takeno:
**Summary**

Adds a `.never()` helper to the TypeScript permissions DSL for explicitly denying operations.

**What changed**

- Added `.never()` for read, insert, update, and delete permission builders
- Made it equivalent to `.where(anyOf([]))`, compiling to a `FALSE` policy
- Updated the permissions docs with TypeScript and SQL examples
- Added behavior and type-inference tests for the new API

### DIFF
--- a/docs/content/docs/permissions.mdx
+++ b/docs/content/docs/permissions.mdx
@@ -29,7 +29,7 @@ Define policies in `schema/permissions.ts` for TypeScript DSL apps, or directly 
   </Tab>
 </Tabs>
 
-Apps that do not need user-scoped filtering can still define explicit allow-all policies (`policy.todos.allowRead.where({})` / `USING (TRUE)`).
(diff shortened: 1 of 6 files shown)


joeinnes:
I like this. How would you feel about a counterpart `always`?

Rationale: 
With the current syntax, as you mention,`.where(anyOf([]))`, compiles to FALSE, **but** `.where({})` evaluates to `TRUE`. From a pure logic level, it does make sense, but it feels like an opportunity for folks to make very annoying mistakes with their permissions set-up. I'd rather just completely disallow empty objects/arrays and force folks to use these helpers.

What do you think?

Takeno:
It would be an explicit way of what the already-default behavior is. Dummy helper, but easier to read.
I do it

gdorsi (approved):

PR #343, opened 2026-03-16
feat: add TestingServer rust module

gdorsi:
This PR adds a ServerBuilder and TestingServer modules to make it eaiser to spin up a server programmatically and run a TestingServer with "fake stuff" around to make it work.

Thinking about adopting the builder pattern for TestingServer as well, to make it easier to reproduce broken scenarios like unrechable JwksServer.


antoniomuso (approved):
LGTM!

PR #345, opened 2026-03-16
feat: split loaded reference fields from FKs

nrainhart:
## Description

_Note for reviewers: this PR is quite large because it renames fields in a lot of example apps. I recommend reviewing commits one by one._

Before this PR, we used the same name for a row's foreign key (i.e. the string field referencing another table) and the resolved reference itself. When using `include` we'd return just the resolved reference, and when not we'd just return the id.

This meant that rows with loaded references were not a subtype of plain rows (because they were missing the id field).

This PR splits both fields, so that we always return the row's id (unless explicitly filtered out by a projection) and only return the loaded reference if using `include`. We use the following convention:
- all scalar ref columns have a `-Id` or `-_id` suffix
- all array ref columns have a `-Ids` or `-_ids` suffix
- resolved reference fields are autogenerated by stripping that suffix. 

This convention is enforced both at the TS type level and at runtime.

Now that there is not risk of collision between FKs and loaded references, I could simplify the type of rows with includes (i.e. I reverted some of the changes introduced in https://github.com/garden-co/jazz2/pull/291).


packages/jazz-tools/src/dsl.ts
```diff
@@ -387,7 +417,28 @@ export const col = {
...
+    : T[K] extends ArrayBuilder<RefBuilder>
+      ? K extends RefArrayColumnKey
+        ? T[K]
+        : ArrayIdColumnError<K>
+      : T[K];
+};
+
+export function table<const T extends Record<string, ColumnBuilder>>(
+  name: string,
+  columns: EnforceReferenceColumnNames<T>,
```
nrainhart, line 440: I think this approach yields the best error feedback:

https://github.com/user-attachments/assets/ec0c2104-d92d-4d09-9dd5-8a297b98977b

gdorsi (approved):
LGTM

PR #347, opened 2026-03-17
Add session.where support for session-only permission predicates

Takeno:
## Summary
- add `session.where(...)` to the TypeScript permissions DSL for literal-only session claim checks
- introduce additive `Session*` policy AST variants and carry them through schema cloning, serialization, hashing, and SQL generation
- extend Rust SQL parsing and runtime evaluation for session-left comparisons, `IN`, `CONTAINS`, and `IS [NOT] NULL`
- preserve existing row-vs-session predicates and support both `session.user_id` and `session.userId`
- add docs and example snippets showing JWT claim checks in permissions

## Testing
- `pnpm --filter jazz-tools exec vitest run src/permissions/index.test.ts src/permissions/type-inference.test.ts src/sql-gen.test.ts src/codegen/codegen.test.ts`
- `cargo test -p jazz-tools`

### DIFF
--- a/crates/jazz-tools/src/query_manager/policy.rs
+++ b/crates/jazz-tools/src/query_manager/policy.rs
@@ -87,17 +87,36 @@ pub enum PolicyExpr {
         value: PolicyValue,
     },
 
+    /// Compare a session value against a literal value.
+    SessionCmp {
+        path: Vec<String>,
+        op: CmpOp,
+        value: Value,
+    },
+
     /// Check if a column is NULL.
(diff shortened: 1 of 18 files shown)


gdorsi (approved):
LGTM

The only improvement point I can suggest is that this policy is going to be evaluated for each row, but we could in theory compute it once per query.

Doing so now would add too much complexity to this PR, which would be incredibly useful even if not "fast".

How about we track this in a TODO/task?

PR #349, opened 2026-03-17
fix(update): allow unsetting optional fields

nrainhart:
## Description

Although this is possible at the Rust level, the TS types prevented unsetting an optional field.

### DIFF
--- a/packages/jazz-tools/src/runtime/db.ts
+++ b/packages/jazz-tools/src/runtime/db.ts
@@ -222,6 +222,14 @@ export interface TableProxy<T, Init> {
   readonly _initType: Init;
 }
 
+type UpdateValue<RowValue, InitValue> = undefined extends RowValue
+  ? InitValue | null
+  : InitValue | undefined;
+
+type UpdateData<T, Init> = {
+  [K in keyof Init]?: K extends keyof T ? UpdateValue<T[K], Init[K]> : Init[K] | undefined;
+};
+
 interface BroadcastChannelLike {
   postMessage(data: unknown): void;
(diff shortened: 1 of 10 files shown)


packages/jazz-tools/src/runtime/db.ts
```diff
@@ -222,6 +222,14 @@ export interface TableProxy<T, Init> {
   readonly _initType: Init;
 }
 
+type UpdateValue<RowValue, InitValue> = undefined extends RowValue
+  ? InitValue | null
+  : InitValue | undefined;
+
+type UpdateData<T, Init> = {
+  [K in keyof Init]?: K extends keyof T ? UpdateValue<T[K], Init[K]> : Init[K] | undefined;
+};
```
gdorsi, line 231: Could you help me understand it a bit better?

I'm not fully clear on what it means for for RowValue to be undefined, and the difference between undefined and null in data.

nrainhart: Rows can only have set or undefined values (never null). So, a RowValue that can be undefined is an optional row value.

For update paylods, undefined means "do not update this field", while null means "unset this field".

It is a bit confusing, because it looks like "updating a field to null sets it to undefined".

nrainhart: It could be less confusing if we always returned `Value | null` for rows (i.e. never undefined), and it'd also align more closely with SQL nullability.

@aeplay wdyt? Did you have any reason to go with undefined in the first place?

aeplay: no good reason, we can always do null

PR #353, opened 2026-03-18
fix(sync_manager): merge ephemeral payload claims into server session

joeinnes:
## Summary

- When a client subscribes with per-subscription claims (e.g. `{ claims: { join_code: "abc" } }`), those claims are now merged into the server-established session rather than being discarded.
- The `user_id` from auth remains authoritative and cannot be overridden — only non-identity claims are merged from the payload.
- This enables invite flows where a join code is presented as a session claim for policy evaluation (e.g. `joinCode = @session.claims.join_code` in a SELECT policy).

## Context

Commit 35e32128 correctly prevented payload sessions from overriding the server-established `user_id`, but it also discarded ephemeral per-subscription claims. This broke invite flows in the chat-react example where `@session.claims.join_code` needs to evaluate against a client-supplied value.

## Test plan

- [ ] Existing sync_manager tests pass
- [ ] chat-react invite flow: User A creates private chat, generates invite link; User B follows link and gains access via join code claim


gdorsi (approved):
LGTM

gdorsi (approved):

PR #354, opened 2026-03-18
Add policy and subscribe-all integration tests and test helper cleanup

gdorsi:
This PR now contains the testing improvements from the original branch after the insert-policy fix landed separately in #329.

Included here:
- policy and subscribe-all integration tests
- TestingServer builder cleanup to simplify test setup
- client connection helper renames/cleanup used by the tests

Removed from this PR:
- the narrower client-originated insert policy fix commit, since #329 already landed the broader fix

This keeps the added coverage and test ergonomics without reintroducing the overlapping fix.

### DIFF
--- a/crates/jazz-tools/tests/policies_integration.rs
+++ b/crates/jazz-tools/tests/policies_integration.rs
@@ -87,7 +87,10 @@ async fn create_document(client: &JazzClient, owner_id: &str, title: &str) -> Ob
 #[tokio::test]
 async fn select_policies_filter_subscription_results_per_client_session() {
     let schema = select_policy_schema();
-    let server = TestingServer::start_with_schema(schema.clone()).await;
+    let server = TestingServer::builder()
+        .with_schema(schema.clone())
+        .start()
+        .await;
     let alice =
         connect_jwt_client(&server, schema.clone(), "alice", "documents", READY_TIMEOUT).await;
     let bob = connect_jwt_client(&server, schema, "bob", "documents", READY_TIMEOUT).await;
@@ -171,7 +174,10 @@ async fn select_policies_filter_subscription_results_per_client_session() {
 #[tokio::test]
 async fn insert_policies_are_enforced_by_server_for_client_sync() {
     let schema = write_policy_schema();
-    let server = TestingServer::start_with_schema(schema.clone()).await;
+    let server = TestingServer::builder()
+        .with_schema(schema.clone())
+        .start()
+        .await;
     let intruder = connect_jwt_client(
         &server,
         schema.clone(),
@@ -262,7 +268,10 @@ async fn insert_policies_are_enforced_by_server_for_client_sync() {
 #[tokio::test]
 async fn update_and_delete_policies_block_unauthorized_server_mutations() {
     let schema = write_policy_schema();
-    let server = TestingServer::start_with_schema(schema.clone()).await;
+    let server = TestingServer::builder()
+        .with_schema(schema.clone())
+        .start()
+        .await;
     let alice =
         connect_jwt_client(&server, schema.clone(), "alice", "documents", READY_TIMEOUT).await;
     let bob = connect_jwt_client(&server, schema.clone(), "bob", "documents", READY_TIMEOUT).await;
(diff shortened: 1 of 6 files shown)


crates/jazz-tools/tests/support/mod.rs
```diff
@@ -1,14 +1,27 @@
...
-use jazz_tools::{DurabilityTier, JazzClient, ObjectId, Query, Value};
+use jazz_tools::server::TestingServer;
+use jazz_tools::{
+    AppContext, DurabilityTier, JazzClient, ObjectId, OrderedRowDelta, Query, QueryBuilder, Schema,
+    SubscriptionStream, Value,
+};
 
 const DEFAULT_POLL_INTERVAL: Duration = Duration::from_millis(250);
 const DEFAULT_QUERY_TIMEOUT: Duration = Duration::from_secs(8);
+#[allow(dead_code)]
```
nrainhart, line 12: Just curious: why is this needed? I see these constants are being used

crates/jazz-tools/tests/policies_integration.rs
```diff
@@ -0,0 +1,350 @@
...
+    let rows = wait_for_rows(
+        &observer,
+        query,
+        "observer sees only accepted row",
+        |rows| (rows.len() == 1 && rows[0].0 == accepted_doc).then_some(rows),
+    )
+    .await;
+    assert_eq!(rows[0].1, document_values("alice", "allowed"));
+
+    intruder.shutdown().await.expect("shutdown intruder");
```
nrainhart, line 238: What happens with the intruder's local db state in this case? Is the invalid insert preserved? is it rolled back? Should we include this in this test?

crates/jazz-tools/tests/policies_integration.rs
```diff
@@ -0,0 +1,350 @@
...
+            .any(|(id, values)| *id == doc_id && *values == document_values("alice", "original")),
+        "unauthorized update should not be persisted at EdgeServer: rows={rows_after_update:?}"
+    );
+    collect_stream_deltas(&mut observer_stream, &mut observer_log, NO_DELTA_WINDOW).await;
+    assert!(
+        !has_updated(&observer_log, doc_id),
+        "unauthorized update should not be broadcast to subscribers: log={observer_log:?}"
+    );
+
+    bob.delete(doc_id).await.expect("optimistic local delete");
```
nrainhart, line 327: Nit: i'd split upload & delete rejections into 2 separate tests

nrainhart (approved):
Love this!! 🚀

crates/jazz-tools/tests/policies_integration.rs
gdorsi: From our last sync with @anselm my understanding is that this is still uncovered, and we don't yet bubble up the error to the client.

From the phone now so I can't confirm this.

My plan was to focus on this problem separately, but I should at least add a TODO.

Thanks for pointing this out.

Takeno:
Can the fix be related to #329?

antoniomuso (approved):
LGTM!

gdorsi:
> Can the fix be related to #329? 

Ah, yes. Looks like the same fix.

I guess this validates your PR.

Approved, going to revise after the merge

crates/jazz-tools/tests/support/mod.rs
gdorsi: Each file in crates/jazz-tools/tests/*.rs is compiled as its own separate test crate, and each one pulls this file in with mod support;. 

So a helper can be used in one test crate but still look “unused” in another, and we need this dead code to turn off the linter complaints.

crates/jazz-tools/tests/policies_integration.rs
gdorsi: Checked, we have errors but there is no rollback.

Going to create a discussion on Discord

aeplay:
I rebased this PR around #329 so we keep the additive parts and drop the overlapping fix commit.

Kept:
- `8cc5f261` `test: add integration tests for policies and subscribe all`
- `bd6e6d61` `refactor: adopt the builder pattern in TestingServer to simplify test setup`
- `700ec526` `refactor: rename client connection functions for clarity and consistency`

Removed:
- `b54bb69c` `fix: apply policy correctly on client-originated inserts`

That fix is now covered by #329, so this PR is just the tests and test-helper cleanup.

PR #366, opened 2026-03-19
test: more policy integration tests + fix UUID string serialization in policy

gdorsi:


crates/jazz-tools/src/query_manager/policy.rs
```diff
@@ -1466,7 +1466,7 @@ pub fn evaluate_in(
         Value::Integer(i) => session_array.iter().any(|v| v.as_i64() == Some(*i as i64)),
         Value::BigInt(i) => session_array.iter().any(|v| v.as_i64() == Some(*i)),
         Value::Uuid(id) => {
-            let id_str = format!("{:?}", id);
```
gdorsi, line 1469: This line was making `in_session_array_policy_gates_visibility_by_membership` fail because the claims UUID matching was going agains a badly serialized id.

Changing to id.to_string() uses the Display impl, which outputs the standard hyphenated UUID format matching how ObjectId serializes to JSON.

Takeno (approved):

PR #367, opened 2026-03-19
Auth Examples

Takeno:
## Summary

Add three auth-focused Jazz chat examples that show how to authenticate Jazz with external JWT providers across common setups: a custom local auth server, Better Auth, and WorkOS AuthKit. This also updates the policy test helper so callers can inject their own `expect` implementation instead of depending directly on Vitest.

## What Changed

- Add `examples/auth-simple-chat`, a React/Vite example backed by a local Express auth server that issues ES256 JWTs via JWKS, falls back to anonymous local auth, and enforces role-based chat permissions.
- Add `examples/auth-betterauth-chat`, showing Better Auth session handling, JWT exchange, JWKS-backed sync server validation, and the same role-based chat flow.
- Add `examples/auth-workos-chat`, showing WorkOS AuthKit integration with hosted JWKS and passing WorkOS access tokens directly to `JazzProvider`.
- Update `packages/jazz-tools/src/testing/policy-test-app.ts` to accept an injected `expect` function, making the helper usable with external test runners instead of hard-wiring Vitest.


gdorsi (approved):
LGTM

PR #369, opened 2026-03-19
Add more integration tests, default to MemoryStorage for TestingClient and TestingServer

gdorsi:
Started getting the following error:
- `Storage("IoError(\"fjall open: FjallError: Io(Os { code: 24, kind: Uncategorized, message: \\\"Too many open files\\\" })\")")`

As fix I've changed the default for integration tests to not use the persistent storage but go for the MemoryStorage.

Added a bunch of more integration tests, including a failing one related to the server not propagating correctly "latest write wins" during a burst of updates.

### DIFF
--- a/crates/jazz-tools/tests/subscribe_all_integration.rs
+++ b/crates/jazz-tools/tests/subscribe_all_integration.rs
@@ -364,20 +364,40 @@ async fn subscribe_all_emits_add_update_remove_and_tracks_current_results() {
 /// same final title.
 ///
 /// TODO:
-/// This is currently ignored because the client path still has two burst-load
-/// bugs in [client.rs](/Users/guidodorsi/workspace/jazz2/crates/jazz-tools/src/client.rs):
-/// server-bound object updates are sent via one detached task per payload, so
-/// the runtime's ordered outbox can arrive at the server out of order; and the
-/// subscription callback forwards deltas with `try_send` into a bounded channel,
-/// so the terminal row-bearing delta can be dropped when the subscriber lags.
+/// Keep this test ignored until the client preserves "latest write wins" during
+/// a burst of updates. The assertion is fine; the client path in src/client.rs
+/// still has two burst-load bugs:
+///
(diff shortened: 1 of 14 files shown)


crates/jazz-tools/tests/subscribe_all_integration.rs
```diff
@@ -321,6 +354,147 @@ async fn subscribe_all_emits_add_update_remove_and_tracks_current_results() {
...
+/// [client.rs](/Users/guidodorsi/workspace/jazz2/crates/jazz-tools/src/client.rs)
+/// still has two burst-load bugs:
+///
+/// 1. Outgoing updates are sent in separate spawned tasks, so a fast sequence of
+///    writes can reach the server in a different order than the writer produced
+///    them.
+/// 2. Incoming subscription updates are pushed with `try_send` into a fixed-size
+///    channel, so a slow subscriber can silently drop updates, including the
+///    final one that should carry the last title.
+///
```
gdorsi, line 377: The analysis from Claudio:

Both bugs in the TODO are accurate. Here's what the code confirms:

  Bug 1 — outgoing send ordering (lines ~160-178 of client.rs):

  tokio::spawn(async move {
      if let Err(e) = conn.push_sync(payload, cid).await { ... }
  });

  Each of the 500 updates spawns an independent task that makes its own HTTP request to /sync. There's no sequencing guarantee between them — they race to the server.

  Bug 2 — channel overflow (lines 314-325 of client.rs):

  let (tx, rx) = mpsc::channel::<OrderedRowDelta>(64);
  // ...
  move |delta| {
      let _ = tx.try_send(delta.ordered_delta);  // silently discarded if full
  }

  The subscription channel capacity is 64. With 500 server-side broadcast updates arriving, try_send will silently drop updates once the consumer (the test's wait_for_subscription_update
  loop) falls behind. The final delta carrying title-500 can be among the dropped ones.

nrainhart (approved):
I really like these integration tests, they are super easy to read

PR #370, opened 2026-03-19
feat: improved compute tuple delta performance

antoniomuso:
`computer_tuple_delta` requires a lot of hashing. The main costs are:

1. `HashMap<Vec<ObjectId>, usize>` - hashes the entire Vec on insert and on every lookup
2. `Vec<ObjectId>` allocations via t.ids() for every tuple

This fix introduces:

1. Hash once — `HashedTupleRef::new()` computes the hash upfront. Every subsequent HashMap insert/lookup just writes the cached u64, never re-hashes the slice.
2. No `Vec<ObjectId>` allocations via `t.ids()` for every tuple; now we return an iterator instead.

## Profiling

**Before**
Screenshot 2026-03-19 alle 13 24 14

**After**

Screenshot 2026-03-20 alle 13 06 10

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph_nodes/tuple_delta.rs
+++ b/crates/jazz-tools/src/query_manager/graph_nodes/tuple_delta.rs
@@ -1,36 +1,39 @@
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 
-use crate::object::ObjectId;
 use crate::query_manager::types::{Tuple, TupleDelta, TupleElement};
 
-/// Pre-hashed ID list: hash is computed once at construction, equality falls back to full comparison.
-#[derive(Clone)]
-struct HashedIds {
-    ids: Vec<ObjectId>,
+/// Borrowed tuple identity key with a precomputed hash.
+///
+/// We keep the "hash once, compare full IDs on collision" behavior from the old
+/// implementation, but we borrow the tuple instead of allocating a temporary
+/// `Vec<ObjectId>` just to key the maps in this function.
+#[derive(Clone, Copy)]
+struct HashedTupleRef<'a> {
+    tuple: &'a Tuple,
     hash: u64,
 }
 
-impl HashedIds {
-    fn new(ids: Vec<ObjectId>) -> Self {
+impl<'a> HashedTupleRef<'a> {
+    fn new(tuple: &'a Tuple) -> Self {
         let mut hasher = std::hash::DefaultHasher::new();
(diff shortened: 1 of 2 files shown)


gdorsi (approved):
Nice job!

crates/jazz-tools/src/query_manager/graph_nodes/tuple_delta.rs
```diff
@@ -1,40 +1,77 @@
...
+/// `Vec<ObjectId>` just to key the maps in this function.
+#[derive(Clone, Copy)]
+struct HashedTupleRef<'a> {
+    tuple: &'a Tuple,
+    hash: u64,
+}
+
+impl<'a> HashedTupleRef<'a> {
+    fn new(tuple: &'a Tuple) -> Self {
+        let mut hasher = std::hash::DefaultHasher::new();
```
nrainhart, line 19: Would this be faster if we use AHash/FxHash?

crates/jazz-tools/src/query_manager/graph_nodes/tuple_delta.rs
```diff
@@ -1,40 +1,77 @@
 use std::collections::HashMap;
+use std::hash::{Hash, Hasher};
 
 use crate::query_manager::types::{Tuple, TupleDelta, TupleElement};
 
+/// Borrowed tuple identity key with a precomputed hash.
+///
+/// We keep the "hash once, compare full IDs on collision" behavior from the old
+/// implementation, but we borrow the tuple instead of allocating a temporary
+/// `Vec<ObjectId>` just to key the maps in this function.
```
nrainhart, line 10:
```suggestion
```

nrainhart (approved):

crates/jazz-tools/src/query_manager/graph_nodes/tuple_delta.rs
gdorsi: I was also about to ask, but from the profiles it looks like this is fast enough to not require a "faster" hashing algorithm.

Correct me if I'm wrong @antoniomuso

antoniomuso: This is the result with FXHash:
Screenshot 2026-03-20 alle 16 57 32

antoniomuso: My benchmarks don't show a huge speedup (90 op/s without) and (92 op/s with).

PR #371, opened 2026-03-20
feat: rn stress test

antoniomuso:
RN stress test example app.


stress-tests/stress-test-expo/src/StressTest.tsx
```diff
@@ -0,0 +1,427 @@
...
+  const [isRunning, setIsRunning] = useState(false);
+  const [logs, setLogs] = useState<LogEntry[]>([]);
+  const [progress, setProgress] = useState({ inserted: 0, total: 0 });
+  const abortRef = useRef(false);
+
+  const log = useCallback((message: string) => {
+    setLogs((prev) => [...prev, { timestamp: Date.now(), message }]);
+  }, []);
+
+  const generateTodos = useCallback(async () => {
```
gdorsi, line 135: It would be nice to have a fixture generation step without any active subscription.

This would let us to test load performance with high numbers more easily, because generating 50k rows would take seconds instead of minutes.

gdorsi (approved):
LGTM

stress-tests/stress-test-expo/src/StressTest.tsx
antoniomuso: There is a button to activate or disactivate subscription.

PR #373, opened 2026-03-20
fix: update & delete do not copy existing row from old schema

nrainhart:
## Description

This PR started off as just adding RuntimeCore integration tests that cover reads and writes on evolving schemas. This uncovered a few bugs:
- updates & deletes where only looking for the current schema, instead of checking branches for potentially old schemas
- `ObjectManager::get_or_load` would cache objects by id and always return the originally-requested branches

This PR fixes both errors. It also removes a CopyOnWriteWriter util that wasn't being used anywhere in our codebase.


crates/jazz-tools/src/query_manager/manager_tests.rs
```diff
@@ -4558,23 +4558,6 @@ fn file_storage_schema() -> Schema {
     schema
 }
 
-fn scalar_fk_schema() -> Schema {
```
nrainhart: Unused

crates/jazz-tools/src/schema_manager/writer.rs
nrainhart, line 1: Unused

crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -11,7 +11,212 @@ use super::policy::{ComplexClause, Operation, evaluate_simple_parts};
...
+                    match &best {
+                        None => {
+                            best = Some((
+                                commit.timestamp,
+                                branch_name.clone(),
+                                commit.content.clone(),
+                                tip_id,
+                            ));
+                        }
+                        Some((best_ts, _, _, _)) if commit.timestamp > *best_ts => {
```
nrainhart: I don't like that we're duplicating what we're already doing for queries, but didn't want to make this PR even larger. Happy to rework this (either in this PR or as a fup)

crates/jazz-tools/src/runtime_core/tests.rs
```diff
@@ -1810,6 +1845,110 @@ fn test_sync_edit_fires_callback_synchronously() {
...
+    assert_eq!(queried_values[name_idx], Value::Text("Alice".to_string()));
+    assert_eq!(
+        queried_values[email_idx],
+        Value::Text(String::new()),
+        "New required column should be backfilled with the lens default",
+    );
+}
+
+#[test]
+fn rc_update_old_schema_row_after_evolution_copies_row_to_current_schema() {
```
nrainhart, line 2017: I suspect deletes are not working correctly either, I'll check that as a FUP

crates/jazz-tools/src/query_manager/writes.rs
nrainhart: Attempt at reducing duplication: [`fe02594` (this PR)](https://github.com/garden-co/jazz2/pull/373/commits/fe025946f71e5936d5ef3c78751aa12b41ab048e)

crates/jazz-tools/src/runtime_core/tests.rs
nrainhart: Fixed in [`b89c87f` (this PR)](https://github.com/garden-co/jazz2/pull/373/commits/b89c87fca2da05238e659c0ceace35255ce1e323)

crates/jazz-tools/src/query_manager/server_queries.rs
```diff
@@ -117,18 +124,83 @@ impl QueryManager {
...
+                    Some((best_ts, _, _, _, _)) if commit.timestamp > *best_ts => {
+                        best = Some((
+                            commit.timestamp,
+                            commit.content.clone(),
+                            tip_id,
+                            branch_name,
+                            is_soft_deleted,
+                        ));
+                    }
+                    _ => {}
```
gdorsi, line 171: We have recently introduced the commit_id as additional tie-breaker for conflict resolution to handle cases when the timestamp is the same: https://github.com/garden-co/jazz2/pull/389/changes#diff-aa173df57b443ceb36a743b212bf10d1e2e2c238c7e1e651faf7cc3e112fc475R749

Shall we do the same here?

nrainhart: Done after fixing conflicts!

gdorsi (approved):

PR #374, opened 2026-03-20
feat: expose TestingServer via NAPI

gdorsi:
Adds a `TestingServer` NAPI binding so Node.js consumers (and Vitest browser tests) can start/stop a real Rust-backed test server without hardcoded ports or secrets

I've also ported all the testing code to use NAPI instead of the CLI


packages/jazz-tools/src/runtime/napi.fjall.test-helpers.ts
```diff
@@ -2,11 +2,12 @@ import { randomUUID } from "node:crypto";
...
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, beforeAll } from "vitest";
+import { NapiRuntime } from "jazz-napi";
+import { afterEach } from "vitest";
+import { serializeRuntimeSchema } from "../drivers/schema-wire.js";
 import type { Value, WasmSchema } from "../drivers/types.js";
 import { JazzClient, type Row } from "./client.js";
 import type { QueryBuilder } from "./db.js";
-import { createPersistentNapiRuntime, loadNapiModule } from "./testing/napi-runtime-test-utils.js";
```
gdorsi: Removed an overenginerred abstraction layer to consume NAPI modules in tests.

nrainhart (approved):

PR #375, opened 2026-03-20
chore: upgrade Vite (and related deps) to 8.0

nrainhart:
## Description

Upgrades all packages to use Vite 8 (and related Vite upgraded deps). This release replaces esbuild & rollup with rolldown, which should give us faster build times.

### DIFF
--- a/packages/jazz-tools/package.json
+++ b/packages/jazz-tools/package.json
@@ -85,7 +85,7 @@
   },
   "devDependencies": {
     "@sveltejs/package": "^2.0.0",
-    "@sveltejs/vite-plugin-svelte": "^5.0.0",
+    "@sveltejs/vite-plugin-svelte": "^7.0.0",
     "@types/node": "^22.0.0",
     "@types/react": "^19.0.0",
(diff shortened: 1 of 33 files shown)


packages/jazz-tools/src/react-native/create-jazz-client.test.ts
nrainhart, line 1: Needed to change how we mock imports after upgrading to vitest 4.1

packages/jazz-tools/package.json
```diff
@@ -96,7 +96,7 @@
     "svelte": "^5.0.0",
     "typescript": "catalog:default",
     "vite-plugin-top-level-await": "^1.6.0",
```
Takeno: TIL: Does it make sense to put it in the catalog as well? Since it might be strictly related to the vite version.

gdorsi (approved):

packages/jazz-tools/package.json
nrainhart: Yup, done!

PR #382, opened 2026-03-23
feat: add a project-management skill and organizes the things in todo  for MVP

gdorsi:
Manually gone through all the MVP tasks, cleaned them up and organized in a way that we know what to pick next.

Added a Project Management skill based on the flow we used for Jazz1, organizing the tasks into:
- issues, self-contained problems that can be handled without any further spec
- ideas, things we know we should solve but scope and approach is not clear yet
- projects, fully shaped ideas. They should not contain too many technical details but should help on defining a high-level solution to the problem and split it in multiple sub-problems which can handled in separate PRs

To validate the shaping skill I've added the Ordered Index Top-K Query Path project.

The shaping flow is not intended to produce an implementation plan, or to go deep technically, but to explore a problem and split it in different scopes that are intended to be shipped in standalone PRs. 

This should help with the problem of taking on big challenges, such as improving the query graph to effectively use the indexes.

The skill automatically updates a [TODO.md](https://github.com/garden-co/jazz2/blob/feat/pm/TODO.md) file that should be useful to navigate through the backlog.

Left launch and later tasks there because by not being a priority they probably don't need to be cleaned up yet.


TODO.md
```diff
@@ -0,0 +1,35 @@
...
+- [**complex-merge-strategies**](todo/ideas/1_mvp/complex-merge-strategies.md) — Per-column/per-table merge strategies beyond LWW (counters, sets, rich text, custom logic).
+- [**explicit-indices**](todo/ideas/1_mvp/explicit-indices.md) — Developer-declared indices in the schema language, replacing auto-index-all-columns.
+- [**optimistic-update-dx**](todo/ideas/1_mvp/optimistic-update-dx.md) — Developer-facing API for mutation settlement state — show pending/confirmed/rejected status on rows and filter queries by settlement tier.
+- [**ordered-index-topk-query-path**](todo/ideas/1_mvp/ordered-index-topk-query-path.md) — Make paginated ordered queries fast by scanning in order and stopping early, instead of loading everything then sorting.
+- [**self-referential-inherits**](todo/ideas/1_mvp/self-referential-inherits.md) — Support bounded self-referential INHERITS relationships (e.g., a folder inheriting permissions from its parent folder).
+- [**storage-limits-and-eviction**](todo/ideas/1_mvp/storage-limits-and-eviction.md) — Bounded storage with LRU eviction of cold data on clients and edge servers, with lazy re-fetch from upstream.
```
gdorsi, line 31: @aeplay 

Is this really something we should handle in the MVP?

PR #383, opened 2026-03-23
feat: default values

nrainhart:
## Description

Allow declaring default values as part of the schema definition, and use them for:
- Insert APIs when a caller omits a field.
	- Omitted field on insert: use schema default if present.
	- `undefined` on insert: treated the same as omitted.
	- Explicit `null`: only allowed for nullable columns; it does not trigger the default.
- Auto-generated lens defaults when a column is added or removed across schema versions.

Today, defaults exist only in the migration DSL (`col.add(... { default })` / `col.drop(... { backwardsDefault })`). Base table schemas do not carry defaults, `CREATE TABLE ... DEFAULT ...` is parsed but discarded, and the current insert flow materializes missing TS fields as `Null` before Rust can distinguish omission from an explicit null.

## Design decisions

We only support default _values_, and not default _expressions_ that are evaluated at insert time (as opposed to Postgres, [that does](https://www.postgresql.org/docs/current/ddl-default.html#:~:text=The%20default%20value%20can%20be%20an%20expression)). This prevents common patterns like adding a `now()` default value for timestamp columns. I did this to keep the scope manageable (otherwise we'd need to add expression parsing and evaluation, and figure out how to provide features like "get current date", since we don't have functions/stored procedures). I don't think this is a huge limitation, as defaults are mostly a convenience and users can always provide values like current timestamp at insert time, but lmk if you differ.


crates/jazz-tools/src/schema_manager/encoding.rs
```diff
@@ -17,9 +17,46 @@ use crate::query_manager::types::{
...
+enum SchemaEncodingVersion {
+    // v1 schemas did not encode policies.
+    V1 = 1,
+    // v2 schemas include policies, but no legacy inherit-policy byte.
+    V2 = 2,
+    // v3 schemas include policies and a legacy inherit-policy byte.
+    V3 = 3,
+    // v4 schemas include column defaults.
+    V4 = 4,
+}
```
gdorsi, line 34: Does it make sense to support all these versions?

crates/jazz-tools/src/schema_manager/sql.rs
gdorsi: Flagged by the AI review, but does this lead to data loss?

gdorsi: This should be `id.to_string()`

crates/jazz-tools/src/schema_manager/manager.rs
```diff
@@ -224,21 +224,51 @@ impl SchemaManager {
...
+                        "cannot set required field `{}` to null",
+                        column.name
+                    )));
+                }
+                aligned_values.push(value);
+                continue;
+            }
+
+            if let Some(default) = &column.default {
+                aligned_values.push(default.clone());
```
gdorsi, line 260: Which values can be assigned to `default`?

Is it validated somewhere before reaching here?

crates/jazz-tools/src/schema_manager/encoding.rs
nrainhart: I guess we could drop old versions, but that'll break schemas that are already stored in the DB

crates/jazz-tools/src/schema_manager/sql.rs
nrainhart: Fixed in [`3b268e7` (this PR)](https://github.com/garden-co/jazz2/pull/383/commits/3b268e7b946fa0a7daf536fe1678e7da251a2cf3)

nrainhart: Yup. Fixed in [`e9ea70a` (this PR)](https://github.com/garden-co/jazz2/pull/383/commits/e9ea70a80bdef971fd9b5ee760cb1cc89e56bff4)

crates/jazz-tools/src/schema_manager/manager.rs
nrainhart: What are you concerned about exactly?

If you're talking about type validation, we're doing it during encoding (see [value_matches_column_type](https://github.com/garden-co/jazz2/blob/feat/default-values/crates/jazz-tools/src/query_manager/encoding.rs#L171)), both for values provided at insert time and for default values

gdorsi (approved):

PR #385, opened 2026-03-23
Feature/history conflict tests

antoniomuso:
## Summary

- Add 15 integration tests covering history (commit DAG) and conflict management (LWW resolution) across three test levels: Rust ObjectManager, Rust E2E, and browser
- Extract shared browser test utilities into `tests/browser/support.ts`, mirroring Rust's `tests/support/mod.rs` pattern
- Refactor `worker-bridge.test.ts` to use shared utilities, removing ~170 lines of duplicated helpers
- Two failing tests - check the last section in the description

## Test plan

**Rust ObjectManager isolated tests** (5 new in `src/object_manager/tests.rs`):
- [x] `lww_selects_highest_timestamp_tip` — highest-timestamp commit wins
- [x] `lww_deterministic_on_equal_timestamps` — equal timestamps produce stable ordering
- [x] `receive_commit_idempotent_during_conflict` — replaying a commit doesn't alter frontier or notify subscribers
- [x] `truncate_with_diverged_tips` — truncation works with multiple diverged tips
- [x] `truncate_rejects_when_tip_not_descendant_of_tail` — safety invariant enforced

**Rust E2E tests** (6 new in `tests/history_conflict.rs`):
- [x] `concurrent_updates_resolve_to_lww_winner` — two clients converge after conflicting updates
- [x] `concurrent_creates_both_survive` — independent objects don't interfere
- [x] `rapid_concurrent_updates_converge` — stress test with 10 interleaved updates per client
- [x] `fresh_client_sees_lww_winner_after_conflict` — late joiner sees converged state
- [x] `subscription_reflects_concurrent_update` — subscription stream delivers remote changes
- [x] `sequential_updates_preserve_latest` — non-conflicting baseline
- [x] `online_user_wins_on_reconnect` - Offline user and Online user update the same object. 
- [ ] `offline_user_wins_on_reconnect` - Offline user and Online user update the same object. This test fails, offline object doesn't sync - ref https://github.com/garden-co/jazz2/blob/main/todo/ideas/1_mvp/sync-protocol-reliability.md

**Browser tests** (4 new in `tests/browser/history-conflict.test.ts`):
- [ ] `concurrent updates converge in browser` — full WASM + Worker + OPFS + sync pipeline - This test fails
- [x] `concurrent creates both visible in browser`
- [x] `subscription fires on remote concurrent update`
- [ ] `fresh db sees converged state` - This test fails

### Failing tests:
Temporarily skipping two browser-side tests. The tests are correct, but are currently failing due to an underlying LWW tie-break bug.

**Symptom**: Concurrent updates to the same row fail to converge on a single final value.

**Scope**: Normal sync propagation is fine. This only breaks on true conflicts where replicas produce competing tips for the same object.

**Root Cause**: Our Last-Writer-Wins (LWW) resolution is timestamp-based, but lacks a canonical tie-breaker for exact-timestamp collisions. When timestamps match, the winner defaults to the local tip iteration order, which differs between runtimes (e.g., Client A keeps alice-edit, Client B keeps bob-edit).

**Impact**: Browser conflict tests time out waiting for convergence, and fresh clients can observe an inconsistent state.

**Why it's happening in the browser**:
This is a core logic bug, not just a browser bug. However, the browser exposes it much more frequently because its `Date.now()` precision is much coarser than other environments, making equal-timestamp concurrent commits highly likely.


packages/jazz-tools/tests/browser/history-conflict.test.ts
```diff
@@ -0,0 +1,388 @@
...
+   *
+   *
+   * KNOWN BUG: the server does not relay concurrent commits between
+   * browser clients. Each client only ever sees its own update
+   * (alice=alice-edit, bob=bob-edit — verified via 40s polling).
+   * The server scope-based forwarding (forward_update_to_clients_except)
+   * appears to work in Rust E2E tests (in-process RuntimeCore) but not
+   * through the HTTP /sync + /events pipeline.
+   */
+  it.skip("concurrent updates converge in browser", async () => {
```
gdorsi: You can use `it.fails` for failing tests.

This way we still run them, but we expect them to fail.

packages/jazz-tools/tests/browser/history-conflict.test.ts
```diff
@@ -0,0 +1,388 @@
...
+      dbBob,
+      allTodos,
+      (rows) => rows.some((row) => row.id === id && row.title === "updated-by-alice"),
+      "Bob sees Alice's update",
+      20000,
+    );
+  }, 60000);
+
+  // -------------------------------------------------------------------------
+  // Test 13: concurrent_creates_both_visible_in_browser
```
gdorsi: nit: these comments aren't very useful

crates/jazz-tools/tests/history_conflict.rs
```diff
@@ -0,0 +1,664 @@
...
+            .expect("alice updates title");
+    });
+    let bob_handle = tokio::spawn(async move {
+        bob2.update(
+            todo_id,
+            vec![("title".to_string(), Value::Text("bob-edit".to_string()))],
+        )
+        .await
+        .expect("bob updates title");
+    });
```
gdorsi, line 102: Are these edits happening sequentially?

In this case we are not simulating a conflict, it would be possible that bob has got already the alice update.

crates/jazz-tools/tests/history_conflict.rs
```diff
@@ -0,0 +1,664 @@
...
+        .table(
+            TableSchema::builder("todos")
+                .column("title", ColumnType::Text)
+                .column("completed", ColumnType::Boolean),
+        )
+        .build()
+}
+
+/// Two clients update the same todo concurrently (no sync wait between writes).
+/// Both must eventually converge to the same final title.
```
gdorsi, line 28: I would also be helpful to have these two cases:
- Alice and Bob concurrently edit two different fields of the same item
- Alice adds 5 update commits, after that Bob (offline) adds 1. How does the conflict resolves?

crates/jazz-tools/tests/history_conflict.rs
antoniomuso: No, they are not sequential. They are spawned in parallel. The line after makes the join of it:
```rs
let (alice_res, bob_res) = tokio::join!(alice_handle, bob_handle);
    alice_res.expect("alice task panicked");
    bob_res.expect("bob task panicked");
```

packages/jazz-tools/tests/browser/history-conflict.test.ts
antoniomuso: I remove them

crates/jazz-tools/tests/history_conflict.rs
antoniomuso: I add them

packages/jazz-tools/tests/browser/history-conflict.test.ts
antoniomuso: Got it, there is another PR to solve them. I won't modify thme just to avoid conflicts

antoniomuso: UPDATE: Sometimes tests succeed, so I just skip them for now.

gdorsi (approved):

PR #386, opened 2026-03-24
Better-Auth database adapter

Takeno:
## Summary

This PR adds Better-Auth database adapter based on Jazz.

## What’s included

- adds a new `jazz-tools/better-auth-adapter` package entry so Better Auth can use Jazz as its database adapter
- adds schema generation helpers so Better Auth can output a Jazz `schema/current.ts` and flow into the normal Jazz build/migration/lens workflow
- adds comprehensive adapter tests covering core adapter behavior plus auth-related `user`, `session`, `account`, and `verification` flows
- adds a `better-auth-nextjs` example app showing a working Next.js setup with Jazz-backed Better Auth
- adds docs for usage, compatibility, and the schema generation / migration / `schema:push` workflow

## TODO
- [x] Rebase with other Auth examples
- [ ] Test multiple plugins directly with the full better-auth setup


packages/jazz-tools/src/better-auth-adapter/index.ts
```diff
@@ -0,0 +1,287 @@
...
+              limit: 1,
+            });
+
+            const existing = (await config
+              .db()
+              .all(checkQb, { tier: durabilityTier })) as JazzRowRecord[];
+            if (existing.length > 0) {
+              throw new Error(
+                `Unique constraint violated: "${table}.${storedFieldName}" already has a row with value "${String(value)}"`,
+              );
```
gdorsi: Looking at the `create` method, both the unique constraint checks and the `insertDurable` call use the adapter's configured `durabilityTier` (which defaults to `"edge"`)

Since Jazz also has a `"global"` tier (strongest consistency), should these operations always use `{ tier: "global" }` instead? The uniqueness read needs the most up-to-date view to avoid missing a concurrent insert for the same unique value, and the insert itself should presumably also be durably committed at the strongest tier.

Is there a reason to prefer the adapter's configured tier here — e.g. the caller explicitly trading off consistency for latency, or the backend context already guaranteeing global durability regardless of the passed tier?

examples/auth-betterauth-chat/scripts/sync-server.ts
```diff
@@ -0,0 +1,52 @@
...
+  backendSecret: BACKEND_SECRET,
+  jwksUrl: `${APP_ORIGIN}/api/auth/jwks`,
+});
+
+await pushSchemaCatalogue({
+  serverUrl: server.url,
+  appId: APP_ID,
+  adminSecret: server.adminSecret,
+  schemaDir: join(import.meta.dirname ?? __dirname, "../schema-better-auth"),
+});
```
gdorsi: Guess you are doing this to not bundle the BetterAuth schema inside the client bundle.

How about having the schema & permissions publishing take multiple schemas from a directory instead of having to do this manually here?

Could be a follow up fix, but worth considering.

examples/auth-betterauth-chat/schema-better-auth/schema.ts
```diff
@@ -0,0 +1,88 @@
+import { schema as s } from "jazz-tools";
+
+const schema = {
+  better_auth_user: s.table({
```
gdorsi, line 4: Can't we import the schema and permissions from the better auth package?

Why generating it here?

gdorsi: Ah, I guess to support different configurations and plugins.

packages/jazz-tools/src/better-auth-adapter/index.ts
```diff
@@ -0,0 +1,287 @@
...
+        async update({ model, where, update }): Promise<any> {
+          const [match] = await findAllRows(model, { where, limit: 1 });
+          if (!match) {
+            return null;
+          }
+
+          const table = getPrefixedModelName(model);
+          const qb = createQueryBuilder(table, wasmSchema);
+
+          await config.db().updateDurable(qb, match.id, update as any, { tier: durabilityTier });
```
gdorsi: Better Auth documents `user.changeEmail` as a supported flow, and by default the user’s stored email is updated once the new address is verified. 

That means `user.email` is not insert-only: it can change through adapter update paths. Here, `field.unique` is only enforced in `create`, so updating a unique column like `email` to a value that already exists can silently create duplicates. 

The same problem affects `updateMany` a few lines below. 

Should this path reuse the unique-field guard from `create`, excluding the current row id from the conflict check?

gdorsi (commented):
Inline adapter feedback.

packages/jazz-tools/src/better-auth-adapter/schema.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  const tableNames = Object.keys(tables).map((modelName) => getModelName(modelName));
+
+  const permissionLines: string[] = [];
+  for (const tableName of tableNames) {
+    if (permissionLines.length > 0) {
+      permissionLines.push("");
+    }
+    const key = formatObjectKey(tableName);
+    permissionLines.push(`  policy.${key}.allowRead.never();`);
```
gdorsi: Better Auth supports custom table names via `modelName`, so `tableName` is not guaranteed to be a valid JS identifier here. `formatObjectKey(...)` is fine for object literal keys, but in this emitted member access it can produce invalid syntax like `policy."user-sessions".allowRead.never()`. Should this generate bracket access instead, e.g. `policy[${formatStringLiteral(tableName)}].allowRead.never()`?

gdorsi (commented):
Inline schema-generation feedback.

packages/jazz-tools/src/better-auth-adapter/index.ts
Takeno: Maybe there is really no reason to let the user configure the tier, which can lead to errors and misunderstandings.

My only doubt: if I use Jazz as a local persistence (sqlite-like) with no remote server url, is the `global` tier working as expected?

examples/auth-betterauth-chat/scripts/sync-server.ts
Takeno: Yes, exactly. I don't want the BA schema published alongside the client schema. From the schema, attackers can guess installed plugins and/or versions. That's something I would never make public.

I agree we need the possibility to have multiple schemas/permissions from a directory. For online apps, we might suggest using two different app IDs to keep them isolated, too.

examples/auth-betterauth-chat/schema-better-auth/schema.ts
Takeno: Yes, exactly. The Better Auth CLI command generates the schema based on the installed plugins.

packages/jazz-tools/src/better-auth-adapter/schema.ts
Takeno: In the meantime, we've changed the default policy to be private-by-default. We can remove the permission files for BA completely, since it will always be used as a backend agent.

PR #387, opened 2026-03-24
feat(svelte): add rune-based test infrastructure

joeinnes:
## Summary
Preparation for deep reactivity work — replaces mock-based Svelte tests with `.svelte.test.ts` files that run through the Svelte compiler, enabling real `$state`/`$effect` verification.

- Dedicated `vitest.config.svelte.ts` (jsdom env, Svelte plugin), wired into `test` script
- Shared context mock in `test-helpers.svelte.ts`
- `use-all.test.ts` → `rune-patterns.svelte.test.ts` (tests callback patterns, not `useAll` itself)

## Test plan
- [ ] `pnpm --filter jazz-tools test` runs both vitest configs
- [ ] `pnpm --filter jazz-tools test:svelte` runs 16 rune tests in isolation


packages/jazz-tools/package.json
```diff
@@ -67,7 +67,8 @@
   "scripts": {
     "build": "tsc && cp src/worker/jazz-worker.ts dist/worker/jazz-worker.ts && pnpm build:test && pnpm build:svelte",
     "build:svelte": "svelte-package -i src/svelte -o dist/svelte --tsconfig tsconfig.svelte.json && node scripts/cleanup-generated-src-types.mjs",
-    "test": "vitest run --config vitest.config.ts",
+    "test": "vitest run --config vitest.config.ts && vitest run --config vitest.config.svelte.ts",
```
gdorsi, line 70: Shall we bring test:browser here as well?

gdorsi (approved):

packages/jazz-tools/package.json
joeinnes: We can, it adds Playwright and the rust build as dependencies to running the test suite. I think probably anyone that is doing dev work that needs to run these tests likely needs those dependencies anyway, so it's not an unreasonable expectation. What do you think?

PR #388, opened 2026-03-24
Add external JWKS URL support to TestingServer

Takeno:
## Summary
- add optional `jwksUrl` support to `jazz-napi` `TestingServer.start(...)`
- keep the embedded JWKS path as the default when no URL is provided
- disable built-in JWT helpers in external-JWKS mode and return a clear error instead
- add test coverage for default embedded JWKS behavior and external JWKS auth flow

## Testing
- `cargo test -p jazz-tools --features test-utils server::testing::tests --lib`
- `pnpm --dir crates/jazz-napi build:debug`
- `pnpm --dir packages/jazz-tools exec vitest run src/testing/index.test.ts`

### DIFF
--- a/crates/jazz-napi/index.d.ts
+++ b/crates/jazz-napi/index.d.ts
@@ -41,7 +41,7 @@ export declare class NapiRuntime {
 }
 
 export declare class TestingServer {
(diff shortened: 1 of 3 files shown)


gdorsi (approved):
LGTM!

PR #393, opened 2026-03-25
feat: Granular reactivity for Svelte & Vue

joeinnes:
## Summary
Adds a `reconcileArray` utility that diffs keyed arrays in-place, matching by `id`, deep-merging changed properties, preserving object identity, and wires it into Svelte and Vue `onDelta` callbacks so reactive proxy systems only trigger signals for properties that actually changed.

- `reconcileArray` utility with `deepMerge`, `valuesEqual`, `isKeyedArray` helpers
- Svelte `QuerySubscription.onDelta` reconciles incrementally instead of replacing the array
- Vue `useAll` `onDelta` does the same
- Unit tests for the utility (identity, reorder, nested arrays, Date/Uint8Array, key deletion)
- Svelte rune tests proving reconciliation works through `$state` proxy with `$effect` observation

## Test plan
- [ ] `pnpm --filter jazz-tools test:svelte` - 20 rune tests pass
- [ ] `npx vitest run --config vitest.config.ts src/reconcile-array.test.ts` - 12 unit tests pass
- [ ] `npx vitest run --config vitest.config.ts src/vue/use-all.test.ts` - 7 vue tests pass

### DIFF
--- a/packages/jazz-tools/src/vue/use-all.ts
+++ b/packages/jazz-tools/src/vue/use-all.ts
@@ -1,12 +1,13 @@
-import { shallowRef, toValue, watchEffect, type MaybeRefOrGetter, type ShallowRef } from "vue";
+import { ref, toValue, watchEffect, type MaybeRefOrGetter, type Ref } from "vue";
 import type { QueryBuilder, QueryOptions } from "../runtime/db.js";
 import type { SubscriptionDelta } from "../runtime/subscription-manager.js";
+import { reconcileArray } from "../reconcile-array.js";
 import type { CacheEntryHandle, UseAllState } from "../subscriptions-orchestrator.js";
 import { useJazzClient } from "./provider.js";
 
 function applyEntryState<T extends { id: string }>(
   state: UseAllState<T>,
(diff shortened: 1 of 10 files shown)


packages/jazz-tools/src/vue/use-all.ts
```diff
@@ -26,7 +27,11 @@ function subscribeToEntry<T extends { id: string }>(
       data.value = nextData;
     },
     onDelta: (delta: SubscriptionDelta<T>) => {
-      data.value = delta.all;
+      if (data.value) {
+        reconcileArray(data.value, delta.all);
```
gdorsi: Why are you not using the delta (added, removed, updated)?

That should help on skipping the "reconcileArray" and just deepMerge the rows.

joeinnes: With batched removals/additions the indexes in the subscription manager are no longer accurate compared to the info in the delta, but I guess we could build a set of IDs of rows which have changed and target those, and then just run one pass of `reconcileArray` at the end to make sure the order is correct?

gdorsi (approved):

PR #395, opened 2026-03-25
feat: add a review-current-branch skill

gdorsi:
This PR adds a review-current-branch skill that is designed to focus on the important bits of Jazz.

By directing the AI toward specific issues to check, and splitting the review in sub-agents, we can get a more in-depth analysis on what needs to be improved.

This review process still outputs false-positives, but it also gives superhuman insights on what can be improved.

We could probably double-down on specialization by splitting the Rust review agent from the TS one.

Curious on your takes/experiences around automated reviews.

### DIFF
--- a/.agents/skills/review-current-branch/SKILL.md
+++ b/.agents/skills/review-current-branch/SKILL.md
@@ -173,36 +173,7 @@ Output format:
 ...
 ```
 
-### Step 3: Conditionally spawn peripheral agents
-
-If changed files include `examples/` paths, spawn an **Examples Agent**:
-
-```
-You are reviewing example application changes for a local-first relational database framework. Examples serve as both documentation and integration tests.
-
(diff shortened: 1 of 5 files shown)


.agents/skills/review-current-branch/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+Read the changed files, test files, and surrounding context when needed. Use the Read tool.
+
+Output format:
+## Integration Test Coverage Review
+
+1. [severity] `file:line` — description of untested behavior and what test is missing
+...
+```
+
+### Step 3: Conditionally spawn peripheral agents
```
joeinnes: I don't think we should limit this. It's possible that changes on core now mean that examples/docs are outdated and need fixing. CI will catch test failures, but not subtle drift/deprecation.

rewinfrey:
Rick's Implement agent started on this issue.

Branch: `rick/395-import-throughput-ladder-rung-2-per`

_Posted automatically by the Opacity Dev agents._

PR #398, opened 2026-03-26
Fix inherited write checks on the payload branch

Takeno:
## Summary

Fix server-side inherited write checks so they evaluate against the mutation's actual branch, not `current_branch()`.

## Example
It happened while testing inheritance permission policies. 
Considering the tables:
`folders { title: string; owner_id: session.user_id}`
`documents { title:string, owner_id: session.user_id, folder: ref(folders)}`

A failing case was an insert into `documents` where Alice wrote:

- `owner_id = alice; folder_id = shared_folder`

The policy was valid because the row owner matched the session, and `folder_id` pointed at a folder Alice owned. But the server still rejected the write with `complex policy check failed`.

The reason was that the inherited parent lookup for `folder_id` was being evaluated on the wrong branch. `INHERITS` compiles to a graph that loads the parent row and evaluates its policy; if that graph scans a different branch than the incoming mutation, it can miss the parent row entirely or read stale state and incorrectly deny an otherwise valid write.

## Fix

Thread the payload branch through complex policy evaluation instead of falling back to `current_branch()`:

- use the mutation branch when building policy graphs for `INHERITS` / `EXISTS`
- keep that branch on deferred policy checks so settlement uses the same branch context
- hydrate the requested branch on cached objects, so settlement doesn't accidentally reuse an unrelated already-loaded branch

### DIFF
--- a/crates/jazz-tools/src/object_manager/mod.rs
+++ b/crates/jazz-tools/src/object_manager/mod.rs
@@ -6,7 +6,7 @@ use smolset::SmolSet;
 
 use crate::commit::{Commit, CommitId, StoredState};
 use crate::object::{Branch, BranchLoadedState, BranchName, Object, ObjectId};
-use crate::storage::{Storage, StorageError};
+use crate::storage::{LoadedBranch, Storage, StorageError};
 
 /// Unique identifier for a subscription.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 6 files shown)


crates/jazz-tools/src/object_manager/mod.rs
```diff
@@ -182,73 +215,56 @@ impl ObjectManager {
         branches: &[String],
     ) -> Option<&Object> {
         let _span = tracing::trace_span!("OM::get_or_load", %id).entered();
-        if self.objects.contains_key(&id) {
-            return self.objects.get(&id);
-        }
+        if let std::collections::hash_map::Entry::Vacant(e) = self.objects.entry(id) {
```
gdorsi: Why this instead of `!self.objects.contains_key(&id)`?

Is there any difference?

gdorsi (approved):

crates/jazz-tools/src/object_manager/mod.rs
Takeno: It's because there is now an `insertion` inside the if-clause, and the Entry::Vacant pattern avoids a double lookup.

`Entry(id)` gives you an Entry that holds the map slot, then `e.insert(...)` writes directly into that already-found slot — no second lookup

Disclaimer: `cargo` reported it explicitely on build

PR #410, opened 2026-03-27
fix: schema evolution edge cases

nrainhart:
## Description

Improves test coverage on lenses, and indentifies/fixes edge cases


crates/jazz-tools/src/schema_manager/transformer.rs
```diff
@@ -171,43 +170,6 @@ impl<'a> LensTransformer<'a> {
...
-    ///
-    /// # Arguments
-    /// * `base_loader` - Base loader function that fetches raw row data
-    /// * `branch_schema_map` - Maps branch names to schema hashes
-    ///
-    /// The returned loader will:
-    /// 1. Call base_loader to get raw row data
-    /// 2. Determine which schema the row is from (based on branch)
-    /// 3. Apply lens transforms if needed
-    pub fn make_loader<F>(
```
nrainhart, line 187: Unused

crates/jazz-tools/src/schema_manager/integration_tests.rs
```diff
@@ -1236,6 +1236,91 @@ mod tests {
         assert_eq!(manager_b.all_branches().len(), 2);
     }
 
+    /// Test catalogue update processing: draft lenses are stored but must not
+    /// activate pending schemas.
```
gdorsi, line 1240: Side note, we probably should extend the rules about how to document tests to the whole Rust crates.

It makes way easier to understand what the tests are covering

gdorsi (approved):

crates/jazz-tools/src/schema_manager/integration_tests.rs
nrainhart: +1

PR #415, opened 2026-03-29
Implement batch branch storage and prefix indexes

aeplay:
## Summary

This implements the batch-branch schema and the first storage/query rewrite behind it.

The branch moves Jazz from the old multi-tip branch model toward explicit batch branches of the form `env-schemaHash-userBranch-batchId`, then threads that model through object storage, query execution, sync state, and runtime bindings.

## What changed

- make batch ids mandatory in composed branch names and carry schema context explicitly on query subscription payloads
- add prefix-aware object and storage indexes for batch leaves, commit ownership, and table-prefix active batches
- move durable branch persistence from per-commit KV entries to compact branch refs, single-blob state, then manifest-plus-segment storage
- add tips-only cold loads and prefix batch catalogs so reads can avoid loading full branch history in common paths
- compact hot in-memory and storage identities with batch-aware keys and dense ordinals
- update query, sync, and provenance handling to operate on batch-aware refs instead of flat branch strings

## Why

This is the implementation follow-up to the batch-branch storage spec. The goal is to make concurrent work happen as concurrent branches instead of concurrent tips, while making large numbers of branches per object materially cheaper to store, load, and query.

## Validation

- `cargo fmt --all`
- `cargo check -p jazz-tools --lib`
- `cargo clippy -p jazz-tools --lib --tests --features fjall -- -D warnings`
- pre-commit hooks passed:
  - workspace clippy
  - workspace cargo check
  - `oxfmt`
  - `oxlint`
  - `rustfmt`


aeplay:
Closing in favor of #473

PR #380, opened 2026-03-23
Restructure and polish documentation site

joeinnes:
## Summary

- Restructure docs into logical sections: auth/, reading/, writing/, reference/, schemas/
- Add new pages: auth sessions, concepts, reading/writing guides, reference pages
- Polish quickstarts: consolidate, add Hono server pattern, mobile LAN IP guidance
- Multiple editorial passes: tighten prose, fix broken links, remove stale TODOs
- Extract inline code snippets into real, type-checked example files
- Expand docs example schemas (expo, server-ts) to support all demonstrated features
- Fix build, lint, and type errors across examples and packages
- Fix duplicate Rust struct fields and invalid QueryBuilder::clone() usage
- Update .oxlintignore for generated and snippet files
- Regenerate lockfile, app.ts, and migration files from updated schemas

## Test plan

- [x] `pnpm build` passes (all 17 packages)
- [x] `pnpm format:check` passes
- [x] `pnpm lint` passes (oxlint + cargo clippy)
- [x] `pnpm test` — 11/12 pass; `todo-client-localfirst-react-docs` has pre-existing failures unrelated to this PR

PR #417, opened 2026-03-29
Align docs restructure with current schema workflow

aeplay:
## What changed

This is a stacked follow-up on top of #380 / `joe/docs-restructure`.

Instead of rebasing Joe's branch onto current `main`, this keeps Joe's docs structure intact and layers the schema-workflow changes from #316 on top as a narrow diff.

Concretely, this updates the docs/examples to the current schema model:

- root-level `schema.ts` and `permissions.ts`
- `npx jazz-tools validate` instead of the old `build` workflow
- reviewed migrations under `migrations/`
- current permissions lifecycle language (`permissions push`, no schema hash churn for permission-only changes)
- example snippet imports that point at the new root schema entrypoints

To keep this PR readable against Joe's branch, the legacy generated `schema/` directories on the base branch are mostly left in place rather than being mass-deleted here.

## Why

Joe asked for a readable stacked diff rather than a `main`-based replacement PR. The earlier restack against `main` made the compare too noisy for review because `joe/docs-restructure` does not contain the already-merged #316 changes.

This branch backports the relevant docs/example layer so reviewers can see just the schema-workflow alignment on top of Joe's structure.

## Validation

- `pnpm --filter docs types:check`
- `pnpm --filter todo-server-ts-docs build`
- `pnpm --filter todo-client-localfirst-react-docs build`
- `pnpm --filter todo-client-localfirst-vue-docs build`
- `pnpm --filter todo-client-localfirst-svelte-docs build`
- `pnpm --filter todo-client-localfirst-ts-docs build`

## Notes

This is intentionally scoped as a stacked docs/example backport. It carries the new schema workflow into Joe's restructure branch without pulling unrelated `main` history into the PR diff.

### DIFF
--- a/docs/content/docs/reference/server-setup.mdx
+++ b/docs/content/docs/reference/server-setup.mdx
@@ -100,10 +100,17 @@ For React Native/Expo clients, make sure `serverUrl` points to a host your runti
 
 ### Backend identity pattern
 
-On the client side, queries automatically run as the logged-in user. On a backend, create the context once at startup, then choose whether a request should run as the server or as the caller.
+On the client side, queries automatically run as the logged-in user. On a backend, create the
+context once at startup, then choose whether work should run as an unscoped local runtime, as an
+explicit backend connection, or as the caller.
 
-- `context.db()` returns a `Db` that runs queries as the configured runtime identity, useful for admin tasks or background jobs that aren't tied to a specific user.
-- `context.forRequest(req)` returns a `Db` that runs queries as the authenticated user who made the request, so permission policies apply to them rather than the server.
+- `context.db()` returns the shared high-level `Db` for the context with no per-request user
+  session attached, useful for local-only jobs or server-owned work that is not impersonating a
+  caller.
+- `context.asBackend()` returns a `Db` that uses `backendSecret` for sync/auth when this process
+  talks to a Jazz server.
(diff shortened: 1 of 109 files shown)


docs/content/docs/reference/server-setup.mdx
```diff
@@ -100,22 +100,26 @@ For React Native/Expo clients, make sure `serverUrl` points to a host your runti
...
+In TypeScript backends, create the context once with both `app` and `permissions`, then use
+`context.asBackend()` for server-owned work and `context.forRequest(req)` when you need a
+bearer-JWT-scoped high-level `Db`.
 
 `forRequest` reads authentication from standard HTTP headers, so `req` can be any object that exposes them: Express, Hono, Fastify, or a Web Fetch API `Request` all work.
 
+<Callout type="info">
+  `context.db()` still exists for embedded or local-only runtime setups, but it is not the primary
+  pattern for server-connected backends.
+</Callout>
```
joeinnes, line 120: I think this reads a bit unusually for someone who hasn't come across it before.

```suggestion
<Callout type="info">
  For embedded or local-only runtime setups where your backend is not connected to a server, use `context.db()` to get an unscoped, unauthenticated handle. This allows read/write access to the database directly, without permissions being evaluated.
</Callout>
```

PR #418, opened 2026-03-30
Fix NULL literal handling in policy evaluation

Takeno:
## Summary

Before this change, comparisons like column = NULL were evaluated through the normal scalar encoding path. That works for ordinary values, but it gives the wrong result for Value::Null, where the engine needs null-aware semantics instead of byte comparison.

- handle `Value::Null` with value-level semantics in row policy comparisons instead of encoded-byte comparison
- normalize null-valued query conditions into `IS NULL` and `IS NOT NULL` predicates and prevent them from being treated as index-scannable scalar filters
- add coverage for null literal comparisons in direct policy evaluation, query predicate lowering, ReBAC inserts, `EXISTS REL`, and select filtering

## Testing
- `cargo test -p jazz-tools null_`

### DIFF
--- a/crates/jazz-tools/src/query_manager/policy.rs
+++ b/crates/jazz-tools/src/query_manager/policy.rs
@@ -880,6 +880,20 @@ pub fn evaluate_cmp(
         None => return false,
     };
 
+    if cmp_value.is_null() {
+        return match column_is_null(descriptor, content, col_index) {
+            Ok(is_null) => match op {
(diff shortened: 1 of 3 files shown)


gdorsi (approved):
LGTM

Could you add an integration test to cover the fix?

Takeno:
> LGTM
> 
> Could you add an integration test to cover the fix?

It is already in #403, where I found the bug. I have a should_panic test that will be fail once merged this

PR #419, opened 2026-03-30
feat(storage): add RocksDB storage backend

gdorsi:
## Summary

Adds RocksDB as an alternative persistent storage backend behind the `rocksdb` feature flag, passing all 27 conformance tests. Fjall remains the default for `client`/`cli` features — RocksDB is opt-in for now.

## Benchmark results

Storage backend comparison on the realistic benchmark suite (Apple Silicon, release build):

| Benchmark | Memory | Fjall | RocksDB | RocksDB vs Fjall |
|---|---|---|---|---|
| R1 CRUD sustained (512 ops/batch) | 15.3 ms | 239 ms | 72.6 ms | ~3.3x faster |
| R2 Reads sustained (512 ops/batch) | 238 ms | 555 ms | 279 ms | ~2x faster |
| R3 Cold load (open + query 200 tasks) | — | 778 ms | 32 ms | ~24x faster |
| R8 Many-branches cold load (5000 branches) | — | 189 ms | 152 ms | ~1.25x faster |

Run with:
```
cargo bench --bench realistic_phase1 -p jazz-tools \
  --features jazz-tools/fjall,jazz-tools/rocksdb \
  -- "crud_sustained|reads_sustained|cold_load"
```

## Design choices

**TransactionDB over WriteBatch.** `storage_core.rs` interleaves reads and writes within operations (e.g., `append_commit_core` reads the current commit list, appends, writes back). `WriteBatch` only applies at the end, so reads would see stale data. `TransactionDB` with pessimistic transactions gives read-your-writes semantics within each Storage method, matching Fjall's `write_tx()` behavior.

**RefCell\<Transaction\> for shared borrow.** Several `storage_core` functions take both a `get` and `set` closure that must borrow the same transaction. Wrapping the transaction in `RefCell` lets both closures borrow it without fighting the borrow checker.

**Direct DB reads for read-only methods.** Read-only Storage methods (`load_object_metadata`, `load_branch`, `index_lookup`, etc.) skip transactions entirely and use `db.get()` / `db.iterator()` directly — no transaction overhead for queries.

**iterate_upper_bound on all scans.** Prefix and range scans set `ReadOptions::set_iterate_upper_bound` so RocksDB's internal iteration stops at the boundary instead of scanning past it and relying on application-side `starts_with` / comparison checks.

**Single column family.** The existing composite key scheme (`obj:`, `idx:`, `ack:`, `catman:`) maps directly to RocksDB's default lexicographic byte ordering. No prefix extractor needed.

## Test plan

- [x] All 27 storage conformance tests pass (`storage_conformance_tests_persistent!` macro)
- [x] Server and client integration tests pass with RocksDB wired in
- [x] Cold-load benchmarks added and run (R3 + R8)
- [x] CRUD and read benchmarks added for both backends (R1 + R2)
- [x] Verify `cargo test --features fjall` still works with Fjall fallback


CONTRIBUTING.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+libclang.dylib not found
+```
+
+**Fix:** add these to your shell profile (`~/.zshrc` or `~/.bash_profile`):
+
+```sh
+export DYLD_LIBRARY_PATH=/opt/homebrew/Cellar/llvm/19.1.7_1/lib${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}
+export LIBCLANG_PATH=/opt/homebrew/Cellar/llvm/19.1.7_1/lib
```
antoniomuso: What about Linux? Do we need to install LLVM there, too? 
Also, if these are build-time variables, could we move them into `config.toml` or `build.rs`? That would let us strip out the manual configuration entirely.

antoniomuso (approved):
LGTM! The code is ok, I added a comment on the build variables. GJ!

CONTRIBUTING.md
gdorsi: Good point.

> Do we need to install LLVM there, too?

Yes, but it's often already available, in fact CI is building without any config changes.

Added the line for completeness.

> Also, if these are build-time variables, could we move them into config.toml or build.rs?

Good point, yes we can!

.cargo/config.toml
```diff
@@ -0,0 +1,7 @@
+[env]
```
antoniomuso: These variables have to be set just on Macos

antoniomuso: These variables have to be set just on Macos

PR #420, opened 2026-03-30
fix: type inference for default values

nrainhart:
## Description

Modifies type inference for the TS DSL to make fields with default values optional when inserting a record.

Also restores tests that were deleted by https://github.com/garden-co/jazz2/pull/316

### DIFF
--- a/packages/jazz-tools/tests/ts-dsl/insert-api.test.ts
+++ b/packages/jazz-tools/tests/ts-dsl/insert-api.test.ts
@@ -1,7 +1,7 @@
 import { createDb, type Db } from "../../src/runtime/db.js";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { app } from "./fixtures/basic/schema";
-import { insertUser, uniqueDbName } from "./factories";
+import { insertProject, insertUser, uniqueDbName } from "./factories";
 
 describe("TS Insert API", () => {
   let db: Db;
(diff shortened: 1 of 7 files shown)


packages/jazz-tools/tests/ts-dsl/insert-api.test.ts
```diff
@@ -83,91 +83,66 @@ describe("TS Insert API", () => {
...
     });
 
-    const result = db.update(app.todos, todo.id, { done: true });
-    expect(result).toBeUndefined();
-
-    const [updated] = await db.all(app.todos.where({ id: { eq: todo.id } }));
-    expect(updated!.done).toBe(true);
-  });
-
-  it("can wait for updates to be persisted up to a specific durability tier", async () => {
```
Takeno, line 105: We have lost these tests on updates/deletes. Is it ok?

nrainhart: They were moved to their own files! [update-api.test.ts](https://github.com/garden-co/jazz2/blob/fix/default-values/packages/jazz-tools/tests/ts-dsl/update-api.test.ts#L6)  & [delete-api.test.ts](https://github.com/garden-co/jazz2/blob/fix/default-values/packages/jazz-tools/tests/ts-dsl/delete-api.test.ts#L6).

These were duplicate tests that ended up here after https://github.com/garden-co/jazz2/pull/316

Takeno (approved):

PR #424, opened 2026-03-30
feat: use `null` instead of `undefined` for nullable columns

nrainhart:
## Description

- Change Jazz read surfaces so selected row fields that currently come back as `undefined` because of runtime `NULL` instead come back as `null`.
- Apply that consistently to `db.all`, `db.one`, `db.insert`, `db.insertDurable`, nested includes, permission magic columns, and `subscribeAll`/delta item payloads.
- Update nullable `where(...)` typing to accept null and `eq/ne: null`

## Motivation

`null` & `undefined` already have distinct semantics in inserts and updates:
- for inserts: undefined means "use default value", null means "set column to null"
- for updates: undefined means "don't update the value", null means "set column to null"

Requiring to use `null` to set the value and returning `undefined` was definitely weird (see https://github.com/garden-co/jazz2/pull/349#discussion_r2948944380), and this change also aligns optional value handling more closely with SQL.

### DIFF
--- a/packages/jazz-tools/tests/ts-dsl/query-api.test.ts
+++ b/packages/jazz-tools/tests/ts-dsl/query-api.test.ts
@@ -71,8 +71,9 @@ describe("TS Query API", () => {
       expect(results.map((todo) => todo.id)).toEqual([todoWithOwner.id]);
     });
 
-    it("filters with explicit null values always return false", async () => {
-      const _todoWithoutOwner = insertTodo(db, {
+    // Note: this is a difference with respect to SQL, when =null checks always return false.
+    it("filters with explicit null values work as isNull:true", async () => {
+      const todoWithoutOwner = insertTodo(db, {
         title: "Todo without owner",
         ownerId: null,
       });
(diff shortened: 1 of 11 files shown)


packages/jazz-tools/tests/ts-dsl/query-api.test.ts
```diff
@@ -54,14 +41,68 @@ describe("TS Query API", () => {
...
+        title: "Todo with owner",
+        ownerId: insertUser(db).id,
+      });
+
+      const results = await db.all(app.todos.where({ ownerId: { isNull: false } }));
+
+      expect(results.map((todo) => todo.id)).toEqual([todoWithOwner.id]);
+    });
+
+    it("filters with explicit null values always return false", async () => {
```
nrainhart: This preserves what is arguably a SQL irk: https://en.wikipedia.org/wiki/Null_%28SQL%29#Common_mistakes

We could alternatively not allow using `eq/ne: null`, and convert `row: null` into an `is null` check. @aeplay what do you think?

nrainhart: Yeah, I think this makes the most sense. Went ahead and implemented it: [`5b14565` (this PR)](https://github.com/garden-co/jazz2/pull/424/commits/5b145659727e2eacedb7b73f94caaaa2a7cbe19e)

gdorsi (approved):

PR #433, opened 2026-03-31
feat: `AddTable`, `RemoveTable` & `RenameTable` lenses

nrainhart:
## Description

- Uses `AddTable` and `RemoveTable` lenses when querying rows from different schemas, to prevent fetching rows from tables that do not exist in that schema.
- Adds a new `RenameTable` lens
	- these lenses are auto-generated if the table is renamed while keeping the exact same structure (the generated lens is considered a draft, like we do for renamed columns)

## Migration DSL

Added/removed/renamed columns are now tracked as part of the migration:

```ts
s.defineMigration({
  fromHash: "aaaaaaaaaaaa",
  toHash: "bbbbbbbbbbbb",
  from: {
    users: s.table({
      email: s.string(),
    }),
    legacyProfiles: s.table({
      bio: s.string().optional(),
    }),
  },
  to: {
    profiles: s.table({
      bio: s.string().optional(),
    }),
    tasks: s.table({
      description: s.string(),
    }),
  },
  createTables: {
    tasks: true,
  },
  dropTables: {
    users: true,
  },
  renameTables: {
    profiles: s.renameTableFrom("legacyProfiles"),
  },
});
```


crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -78,6 +91,40 @@ impl QueryManager {
...
+        self.find_schema_by_short_hash(&composed.schema_hash)
+    }
+
+    fn row_object_metadata_for_branch(&self, table: &str, branch: &str) -> HashMap<String, String> {
+        let mut metadata = HashMap::from([(MetadataKey::Table.to_string(), table.to_string())]);
+        if let Some(origin_schema_hash) = self.origin_schema_hash_for_branch(branch) {
+            metadata.insert(
+                MetadataKey::OriginSchemaHash.to_string(),
+                origin_schema_hash.to_string(),
+            );
```
nrainhart, line 123: Now we include the object's original schema hash as part of its metadata. This lets us convert the original table name into the current name, assuming there's a lens path between the original schema and the current schema.

I think this is actually more efficient than storing table name changes, as the object storage cost does not increase with each rename, and we're already fetching the object metadata (so there are no extra loads from disk)

packages/jazz-tools/src/cli.ts
```diff
@@ -276,6 +276,57 @@ function changedTableNames(fromSchema: WasmSchema, toSchema: WasmSchema): string
...
+      )
+      .filter((addedIndex) => addedIndex >= 0),
+  );
+  const addedCandidates = addedTables.map((newTableName) =>
+    removedTables
+      .map((oldTableName, removedIndex) =>
+        tableSchemasEqual(fromSchema[oldTableName], toSchema[newTableName]) ? removedIndex : -1,
+      )
+      .filter((removedIndex) => removedIndex >= 0),
+  );
```
gdorsi: Why is this scanning both sets?

crates/jazz-tools/src/schema_manager/integration_tests.rs
```diff
@@ -1023,6 +1145,726 @@ mod tests {
...
+        assert_eq!(results.len(), 1);
+
+        let (_, row) = &results[0];
+        assert_eq!(row.len(), 2);
+        assert_eq!(row[0], Value::Uuid(row_id));
+        assert_eq!(row[1], Value::Text("alice@example.com".to_string()));
+    }
+
+    /// Table renames must also keep existing subscriptions reactive when old-schema
+    /// rows arrive after the query graph has already been compiled.
```
gdorsi, line 1234: how about adding a test about:
1. V1 is created
2. A subscrition on V1 is done
3. V2 is created with RenameTable lens
4. A row with v2 is pushed, the subscription should get an update in v1

crates/jazz-tools/src/schema_manager/integration_tests.rs
```diff
@@ -1023,6 +1145,726 @@ mod tests {
...
+        assert_eq!(results.len(), 1);
+        assert_eq!(results[0].0, row_id);
+        assert_eq!(
+            results[0].1[1],
+            Value::Text("alice@example.com".to_string())
+        );
+    }
+
+    #[test]
+    fn table_rename_update_and_delete_copy_on_write() {
```
gdorsi, line 1409: Could you add a doc comment to explain what this test does cover?

crates/jazz-tools/src/schema_manager/integration_tests.rs
gdorsi: One simple and one that subscribes with a join

crates/jazz-tools/src/schema_manager/integration_tests.rs
```diff
@@ -829,6 +837,118 @@ mod tests {
         assert_eq!(charlie_row.1[3], Value::Text("admin".to_string()));
     }
 
+    #[test]
+    fn removed_table_then_readded_does_not_resurface_old_rows() {
```
gdorsi, line 841: How about having dedicated tests for Adding a table on an existing schema and dropping a table with a lens?

gdorsi (approved):

packages/jazz-tools/src/cli.ts
nrainhart: Simplified this in https://github.com/garden-co/jazz2/pull/433/changes/3faa7adf4ee7b331aaf23bef247b3fcd0892071e

crates/jazz-tools/src/schema_manager/integration_tests.rs
nrainhart: There's not really much we can test in these cases:
- if you drop a table from v1, you cannot query against it in the v2 schema (trying to query a table that does not exist throws an error)
- if you add a table, there won't be any data from that table in a previous schema

crates/jazz-tools/src/schema_manager/integration_tests.rs
nrainhart: Done in https://github.com/garden-co/jazz2/pull/433/commits/43ed07a1b5a9ec58cbcd333fea6f786256beb1ba

PR #453, opened 2026-04-02
feat(storage): SQLite backend

gdorsi:
## Design

**Schema:** single `kv (key BLOB PRIMARY KEY, value BLOB) WITHOUT ROWID` — a pure B-tree with no rowid indirection. All storage abstractions (commits, branches, indexes, catalogue) are encoded as prefixed keys by `storage_core`.

**WAL + synchronous=NORMAL:** writers don't block readers; no extra fsync after each commit, though checkpoints may sync. Considering the mobile-first design we think that is an acceptable durability tradeoff.

**Batched write transaction:** the first mutating call opens a `BEGIN` that stays open across subsequent writes. `flush_wal()` commits the open transaction then runs `PRAGMA wal_checkpoint(PASSIVE)` to merge WAL pages into the main db file. `flush()` delegates to `flush_wal()`. This turns N inserts into one transaction boundary instead of N — the main reason SQLite was initially 2.5× slower than fjall on mobile. Multi-step mutations use nested `SAVEPOINT`s inside the open transaction for individual rollback semantics.

**Durability tradeoffs:**
- Writes since the last `flush_wal()` are lost on process crash — they are not even committed to SQLite's WAL yet. This is stricter than fjall's default, where each write goes to OS buffers immediately and is only vulnerable to power loss or kernel crash.
- With `synchronous=NORMAL` there is no extra fsync after each commit; checkpoints may sync. Committed writes survive process crashes but NORMAL does not guarantee durability across power loss or system crash.

Callers must invoke `flush_wal()` at meaningful checkpoints (sync cycle complete, app backgrounded). `close()` always commits, so clean shutdown is safe. Single-connection: reads within the open transaction see their own uncommitted writes.

## Benchmarks

**On-device (stress-test/expo) — 15k todo inserts:**

| Backend | Time  | Inserts/sec | vs SQLite |
|---------|-------|-------------|-----------|
| SQLite  | 1.90s | 7885        | baseline  |
| fjall   | 2.06s | 7293        | 1.08× slower |

**Microbenchmark (R1 — crud_sustained, median):**

| Backend | Time   | Throughput   | vs SQLite |
|---------|--------|--------------|-----------|
| SQLite  | 61.6ms | 8.31 Kelem/s | baseline  |
| RocksDB | 74.8ms | 6.85 Kelem/s | 1.21× slower |
| fjall   | 262ms  | 1.95 Kelem/s | 4.25× slower |

**Microbenchmark (R2 — reads_sustained, median):**

| Backend | Time  | Throughput   | vs SQLite |
|---------|-------|--------------|-----------|
| SQLite  | 348ms | 1.47 Kelem/s | baseline  |
| RocksDB | 353ms | 1.45 Kelem/s | 1.01× slower |
| fjall   | 740ms | 0.69 Kelem/s | 2.13× slower |


crates/jazz-tools/src/server/builder.rs
```diff
@@ -14,10 +14,12 @@ use crate::routes;
 use crate::runtime_tokio::TokioRuntime;
 use crate::schema_manager::{AppId, SchemaManager, rehydrate_schema_manager_from_manifest};
 use crate::server::{CatalogueAuthorityMode, DynStorage, ExternalIdentityStore, ServerState};
-#[cfg(all(feature = "fjall", not(feature = "rocksdb")))]
+#[cfg(all(feature = "fjall", not(feature = "rocksdb"), not(feature = "sqlite")))]
```
gdorsi, line 17: Gonna remove fjall after this PR, and try to consolidate the feature flags because now it is a mess.

CONTRIBUTING.md
```diff
@@ -2,42 +2,40 @@
...
+Then add to your shell profile (`~/.zshrc`, `~/.bashrc`, etc.):
 
 ```sh
-brew --prefix llvm
+export RUSTC_WRAPPER=sccache
 ```
 
-Then use that path in either your shell profile or your personal
-`~/.cargo/config.toml`:
+### libclang for RocksDB builds
```
gdorsi, line 19: Found these instructions to be more simple and effective than hand-writing the env vars.

turbo.json
```diff
@@ -1,10 +1,9 @@
 {
   "$schema": "https://turbo.build/schema.json",
```
gdorsi, line 2: Done some improvements here using the turbo skill

crates/jazz-tools/src/storage/sqlite.rs
```diff
@@ -0,0 +1,580 @@
...
+
+    pub fn open(path: impl AsRef<Path>) -> Result<Self, StorageError> {
+        let path = path.as_ref();
+        let conn = rusqlite::Connection::open(path)
+            .map_err(|e| StorageError::IoError(format!("sqlite open: {e}")))?;
+
+        conn.execute_batch(
+            "PRAGMA journal_mode = WAL;
+             PRAGMA synchronous = NORMAL;
+             PRAGMA cache_size = -65536;
```
gdorsi, line 89: Ended up using the same cache size as fjall and rocksdb.

Got a sensible performance improvement over 2mb cache.

antoniomuso (approved):

PR #459, opened 2026-04-02
[codex] Show full schema file and require schema namespace imports

aeplay:
## What changed

This PR now has two related cleanup passes:

- updated `/docs/schemas/defining-tables` to show a complete `schema.ts` example instead of separate partial snippets
- removed the legacy root schema authoring exports from `jazz-tools` so schema code uses `import { schema as s } from "jazz-tools"`
- added tests that lock the public export surface to the `schema` namespace path
- migrated the remaining in-repo callers that still used the old root schema imports

## Why

The schema docs page felt incomplete compared to the quickstarts because it showed only the table body first and made readers reconstruct the full file shape themselves.

At the same time, `jazz-tools` still exposed the older root schema exports even though the intended API is the namespaced `schema as s` style. Tightening that surface makes the docs, quickstarts, and package API all point at the same pattern.

## Impact

Readers now see a copy-pasteable full `schema.ts` example up front.

Library consumers must use the namespaced schema API for schema authoring and helper types, for example:

```ts
```

## Validation

- `pnpm exec oxfmt --check docs/content/docs/schemas/defining-tables.mdx examples/docs/schema-defining-tables/schema.ts`
- `pnpm --filter docs run types:check`
- `pnpm --filter docs run build`
- `pnpm exec oxfmt --check packages/jazz-tools/src/index.ts packages/jazz-tools/src/index.test.ts packages/jazz-tools/tests/ts-dsl/schema-namespace-only.typecheck.ts packages/jazz-tools/src/permissions/type-inference.test.ts packages/inspector/tests/browser/schema.ts examples/wequencer/schema/migration_v2_v3_942c3359fce7_f9bfb520269b.ts`
- `pnpm --filter jazz-tools exec vitest run src/index.test.ts src/permissions/type-inference.test.ts --config vitest.config.ts`
- `pnpm --filter jazz-tools exec tsc --project tsconfig.tests.json`

### DIFF
--- a/docs/content/docs/schemas/defining-tables.mdx
+++ b/docs/content/docs/schemas/defining-tables.mdx
@@ -24,10 +24,12 @@ separate file. The `migrations/` directory holds reviewed migration stubs — se
 
 ## Table definitions
 
-Tables are defined in `schema.ts` using the Jazz DSL. Each `s.table(...)` call registers a table and `s.ref(...)` defines typed relations between them.
+Tables are defined in `schema.ts` using the Jazz DSL. A typical schema file imports the DSL,
+defines its tables, exports the typed `app`, and can derive helper types in the same module. Each
+`s.table(...)` call registers a table and `s.ref(...)` defines typed relations between them.
 
 <include cwd lang="ts" meta='title="schema.ts"'>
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Closing because the integration branch already adopted schema namespaces throughout examples and documentation (`87542093`). This old branch should not be ported.

PR #460, opened 2026-04-03
 fix: permissions push

nrainhart:
## Description

`jazz-tools permissions push` was failing even when the local schema matched the server schema, because the schemas were not considered equal:
- the server incorrectly returned `references: null` for each column without a reference
- columns could be reordered, and the CLI didn't account for this

### DIFF
--- a/packages/jazz-tools/tsconfig.tests.json
+++ b/packages/jazz-tools/tsconfig.tests.json
@@ -6,7 +6,6 @@
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "noEmit": true,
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/tsconfig.tests.json
```diff
@@ -6,7 +6,6 @@
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "noEmit": true,
-    "noUncheckedIndexedAccess": true,
```
nrainhart, line 9: Aligns type-checking for tests to our regular tsconfig settings

PR #461, opened 2026-04-03
feat: add `migrations create-missing` CLI command

nrainhart:
## Description

Adds a new `migrations create-missing` CLI command to generate migration stubs relative to the server's current schema (i.e. the explicitly set server schema, or else the schema used for enforcing permissions). It works the following way:
- If you pass `--schema-dir`, it compiles that local schema and creates a single migration stub from the current server schema to the local schema
	- Workflow: dev starts from current schema, makes changes locally, and then generates the migration file
- If you omit `--schema-dir`, it asks the server which stored schemas are not yet reachable from the current server schema and generates one direct migration stub from each of those schemas to the current server schema.
	- Workflow: dev realizes there's missing data from old schemas, and generates the required migrations to make that data accessible again

### DIFF
--- a/todo/projects/new-schema-migration-flow/spec.md
+++ b/todo/projects/new-schema-migration-flow/spec.md
@@ -1,0 +1,168 @@
+# New Schema Migration Flow
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


nrainhart:
Closed in favor of https://github.com/garden-co/jazz2/pull/471

PR #463, opened 2026-04-03
fix: update PolicyTestApp to accept custom expect function

Takeno:


gdorsi (approved):

.changeset/policy-test-custom-expect.md
```diff
@@ -0,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Fix `createPolicyTestApp(...)` so policy test helpers no longer hard-code Vitest's `expect`.
+
+Callers now pass the `expect` function explicitly, which keeps `jazz-tools/testing` policy assertions working when the test harness provides its own assertion context.
```
gdorsi, line 7: Do we need a changeset?

Takeno: It is an exposed method for end users, even if probably nobody is using it yet. Do we just let TS report / JS crash?

PR #464, opened 2026-04-03
Update auth examples to use $createdBy mutation policies

Takeno:
## Summary
- migrate auth chat examples from field-based mutation ownership checks to `$createdBy`


joeinnes (approved):
LGTM!

PR #466, opened 2026-04-05
Introduce FFIValue & fix NAPI file creation

aeplay:
## What changed
This PR turns the original `s.bytes()` failure into a boundary refactor: mutation writes now cross runtime boundaries through an explicit `FFIValue` shape instead of depending on whatever each runtime's default JSON serialization does to nested values.

Concretely:
- `JazzClient` now preserves `Uint8Array` in a named `FFIValue` mutation boundary instead of silently coercing `Bytea` to `number[]`
- React Native writes now use an explicit FFI JSON codec for `Bytea` instead of raw `JSON.stringify(values)`
- N-API writes now decode through a dedicated FFI reader that accepts native JS binary inputs, including `Uint8Array`
- the original red repros remain in place and now pass unchanged

## Why
The trigger was the direct `s.bytes()` bug:
- `db.insert(...)` into a `Bytea` column failed with a real `Uint8Array`
- `db.update(...)` showed the same behavior
- `db.createFileFromBlob(...)` failed on the first `file_parts.data` insert

That bug exposed a larger design issue: our mutation boundaries in N-API and RN were too close to "just squeeze JSON through it," which meant binary values depended on transport quirks instead of on an intentional boundary contract.

## Design
`FFIValue` is the logical mutation boundary.

That boundary is now kept separate from transport details:
- browser/WASM continues to receive native JS values
- N-API decodes native JS values directly into Rust-side mutation values
- RN still uses a JSON transport in this PR, but it is now an explicit FFI transport codec rather than implicit default JSON behavior

This keeps direct callers on `Uint8Array` while making the runtime-specific transport logic explicit and local.

## Validation
Passed locally:

```sh
cd packages/jazz-tools
node_modules/.bin/vitest run --config vitest.config.ts src/runtime/client.mutations.test.ts src/react-native/jazz-rn-runtime-adapter.test.ts
node_modules/.bin/vitest run --config vitest.config.ts src/runtime/napi.integration.test.ts -t "accepts Uint8Array|stores Blob chunks"
cd /Users/anselm/.codex/worktrees/8405/jazz2
cargo test -p jazz-rn --lib --no-default-features --quiet
pnpm --dir crates/jazz-napi build:debug
```

I also tried a plain `cargo test -p jazz-napi --lib --quiet`, but in this workspace it appeared to hang after the relevant build/test work had already completed, so I relied on the rebuilt-addon integration repros above for the end-to-end N-API signal.


gdorsi (approved):

crates/jazz-rn/rust/src/lib.rs
```diff
@@ -96,12 +97,77 @@ where
...
+}
+
+fn decode_ffi_json_record(values_json: &str) -> Result<HashMap<String, Value>, JazzRnError> {
+    let values: HashMap<String, FfiJsonValue> =
+        serde_json::from_str(values_json).map_err(json_err)?;
+    values
+        .into_iter()
+        .map(|(key, value)| decode_ffi_json_value(value).map(|value| (key, value)))
+        .collect()
+}
```
gdorsi, line 163: 🤖 Add a round-trip test for RN hex encoding. At minimum, test that encodeFFIRecordToJson → Rust decode_ffi_json_record round-trips a Bytea value. This could be a unit test against the RN
 adapter or a Rust unit test.

antoniomuso (approved):

gdorsi (approved):

PR #467, opened 2026-04-05
[codex] Fix inspector data grid freeze loop

aeplay:
## What changed

This fixes the inspector data-grid freeze when switching to the next page or sorting a column.

It also adds a browser repro harness and gated diagnostics so the freeze path is easy to inspect if it regresses.

## Why it changed

The inspector was repeatedly re-rendering while a newly selected query was still pending. In the failing path, we were driving `useAll(...)` and `useReactTable(...)` from the same component. That let TanStack table state churn and pending-query fallback values feed back into the query-owning render path on the main thread.

## Root cause

The freeze was not one giant query task alone.

The main loop was caused by a feedback path between:
- `useReactTable(...)` internal state updates and option writes
- pending query transitions that temporarily produced empty rows
- the query hook and table hook living in the same component render

In the failing repro, `Next page` and `Sort by title` would switch to a new pending query and then spin `makeQueryKey` / `reuse-entry` over and over while the second subscription remained pending.

## Fix

- move the TanStack table hook behind a child component boundary
- keep pending rows stable with a shared `EMPTY_ROWS` value
- memoize the `useAll` query options
- keep the isolated browser repro and diagnostics coverage

## Impact

The inspector stays responsive when paging and sorting, and the new repro spec protects the fix.

## Validation

- `pnpm lint`
- `pnpm --filter inspector test -- src/components/data-explorer/TableDataGrid.test.tsx`
- `pnpm --filter inspector test:browser -- tests/browser/pagination-freeze-repro.spec.ts tests/browser/standalone-inspector.spec.ts`

### DIFF
--- a/.changeset/olive-buttons-repair.md
+++ b/.changeset/olive-buttons-repair.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
-Improve inspector data-grid stability and diagnostics around pending query transitions.
+Fix the inspector data grid freezing the browser tab when paging or sorting, and improve diagnostics around pending query transitions.
--- a/.changeset/olive-buttons-repair.md
+++ b/.changeset/olive-buttons-repair.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Improve inspector data-grid stability and diagnostics around pending query transitions.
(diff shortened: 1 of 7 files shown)


Takeno:
Closed as managed in #470

PR #471, opened 2026-04-06
feat: new schema migration flow

nrainhart:
## Problem

The current migration authoring flow is server-led: a developer changes `schema.ts`, runs the app, notices old data is missing, digs through server logs to discover schema hashes, then runs `jazz-tools migrations create <fromHash> <toHash>`. That makes migration authoring reactive instead of part of normal schema work. It also couples the default workflow to server state, runtime warnings, and network access. In practice, developers cannot create a migration until they have seen a warning.

The default workflow should instead be local and proactive. A developer who changes `schema.ts` should be able to run one command, offline, and get the next migration stub by diffing the current schema against the latest committed local schema snapshot. When recovery from server state is needed, the primitive should be "materialize the missing structural schema locally" rather than a special migration command.

## Solution

This project changes the migration flow from "discover hashes from the server, then author a migration" to "diff the current schema against the latest local snapshot, then author a migration". It introduces committed JSON snapshot files, changes `jazz-tools migrations create` to not require positional from/to hashes, and extends `jazz-tools schema export` so a developer can ask for a specific schema hash and have it loaded from the server.

### File Layout

```text
app-root/
├── schema.ts
├── permissions.ts
└── migrations/
    ├── 20260406T153045Z-rename-todo-title-a1b2c3d4e5f6-bbccddeeff00.ts
    └── snapshots/
        ├── 20260401T101500Z-a1b2c3d4e5f6.json
        └── 20260406T153045Z-bbccddeeff00.json
```

- `schema.ts`
  - remains the only structural schema file a developer edits directly
- `migrations/*.ts`
  - remains the reviewed migration edge files that are pushed to servers
- `migrations/snapshots/*.json`
  - are machine-generated, committed schema snapshots
  - are ordered by a sortable UTC timestamp prefix
  - contain the exact JSON emitted by `jazz-tools schema export --format json`

Committed snapshots are part of the repository and are reviewed alongside migrations.

The CLI defines the "latest snapshot" as the file in `migrations/snapshots/` with the lexicographically greatest timestamped filename.

#### Normal Flow

1. A developer edits `schema.ts`.
2. The developer runs `jazz-tools migrations create`.
3. The CLI compiles the current schema from `schema.ts` into canonical structural JSON.
4. The CLI loads the latest local committed snapshot from `migrations/snapshots/`.
5. The CLI diffs latest snapshot JSON -> current schema JSON.
6. The CLI writes:
   - a new migration stub in `migrations/`
   - a new committed snapshot for the current schema in `migrations/snapshots/`
7. The developer reviews the migration and snapshot, commits both files, and later runs `jazz-tools migrations push`.

#### First Run in a Repo

1. A developer runs `jazz-tools migrations create`.
2. The CLI finds no existing committed snapshots.
3. The CLI writes a single initial snapshot for the current schema.
4. The CLI prints that no migration was created because there was no previous local schema baseline.

This creates the baseline for future diffs.

#### Historical Recovery

1. A developer sees that the server contains rows under schema hash `a1b2...`.
2. The developer runs `jazz-tools schema export --schema-hash a1b2...`.
3. If that schema is not already available locally, the CLI fetches it from the server saves the JSON snapshot.
4. The developer decides how to connect the exported schema to an existing schema and runs `jazz-tools migrations create --fromHash a1b2...`.
5. The CLI diffs the fetched historical schema -> current schema.
6. The CLI writes a normal migration stub. The snapshot is not generated, as it's already present from 2.

### Runtime Warnings

Missing-schema detection remains valuable, but it changes role. It is no longer responsible for teaching developers the normal migration authoring flow. Instead, it acts as a recovery warning when stored server data is incompatible with the currently queried schema graph.

When the runtime or server detects rows that are not visible because no lens path exists, it should emit an actionable warning that points to `schema export --schema-hash`, including the missing hash or hashes.


todo/issues/lens-transform-silent-degradation.md
nrainhart, line 1: This issue is stale, it has already been addressed

aeplay (approved):
Comments from codex - these can all be addressed in followups

1. [warning] `packages/jazz-tools/src/cli.ts` (around lines 428, 1120, 1224, 1276) is not a blocker under the greenfield assumption, because new projects do not need to preserve any pre-PR snapshot history. The remaining issue is narrower: once a project uses the new historical recovery flow, recovered schemas are written into the same timestamp-ordered snapshot pool that plain `jazz-tools migrations create` uses as its implicit baseline, so a recovery export can later cause the next no-flags migration to diff from an old historical schema instead of the latest local schema. If historical recovery is intentionally provisional for now, this can be deferred.

2. [warning] `packages/jazz-tools/src/cli.ts` (around lines 337, 398, 1174, 1237) splits snapshot storage across two locations. `schema export --schema-hash` always persists into `schemaDir/migrations/snapshots`, while `createMigration()` only reads `options.migrationsDir/snapshots`. In any repo using `--migrations-dir`, the advertised offline recovery flow stops working: export succeeds, then `migrations create --fromHash ...` still has to hit the server.

3. [warning] `crates/jazz-tools/src/sync_manager/mod.rs` (around line 122), `packages/jazz-tools/src/runtime/sync-transport.ts` (around line 114), and `packages/jazz-tools/bin/docs-index.txt` (around line 3970) still tell users to run `migrations create --fromHash ... --toHash <targetHash>`. That brings back manual target-hash discovery in the exact workflow this PR is trying to remove; the normal recovery path should be able to stop at `--fromHash` and default the target to the current local schema.

4. [warning] `TODO.md` is stale after deleting `todo/issues/lens-transform-silent-degradation.md`. This repo regenerates `TODO.md` from `todo/` via `scripts/update-todo.sh`, and the deleted issue is still listed there.

nrainhart:
> 3. [warning] crates/jazz-tools/src/sync_manager/mod.rs (around line 122), packages/jazz-tools/src/runtime/sync-transport.ts (around line 114), and packages/jazz-tools/bin/docs-index.txt (around line 3970) still tell users to run migrations create --fromHash ... --toHash <targetHash>. That brings back manual target-hash discovery in the exact workflow this PR is trying to remove; the normal recovery path should be able to stop at --fromHash and default the target to the current local schema.

This is expected. When there are state disconnected schemas, the user won't necessarily want to connect them to the current schema. I don't think it makes sense to design a heuristic for this case, so I just left it up to the user to decide.

PR #475, opened 2026-04-07
Add From impls on Value and row_input! macro

joeinnes:
## Summary

- Add `From<T>` impls on `Value` for all common types (`bool`, `i32`, `i64`, `f64`, `&str`, `String`, `ObjectId`, `Vec<u8>`, `Vec<Value>`, `Option<T>`)
- Add `row_input!` macro for ergonomic `HashMap<String, Value>` construction with heterogeneous types
- Migrate all test and example call sites, deleting duplicate helpers and verbose `HashMap::from` / `map/unwrap_or` boilerplate (~250 lines removed)

## Test plan

- [x] 17 unit tests for all `From` impls and `row_input!` macro (red-green TDD)
- [x] 894 jazz-tools lib tests pass
- [x] todo-server and todo-server-docs examples compile and pass
- [x] clippy and rustfmt clean

### DIFF
--- a/crates/jazz-tools/tests/policies_integration/complex_policies.rs
+++ b/crates/jazz-tools/tests/policies_integration/complex_policies.rs
@@ -32,7 +32,7 @@ fn row_changes<const N: usize>(pairs: [(&str, Value); N]) -> Vec<(String, Value)
 }
 
 fn title_document_values(title: &str) -> Vec<Value> {
-    vec![Value::Text(title.to_string())]
+    vec![title.into()]
 }
 
 fn complex_document_values(
@@ -42,9 +42,9 @@ fn complex_document_values(
     folder_id: Option<ObjectId>,
 ) -> Vec<Value> {
     vec![
-        Value::Text(team_slug.to_string()),
-        Value::Boolean(published),
-        Value::Text(title.to_string()),
+        team_slug.into(),
+        published.into(),
+        title.into(),
         folder_id.into(),
     ]
 }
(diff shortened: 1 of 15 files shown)


crates/jazz-tools/tests/policies_integration/complex_policies.rs
```diff
@@ -53,7 +45,7 @@ fn complex_document_values(
         Value::Text(team_slug.to_string()),
         Value::Boolean(published),
         Value::Text(title.to_string()),
-        folder_id.map(Value::Uuid).unwrap_or(Value::Null),
+        folder_id.into(),
```
nrainhart, line 48: You could also use into in all of the above elements, right?

crates/jazz-tools/tests/policies_integration/inherited_policies.rs
nrainhart: You could just use the macro here

nrainhart: Ditto

nrainhart (approved):
Nice! Love how the macro reduces visual clutter.

This also gave me an excuse to learn a bit about macros, so thanks! 😝

crates/jazz-tools/tests/policies_integration/complex_policies.rs
joeinnes: Good idea!

antoniomuso (approved):
Nice Idea! LGTM!

gdorsi (approved):

PR #480, opened 2026-04-07
Add creator policy shorthands for permissions DSL

Takeno:
## Summary
- add `policy.<table>.managedByCreator()` as shorthand for creator-scoped read/insert/update/delete policies
- expose `isCreator` in `definePermissions(...)` so creator checks can be reused inside `anyOf(...)` and other composed rules
- add DSL, type inference, integration, and docs/example coverage for the new shorthand APIs
- document the shorthand examples separately from the composable `isCreator` example
- include unrelated `pnpm-lock.yaml` cleanup for the excluded `examples/world-tour` workspace


gdorsi (approved):
Nice!

PR #481, opened 2026-04-07
chore: remove project skill, clean up CLAUDE.md

gdorsi:
Noticed that having a standard for projects doesn't bring much value, as it's better to leave everyone to their personal processes there without prescripting too much.

Also spec/project content becomes stale quite quickly (often even in the same PR where the spec has been created) so keeping them around is not much useful.

Issue/idea instead are quite useful to capture a need or an idea without too much structure, so we can easily work on it when the time is right without the risk of the content becoming stale easily.


CLAUDE.md
```diff
@@ -14,92 +14,50 @@ Communicate tersely without losing precision or warmth.
...
-Concretely:
-
-- Intermediate breakage during complex work is fine. Don't add shims or workarounds to make each micro-step compile.
-- Never weaken assertions, add `#[cfg(test)]` backdoors, or explain away failures to reach "done." If a test fails, that's information — investigate or leave it red.
-- Ask before moving to the next phase. "It compiles" is not "it works."
-- It's OK to stop mid-plan with a clear summary. Don't rush to a false finish.
-
-## After hard problems
-
-When something was harder than expected, pause and reflect 5-whys-style: where did the difficulty actually come from? Was it a wrong assumption, a missing spec, an architectural gap? Write it down (in memory or as a spec update) so we don't repeat it.
```
gdorsi, line 45: Don't think that these instructions are needed anymore, LLMs got way better

PR #482, opened 2026-04-07
Run native benchmarks on RocksDB

aeplay:
## What changed
- switched the native realistic example runner to `--features client,rocksdb`
- switched the native Criterion runner and workflow prebuild step to `--features rocksdb`
- renamed the R3 cold-load benchmark catalog/history mapping from Fjall to RocksDB
- added a regression test that locks the native benchmark catalog, cargo commands, and workflow prebuild flags to RocksDB

## Why
The native benchmark plumbing had drifted after the backend switch. The realistic example runner still built with only `client`, which meant `ClientStorage::Persistent` did not actually use RocksDB there, and the Criterion cold-load slot plus workflow still referenced Fjall.

## Impact
Native benchmark runs and the benchmark workflow now target RocksDB consistently, and the cold-load benchmark metadata is labeled and mapped to the RocksDB-backed group.

## Checks
- `node --test benchmarks/realistic/ci_benchmarks.test.mjs`
- started `cargo run -p jazz-tools --features client,rocksdb --example realistic_bench -- --help`
- started `cargo bench -p jazz-tools --features rocksdb --bench realistic_phase1 --no-run`

The RocksDB cargo commands entered `librocksdb-sys` native compilation locally, but I did not wait for full completion before opening this draft.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -96,8 +96,8 @@ jobs:
       - name: Build native benchmark binaries
         run: |
           set -euo pipefail
-          cargo build --release -p jazz-tools --features client --example realistic_bench
-          cargo bench -p jazz-tools --bench realistic_phase1 --no-run
+          cargo build --release -p jazz-tools --features client,rocksdb --example realistic_bench
+          cargo bench -p jazz-tools --features rocksdb --bench realistic_phase1 --no-run
 
       - name: Run native benchmark suite
         run: |
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -222,6 +222,7 @@ jobs:
 
       - run: corepack enable
       - run: pnpm install --frozen-lockfile
+      - run: pnpm --filter jazz-napi build
       - run: pnpm exec playwright install chromium
 
       - name: Collect host metadata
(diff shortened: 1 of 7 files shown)


aeplay:
Follow-up for the browser benchmark failures:

The browser harness starts a Node-side `TestingServer` from `packages/jazz-tools/tests/browser/testing-server-node.ts`, and that host dependency comes from `jazz-napi`. Regular CI already builds `jazz-napi` via `pnpm build:ci`, but the benchmark workflow went from `pnpm install` straight to Playwright/browser benchmarks, so the native binding was never built there.

This branch now:
- adds `pnpm --filter jazz-napi build` to the browser benchmark job
- documents the same requirement in the realistic benchmark README
- wraps the `jazz-napi` load in `testing-server-node.ts` so failures point at `pnpm --filter jazz-napi build` instead of the generic npm optional-deps message
- locks the workflow step in with a regression test

Local checks:
- `node --test benchmarks/realistic/ci_benchmarks.test.mjs`
- `pnpm --dir packages/jazz-tools exec tsx -e "import('./tests/browser/testing-server-node.ts').then(() => console.log('testing-server-node ok'))"`
- `pnpm --dir packages/jazz-tools exec node -e "require('jazz-napi'); console.log('jazz-napi ok')"`

gdorsi (approved):

PR #488, opened 2026-04-08
Vite plugin for zero-config Jazz dev server

gdorsi:
## Summary

Adds a `jazzPlugin()` Vite plugin that eliminates manual dev server setup.

**DX choices:**
- Server binds to **port 0**
- **appId is auto-generated** if not provided — temporary until the setup flow lands
- Server options (`dataDir`, `inMemory`, `allowAnonymous`, `jwksUrl`, etc.) **mirror the CLI flags**
- **SQLite storage by default** — preferred over RocksDB because for dev it's more simple
- Schema and permissions are **pushed on startup and re-pushed on file change** via fs watcher
- Permissions build errors are **surfaced in the browser error overlay** using Vite's WebSocket API — thanks to Matteo for the suggestion!
- `JAZZ_APP_ID` and `JAZZ_SERVER_URL` are **injected as env vars** automatically — app code reads them without any `.env` file
- If `JAZZ_SERVER_URL` is already set in `.env`, the plugin **skips starting a local server** and uses that URL instead; `JAZZ_APP_ID` from `.env` is also respected when present

**Other changes:**
- Expose `DevServer` in `jazz-napi` (extracted shared `JazzHostedServer` from `TestingServer`) so the plugin runs native Rust, not a subprocess
- Migrated two apps (`auth-simple-chat`, `todo-react`) to the plugin — remaining apps and Next.js support in follow-up PRs


stress-tests/todo-react/src/App.tsx
```diff
@@ -91,11 +63,31 @@ function Router() {
...
+
+  if (!appId) {
+    throw new Error("JAZZ_APP_ID is required");
+  }
+
+  if (!serverUrl) {
+    throw new Error("JAZZ_SERVER_URL is required");
+  }
+
+  const active = getActiveSyntheticAuth(appId, { defaultMode: "demo" });
```
nrainhart: Should we React.memo this?

examples/auth-simple-chat/package.json
```diff
@@ -5,10 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "dev:auth": "tsx server/main.ts",
+    "dev:auth": "node server/main.ts",
```
nrainhart, line 8: What's the reason for making this change? Also, is this why we needed to use .ts extensions in imports now?

packages/jazz-tools/src/dev/vite.ts
```diff
@@ -0,0 +1,160 @@
...
+  server?: boolean | string | JazzServerOptions;
+  adminSecret?: string;
+  schemaDir?: string;
+  appId?: string;
+}
+
+const DEFAULT_PORT = 0;
+const LOG_PREFIX = "[jazz]";
+
+interface ViteDevServer {
```
nrainhart, line 33: Vite does not export this type?

packages/jazz-tools/src/dev/vite.ts
```diff
@@ -0,0 +1,160 @@
...
+      if (serverOpt === false) {
+        return;
+      }
+
+      if (typeof serverOpt === "string") {
+        serverUrl = serverOpt;
+        adminSecret = options.adminSecret ?? "";
+        if (!adminSecret) {
+          throw new Error(
+            `${LOG_PREFIX} adminSecret is required when connecting to an existing server`,
```
nrainhart, line 85: Shouldn't an appId be required in this case as well?

nrainhart (approved):
Looks reasonable from looking at the code. I'll try it out tomorrow!

packages/jazz-tools/src/dev/vite.ts
gdorsi: We don't have the Vite dependency in jazz-tools, and I think it's better to not have it.

examples/auth-simple-chat/package.json
gdorsi: One less dependency, nodejs already handles typescript.

The ts ext is due to how Node.js handles ts files import.

stress-tests/todo-react/src/App.tsx
gdorsi: Moved it outside of the component

PR #496, opened 2026-04-09
Enforce deny-by-default row policies across runtime and examples

Takeno:
## Summary
- make session-scoped row reads and writes fail closed when explicit policies are missing
- update inherited policy evaluation and server-side permission checks to deny when parent or current policies are absent
- add runtime and CLI coverage for deny-by-default behavior, including validation warnings for missing explicit policies
- update example permission bundles to declare explicit allow or deny policies under the new model

## Testing
- `pnpm --filter todo-client-localfirst-ts-docs test -- tests/browser/files-and-blobs-snippets.test.ts`
- validated example roots with `pnpm --filter jazz-tools exec tsx src/cli.ts validate --schema-dir ...`


aeplay (approved):

Takeno:
Closed in favor of #513

PR #501, opened 2026-04-09
feat: log error/warning for missing migrations when connecting to server

nrainhart:
## Description

When a client connects to a sync server, the client's schema may not be able to interoperate with the schema currently used for permissions enforcement by the server, or that server may know schemas (and rows) that are unreachable from the client's schema. Today that state is mostly invisible until reads quietly return less data than expected or writes fail later under policy evaluation. The result is user confusion: the connection succeeds, but the app behaves as if data is missing or permissions are broken without a clear explanation.

This PR adds a connection-time log making this situation clearer:
- if the client's schema is disconnected from the permissions schema, the client logs an error, because reads and writes are effectively blocked
- if the server knows additional unreachable schemas, the client logs a warning so developers understand why some historical data may stay invisible until migrations are published.

### Error

error

### Warning

warning

### DIFF
--- a/crates/jazz-tools/src/schema_manager/manager.rs
+++ b/crates/jazz-tools/src/schema_manager/manager.rs
@@ -7,7 +7,10 @@
 //! - Integrated QueryManager for query/insert/update/delete operations
 //! - Catalogue persistence for schema/lens discovery via sync
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet, VecDeque},
+    sync::Arc,
+};
 
 use blake3::Hasher;
 
@@ -23,7 +26,7 @@ use crate::query_manager::writes::{RowBranchDelete, RowBranchWrite};
 use crate::row_format::decode_row;
 use crate::schema_manager::rehydrate::latest_catalogue_content;
 use crate::storage::Storage;
-use crate::sync_manager::SyncManager;
+use crate::sync_manager::{ConnectionSchemaDiagnostics, SyncManager};
 use uuid::Uuid;
 
 use super::auto_lens::generate_lens;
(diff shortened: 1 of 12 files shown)


packages/jazz-tools/src/runtime/sync-transport.ts
```diff
@@ -823,7 +810,6 @@ export async function readBinaryFrames(
             event.next_sync_seq ?? null,
           );
         } else if (event.type === "SyncUpdate") {
-          logSchemaWarningPayload(event.payload, logPrefix);
```
nrainhart, line 846: Row-level warnings are not working right now, but they were also duplicated in JS and Rust

gdorsi (approved):

crates/jazz-tools/src/schema_manager/manager.rs
```diff
@@ -539,6 +542,68 @@ impl SchemaManager {
...
+            .filter(|hash| !reachable_hashes.contains(hash))
+            .filter(|hash| Some(*hash) != disconnected_permissions_schema_hash)
+            .collect();
+        unreachable_schema_hashes.sort_by(|left, right| left.as_bytes().cmp(right.as_bytes()));
+
+        ConnectionSchemaDiagnostics {
+            client_schema_hash,
+            disconnected_permissions_schema_hash,
+            unreachable_schema_hashes,
+        }
```
gdorsi, line 574: Would it be possible to add the tabkle name to the diagnostic?

It would be useful to help people understand where the problem is

nrainhart: What table name do you mean exactly? I mean we _could_ fetch both schemas and diff them, but I feel like the output would be super verbose for a console warning/error. And folks can do it themselves by creating the migration

PR #503, opened 2026-04-10
fix: update WASM URL handling to being used in next.js

Takeno:
#478 changed how WASM modules are loaded. In Next.js, during development, it caused them to be loaded from the root instead of the static assets directory

## Testing
- `cd examples/nextjs-csr-ssr && pnpm run test:e2e`

### DIFF
--- a/packages/jazz-tools/src/runtime/client.browser-assets.test.ts
+++ b/packages/jazz-tools/src/runtime/client.browser-assets.test.ts
@@ -104,7 +104,10 @@ describe("loadWasmModule runtimeSources bootstrap", () => {
     });
   });
 
-  it("uses an HTTP wasm URL when the runtime module itself is loaded from file://", async () => {
+  it("lets wasm-bindgen self-resolve the URL when the page is web-hosted and module is file://", async () => {
+    // Covers bundlers (Turbopack, webpack) that compile import.meta.url of client.ts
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/src/runtime/client.browser-assets.test.ts
```diff
@@ -113,8 +116,6 @@ describe("loadWasmModule runtimeSources bootstrap", () => {
     await loadWasmModule();
 
     expect(wasmDefaultInit).toHaveBeenCalledTimes(1);
-    expect(wasmDefaultInit).toHaveBeenCalledWith({
-      module_or_path: "http://localhost:3000/jazz_wasm_bg.wasm",
-    });
+    expect(wasmDefaultInit).toHaveBeenCalledWith();
```
nrainhart, line 119: Is this explicitly asserting init is called without params?

nrainhart (approved):

packages/jazz-tools/src/runtime/client.browser-assets.test.ts
Takeno: Yes, it is. Because those cases now follow the default import path. Instead, the worker now follows another path (resolveWorkerBootstrapWasmUrl). Next.js was receiving `http://localhost:3000/jazz_wasm_bg.wasm`, instead of resolving the correct _next/static path.

PR #508, opened 2026-04-10
feat: add local-first auth flow

gdorsi:
## Self-signed auth + signup continuity

Replace the identity linking system with a self-signed JWT flow that gives every user an identity from the moment they open the app — no sign-up required.

### How it works

**Anonymous-first sessions.** The client generates a random seed on first visit and derives a stable `userId` from it. A self-signed JWT (Ed25519) proves ownership of that identity to the server. The user can interact with the app immediately.

**Signup preserves identity.** When a user signs up (e.g. via Better Auth), the client mints a short-lived proof token from the same seed. The auth server validates the token and sets the proved `userId` as the new account's id. The user keeps all their data — no migration, no linking.

### User-facing API vocabulary

| Term | What it is |
|------|-----------|
| `auth.seed` | Config key on `createDb` / `DbConfig` — pass a seed to opt into self-signed sessions with automatic token refresh |
| `AuthSecretStore` | Interface for seed persistence. `LocalStorageAuthSecretStore` is the bundled implementation; apps can supply their own |
| `db.getSelfSignedToken({ ttlSeconds, audience })` | Mints a short-lived Ed25519 proof token tied to the current identity. Returns `null` when the session isn't self-signed |
| `verifySelfSignedToken(token, audience)` | Server-side NAPI binding — validates a proof token and returns `{ ok, id }` |
| `allowSelfSigned` | Per-app server flag — whether to accept self-signed JWT sessions (default: `true`) |

### Setting up a self-signed session

Full example in [`stress-tests/todo-react/src/App.tsx`](https://github.com/garden-co/jazz2/blob/feat/auth/stress-tests/todo-react/src/App.tsx).

```ts
const authSecretStore = new LocalStorageAuthSecretStore();
const secret = await authSecretStore.getOrCreateSecret();

const config: DbConfig = {
  appId,
  serverUrl,
  auth: { seed: secret },  // enables self-signed mode
};
```

That's the whole client setup. The seed is persisted across reloads; `userId` is derived deterministically from it.

### Wiring signup continuity with Better Auth

Full server setup in [`examples/auth-betterauth-chat/src/lib/auth.ts`](https://github.com/garden-co/jazz2/blob/feat/auth/examples/auth-betterauth-chat/src/lib/auth.ts).

**Client — mint a proof token before sign-up:**

```ts
const proofToken = await db.getSelfSignedToken({
  ttlSeconds: 60,
  audience: "betterauth-signup",
});

await authClient.signUp.email({ email, name, password, proofToken });
```

**Server — validate and pin the userId:**

```ts
hooks: {
  before: createAuthMiddleware(async (ctx) => {
    if (ctx.path !== "/sign-up/email") return;
    const { verifySelfSignedToken } = await import("jazz-napi");
    const { ok, id: provedUserId } = verifySelfSignedToken(ctx.body.proofToken, "betterauth-signup");
    if (!ok) throw new APIError("UNAUTHORIZED", { message: "Invalid proof token" });
    return { context: { ...ctx, body: { ...ctx.body, provedUserId } } };
  }),
},
databaseHooks: {
  user: {
    create: {
      before: async (user, ctx) => {
        const provedUserId = ctx?.body?.provedUserId;
        if (provedUserId) return { data: { ...user, id: provedUserId } };
      },
    },
  },
},
```

### Design choices

**Seed generation and storage live outside the DB.** `AuthSecretStore` is an interface — the bundled `LocalStorageAuthSecretStore` is one implementation, but apps can persist the seed however they want (secure enclave, server-side, etc). The DB never owns the seed lifecycle.

**Seed → JWT lives inside the DB.** `db.getSelfSignedToken()` and the automatic token refresh are cross-runtime (WASM + NAPI) because the JWT minting depends on the Ed25519 signing key derived from the seed, which is implemented in Rust. Packaging this inside the DB means it works identically in browsers, Node, and React Native without the app wiring up crypto.

**No batteries-included auth flow yet.** Unlike Jazz1's opinionated auth, this branch exposes composable primitives: `auth.seed` config, `getSelfSignedToken()`, `verifySelfSignedToken()`. The Better Auth example shows how to wire these together, but the ceremony is still on the app developer. A higher-level `useAuth()` hook that handles the full anonymous → signup → authenticated transition is a natural next step, but we want to validate the primitives first.

### What's in the branch

- **Identity crypto module** — Ed25519 key derivation, UUIDv5 userId, self-signed JWT mint/verify
- **`auth.seed` config path** — pass a seed to `createDb` for self-signed sessions with automatic token refresh
- **`AuthSecretStore`** — interface + `LocalStorageAuthSecretStore` for seed persistence
- **`db.getSelfSignedToken()`** — mint proof tokens with configurable TTL and audience
- **`verifySelfSignedToken` NAPI binding** — server-side token verification
- **Better Auth signup continuity** — hooks.before validates proof, databaseHooks sets userId
- **Server support** — `allow_self_signed` flag on server builder (default: true)
- **Removed** — identity linking system (`useLinkExternalIdentity`, external identity store, link routes)

### Testing

- Rust unit tests for identity crypto (mint, verify, audience, TTL, auth_mode rejection)
- Integration tests for self-signed auth against the cloud server
- Vitest tests for `getSelfSignedToken` (self-signed returns token, non-self-signed returns null)
- Playwright E2E tests for signup continuity and token rejection


aeplay (changes requested):
The mechanism looks great to me, but I'd like the final API to be easier to use without understanding/exposing internals, mostly by naming concepts not according to their cryptographic details, but by their "purpose". I also think we can skip some wiring-up steps by wrapping your API and introducing defaults.

rough suggestion:

```ts
const config: DbConfig = {
  appId,
  serverUrl,
  auth: { asDevice: BrowserDeviceAuth }
};
```

```ts
await authClient.signUp.email({
  email,
  name,
  password,
  linkDevice: BrowserDeviceAuth.linkToken("betterauth-signup")
});
```

Everything under "Server — validate and pin the userId" is just part of our default plugin config.

aeplay (approved):
The nicer API can be done in a followup

PR #511, opened 2026-04-12
Handle no-op schema hash changes in migrations

aeplay:
## What changed

This updates the TypeScript migration CLI to recognize a narrow class of schema-hash changes that do not require row transformations and therefore should not force developers to create or hand-edit migration files.

It also fixes migration loading for CommonJS-ish module interop so valid `export default defineMigration(...)` files do not get rejected when the runtime import shape is nested under repeated `default` wrappers.

## Why it changed

We want a first-class concept of "hash changed, but no reviewed migration file is needed" before order-dependent schema hashes land.

The immediate bug that prompted this was a default-only schema change producing a new hash while the generated migration said there were no differences, plus a separate report where a valid migration file failed with:

`Invalid migration export in ... Export default defineMigration(...).`

## User and developer impact

- Default-only schema changes no longer force a reviewed migration file.
- Pure column reordering can follow the same no-op path.
- Nullability-only and reference-only changes still require reviewed migration files.
- `migrations push` can publish an empty lens when a hash-only/no-op transition has no migration file.
- Existing empty migration files are accepted for those no-op transitions.
- CommonJS-compiled migration modules now load correctly.

## Root cause

Two issues were involved:

1. The CLI had no strong notion of a no-op schema transition, so some hash changes still pushed developers into the migration-file workflow even when no row transform was needed.
2. The migration loader only unwrapped one level of `default` export interop, so CommonJS-ish module loading could hide a valid migration object behind `default.default`.

## Validation

- `pnpm --filter jazz-wasm build`
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/cli.test.ts`


nrainhart (approved):
I'm somewhat worried that having some changes that require a migration and some others that don't will confuse users, but for now I'm ok with trying to reduce the amount of migrations users need to write.

PR #521, opened 2026-04-14
test(jazz-tools): local-first auth integration suite

gdorsi:
## Summary

- Adds `crates/jazz-tools/tests/local_first_auth_integration.rs`, an end-to-end suite that drives a real `JazzClient` through `TestingServer` using Ed25519 seed-derived Bearer tokens.
- Covers the gap between `local_first_auth.rs` (middleware-only `extract_session` unit tests) and `auth_test.rs` (HTTP-level JWT/backend/admin coverage). Nothing currently exercises local-first auth through the client stack.
- Threads a new `with_local_first_auth(bool)` passthrough on `TestingServerBuilder` so the rejection scenario can flip the flag without duplicating the test-server plumbing. Default behavior is unchanged.

## Scenarios

- `same_seed_syncs_across_devices` — Alice's seed on two memory-backed clients; device B converges on device A's write purely via server-side principal recognition.
- `different_seeds_produce_distinct_principals` — Alice and Bob derive distinct user IDs; both clients converge on both rows.
- `persistent_seed_reconnects_as_same_principal` — persistent storage + same seed: reconnecting reads local state and still authenticates at EdgeServer durability.
- `server_with_local_first_disabled_rejects` — HTTP `POST /sync` with a valid Ed25519 Bearer token returns `401` when the server is built with `with_local_first_auth(false)`.

## Test plan

- [x] `cargo check -p jazz-tools --features test --tests`
- [x] `cargo test -p jazz-tools --features test --test local_first_auth_integration` — 4/4 pass (~0.04s)
- [x] CI: clippy + rustfmt + full crate test run

### DIFF
--- a/crates/jazz-tools/tests/local_first_auth_integration.rs
+++ b/crates/jazz-tools/tests/local_first_auth_integration.rs
@@ -1,0 +1,325 @@
+#![cfg(feature = "test")]
+
+//! End-to-end integration coverage for local-first (Ed25519 seed) auth.
+//!
+//! The in-process `TestingServer` accepts local-first Bearer tokens by default.
+//! These tests prove that a `JazzClient` carrying such a token syncs correctly,
+//! that seeds map deterministically to principals across devices, that state
+//! persists across reconnects, and that a server configured with local-first
+//! auth disabled rejects the same tokens.
+//!
+//! Narrow middleware-level assertions (wrong audience, expired token, disabled
+//! flag at the `extract_session` layer) live in `local_first_auth.rs`.
+
+mod support;
+
+use std::collections::HashMap;
+
+use jazz_tools::commit::CommitId;
+use jazz_tools::jazz_transport::SyncBatchRequest;
+use jazz_tools::metadata::RowProvenance;
+use jazz_tools::row_histories::{RowState, StoredRowVersion};
+use jazz_tools::server::TestingServer;
+use jazz_tools::sync_manager::{ClientId, SyncPayload};
+use jazz_tools::{
+    AppContext, ClientStorage, ColumnType, JazzClient, ObjectId, QueryBuilder, Schema,
+    SchemaBuilder, TableSchema, Value, identity,
+};
+use reqwest::StatusCode;
+use uuid::Uuid;
+
+use support::{has_row, wait_for_rows};
+
+const TOKEN_TTL_SECS: u64 = 3600;
+
+fn alice_seed() -> [u8; 32] {
+    let mut seed = [0u8; 32];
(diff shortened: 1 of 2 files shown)


crates/jazz-tools/tests/local_first_auth_integration.rs
```diff
@@ -0,0 +1,451 @@
+#![cfg(feature = "test")]
+
+//! End-to-end integration coverage for local-first (Ed25519 seed) auth.
+//!
+//! These tests prove that a `JazzClient` carrying an Ed25519-minted Bearer
+//! token syncs correctly, that seeds map deterministically to principals
+//! across devices, and that state persists across reconnects.
+//!
+//! Narrow middleware-level assertions (wrong audience, expired token,
+//! disabled flag) live in `local_first_auth.rs` and `src/middleware/auth.rs`.
```
nrainhart, line 10: These comments are gold, thanks for adding them 🫶

crates/jazz-tools/tests/local_first_auth_integration.rs
```diff
@@ -0,0 +1,451 @@
...
+    ctx.admin_secret = None;
+    ctx.storage = storage;
+    ctx
+}
+
+fn todo_values(title: &str, completed: bool) -> HashMap<String, Value> {
+    HashMap::from([
+        ("title".to_string(), Value::Text(title.to_string())),
+        ("completed".to_string(), Value::Boolean(completed)),
+    ])
```
nrainhart, line 76: Nit: you can use the `row_input!` macro Joe added to make this more concise

crates/jazz-tools/tests/local_first_auth_integration.rs
```diff
@@ -0,0 +1,451 @@
...
+    wait_for_rows(
+        &client,
+        QueryBuilder::new("todos").build(),
+        "pre-expiry todo settles at edge server",
+        |rows| has_row(&rows, pre_id, &pre_values).then_some(()),
+    )
+    .await;
+
+    // Let the token lapse. 1s TTL + 1500ms yields a definitively-expired token.
+    tokio::time::sleep(Duration::from_millis(1500)).await;
```
nrainhart, line 415: This is an actual 1.5s wait right? We're not using fake timers

nrainhart (approved):
The tests are super clear! 🚀

crates/jazz-tools/tests/local_first_auth_integration.rs
gdorsi: Yeah, you are tight.

Looking into adding a fake timer here.

PR #522, opened 2026-04-14
feat(jazz-tools): add jazzSvelteKit() Vite plugin for SvelteKit projects

joeinnes:
## Summary

- Adds `jazzSvelteKit()` — a new Vite plugin exported from `jazz-tools/dev/sveltekit` that manages the Jazz dev-server lifecycle for SvelteKit (and Vite+Svelte) projects. It mirrors the existing `jazzPlugin()` for vanilla Vite but adapts defaults and env-var shape to SvelteKit conventions: `src/lib/` for schema/permissions, `PUBLIC_` prefix for client-bound env, `BACKEND_SECRET` injection for server-side code, and JWKS URL derived from Vite's configured host/port.
- Supports three connection modes out of the box: embedded local Jazz server (default), connect-by-URL via `server: "https://…"`, or connect to a server described in `PUBLIC_JAZZ_SERVER_URL` / `PUBLIC_JAZZ_APP_ID` env vars.
- Extends `schema-loader.ts` to resolve `src/lib/schema.ts` as a third candidate location (alongside `<root>/schema.ts` and `<root>/src/schema.ts`), and throws "Ambiguous schema location" if multiple candidates exist simultaneously so `jazz-tools validate` and friends don't silently pick the wrong file.
- Migrates `examples/todo-client-localfirst-svelte` to use the new plugin: schema and permissions move into `src/lib/`, the hardcoded app-id fallback is removed, and the example now reads `PUBLIC_JAZZ_APP_ID` / `PUBLIC_JAZZ_SERVER_URL` from the injected Vite env.
- Updates docs (`quickstarts/client.mdx`, `quickstarts/typescript-server.mdx`, `auth/permissions.mdx`) to note the SvelteKit convention for `schema.ts` / `permissions.ts` living in `src/lib/`.

## Design notes

- **Plugin state is per-instance.** All runtime state (server handle, watcher, runtime, init promise) lives in a closure returned by `jazzSvelteKit()`. Each call produces an independent plugin, with no cross-instance coordination. The only module-level state is a shared shutdown-hook registry (`activeDisposers` + a one-time SIGINT/SIGTERM handler) that keeps the process listener count bounded at 3 regardless of how many plugins are active.
- **Shared types with `vite.ts`.** `JazzServerOptions` and `JazzPluginOptions` `extend` the base types exported from `./vite.js`; `ViteDevServer` is exported from `vite.ts` and imported here so the shape is maintained in one place. The only SvelteKit-specific addition is `backendSecret`, which exists here (and not on `vite.ts`) because it injects into `process.env.BACKEND_SECRET` for server-side SvelteKit code to read — a Vite+React SPA has no server side and nothing to consume it.
- **JWKS URL threads through Vite's config.** Precedence: explicit `serverConfig.jwksUrl` → `process.env.APP_ORIGIN` → `viteServer.config.server.{host,port,https}` → `http://localhost:5173`. Known limitation: if Vite auto-increments to a free port, the configured port still differs from the actual listening port; affected users can set `APP_ORIGIN` explicitly.
- **Errors surface via HMR.** Schema-push failures are reported via `viteServer.ws.send({ type: "error", err })` so they appear in the browser overlay, and also re-thrown to reject `configureServer` and tear down the partial runtime. Missing-`schema.ts` is treated as a warning (not an error) with a visible `console.warn`, since it's a common early-project state.

## Test plan

- [ ] `pnpm --filter jazz-tools test` — dev-dir suites (vite, next, sveltekit, schema-watcher, dev-server)
- [ ] `pnpm --filter jazz-tools vitest run src/dev/sveltekit.test.ts` — 13 plugin tests including real integration against an embedded server
- [ ] `pnpm --filter jazz-tools vitest run src/cli.test.ts` — including the new ambiguous-schema-location test
- [ ] `pnpm --filter todo-client-localfirst-svelte build` — example validates and builds via the new plugin
- [ ] `pnpm --filter todo-client-localfirst-svelte test` — 7 browser tests exercise the migrated `src/lib/` layout end-to-end
- [ ] Manually spin up `examples/todo-client-localfirst-svelte` via `pnpm dev` and confirm schema publish logs and HMR error-surfacing behave correctly


packages/jazz-tools/src/dev/sveltekit.ts
```diff
@@ -0,0 +1,284 @@
+import { randomUUID } from "node:crypto";
+import { join } from "node:path";
```
gdorsi, line 1: Centralized most of this logic in managed-runtime.ts

gdorsi (approved):
LGTM

gdorsi (approved):

PR #529, opened 2026-04-15
Transport protocol WS

antoniomuso:
## Summary

Replaces HTTP POST `/sync` + SSE `/events` with a single Rust-owned WebSocket `/ws` endpoint. Transport logic (framing, auth handshake, reconnect, heartbeat) moves from TypeScript into shared Rust.

## Motivation

- **Ordering**: HTTP POSTs arrived out of order, corrupting subscriber state. Two tests were `#[ignore]`d because of it.

- **Split ownership**: JS owned network I/O, Rust owned sync logic. Same code lived in `sync-transport.ts`, three Rust `SyncSender` shims, and a reqwest client.

## Architecture

```
┌────────────────────────────────────────────┐
│ RuntimeCore<S: Storage, Sch: Scheduler>    │
│  transport: Option<TransportHandle>        │
│  sync_sender: Option<Box<dyn SyncSender>>  │  ← server fanout + worker bridge
└──────────────┬──────────────▲──────────────┘
               │ outbox        │ inbound
               ▼ channel       │ channel
┌─────────────────────────────┴──────────────┐
│ TransportManager<W: StreamAdapter,         │
│                  T: TickNotifier>          │
│  - perform_auth_handshake                  │
│  - run_connected (tokio::select! loop)     │
│  - reconnect w/ exponential backoff        │
└────────────────────────────────────────────┘
```

**Invariants**
- `RuntimeCore` is fully synchronous (zero `.await`). Runs on whichever thread calls `batched_tick()`.
- `TransportManager` holds no `RuntimeCore` reference. Communicates only through three channels (outbox, inbound, control). No locks, no borrows.
- Wire format: `[4 bytes u32 BE length][N bytes JSON]`. Binary frames.

**Platform matrix**

| Platform | Stream adapter | Spawn |
|---|---|---|
| Native / Tokio | `NativeWsStream` (tokio-tungstenite + rustls) | `tokio::spawn` |
| NAPI | same | `Handle::try_current()` + fallback thread |
| React Native | same | dedicated thread + own runtime |
| Browser WASM | `WasmWsStream` (ws_stream_wasm) | `spawn_local` |

## Auth flow

Server precedence (`authenticate_ws_handshake`): JWT wins when present; otherwise admin_secret; otherwise backend_secret; otherwise session extraction.

On auth failure, the server sends `ServerEvent::Error { Unauthorized }`. The client:
1. Detects it in `perform_auth_handshake`
2. Pushes `TransportInbound::AuthFailure` and **parks** waiting for `TransportControl::UpdateAuth` (no busy-retry)
3. `batched_tick` invokes `RuntimeCore::auth_failure_callback` → TS `onAuthFailure`
4. `updateAuthToken(newJwt)` sends `TransportControl::UpdateAuth` → transport retries with new credentials

Heartbeat: server emits `ServerEvent::Heartbeat` every 30s via `tokio::time::interval` in the WS select loop.

## Schema push (unchanged contract)

In production, user-role clients cannot push schema over WS — the server denies with `CatalogueWriteDenied`. Canonical push path: `POST /admin/schemas` with `X-Jazz-Admin-Secret`. Matches `origin/main` behavior.

## What's deleted

- `transport.rs` (reqwest), `POST /sync` + `GET /events` handlers
- `NapiSyncSender`, `RnSyncSender`, `SyncMessageCallback` UniFFI trait
- `sync-transport.ts` network code, `setupSync` in `client.ts`, `server-payload-batcher.ts`
- `transport-http` and `transport` Cargo features

**Kept on purpose**: `SyncSender` trait (test spies), `JsSyncSender` (browser main↔worker `postMessage` bridge), `CallbackSyncSender` (server fan-out via `ConnectionEventHub`).

## Breaking changes

- `POST /sync` and `GET /events` routes deleted — 404 for external callers
- `RuntimeCore<S, Sch, Sy>` → `RuntimeCore<S, Sch>`
- `NapiSyncSender`, `RnSyncSender` deleted — use `runtime.connect(url, authJson)`
- `TokioRuntime::new` drops the 4th arg
- Cargo `transport` + `transport-http` features removed

## Status

| Suite | Result |
|---|---|
| `cargo test -p jazz-tools --lib --features transport-websocket` | 917 pass |
| `cargo test -p jazz-tools --all-features` | ~1175 pass (4 pre-existing `sync_tracer_integration` fails match main) |
| `pnpm --filter jazz-tools test` | 70 files pass |
| `pnpm --filter jazz-tools test:browser` | 12 files / 152 tests pass |
| `pnpm build:all` | 26 turbo tasks green |

Un-ignored: `subscription_reflects_final_state_after_rapid_bulk_updates`, `single_client_operations_reach_server_in_causal_order` — both passed after fixing a debounce race in `TokioScheduler::schedule_batched_tick` (flag was cleared after `batched_tick` instead of before, losing messages parked mid-tick).

## Follow-ups (not blockers)

- CI `expo-android-maestro-e2e` fails, I suspect the Vercel Sandbox proxy doesn't support ws.
- `PostMessageStream` adapter would unify the worker bridge under `StreamAdapter` and delete `sync_sender`.
- Reconnect outbox dedup not implemented.
- `tests/websocket_transport.rs` (concurrent writers, reconnect resume, auth failure) still to author.

## Review focus

1. **Rust core** — `transport_manager.rs`, `install_transport` factory, `flush_outbox` destination routing in `ticks.rs`
2. **Server handler** — `routes.rs::handle_ws_connection`, auth precedence, `ConnectionEventHub` fan-out
3. **Platform bindings** — NAPI/RN/WASM `connect()` wrappers; auth failure + `updateAuth` wiring
4. **TS layer** — thinned `client.ts`, `worker-bridge.ts`, `jazz-worker.ts`
5. **Tests** — `auth_test.rs` migration, browser E2E coverage


aeplay (changes requested):
🤖

1. [critical] `packages/jazz-tools/src/runtime/db.ts:567` — direct, non-worker clients now default remote reads to `edge`, but they only kick off `connectTransport()` asynchronously at `:591-595` and there’s no replacement for the old remote-read wait gate. That matters because `QueryManager` treats `!has_servers()` as “frontier complete” (`crates/jazz-tools/src/query_manager/subscriptions.rs:127-129`), so a one-shot `db.query()`/`db.all()` issued during startup can resolve from stale local state before the WS handshake finishes.

2. [warning] `packages/jazz-tools/src/worker/jazz-worker.ts:251` — `performUpstreamConnect()` posts `upstream-connected` as soon as `runtime.connect()` returns, but the Rust transport connects asynchronously. The worker only posts `upstream-disconnected` on auth failure (`:313-315`) or explicit manual disconnect (`:513-514`), so ordinary connect failures / reconnect drops leave `WorkerBridge.waitForUpstreamServerConnection()` believing the upstream is live when it isn’t.

3. [warning] `packages/jazz-tools/src/runtime/client.ts:1344` — `connectTransport()` always calls `httpUrlToWs(url)` without the configured `serverPathPrefix`. Worker init fixed prefix handling, but direct browser/RN clients still dial root `/ws`, so `/apps/<id>` style deployments won’t attach upstream sync on those paths.

4. [warning] `crates/jazz-tools/src/transport_manager.rs:161` — the wasm backoff path computes `delay_ms` and then does `futures::future::ready(()).await` instead of a real timer. If `WasmWsStream::connect()` fails synchronously (bad URL, offline, immediate TLS reject), the reconnect loop busy-spins with effectively zero delay.

Reviewed current head `f19c55d`. I didn’t run the full suite locally.

antoniomuso:
Thanks I have verified with claude:

1. **[critical] `packages/jazz-tools/src/runtime/db.ts:567`** — confirmed, but **pre-existing on main**, not a regression from this PR. Checked `main:db.ts` and `main:subscriptions.rs`: `defaultDurabilityTier: "edge"`, the absent direct-path gate (`ensureQueryReady` at db.ts:611-619 only awaits `waitForUpstreamServerConnection()` on the worker-bridge path), and the `!has_servers()` frontier-complete check in `subscriptions.rs:127-129` are all unchanged. The `client.connectTransport(...)` added at :591-595 actually *narrows* the window — on main, edge queries on the direct path simply never synced upstream; on this branch they race the WS handshake. Folded into the existing `worker-upstream-connected-is-optimistic` issue (same root cause — no Rust-side connect-state signal). Real fix needs an `onConnectStateChanged` callback (or awaitable `connect()`) so both call sites can gate on it.

2. **[warning] `packages/jazz-tools/src/worker/jazz-worker.ts:251`** — confirmed. Already tracked at `todo/issues/worker-upstream-connected-is-optimistic.md`; same architectural gap as (1), shared fix.

3. **[warning] `packages/jazz-tools/client.ts:1344`** — confirmed. Prefix is plumbed into `JazzClient.connectSync` via `db.ts:560` but never reaches `connectTransport` at `db.ts:592`. Also noticed `db.transport.test.ts` DBRT-U01 locks in the buggy shape — passes `serverPathPrefix: "/apps/app"` but asserts the bare URL. New issue opened at `todo/issues/connect-transport-drops-server-path-prefix.md` (priority: high); test needs updating with the fix.

4. **[warning] `crates/jazz-tools/src/transport_manager.rs:161`** — confirmed. Already tracked at `todo/issues/wasm-transport-backoff-has-no-real-sleep.md` (priority: high). Fix: `gloo-timers` or `setTimeout` via `wasm-bindgen-futures`.

(1) is pre-existing; (2) and (4) were already on the follow-up list; (3) is now tracked.

aeplay (approved):

PR #538, opened 2026-04-16
[codex] add polymorphic contract table spec

aeplay:
## What changed

- added a new MVP TODO spec for polymorphic contract tables and references
- linked the new spec from the architecture TODO index in `specs/README.md`

## Why

We explored polymorphic tables/references and landed on a contract-plus-variant design that fits Jazz's migration-lens model better than ref-side `table_id + object_id` routing. This PR captures the proposed high-level API and the low-level implementation work needed across schema metadata, writes, relation planning, query lowering, result shaping, policies, and migrations.

## Impact

- gives the team a concrete design target for nominal `contract` / `variant` tables
- assumes the newly landed caller-supplied create/upsert ids on `main`
- keeps refs single-target by pointing them at contract tables rather than storing per-edge table identifiers

## Validation

- `git diff --check -- specs/README.md specs/todo/a_mvp/polymorphic_contract_tables_and_references.md`
- repo pre-commit hook ran `oxfmt` during commit
- no automated tests were run because this is a docs-only change

### DIFF
--- a/specs/README.md
+++ b/specs/README.md
@@ -77,6 +77,7 @@ If you want one sentence to hold onto while reading the rest:
 
 Design work that has not landed yet lives in [`specs/todo/`](todo/). Useful entry points:
 
+- **[Polymorphic Contract Tables and References](todo/a_mvp/polymorphic_contract_tables_and_references.md)** — nominal contract/variant tables with contract refs and coordinated variant writes
 - **[Opt-In Transactions with Replayable Reconciliation](todo/a_mvp/opt_in_transactions_replayable_reconciliation.md)** — transaction-shaped write semantics on top of the row-history foundation
 - **[Protocol and Storage Version Tags](todo/b_launch/protocol_and_storage_version_tags.md)** — explicit format/versioning strategy for on-disk state and wire payloads
 - **[Sharding Design Sketch](todo/b_launch/sharding_design_sketch.md)** — future distribution work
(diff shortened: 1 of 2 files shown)


Gabrola:
@aeplay I thought about this and experimented a little bit. I think allowing the contract and variant row IDs to share the same UUID is going to require a lot of underlying changes due to the fact that each ID -> unique storage row. I was thinking maybe using `uuidv5(tag, contract_row_id)` could be a good idea.

The other idea which I think is simpler in overall implementation would be to model the contract + variants to live in a single table with all the data living in a single row. The co-location of the data should make the implementation simpler, faster, and it gets rid of the problem of half-written or half-synced polymorphic rows. The only minor downside to this approach is that a ref to a specific variant by ID could actually be a different variant, so it will need to read/sync the row first before the discriminator value can be checked. 

It could end up looking something like this in terms of schema shape.

```ts

export const app = s.defineApp({
  ...s.polymorphic({
    entities: s.contract({
      ownerId: s.ref("users"),
      createdAt: s.timestamp(),
    }),
    posts: s.variant({
      tag: "post",
      columns: {
        title: s.string(),
        body: s.string(),
      },
    }),
    images: s.variant({
      tag: "image",
      columns: {
        url: s.string(),
        alt: s.string().optional(),
      },
    }),
    comments: s.variant({
      tag: "comment",
      columns: {
        body: s.string(),
      },
    }),
  }),
});
```

On a type-level this spreads into 4 virtual tables and it can be typed such that an `entities` row gets typed as a discriminated union with all columns resolvable just through type narrowing. A `ref` to any of the variant "tables" only resolves to the variant's columns + contract columns.

PR #540, opened 2026-04-17
fix(runtime): use static bundler-detected URL pattern for worker

joeinnes:
## What

Previously, `Db.spawnWorker` always resolved the worker URL dynamically at runtime — computing paths relative to `location.href` and appending a WASM URL query param. This works when `runtimeSources` provides an explicit `workerUrl` or `baseUrl`, but breaks in bundled apps (Next.js/Turbopack, Vite, webpack) where those tools need a static `new URL(...)` expression at compile time to detect and co-bundle the worker script and its WASM dependency.

When no `workerUrl` or `baseUrl` is provided, the worker is now spawned with:

```js
new Worker(new URL("../worker/jazz-worker.js", import.meta.url), { type: "module" })
```

The dynamic path is preserved for the explicit-runtimeSources case (CDN, custom deployments, etc.).

A computed `fallbackWasmUrl` is also threaded through the worker init message so non-bundled (static HTML) deployments still receive an explicit WASM path as a last resort if `wasmModule.default()` fails. Bundled contexts are unaffected — they always go via `wasmModule.default()` first.

## Why now

This was the root cause of worker loading failures in the Next.js/Turbopack starter — the worker and its WASM weren't being bundled because Turbopack couldn't statically analyse the URL.

## Test plan

- [x] Existing worker bootstrap tests pass
- [x] Updated test asserts static URL pattern (`/worker/jazz-worker\.js$/`) rather than the old dynamic URL with `?jazz-wasm-url=` appended
- [x] New test asserts `fallbackWasmUrl` is present in bridge options when no `runtimeSources` are provided

### DIFF
--- a/.changeset/worker-bundler-url-static.md
+++ b/.changeset/worker-bundler-url-static.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Use static `new URL(...)` import for worker when no explicit runtime sources are configured, allowing bundlers (Turbopack, webpack, Vite) to detect and co-bundle the worker script and its WASM dependency automatically.
+
+Also passes a computed `fallbackWasmUrl` in the worker init message so non-bundled (static HTML) deployments still receive an explicit WASM path as a last resort if `wasmModule.default()` fails.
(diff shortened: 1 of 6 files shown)


gdorsi (commented):
LGTM

PR #555, opened 2026-04-17
[codex] Isolate browser storage by app and user

aeplay:
## What changed

This isolates browser-local Jazz state by default instead of letting different principals reopen the same local storage namespace.

- scope the default browser persistent `dbName` to `appId` plus the resolved authenticated principal when callers do not provide an explicit `dbName`
- extend `BrowserAuthSecretStore` so callers can namespace stored secrets with `appId`, `userId`, and `sessionId`
- update browser examples and docs snippets to use app-scoped secret storage instead of a single global `jazz-auth-secret` key
- add a changeset plus focused tests for auth-secret scoping and default persistent namespace derivation
- add a browser regression test covering cross-user isolation and same-user continuity

## Why

The root issue was that browser persistence defaulted to app scope only. That meant a second session or different `user_id` could reopen OPFS state that belonged to the previous principal, and examples using the default `BrowserAuthSecretStore` could also share one global local-first secret across unrelated apps.

## Impact

Browser clients get safer default isolation across users while preserving continuity for the same authenticated principal. Apps that intentionally want shared persistence can still opt into that by passing an explicit `dbName` or custom storage key.

## Validation

- `pnpm lint`
- `cd packages/jazz-tools && pnpm exec vitest run --config vitest.config.ts src/runtime/auth-secret-store.test.ts src/runtime/db.browser-storage-scope.test.ts`
- `cd packages/jazz-tools && pnpm exec tsc --noEmit`

## Notes

I also added `packages/jazz-tools/tests/browser/db.storage-isolation.test.ts`, but I could not execute it locally because the existing browser Vitest harness currently fails earlier on a repo-wide `jazz-wasm` import-resolution issue (`Failed to resolve import "jazz-wasm" from "src/runtime/client.ts"`).


aeplay:
Superseded by #557. This draft was opened from a detached-HEAD ancestor and pulled in an oversized commit range, so I opened a clean replacement PR rooted on `main`.

PR #574, opened 2026-04-19
[codex] require transactions for selected tables

aeplay:
## What changed
This adds a schema-level transaction-required flag for specific tables and enforces it across the TypeScript schema APIs, runtime direct-write APIs, and Rust core write paths.

- add `requiresTransaction` schema metadata and schema encoding support
- add `defineTable(...).requireTransaction()` and `table(..., columns, { requiresTransaction: true })`
- reject direct writes and direct batches for flagged tables with a clear error
- keep transactional writes working normally for those same tables
- add TS and Rust regression coverage plus a `jazz-tools` changeset

## Why
Some tables represent side-effectful or invariant-sensitive data that should only be mutated transactionally. Before this change, the schema could not express that requirement, and direct update/delete paths could fall through to row lookup behavior instead of rejecting immediately.

## API
Two schema entry points now mark a table as transaction-required:

```ts
const auditLogs = defineTable({
  actor: text(),
  action: text(),
}).requireTransaction();
```

```ts
table(
  "audit_logs",
  {
    actor: text(),
    action: text(),
  },
  { requiresTransaction: true },
);
```

Behavior:

- direct writes now throw immediately for marked tables
- this includes `db.insert(...)`, `db.update(...)`, `db.delete(...)`, and direct batch entry points like `db.beginDirectBatch(...)`
- transactional writes still work for those tables
- tables that do not opt in keep existing behavior and can still participate in transactions

## Validation
- `cargo clippy --workspace -- -D warnings`
- `pnpm --filter jazz-tools exec vitest run src/dsl.test.ts src/typed-app.test.ts src/runtime/db.transaction.test.ts`
- `cargo test -p jazz-tools rc_transaction_required_tables_reject_direct_writes_but_allow_transactions -- --nocapture`

### DIFF
--- a/.changeset/transaction-required-tables.md
+++ b/.changeset/transaction-required-tables.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Add schema support for transaction-required tables in `jazz-tools`.
+
+You can now mark a table as requiring transactional writes with either `defineTable(...).requireTransaction()` or `table(..., columns, { requiresTransaction: true })`. When a table is marked this way, direct writes like `db.insert(...)`, `db.update(...)`, `db.delete(...)`, and direct batches now fail immediately with a clear error, while transactional writes continue to work normally. Non-transactional tables are still allowed inside transactions.
--- a/crates/jazz-tools/src/query_manager/manager.rs
+++ b/crates/jazz-tools/src/query_manager/manager.rs
@@ -61,6 +61,10 @@ pub enum QueryError {
         table: TableName,
         operation: Operation,
     },
+    /// Table requires transactional write mode.
+    TransactionRequired {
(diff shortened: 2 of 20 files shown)


aeplay:
🤖 Closing this stale pre-core-reconstruction draft. Its TypeScript runtime and former Rust write and schema paths no longer exist. The current core exposes mergeable and serializable exclusive transactions directly, with atomicity, read-your-writes, rollback, and conflict validation specified and tested centrally. `requiresTransaction` was an additional schema-level ergonomics guard, not a missing transactional primitive, and no current consumer or active issue requires it. If that guard becomes a product requirement, it needs a fresh public-schema and catalogue design with core-owned enforcement across current bindings.

PR #585, opened 2026-04-20
feat: add `.wait({tier})` handle to write APIs & global `onMutationError`

nrainhart:
## Description

This PR modifies the Db write APIs to return a `WriteHandle`, which allows callers to wait for a given durability tier to acknowledge the write or reject it. Also introduces a global `Db.onMutationError` handler to receive errors that aren't explicitly handled with `WriteHandle.wait`.

This paves the way to unifying our `insert`/`insertDurable`/`insertPersisted` APIs into a single method (which will be done in a follow-up PR).

TODO: update docs (will push asap, either in this PR or as a fup).

## Discussion

I think it's worth discussing how we should name the object returned by write APIs. I went with Write*Handle*, but it could be a Write*Result*, or something else. If folks have thoughts about this, feel free to suggest an alternative.


crates/jazz-tools/src/runtime_core/writes.rs
```diff
@@ -71,40 +71,29 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
         row_id: ObjectId,
         batch_id: BatchId,
         mode: BatchMode,
-        requested_tier: DurabilityTier,
```
nrainhart, line 74: Now all changes are tracked until they're globally ack-ed (or rejected), so that we can forward errors to `onMutationError` if necessary

aeplay (approved):

PR #592, opened 2026-04-20
Starters: add react SPA variants; move provisioning to create-jazz

joeinnes:
## Summary

- Add three plain-Vite React starters (`react-localfirst`, `react-hybrid`, `react-betterauth`) matching the existing next/sveltekit shape. Hybrid / betterauth sequence Hono + Vite via `concurrently` with a `wait-on` gate on a new `/health` endpoint.
- Move Jazz-Cloud provisioning + `BETTER_AUTH_SECRET` generation out of per-starter postinstall scripts and into `create-jazz`. A new `--hosting hosted` flag writes the four cloud env vars (with the `/apps/<appId>/` suffix baked into `SERVER_URL`). Per-starter `scripts/` folders deleted.
- `jazz-tools/dev`: Vite + SvelteKit plugins now load `.env` into `process.env` at plugin startup (Vite doesn't do this for non-prefixed keys); Vite plugin switches to `VITE_*`-prefixed env-key names so `import.meta.env` reads match what the scaffolder writes; `managed-runtime` falls back to `process.env.JAZZ_ADMIN_SECRET`.
- Unify the signed-in header across the six auth-using starters — `Hello, {user.name}` + Sign out side by side via `.auth-nav` row layout; hybrid's Sign up / Sign in rendered as `.link`.
- Add `next-env.d.ts` to gitignores and untrack the generated copies.
- Playwright `webServer.env` injects a test \`BETTER_AUTH_SECRET\` so e2e doesn't require a pre-existing `.env`.
- README refresh across all nine starters for the new flow.

## Test plan

- [ ] \`pnpm --filter jazz-tools exec vitest run src/dev\` — 56 passing
- [ ] \`pnpm --filter create-jazz exec vitest run\` — 55 passing
- [ ] \`node scripts/check-starters-parity.mjs\` — 9 starters, OK
- [ ] Scaffold via \`create-jazz --hosting selfhosted\` and run \`pnpm dev\`
- [ ] Scaffold via \`create-jazz --hosting hosted\` and verify \`.env\` is populated + schema pushes (blocked on the prod cloud \`/admin/*\` route availability — noted in the branch discussion)
- [ ] \`pnpm --filter <starter> test:e2e\` for each of the nine starters


packages/create-jazz/src/cloud-init.ts
```diff
@@ -0,0 +1,101 @@
...
+      values: {
+        [envKeys.appId]: appId,
+        [envKeys.serverUrl]: appServerUrl,
+        [envKeys.adminSecret]: adminSecret,
+        [envKeys.backendSecret]: backendSecret,
+      },
+      keys,
+    });
+
+    console.log("Jazz app provisioned successfully.");
```
gdorsi: Mention that `.env` was just written — currently reads like a value dump, not the outcome of an action.

packages/create-jazz/src/index.ts
```diff
@@ -99,14 +174,37 @@ async function main() {
...
+
+  const preInstall: ((dir: string) => Promise<void>) | undefined =
+    hosting === "hosted" || needsBetterAuthSecret
+      ? async (dir: string) => {
+          if (needsBetterAuthSecret) {
+            writeBetterAuthSecret(dir);
+          }
+          if (hosting === "hosted") {
+            const envKeys = envKeysForStarter(starter);
+            if (envKeys) {
```
gdorsi: If the starter name doesn't match `next-` / `sveltekit-` / `react-`, this quietly skips the whole cloud setup. The user asked for `--hosting hosted` and gets no `.env` and no warning. Either warn here, or only set `preInstall` for starters we recognise.

starters/react-betterauth/server/auth.ts
```diff
@@ -0,0 +1,51 @@
...
+  session: [],
+  user: [],
+  verification: [],
+};
+
+export const auth = betterAuth({
+  baseURL: APP_ORIGIN,
+  secret: BETTER_AUTH_SECRET,
+  database: memoryAdapter(authMemoryDb),
+  trustedOrigins: [APP_ORIGIN, "http://localhost:5173"],
```
gdorsi: `http://localhost:5173` is hardcoded, but Vite auto-jumps to 5174+ if 5173 is busy → CORS silently fails. In dev, accept any localhost origin so the port doesn't matter:

```suggestion
  trustedOrigins: (req) => {
    const origin = req.headers.get("origin");
    const list = [APP_ORIGIN];
    if (
      process.env.NODE_ENV !== "production" &&
      origin &&
      /^https?:\/\/(localhost|127\.0\.0\.1)(:\d+)?$/.test(origin)
    ) {
      list.push(origin);
    }
    return list;
  },
```

starters/react-hybrid/server/auth.ts
```diff
@@ -0,0 +1,87 @@
...
+  session: [],
+  user: [],
+  verification: [],
+};
+
+export const auth = betterAuth({
+  baseURL: APP_ORIGIN,
+  secret: BETTER_AUTH_SECRET,
+  database: memoryAdapter(authMemoryDb),
+  trustedOrigins: [APP_ORIGIN, "http://localhost:5173"],
```
gdorsi: `http://localhost:5173` is hardcoded, but Vite auto-jumps to 5174+ if 5173 is busy → CORS silently fails. In dev, accept any localhost origin so the port doesn't matter:

```suggestion
  trustedOrigins: (req) => {
    const origin = req.headers.get("origin");
    const list = [APP_ORIGIN];
    if (
      process.env.NODE_ENV !== "production" &&
      origin &&
      /^https?:\/\/(localhost|127\.0\.0\.1)(:\d+)?$/.test(origin)
    ) {
      list.push(origin);
    }
    return list;
  },
```

PR #601, opened 2026-04-21
Reduce settlement fanout and repeated sync-backed re-settles after local write bursts

aeplay:
## Summary
- add an in-repo abstract bench for the slow first `full` query after a large local write burst
- batch outbound client->server sync payloads into `SyncBatchRequest`s instead of sending one WebSocket frame per payload
- stop replaying client-originated row batches when synced query scope bootstraps
- keep frontier-gated sync-backed subscriptions from repeatedly reloading / re-settling visible rows before upstream `QuerySettled`
- merge current `main` into the branch so CI reruns against the latest browser-test baseline before final review

## Why
After `#599` removed the `getSchema()` / schema-marshalling bottleneck, the remaining long first-read stall turned out to be settlement churn in two places:

- transport + sync fanout: a large local write burst produced a large outbound payload/frame backlog, and synced query scope bootstrap could replay row batches back to the same client that had just uploaded them
- frontier-gated query churn: a sync-backed subscription that was still waiting on its upstream frontier could keep getting poked by tier confirmations and repeatedly reload / re-settle visible rows before the final `QuerySettled`

This branch attacks both. It cuts redundant sync traffic, and it also stops held sync-backed subscriptions from doing repeated downstream work while they are still gated on the upstream frontier.

The linked BoreDM follow-up is also important: after this change, the remaining slowdown no longer looks settlement-specific. The same query with `propagation: "local-only"` is still slow, which points to the underlying permission-heavy read as the next bottleneck rather than settlement fanout.

## Validation
- `cargo test -p jazz-tools sync_manager::tests:: --lib`
- `cargo test -p jazz-tools connected_transport_batches_outbox_payloads_into_one_sync_request --lib --features runtime-tokio`
- `pnpm --filter jazz-tools exec vitest run src/runtime/client.test.ts --reporter=verbose`
- `pnpm --filter jazz-napi run build:debug`
- `JAZZ_ABSTRACT_BENCH=1 pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/query-settlement.abstract-bench.test.ts --disableConsoleIntercept`

## Before / After
In-repo node/N-API abstract bench (`1200` inserts, `full` query):
- before: immediate `~6864 ms`, warmed `~5481 ms`
- after: immediate `~1602-1630 ms`, `local-only ~223-228 ms`, warmed `~438-442 ms`

Linked BoreDM Bun repro:
- before: immediate `~8460 ms`, warmed `~311 ms`
- after: immediate `~2941 ms`, warmed `~2789 ms`
- `propagation: "local-only"` remains `~2946 ms`, which is why the remaining time now looks permission-read-bound instead of settlement-bound


aeplay:
Closing in favor of #774

PR #602, opened 2026-04-21
Bypass local policies for admin-secret clients

Takeno:
## Summary
- pass app-level permissions into the inspector dev sync server and include appId when fetching schema hashes
- define browser test permissions so the standalone inspector exercises real policy metadata
- skip anonymous session minting and strip local schema policies for clients that use an admin secret
- tighten inspector tests to wait for loaded table rows and assert permissions are rendered

## Testing
- Added runtime tests for anonymous session handling and schema policy stripping
- Updated browser coverage for the standalone inspector data explorer
- Not run (not requested)

### DIFF
--- a/packages/jazz-tools/src/runtime/db.ts
+++ b/packages/jazz-tools/src/runtime/db.ts
@@ -125,6 +125,22 @@ function resolveStorageDriver(driver?: StorageDriver): StorageDriver {
   return driver ?? { type: "persistent" };
 }
 
+function shouldBypassLocalPolicies(config: DbConfig): boolean {
+  return resolveStorageDriver(config.driver).type === "memory" && !!config.adminSecret;
+}
+
+function stripSchemaPolicies(schema: WasmSchema): WasmSchema {
+  return Object.fromEntries(
+    Object.entries(schema).map(([tableName, tableSchema]) => [
+      tableName,
+      {
+        ...tableSchema,
+        policies: undefined,
+      },
+    ]),
+  ) as WasmSchema;
+}
+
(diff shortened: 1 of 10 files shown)


packages/jazz-tools/src/runtime/db.ts
```diff
@@ -125,6 +125,22 @@ function resolveStorageDriver(driver?: StorageDriver): StorageDriver {
   return driver ?? { type: "persistent" };
 }
 
+function shouldBypassLocalPolicies(config: DbConfig): boolean {
+  return resolveStorageDriver(config.driver).type === "memory" && !!config.adminSecret;
```
gdorsi: Why Policy stripping is gated on `driver.type === "memory"`?

Takeno: We agreed to keep adminSecret for very few use cases, such as inspector or custom data explorers. The combination of "memory" and "adminSecret" was to target the specific use case. Gating on adminSecret only can lead to unwanted behavior.

But since adminSecret on clients is not recommended, not documented and tomorrow might be renamed into `__dangerousAdminSecretUsage`, we can strip the condition on driver.

gdorsi (approved):
Reviewed end-to-end. Core runtime change is clean: bypass gated on memory + adminSecret, anon-session skip when adminSecret is present. Left one question inline about the memory-only scope of policy stripping vs the always-on anon skip. Test coverage is solid through the inspector e2e; the one nit is that the transport unit test uses `toMatchObject`, so it doesn't explicitly assert policies were stripped.

PR #603, opened 2026-04-21
feat(cli): warn on missing explicit permission policies during deploy

joeinnes:
## Summary

- `jazz-tools deploy` now calls `collectMissingExplicitPolicyDiagnostics` after loading the schema, emitting the same yellow warnings as `jazz-tools validate` for tables with no explicit permission policy
- `deploy` does not block on these warnings — they are informational only
- Previously the only way to surface these warnings was to run `validate` separately before deploying

## Test plan

- [ ] Red test added first, confirmed failing before implementation
- [ ] Green implementation added, test confirmed passing
- [ ] Changeset included (`patch` on `jazz-tools`)

### DIFF
--- a/.changeset/deploy-missing-policy-warnings.md
+++ b/.changeset/deploy-missing-policy-warnings.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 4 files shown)


Takeno (approved):
LGTM

PR #605, opened 2026-04-21
feat(worker): fall back to ephemeral storage when OPFS is blocked (Firefox private browsing)

joeinnes:
## Summary

- Firefox private browsing (and Safari in some private-mode configs) blocks `navigator.storage.getDirectory()` with a `SecurityError`, causing Jazz to fail to initialise entirely
- Adds a `SecurityError` detection path through the full Rust stack (`opfs-btree` → `jazz-tools` storage layer → `jazz-wasm` WASM boundary)
- Adds `WasmRuntime.openEphemeral` (in-memory storage, no persistence) as a fallback when `openPersistent` is blocked
- Worker catches the `SecurityError` and transparently falls back to ephemeral storage so the app still loads

## Changes

- **`crates/opfs-btree`** — new `BTreeError::SecurityError` variant; `map_js_error` detects `DOMException(SecurityError)`; `is_retryable_handle_conflict` excludes `SecurityError` from retries
- **`crates/jazz-tools/storage`** — new `StorageError::SecurityError` variant propagated from `BTreeError`
- **`crates/jazz-wasm`** — `open_persistent` maps `SecurityError` to a `js_sys::Error` with `name = "SecurityError"` at the WASM boundary; new `open_ephemeral` entry point using `OpfsBTreeStorage::memory`; shared `build_schema_manager` and `assemble_wasm_runtime` helpers eliminate duplication between both constructors
- **`packages/jazz-tools/worker`** — `handleInit` catches `e?.name === "SecurityError"` from `openPersistent` and falls back to `openEphemeral`
- **`crates/jazz-napi`** — regenerated bindings (alpha.36), removed `returnsDeclaredSchemaRows`
- **`starters/*`** — `prebuild` script auto-generates `.env` with a random `BETTER_AUTH_SECRET` when absent
- **`examples/todo-react`** — corrected env var prefix to `VITE_JAZZ_*`

## Test plan

- [ ] New worker unit tests cover: SecurityError fallback (happy path), non-SecurityError re-throw, `openEphemeral` itself failing
- [ ] Run `pnpm test` — all passing
- [ ] Manual test in Firefox private browsing: app should load with ephemeral storage rather than showing an error

### DIFF
--- a/crates/jazz-wasm/src/runtime.rs
+++ b/crates/jazz-wasm/src/runtime.rs
@@ -1766,7 +1766,7 @@ impl WasmRuntime {
         db_name: &str,
         tier: Option<String>,
         use_binary_encoding: bool,
-    ) -> Result<WasmRuntime, JsError> {
+    ) -> Result<WasmRuntime, JsValue> {
         #[cfg(feature = "console_error_panic_hook")]
         console_error_panic_hook::set_once();
         init_tracing();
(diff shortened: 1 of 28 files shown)


crates/jazz-wasm/src/runtime.rs
```diff
@@ -376,6 +376,71 @@ fn tier_label_for_node_tier(tier: Option<&str>) -> &'static str {
...
+    let scheduler = WasmScheduler::new();
+    let sync_sender = JsSyncSender::new(use_binary_encoding);
+    let mut core = RuntimeCore::new(schema_manager, storage, scheduler);
+    core.set_tier_label(tier_label);
+    core.set_sync_sender(Box::new(sync_sender.clone()));
+    let core_rc = Rc::new(RefCell::new(core));
+    {
+        let mut core_guard = core_rc.borrow_mut();
+        core_guard
+            .scheduler_mut()
```
gdorsi, line 432: nit: how about using MemoryStorage here?

Should be faster than OpfsBTreeStorage:memory

gdorsi (approved):
Looks good to me

PR #608, opened 2026-04-21
fix: reliable wasm load in Next.js Turbopack workers

antoniomuso:
## Summary

- Jazz apps running on Next.js 16 (Turbopack) crash in the worker with `TypeError: Failed to construct 'URL': Invalid URL` because Turbopack doesn't transform wasm-bindgen's `new URL('jazz_wasm_bg.wasm', import.meta.url)` pattern inside nested `node_modules` worker chunks, and the main-thread fallback was gated behind an `isHttpUrl(runtimeModuleUrl)` check that is always true in a real app.
- `withJazz` now copies `jazz_wasm_bg.wasm` into the host app's `public/_jazz/` (dev + production-build phases) and `buildWorkerBridgeOptions` defaults `runtimeSources.wasmUrl` to an absolute URL against `location.origin`, so the worker receives a fully-qualified URL and skips the broken bundler-default path.
- `resolveConfiguredUrl` tries absolute-first so a bogus `self.location.href` (as emitted by Turbopack in some worker contexts) can't poison an already-absolute URL.

## Test plan

- [ ] `pnpm --filter auth-betterauth-chat dev` — app loads without `Invalid URL` in the browser console.
- [ ] `pnpm --filter auth-betterauth-chat build && pnpm --filter auth-betterauth-chat start` — wasm served from `/_jazz/jazz_wasm_bg.wasm`, app loads.
- [ ] Other examples that use `withJazz` still work (no regressions in `nextjs-csr-ssr`, starters).
- [ ] Consumers who pass `runtimeSources.wasmUrl` themselves still have their value respected.

### DIFF
--- a/packages/jazz-tools/src/dev/next.ts
+++ b/packages/jazz-tools/src/dev/next.ts
@@ -1,3 +1,6 @@
+import { createRequire } from "node:module";
+import { copyFile, mkdir } from "node:fs/promises";
+import { dirname, join } from "node:path";
 import { buildInspectorLink } from "./inspector-link.js";
 import { ManagedDevRuntime } from "./managed-runtime.js";
 import type { JazzPluginOptions, JazzServerOptions } from "./vite.js";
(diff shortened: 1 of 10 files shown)


packages/jazz-tools/src/dev/next.ts
```diff
@@ -65,6 +79,17 @@ export function withJazz(
       serverExternalPackages: mergeServerExternalPackages(resolved.serverExternalPackages),
     };
 
+    // Copy jazz-wasm bytes into the host app's public/ dir so they're served
+    // at a stable origin-root URL. Works around bundlers (Turbopack) that
+    // don't transform wasm-bindgen's `new URL('*.wasm', import.meta.url)`
+    // pattern inside worker chunks. Runs in dev and production-build so the
+    // asset is present in the built output.
+    if (phase === DEVELOPMENT_PHASE || phase === PRODUCTION_BUILD_PHASE) {
+      await copyWasmToPublic(options.schemaDir ?? process.cwd());
```
gdorsi: Why `options.schemaDir`?

antoniomuso: This has to be the app root. I am fixing it.

gdorsi (approved):

PR #632, opened 2026-04-22
feat(jazz-plugin): add config hook for worker format, and optimizeDeps

joeinnes:
## Summary

- `jazzPlugin` and `jazzSvelteKit` now inject `worker.format: "es"`, and `optimizeDeps.exclude: ["jazz-wasm"]` via a Vite `config` hook, so consumers no longer need to set these manually
- Removes the now-redundant explicit settings from all example `vite.config.ts` files
- Adds tests for the new `config` hook behaviour in both plugins

## Test plan

- [ ] `pnpm --filter jazz-tools test` passes (vite and sveltekit plugin tests)
- [ ] Example apps build without errors

### DIFF
--- a/.changeset/vite-plugin-config-hook.md
+++ b/.changeset/vite-plugin-config-hook.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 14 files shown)


antoniomuso (approved):
Nice!

Takeno (approved):

PR #650, opened 2026-04-22
fix: generated schema snapshots' timestamps

nrainhart:
## Description

When generating migrations for a new app, I noticed snapshots had weird names (e.g. `582761109T032422-268a4ad77008.json`) instead of the expected timestamp (`20260422T160045-8bde091420f7.json`).

Turns out the server returns the schema timestamp with microsecond precision, and the CLI expects in in millisecond precision. This PR adjusts the timestamp to generate correct snapshot filenames.


antoniomuso (approved):

antoniomuso:
Maybe we should evaluate changing the timestamp to `ms` even on the server.

PR #653, opened 2026-04-23
Fix provenance magic timestamp conversion

aeplay:
## What changed
- fix `jazz-tools` provenance magic timestamp decoding so `$createdAt` and `$updatedAt` come back as real JS dates instead of far-future values
- convert query filter inputs for those magic columns onto the same internal timestamp scale used by the runtime
- add targeted coverage for row transformation, query translation, end-to-end query behavior, and a patch changeset

## Why
Selecting provenance magic timestamps was returning internal microsecond values as if they were JavaScript milliseconds. That produced far-future `Date` objects and made `where({ $updatedAt: ... })` filters inconsistent with selected results.

## Impact
Callers using `jazz-tools` can now safely select and filter on `$createdAt` and `$updatedAt` with normal JavaScript `Date` and millisecond values.

## Root cause
The Rust/runtime boundary stores provenance timestamps in microseconds, but the TypeScript magic-column path was treating all `Timestamp` payloads as ordinary millisecond timestamps.

## Validation
- `pnpm lint`
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/row-transformer.test.ts src/runtime/query-adapter.test.ts tests/ts-dsl/query-api.test.ts`

### DIFF
--- a/.changeset/fix-provenance-magic-timestamps.md
+++ b/.changeset/fix-provenance-magic-timestamps.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Fix `$createdAt` and `$updatedAt` provenance magic columns in `jazz-tools` so they round-trip as real JavaScript dates instead of far-future timestamps.
+
+Queries now convert JS millisecond `Date` and numeric filter values to the internal provenance timestamp format consistently, so selecting and filtering on those magic columns uses the same time scale.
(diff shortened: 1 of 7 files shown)


gdorsi (approved):
Looks good to me.

One small design flag: the overall timestamp decoding/conversion path is starting to look a little verbose for handling narrow cases like provenance magic columns. This fix is reasonable, but we may want to keep an eye on whether the decoding system wants a simpler centralized shape if more column-specific runtime encodings appear.

PR #660, opened 2026-04-23
fix: remove stale jazz_principal_id JWT claims from starters

joeinnes:
## Summary

- Remove the `jazz_principal_id` extra claim from the JWT `definePayload` config in `react-betterauth` and `react-hybrid` starters. Jazz resolves `session.user_id` from `sub` only; the claim was no longer read and was misleading new adopters.
- Extract `jwtPayload` into a standalone `server/jwt-payload.ts` module so it can be unit-tested independently of the server startup (which requires env vars).
- Add vitest to both starters with a regression test that asserts the JWT payload does not contain `jazz_principal_id`. The test was written first (red) and passes after the claim removal (green).

## Test plan

- [ ] `pnpm -C starters/react-betterauth test` → 1 passed
- [ ] `pnpm -C starters/react-hybrid test` → 1 passed
- [ ] `BETTER_AUTH_SECRET=x pnpm -C starters/react-betterauth build` → clean build
- [ ] `BETTER_AUTH_SECRET=x pnpm -C starters/react-hybrid build` → clean build
- [ ] Grep for `jazz_principal_id` in `starters/**/*.ts` returns only test-file assertion strings, no implementation code


.changeset/remove-jazz-principal-id-starters.md
```diff
@@ -0,0 +1,7 @@
+---
+---
```
Takeno: Is this working even without packages selected?

Takeno (approved):

.changeset/remove-jazz-principal-id-starters.md
joeinnes: No :)

PR #663, opened 2026-04-23
docs, example(expo): drop babel.config.cjs shim + type:module

antoniomuso:
## Summary

- `@expo/metro-config/loadBabelConfig.js` only matches `.babelrc`, `.babelrc.js`, and `babel.config.js` — a project-level `babel.config.cjs` is invisible to Expo, and `config.transformer.extendsBabelConfigPath` is a no-op on this pipeline. That means preset options like `unstable_transformImportMeta` never reach the `babel-preset-expo` invocation that processes `node_modules`, and Hermes rejects `import.meta.url` coming from `jazz-tools` with:

  > `import.meta` is not supported in Hermes. Enable the polyfill `unstable_transformImportMeta` in babel-preset-expo to use this syntax.

- Rewrite the Expo install docs and the `todo-client-localfirst-expo` example to the plain CJS `babel.config.js` that Expo picks up automatically, drop the `.babelrc → .cjs` shim, remove the dead `extendsBabelConfigPath` wiring, and stop declaring `"type": "module"` (it wasn't actually needed — `create-expo-app` doesn't set it).
- `metro.config` stays ESM — renamed to `metro.config.mjs` so it can still top-level-`await withJazz(...)` without flipping the whole package to ESM.

Verified end-to-end with `pnpm exec expo export --platform ios`: bundle succeeds on SDK 54, transformed output contains `__ExpoImportMetaRegistry` (i.e. the preset option actually propagated) and zero `import.meta`.

## Test plan

- [ ] `pnpm --filter todo-client-localfirst-expo exec expo export --platform ios` finishes cleanly
- [ ] `pnpm --filter todo-client-localfirst-expo exec expo export --platform android` finishes cleanly
- [ ] Docs render — confirm the `metro.config.mjs` + `babel.config.js` snippets are the ones shown on `/docs/install/client`


joeinnes (approved):
LGTM!

PR #665, opened 2026-04-23
Fix .default schema hash

### LINKED ISSUE #664: CLI `loadCompiledSchema` drops column `.default(...)`, causing schema hash mismatch
### Summary
The CLI's AST-based `loadCompiledSchema` strips `.default(...)` from column definitions, but the runtime `defineApp` preserves them. Because the published schema hash is computed from the CLI's wasmSchema while the client hash is computed from `defineApp`'s wasmSchema, they diverge — so every client hits:

```
Your declared schema <A> is disconnected from the schema used to enforce permissions: <B>
```

### Repro
`schema.ts`:

```ts

const schema = {
  messages: s.table({
    content: s.string(),
    done: s.boolean().default(false), // <-- the trigger
  }),
};

export const app = s.defineApp(schema);
```

Compare the two wasmSchemas:

```ts
import { loadCompiledSchema } from "jazz-tools/dist/schema-loader.js";
import { app } from "./schema.ts";

const cli = await loadCompiledSchema(".");
console.log(JSON.stringify(cli.wasmSchema) === JSON.stringify(app.wasmSchema));
// false — 596 vs 639 bytes
```

The 43-byte delta is:

```json
"default":{"type":"Boolean","value":false}
```

present only in `defineApp`'s output.

### Impact
`jazz-tools deploy` publishes hash `A`; browser/server both compute `B`; permissions can never match. The error is opaque — nothing points at `.default()` as the cause.

### Workaround
Remove `.default(...)` from all column definitions and supply defaults in application code.

### Fix
Either have `loadCompiledSchema` preserve `.default(...)` chains, or have `defineApp` drop them, so the two paths produce byte-identical wasmSchema for identical source.

### Environment
- `jazz-tools@2.0.0-alpha.39`
- Node 22.13.1, macOS

tobiaslins:
### Summary
The CLI's AST-based `loadCompiledSchema` strips `.default(...)` from column definitions, but the runtime `defineApp` preserves them. Because the published schema hash is computed from the CLI's wasmSchema while the client hash is computed from `defineApp`'s wasmSchema, they diverge — so every client hits:

```
Your declared schema <A> is disconnected from the schema used to enforce permissions: <B>
```

### Repro
`schema.ts`:

```ts

const schema = {
  messages: s.table({
    content: s.string(),
    done: s.boolean().default(false), // <-- the trigger
  }),
};

export const app = s.defineApp(schema);
```

Compare the two wasmSchemas:

```ts
import { loadCompiledSchema } from "jazz-tools/dist/schema-loader.js";
import { app } from "./schema.ts";

const cli = await loadCompiledSchema(".");
console.log(JSON.stringify(cli.wasmSchema) === JSON.stringify(app.wasmSchema));
// false — 596 vs 639 bytes
```

The 43-byte delta is:

```json
"default":{"type":"Boolean","value":false}
```

present only in `defineApp`'s output.

### Impact
`jazz-tools deploy` publishes hash `A`; browser/server both compute `B`; permissions can never match. The error is opaque — nothing points at `.default()` as the cause.

### Workaround
Remove `.default(...)` from all column definitions and supply defaults in application code.

### Fix
Either have `loadCompiledSchema` preserve `.default(...)` chains, or have `defineApp` drop them, so the two paths produce byte-identical wasmSchema for identical source.

### Environment
- `jazz-tools@2.0.0-alpha.39`
- Node 22.13.1, macOS

fixes #664

### DIFF
--- a/.changeset/cli-preserve-column-defaults.md
+++ b/.changeset/cli-preserve-column-defaults.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix CLI `loadCompiledSchema` dropping column `.default(...)` values, which caused `deploy` to publish a different schema hash than the runtime computed and broke permission checks with `Your declared schema <A> is disconnected from the schema used to enforce permissions: <B>`. The original `wasmSchema` from `defineApp` is now preserved instead of being round-tripped through a lossy AST.
--- a/TODO.md
+++ b/TODO.md
@@ -38,6 +38,7 @@
 
 - [**postmessage-stream-adapter-for-worker-bridge**](todo/issues/postmessage-stream-adapter-for-worker-bridge.md) — The browser main-thread ↔ worker bridge currently uses a bespoke `postMessage` path and the `JsSyncSender` Rust-side shim. A `PostMessageStream` adapter implementing `StreamAdapter` would let the worker bridge reuse the same `TransportManager` + run-loop code path as the network transport, eliminating the `sync_sender` field and `JsSyncSender` entirely.
 - [**text-encoded-storage-enums**](todo/issues/text-encoded-storage-enums.md) — Flat row storage currently encodes enum-like fields as text, which is larger than necessary and adds avoidable decode overhead on hot storage paths.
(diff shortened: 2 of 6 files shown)


joeinnes (approved):
Thank you!

aeplay:
Closing in favor of #675 (more fundamental fix)

Thanks so much for spotting this!

PR #666, opened 2026-04-24
fix: use output schema from JSON standard schema

Gabrola:
## Summary
`col.json(schema)` stored a JSON Schema that didn't match the inferred TS column type, causing the Rust layer to reject valid values.
The TS type came from Standard Schema's **output** (`InferOutput<Schema>`), but `normalizeJsonSchema` pulled the stored schema from `~standard.jsonSchema.input(...)`. For any schema with a transform these diverge:
```ts
col.json(z.string().transform((input) => input.length))
// TS type:        number                 (from output)
// Stored schema:  { type: "string" }     (from input)
// Rust validates a number against a string schema → error
```
Fix: `normalizeJsonSchema` now reads `~standard.jsonSchema.output(...)`, so the stored schema matches the TS type the user sees.
### Bonus: some validators reject transforms at definition time
Validation only runs in the Rust layer today (the JS layer doesn't validate at all), and JSON Schema can't represent transforms in general. By switching to `.output(...)`, libraries that refuse to emit JSON Schema for transformed shapes now surface that at `col.json(...)` time instead of silently storing a contract the Rust validator can't honor. Zod, for example, throws `Transforms cannot be represented in JSON Schema` for the snippet above. Other Standard Schema implementations may behave differently.


joeinnes (approved):
Looks good! Thank you!

aeplay:
Moved to #683 for trusted CI and adding some tests. Thanks so much for spotting this and contributing a fix!

PR #667, opened 2026-04-24
fix(jazz-tools): prevent tokio scheduler task pileup under contention

antoniomuso:
## Summary

- `TokioScheduler::schedule_batched_tick` cleared the debounce flag at the start of the spawned task, before acquiring the core lock. Under contention every caller that arrived while the task was blocked on the mutex observed `scheduled=false` and spawned another task, piling up behind the same lock.
- Acquire the core lock first, then clear the flag immediately before running `batched_tick`. The lost-wakeup fix is preserved — messages arriving during the tick still find `scheduled=false` and can schedule a follow-up — while the pending queue is capped at one tick.

## Test plan

- [ ] `cargo check -p jazz-tools`
- [ ] `cargo test -p jazz-tools`

### DIFF
--- a/.changeset/steady-tides-tick.md
+++ b/.changeset/steady-tides-tick.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Prevent TokioScheduler task pileup under contention.
+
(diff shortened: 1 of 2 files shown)


gdorsi (approved):
Looks good. The new ordering keeps the debounce flag set while the spawned task is blocked on the core mutex, preventing queued task pileup, while still clearing it before `batched_tick()` so follow-up work can schedule another tick.

PR #670, opened 2026-04-24
Fix websocket row-batch broadcast leakage

Takeno:
## Bug, Cause, And Fix
- Bug: when client B received a row via `RowBatchNeeded` and answered the server with `RowBatchStateChanged`, client A could receive B's ack and the `BatchSettlement` response meant for B.
- Cause: the client-origin `RowBatchStateChanged` handler reused relay logic intended for server-origin durability updates, walking `row_batch_interest` and forwarding the ack plus settlement to other interested clients.
- Why this fix: client-origin row-batch acks are per-client/server bookkeeping, so the server should apply them locally only. Authoritative server-origin `RowBatchStateChanged` messages still use the existing interested-client relay path.

## Summary
- Stop relaying client-origin `RowBatchStateChanged` messages to other clients.
- Keep the server-side row-batch state handling local so Bob’s ack no longer fans out to Alice.
- Add a regression test covering two subscribers on the same row batch.

## Testing
- `cargo fmt --check`
- `cargo test -p jazz-tools client_row_batch_state_ack_does_not_leak_to_other_subscribers`
- `cargo test -p jazz-tools sync_manager::tests`
- `cargo test -p jazz-tools --features test --test sync_tracer_integration alice_write_bob_read`

### DIFF
--- a/.changeset/fix-websocket-rowbatch-ack-leak.md
+++ b/.changeset/fix-websocket-rowbatch-ack-leak.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fixed the flood of ws messages by stop forwarding client-origin `RowBatchStateChanged` acknowledgements to other subscribers. This keeps per-client row-batch durability bookkeeping local to the server/client pair and avoids leaking `BatchSettlement` echoes to unrelated WebSocket clients.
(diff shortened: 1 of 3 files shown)


antoniomuso (changes requested):
The changeset is missing.

gdorsi (approved):

PR #671, opened 2026-04-27
fix(jazz-tools): SvelteKit dev plugin restarts Vite after first appId allocation

joeinnes:
## Summary
- Fixes the SvelteKit dev plugin so the first-ever cold start no longer needs a manual restart. After `runtime.initialize` allocates an appId, `configureServer` calls `viteServer.restart()` so SvelteKit's `config: { order: 'pre' }` hook re-runs and re-reads the now-populated `.env`. Warm starts (where `.env` already has the appId) skip the restart.
- Removes the `httpServer.once("close", dispose)` handler — it would dispose the cached runtime on Vite restart, defeating the cache. Real shutdown is still covered by SIGINT/SIGTERM/exit handlers in `installShutdownHooks`.
- Plugin order in `vite.config.ts` is no longer load-bearing. Removes the "must come before sveltekit" guidance from docs and SvelteKit starter `vite.config.ts` files. Configs return to natural framework-first order (the swap from the first commit is reverted).

## Test plan
- [x] `pnpm exec vitest run src/dev/sveltekit.test.ts` — added `restarts the dev server on first-ever cold start...` and `does not restart when appId is already persisted...`. Both green. Existing tests still pass; the close-hook test was deleted (its asserted behaviour is intentionally removed).
- [ ] Smoke: scaffold a fresh SvelteKit app, run `pnpm dev`, confirm: first run logs the restart message and the page loads without manual intervention; subsequent runs do not restart.
- [ ] Smoke: reverse plugin order in `vite.config.ts` (`[jazzSvelteKit(), sveltekit()]`) — first run still self-heals via restart.

### DIFF
--- a/.changeset/fix-sveltekit-cold-start-env.md
+++ b/.changeset/fix-sveltekit-cold-start-env.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix the SvelteKit dev plugin so the first-ever cold start no longer needs a manual restart. The plugin now triggers a Vite restart immediately after allocating a fresh app ID, so SvelteKit's `$env/*` capture re-reads the now-populated `.env` on the second pass. Plugin order in `vite.config.ts` is no longer load-bearing.
--- a/docs/content/docs/getting-started/client-setup.mdx
+++ b/docs/content/docs/getting-started/client-setup.mdx
@@ -172,7 +172,7 @@ Jazz ships bundler plugins that remove boilerplate in development. They start a 
     import { jazzSvelteKit } from "jazz-tools/dev/sveltekit";
 
     export default defineConfig({
-      plugins: [jazzSvelteKit(), svelte()],
+      plugins: [svelte(), jazzSvelteKit()],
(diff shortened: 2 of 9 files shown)


Takeno (approved):
Should we add an explicit warning in the docs to remark this plugins order? It would be a very common mistake.
Otherwise, is there a way inside our plugin to understand if `svelte()` plugin already ran?

joeinnes:
> Should we add an explicit warning in the docs to remark this plugins order? It would be a very common mistake. Otherwise, is there a way inside our plugin to understand if `svelte()` plugin already ran?

It's already there as a comment in the sibling tab for SvelteKit (which is where it's necessary):

image

The ordering doesn't matter in pure Vite, this is just to avoid people being confused.

I'm not sure whether we can detect whether the Svelte plugin ran or not, but I'll check.

joeinnes:
Ended up taking this in a completely different direction so that we can avoid making folks specify things in a particular order

Takeno (approved):
Great work!

PR #672, opened 2026-04-27
feat: accept getter functions in Svelte QuerySubscription

joeinnes:
## Summary
- Widens `QuerySubscription` (Svelte) to accept `() => QueryBuilder | undefined` for both the query and the options. Reactive reads inside the getter are tracked, so the subscription re-runs on dependency changes.
- Bare-value forms continue to work unchanged.
- The published Svelte conditional-query docs example (`ConditionalQueryExample.svelte`) already used this form; the implementation now matches the docs.
- Vue's `useAll` already supports `MaybeRefOrGetter` via `toValue`, so this gap was Svelte-only.

## Test plan
- [x] New `use-all.svelte.test.ts` covers: plain query (regression), undefined (regression), getter returning a query, getter returning undefined, getter reading `\$state` re-subscribes on change, getter flipping undefined → query, getter flipping query → undefined, options-as-getter.
- [x] Red-then-green TDD: 6 getter tests failed before the implementation; all 33 Svelte tests pass after.
- [x] Full `pnpm test` in `jazz-tools` passes (93 test files).
- [x] `pnpm build` in `jazz-tools` and the Svelte docs example build cleanly.

### DIFF
--- a/packages/jazz-tools/src/svelte/use-all.svelte.ts
+++ b/packages/jazz-tools/src/svelte/use-all.svelte.ts
@@ -55,6 +55,7 @@ export class QuerySubscription<T extends { id: string }> {
       if (!resolvedQuery) {
         this.current = undefined;
         this.loading = false;
+        this.error = null;
         return;
       }
 
--- a/packages/jazz-tools/src/svelte/use-all.svelte.ts
+++ b/packages/jazz-tools/src/svelte/use-all.svelte.ts
@@ -4,12 +4,21 @@ import type { SubscriptionDelta } from "../runtime/subscription-manager.js";
 import { applyDelta } from "../reconcile-array.js";
 import { getJazzContext } from "./context.svelte.js";
 
+type MaybeGetter<T> = T | (() => T);
+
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/svelte/use-all.svelte.ts
```diff
@@ -34,12 +43,16 @@ export class QuerySubscription<T extends { id: string }> {
...
     const ctx = getJazzContext();
-    this.current = options?.tier ? undefined : [];
+    this.current = resolve(options)?.tier ? undefined : [];
 
     $effect(() => {
-      if (!query) {
+      const resolvedQuery = resolve(query);
+      if (!resolvedQuery) {
         this.current = undefined;
         this.loading = false;
```
Takeno, line 57: If the query becomes undefined, should we reset the error state? `this.error = null;`

Takeno (approved):

packages/jazz-tools/src/svelte/use-all.svelte.ts
joeinnes: Done

PR #673, opened 2026-04-27
fix(dev): honor root adminSecret in managed dev server

antoniomuso:
## Summary

- The dev plugins (`vite`, `next`, `sveltekit`, `expo`) accept `adminSecret` both at the root of `JazzPluginOptions` and nested under `server`. Only the external-server branches in `ManagedDevRuntime.initialize` consulted the root field; the managed local-server branch read `server.adminSecret` exclusively, silently falling back to a random `jazz-dev-XXXXXXXX` when callers set the root option. This made root `adminSecret` look like a no-op.
- Mirrors the precedence already used a few lines below for `appId`: nested wins, root falls through, random as last resort.
- Surfaces the resolved admin secret in the managed-server startup banner and prints the inspector link from `expo.ts` so the value is visible the same way it is for `vite` / `next`.

## Test plan

- [x] New regression test in `packages/jazz-tools/src/dev/expo.test.ts` spawns a managed server with only the root `adminSecret`, hits `/apps/:id/schemas` with that secret, and asserts the server accepts it. Red before the fix, green after.
- [x] Full `pnpm vitest run src/dev/` suite (60 tests) still passes.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/.changeset/expo-managed-admin-secret-root.md
+++ b/.changeset/expo-managed-admin-secret-root.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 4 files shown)


Takeno (approved):
Does it make sense to unify the parameters and have a single adminSecret parameter?

PR #674, opened 2026-04-27
fix(jazz-tools): use bundled webpki roots for native WS TLS

antoniomuso:
## Summary

- `NativeWsStream` was using `tokio-tungstenite`'s default `Connector`, which builds rustls from `rustls-native-certs`. On Android (and frequently iOS) the OS trust store is unreachable from a Rust binary, so the `RootCertStore` came back empty and every `wss://` handshake failed with `UnknownIssuer` — breaking RN clients connecting to cloud and edge tier.
- Construct an explicit `Connector::Rustls`. The trust store is built **strictly from the OS native store**, with `webpki-roots` used **only as a fallback when the native store yields zero usable anchors**. The `ClientConfig` is built once via `OnceLock` to avoid re-parsing certs per connection.
- `rustls` / `rustls-native-certs` / `webpki-roots` are pinned to the same versions `tokio-tungstenite 0.24` already pulls in, so there is no duplicate `rustls` in the binary and `Connector::Rustls(Arc<ClientConfig>)` types match.

## Why fallback, not union

An earlier draft unconditionally appended `webpki-roots` on top of the native store. That subtly broke security guarantees on desktop: environments that rely on OS trust restrictions (enterprise-managed trust stores, administrator-distrusted roots) would silently re-accept CAs because the bundled Mozilla set re-introduced them.

The fallback shape preserves OS trust policy wherever the OS is reachable, and only kicks in when the native store is genuinely unusable (mobile, or stripped-down environments).

## Why not just enable `rustls-tls-webpki-roots` on tokio-tungstenite?

That feature only compiles the Mozilla CA bundle into the binary; it doesn't change which root source the default connector picks. The default still chose native roots first, returned an empty store on RN, and never consulted webpki. Crates don't chain root sources automatically — you have to build a `ClientConfig` whose store is constructed by your own logic.

## Test plan

- [x] `cargo test -p jazz-tools --features transport-websocket --no-default-features --lib ws_stream::native` — 3/3 pass (incl. new `empty_native_store_falls_back_to_bundled_webpki_roots`).
- [x] `cargo build -p jazz-tools --features client` and `--features server` — clean.
- [x] Verify on a real Android RN build that `wss://` to cloud/edge succeeds.
- [x] Verify on a real iOS RN build that `wss://` to cloud/edge succeeds.
- [ ] Verify on desktop with an admin-distrusted CA that the previously-rejected cert is **still rejected** (confirms no webpki re-introduction).

### DIFF
--- a/crates/jazz-tools/src/ws_stream/native.rs
+++ b/crates/jazz-tools/src/ws_stream/native.rs
@@ -11,17 +11,44 @@ pub struct NativeWsStream {
     inner: WebSocketStream<MaybeTlsStream<TcpStream>>,
 }
 
-// Build a rustls root store that combines the OS trust store (when reachable)
-// with Mozilla's bundled webpki-roots. On Android/iOS the native store is
-// frequently empty or unreadable from a Rust binary, so the bundled roots are
-// what actually let mobile clients reach the cloud/edge over wss://.
+// Build a rustls root store from the OS trust store, falling back to Mozilla's
+// bundled webpki-roots only when the native store yields zero usable anchors.
+//
+// We do NOT union webpki-roots with a populated native store: doing so would
+// silently re-introduce CAs that an administrator or enterprise policy has
+// distrusted, undermining OS-level trust restrictions on desktop.
+//
+// The fallback exists for platforms where `rustls-native-certs` cannot reach
+// the OS trust store at all — most notably Android (anchors live in
+// `AndroidCAStore`, accessible only via JNI), and frequently iOS depending on
+// linkage. Without the fallback, every `wss://` handshake on those platforms
+// fails with `UnknownIssuer`.
 fn root_store() -> rustls::RootCertStore {
-    let mut roots = rustls::RootCertStore::empty();
     let native = rustls_native_certs::load_native_certs();
-    for cert in native.certs {
-        let _ = roots.add(cert);
+    let native_error_count = native.errors.len();
+    build_root_store_from_native(native.certs, native_error_count)
+}
+
+fn build_root_store_from_native(
+    native_certs: Vec<rustls::pki_types::CertificateDer<'static>>,
+    native_error_count: usize,
+) -> rustls::RootCertStore {
+    let mut roots = rustls::RootCertStore::empty();
+    let mut native_added = 0usize;
+    for cert in native_certs {
+        if roots.add(cert).is_ok() {
+            native_added += 1;
+        }
     }
-    roots.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
+
+    if native_added == 0 {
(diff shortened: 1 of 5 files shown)


crates/jazz-tools/src/ws_stream/native.rs
```diff
@@ -1,18 +1,76 @@
...
+// The fallback exists for platforms where `rustls-native-certs` cannot reach
+// the OS trust store at all — most notably Android (anchors live in
+// `AndroidCAStore`, accessible only via JNI), and frequently iOS depending on
+// linkage. Without the fallback, every `wss://` handshake on those platforms
+// fails with `UnknownIssuer`.
+fn root_store() -> rustls::RootCertStore {
+    let native = rustls_native_certs::load_native_certs();
+    let native_error_count = native.errors.len();
+    build_root_store_from_native(native.certs, native_error_count)
+}
```
gdorsi, line 30: Would it make sense to gate this fallback behind a feature flag?

crates/jazz-tools/src/ws_stream/native.rs
```diff
@@ -49,6 +107,20 @@ mod tests {
...
 
+    #[test]
+    fn empty_native_store_falls_back_to_bundled_webpki_roots() {
+        // Mobile case: OS trust store unreachable / empty. The fallback must
+        // populate the store with the bundled Mozilla anchors so wss:// works.
+        let roots = build_root_store_from_native(vec![], 3);
+        assert_eq!(roots.len(), webpki_roots::TLS_SERVER_ROOTS.len());
+    }
+
+    // The "populated native store does not pull in webpki-roots" branch is
```
gdorsi: Could we add a regression test for the populated native store path too, so we lock in that bundled `webpki` roots are not added when native roots load successfully?

crates/jazz-tools/src/ws_stream/native.rs
antoniomuso: Probably not worth a feature flag - the fallback only fires when the native store yields zero anchors, which on desktop already means TLS is broken, so toggling it off doesn't buy more security,
  just a confusing failure mode. If the goal is shaving the ~250KB `webpki-roots` blob from desktop builds, a `cfg(target_os = "android")`-gated fallback (or a mobile-roots cargo feature defaulted on   
  for the mobile-targeting consumers) is cleaner than a runtime flag, since it removes the consumer-side footgun of "forgot to enable it and wss:// silently breaks on Android."

gdorsi (approved):

crates/jazz-tools/src/ws_stream/native.rs
antoniomuso: done!

PR #675, opened 2026-04-27
Preserve schema loader defaults

aeplay:
## Summary

Make the CLI schema loader's wasm-to-AST conversion preserve column fields that the AST already supports:

- column defaults, including arrays, JSON, byte arrays, and null
- counter merge strategies

This keeps `LoadedSchemaProject.schema`, `schemaToWasm(loaded.schema)`, and the typed app's original `wasmSchema` aligned instead of bypassing the lossy AST round-trip.

## Root Cause

`wasmColumnToAst()` only projected `name`, `sqlType`, `nullable`, and `references` from `ColumnDescriptor`. When loading a typed app export, `.default(...)` and `merge("counter")` survived in the app's runtime `wasmSchema`, but were dropped from the AST before the CLI recomputed the schema hash.

## Validation

- `corepack pnpm install`
- `corepack pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/schema-loader.test.ts` (red before the fix, green after)
- `corepack pnpm --filter jazz-tools build`
- `corepack pnpm --filter jazz-tools test`

Note: the first full test run failed before build artifacts existed (`jazz-wasm` / `jazz-napi` resolution failures). After running the package build, the full `jazz-tools` test script passed.

### DIFF
--- a/.changeset/lossless-schema-loader-roundtrip.md
+++ b/.changeset/lossless-schema-loader-roundtrip.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Preserve column defaults and counter merge strategies when loading typed-app schemas through the CLI schema loader, keeping the AST and wasm schema round-trip lossless for fields both representations support.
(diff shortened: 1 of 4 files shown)


gdorsi (approved):
LGTM

PR #679, opened 2026-04-27
Seal direct batches explicitly

aeplay:
## Summary
- make local direct batch records sealable/frozen using the existing sealed submission storage shape
- infer direct vs transactional sealed batch mode from stored row state on the authority side
- retransmit sealed direct submissions after Missing settlements and update the transaction reconciliation spec
- align TypeScript APIs around explicit batch commit handles: open transaction/direct-batch writes return rows or void, while `tx.commit()` and `batch.commit()` return the waitable handle
- keep simple CRUD ergonomic by sealing immediately and returning waitable handles
- add `beginBatch(...)` aliases for explicit direct batches
- locally upgrade retained legacy direct batch records on runtime open by synthesizing the missing direct sealed submission once; already-upgraded records are skipped via the existing `sealed_submission` check

## New API shape

Simple CRUD seals immediately:

```ts
await db.insert(app.todos, { title: "Ship", done: false }).wait({ tier: "edge" });
await db.update(app.todos, todoId, { done: true }).wait({ tier: "edge" });
await db.delete(app.todos, todoId).wait({ tier: "edge" });
```

Explicit direct batches are visible optimistically while open, then frozen by `commit()`:

```ts
const batch = db.beginBatch(app.todos);

const created = batch.insert(app.todos, { title: "Grouped", done: false });
batch.update(app.todos, created.id, { done: true });

await batch.commit().wait({ tier: "edge" });
```

Transactions stage writes until accepted by the authority; only the committed batch is waitable:

```ts
const tx = db.beginTransaction(app.todos);

const created = tx.insert(app.todos, { title: "Transactional", done: false });
tx.update(app.todos, created.id, { done: true });

await tx.commit().wait({ tier: "global" });
```

## Compatibility
- Existing visible direct rows remain visible.
- On local runtime open, retained direct `LocalBatchRecord`s with members and no sealed submission are upgraded by computing the direct `SealedBatchSubmission` from retained members, marking the record sealed, and persisting it in the existing storage shape.
- Already-upgraded records are no-ops because `sealed_submission` is present.
- Currently open direct batches in a running runtime are not upgraded by inspection; they still require explicit `commit()`.

## Tests
- pnpm lint
- pnpm format:check
- pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts src/runtime/db.transaction.test.ts src/runtime/client.test.ts src/runtime/client.mutations.test.ts src/runtime/client.for-request.test.ts
- cargo test -p jazz-tools seal_batch --lib
- cargo test -p jazz-tools direct_batch --lib
- cargo test -p jazz-tools rc_startup_upgrades_legacy_direct_batch_records_with_sealed_submission --lib
- cargo test -p jazz-tools rc_local_batch_record_does_not_seal_current_open_direct_batch --lib
- cargo test -p jazz-tools rc_sealed_direct_batch_rejects_further_writes --lib
- cargo test -p jazz-tools rc_persisting_invalid_multibranch_sealed_batch_submission_fails --lib

## Local limitations
- Full TypeScript/build checks are blocked locally by missing `jazz-wasm` resolution/native artifacts.


gdorsi:
Nit: `visible_members` has been a little hard for me to parse. “Members” reads a bit user/member-ish at first glance, but this field is really describing the row contents covered by a batch/transaction settlement.

I don’t have a perfect rename, but something like `settled_versions`, `settled_content`, or maybe a `visible_*` name that points more directly at row versions/content might make the shape easier to understand.

gdorsi:
Do you think it makes sense to add a test for the mixed-state rejection case?

The code rejects a sealed batch if its referenced rows infer different modes, e.g. one row is `VisibleDirect` and another is `StagingPending`, with reason `sealed batch mixes direct and transactional rows`. I see coverage for all-direct and all-transactional batches, plus other invalid submissions, but not this specific invariant.

gdorsi (approved):
Looks good to me. The direct/transactional sealing model is coherent, and the API changes line up with the new “commit returns the waitable handle” contract. I left a couple of small naming/test-coverage comments, but nothing blocking.

gdorsi:
CI looks like real failures rather than flaky tests.

- `test-ts` fails during `pnpm build:ci` in `packages/jazz-tools` before tests run:
  ```text
  src/runtime/client-tests/support.ts(47,14): error TS2883: The inferred type of 'runtimeBatchRecordStubs' cannot be named without a reference to 'Procedure' from '.pnpm/@vitest+spy@4.1.0/node_modules/@vitest/spy'. This is likely not portable. A type annotation is necessary.
  ```
  `expo-android-e2e` is failing for the same reason while running `pnpm --filter jazz-tools run build:runtime`, so it never reaches the Android E2E part.

- `test-rust` has one deterministic assertion mismatch:
  ```text
  runtime_core::tests::write_batch::rejection_recovery::rc_persisting_invalid_multibranch_sealed_batch_submission_fails
  left:  reason: "sealed batch rows must belong to the declared target branch"
  right: reason: "sealed transactional batch rows must belong to the declared target branch"
  ```
  This looks like the test expectation still has the old transactional-only wording after the error was generalized.

There is also an external Vercel `jazz2-inspector` failure, but I didn’t inspect provider logs here; the GitHub Actions failures above are enough to explain the red CI.

PR #682, opened 2026-04-27
Add support for multiple connections in inspector

tobiaslins:
image


Takeno (approved):
Thanks for your contribution!

PR #693, opened 2026-04-28
Add transformed columns

aeplay:
## What changed

Adds TypeScript DSL support for transformed columns via `.transform({ from, to })` on normal column definers such as `s.string()`, `s.boolean()`, `s.int()`, refs, arrays, JSON, and enums.

- Tracks transformed column value types through `RowOf` and `InsertOf`.
- Applies `from` when rows are returned from reads/subscriptions/inserts.
- Applies `to` before inserts and updates.
- Keeps `where` inputs based on the stored/raw column type, since arbitrary transforms cannot be translated into SQL predicates.
- Documents transformed columns at the end of the Column Types reference page.

## Validation

- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts tests/ts-dsl/typed-app.test.ts`
- Pre-commit hook passed: `oxfmt`, `oxlint`

## Notes

- Merge conflicts with `main` are resolved.
- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts tests/ts-dsl/transformed-columns.test.ts` is still blocked in this local checkout because `jazz-wasm` package resolution fails before the test body executes.


docs/content/docs/reference/column-types.mdx
```diff
@@ -9,3 +9,49 @@ import AvailableColumnTypes from "../../partials/available-column-types.mdx";
...
+
+db.update(app.tasks, task.id, {
+  priority: "medium",
+});
+
+const task = await db.one(app.tasks.where({ id: task.id }));
+task?.priority; // "low" | "medium" | "high"
+```
+
+Filters still use the stored column value, because arbitrary transforms cannot be translated into SQL predicates:
```
nrainhart, line 51: This could be quite confusing. I feel like ideally we should restrict the supported operations to something like https://cel.dev/ (I'd love to support a subset of TS that compiles to CEL), and used the transformed value everywhere (except for storage).

Still, we can keep this as is for now and explore changing it in the future.

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -2206,7 +2272,8 @@ export class Db {
     options: UpsertOptions,
   ): WriteHandle {
     const client = this.getClient(table._schema);
-    const values = toUpdateRecord(data as Record<string, unknown>, table._schema, table._table);
+    const transformedData = transformUpdateInput(table, data);
```
nrainhart, line 2275: Do we want to keep this transforms scoped to the `Db` layer? Wouldn't it make sense to move this into the `JazzClient`?

nrainhart (approved):

PR #703, opened 2026-04-28
[codex] Add callback Db batch APIs

aeplay:
## What changed

- Replaced the typed `Db` table-seeded transaction and batch entry points with callback-scoped `db.transaction(...)` and `db.batch(...)` helpers.
- The callback handles lazily bind to the first table/query used, then automatically commit and return the committed `WriteHandle` so callers can use `.wait(...)` directly.
- Updated transaction/batch tests and status-quo docs to describe the new typed API.

## Why

`Db.beginTransaction(table)` and `Db.beginBatch(table)` implied transactions and batches were table-scoped, but both can cover multiple tables. The new callback API reflects the real scope: one database/runtime context, many table writes.

## Validation

- `pnpm exec vitest run src/runtime/db.transaction.test.ts --config vitest.config.ts`
- `pnpm exec vitest run src/runtime/client.test.ts --config vitest.config.ts`
- `pnpm exec tsc --noEmit` currently stops on existing unresolved local `jazz-wasm` module types in `src/runtime/client.ts` and `src/worker/jazz-worker.ts`.

### DIFF
--- a/.changeset/codex-db-callback-batches.md
+++ b/.changeset/codex-db-callback-batches.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Replace the typed `Db` table-seeded transaction and batch APIs with callback-scoped helpers that automatically commit and return a waitable write handle.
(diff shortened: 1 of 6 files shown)


aeplay:
Closing in favor of #724

PR #706, opened 2026-04-29
create-jazz spinner UX + jazz-tools CLI dotenv support

joeinnes:
## Summary

Enhancements to CLI experience.

### `create-jazz` provisioning UX
- Advance the spinner to "Provisioning Jazz Cloud app" before the dashboard HTTP round-trip, so it no longer appears stuck on "Initialising git" during the network wait.
- Route credential/warning output through an injected `onLog` hook that the CLI replays via clack's `log.info` / `log.warn` after the spinner stops — the success banner now lands on its own line.
- Gate `pnpm test` on `tsc --noEmit` so type errors fail CI (tsup and vitest were both transpile-only).

### `jazz-tools` CLI dotenv support
- The CLI now loads `.env` from the working directory at startup, so `deploy` (and any other command that reads `JAZZ_ADMIN_SECRET` / `JAZZ_SERVER_URL` / framework-prefixed equivalents) picks up values from the same dotenv file your app uses.
- Real environment variables still take precedence; a missing `.env` is a no-op.

## Test plan
- [ ] `pnpm --filter create-jazz test` passes (typecheck + 64 tests).
- [ ] `pnpm --filter jazz-tools vitest run --config vitest.config.ts src/cli.test.ts -t loadDotEnv` passes (5 tests).
- [ ] In a project containing a `.env` with `JAZZ_ADMIN_SECRET=…`, run `npx jazz-tools deploy` without exporting the var — confirm the deploy succeeds.
- [ ] Run `pnpm create jazz` end-to-end and confirm the spinner advances during provisioning and the success banner prints on its own line.

### DIFF
--- a/packages/create-jazz/package.json
+++ b/packages/create-jazz/package.json
@@ -22,7 +22,8 @@
   "scripts": {
     "build": "tsup src/index.ts --format esm --out-dir dist",
     "prepublishOnly": "pnpm build",
-    "test": "vitest run"
+    "typecheck": "tsc --noEmit",
+    "test": "pnpm typecheck && vitest run"
   },
   "dependencies": {
     "@clack/prompts": "^0.10.0",
--- a/packages/create-jazz/package.json
+++ b/packages/create-jazz/package.json
@@ -22,8 +22,8 @@
   "scripts": {
     "build": "tsup src/index.ts --format esm --out-dir dist",
     "prepublishOnly": "pnpm build",
-    "typecheck": "tsc --noEmit",
-    "test": "pnpm typecheck && vitest run"
+    "check": "tsc --noEmit",
+    "test": "pnpm check && vitest run"
   },
   "dependencies": {
     "@clack/prompts": "^0.10.0",
--- a/packages/jazz-tools/src/cli.ts
+++ b/packages/jazz-tools/src/cli.ts
@@ -2,6 +2,7 @@
 
 // CLI for jazz-tools schema tooling
 
+import { existsSync, readFileSync } from "fs";
 import { access, mkdir, readFile, readdir, rm, writeFile } from "fs/promises";
 import { basename, dirname, join, resolve } from "path";
 import { pathToFileURL } from "url";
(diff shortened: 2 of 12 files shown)


packages/create-jazz/package.json
```diff
@@ -22,7 +22,8 @@
   "scripts": {
     "build": "tsup src/index.ts --format esm --out-dir dist",
     "prepublishOnly": "pnpm build",
-    "test": "vitest run"
+    "typecheck": "tsc --noEmit",
```
nrainhart: Super nit: in Jazz1 we usually named this command just `check`

packages/jazz-tools/src/cli.ts
```diff
@@ -201,6 +202,33 @@ export const APP_ID_ENV_VARS = [
   "EXPO_PUBLIC_JAZZ_APP_ID",
 ] as const;
 
+// Load values from `.env` in the given directory into `env` without
+// overriding entries that are already set. Real environment variables
+// always win — `.env` is a fallback only. No-op when the file is absent.
+export function loadDotEnv(
```
nrainhart: Nit: should we use https://nodejs.org/api/process.html#processloadenvfilepath instead?

nrainhart (approved):
This is great! Left only a few minor comments.

I'm also wondering if we should allow setting a custom `.env` file (e.g. `.env.local`) as well, for users that work with different Jazz servers in different envs. Actually, maybe we can just use Node's `--env-file` when running the CLI with node/npx?

joeinnes:
Technically we support `"node": ">=20"` and `process.loadEnvFile` wasn't added until 20.12.0, so I've kept the custom parser as a fallback but prefer `process.loadEnvFile` when it's available. 

Also, because users run this through a bin shim, we'd have to do extra work to pass `--env-file` through to the underlying node process, and their version may not even accept it, that flag was introduced in 20.6.0. I've added a flag that goes through the same loader instead (built-in when possible, custom parser otherwise).

PR #707, opened 2026-04-29
Carry resolved table name into MaterializeNode element

joeinnes:
## Summary
- MaterializeNode's output-descriptor element table propagates as the table_hint to the row loader and into LensTransformer::new. When that element was an empty string, old-branch rows fell into TableNotFound("") and were silently dropped — appearing to users as data vanishing after a schema migration.
- MemoryStorage masked the bug because it keys row history by table name, so the empty hint short-circuited to the locator fallback. Storage backends that resolve rows by locator (RocksDB) hit it directly.
- Pass the resolved table name in the schema-aware compile path and in the policy USING/EXISTS subgraphs. Adds a graph-compile regression test that catches the empty hint regardless of storage backend.

## Test plan
- [ ] `cargo test -p jazz-tools --lib materialize_nodes_carry_concrete_table_name` — passes against this branch, fails against main
- [ ] `cargo test --workspace` — full Rust workspace stays green (1322 passed locally)
- [ ] Reproduce the original symptom: ship a v1 → v2 schema migration on a Jazz app with persistent old-schema rows on a RocksDB-backed server; confirm old rows now project through the lens instead of being dropped with the misleading "run schema export … migrations create" warning


crates/jazz-tools/src/schema_manager/integration_tests/tests/migration.rs
```diff
@@ -442,3 +442,67 @@ fn end_to_end_multi_hop_transform() {
...
+    let mut materialize_count = 0;
+    for compact in &graph.nodes {
+        let GraphNode::Materialize(node) = &compact.node else {
+            continue;
+        };
+        materialize_count += 1;
+        let desc = node.output_tuple_descriptor();
+        for index in 0..desc.element_count() {
+            let element = desc.element(index).expect("element index in range");
+            assert!(
```
nrainhart, line 491: This test is pretty implementation dependent. Do you think we could have a test that repros the end-user bug we were previously experiencing?

nrainhart (approved):
Good catch!

joeinnes:
Added a more thorough behavioural test.

PR #712, opened 2026-04-29
refactor(jazz-tools): Phase 8 (row_histories) — split mod.rs along type/codec/apply seams

antoniomuso:
## Summary

Phase 8 of [crate structure cleanup](specs/todo/c_later/crate_structure_cleanup.md), `row_histories` half. `row_histories/mod.rs` was ~3.4K LOC mixing data types, descriptor codecs, visibility resolution, and storage-mutating verbs; this splits it along natural seams. **Independent of every other phase.**

| File | Size | Contents |
| --- | --- | --- |
| `row_histories/mod.rs` | ~1.2K (mostly tests) | module docs, sibling re-exports, the existing tests block |
| `row_histories/types.rs` | ~660 | `BatchId`, `RowState`, `HistoryScan`, `RowVisibilityChange`, `ApplyRowBatchResult`, `RowHistoryError`, `QueryRowBatch`, `StoredRowBatch`, `RowMetadata`, `VisibleRowEntry` (and impl) |
| `row_histories/codecs.rs` | ~910 | descriptor builders + flat-row encoders/decoders + shared utilities (`compute_row_digest`, `tier_satisfies`, `malformed`, `flat_user_values`, `project_user_row_data_from_physical`) |
| `row_histories/resolution.rs` | ~460 | pure visibility/merge math: frontier walk, latest common ancestor, per-column merge under `ColumnMergeStrategy`, delete-winner overlay, computed visible preview, `branch_frontier`, `latest_visible_version_for_tier`. No `Storage` access. |
| `row_histories/mutations.rs` | ~365 | the storage-mutating verbs `apply_row_batch` and `patch_row_batch_state`, plus their direct support (`load_branch_history`, `rebuild_visible_entry_from_history`, `supersede_older_staging_rows_for_batch`, `visibility_change_from_applied`) and the `AppliedRowBatch` capture struct |

The split happened in two passes on this branch: first the original three-way carve (`types` / `codecs` / `apply`), then `apply.rs` → `resolution.rs` + `mutations.rs` once it was clear the file's contents fell into two genuinely different concerns — pure data → data resolution math, and storage orchestration around the public verbs. As part of the second pass the internal `RowBatchApply` capture struct was renamed to `AppliedRowBatch` (it's a captured outcome, not the operation).

Dependency direction is one-way: `mutations` → `resolution`, and `types` also pulls a few helpers from `resolution` for `VisibleRowEntry::rebuild_*`.

Cross-file helpers are exposed as `pub(super)`, so the surface beyond the module is unchanged. Public re-exports (`row_histories::apply_row_batch`, `::compute_row_digest`, `::StoredRowBatch`, etc.) and `pub(crate)` items (`flat_row_codecs`, `decode_flat_*_with_codecs`, `FlatRowCodecs`, `visible_row_preview_from_history_rows`) all keep their existing visibility.

The `routes.rs` half of Phase 8 is deferred — it depends on Phase 1 ([#709](https://github.com/garden-co/jazz2/pull/709)) landing the `routes.rs` → `server/routes.rs` move first.

## Test plan

- [x] `cargo build --all-features -p jazz-tools` (clean)
- [x] `cargo build -p jazz-napi -p jazz-wasm -p jazz-rn`
- [x] `cargo test --all-features -p jazz-tools` — 1113 lib tests + integration tests all green
- [x] `cargo clippy -p jazz-tools --lib` (clean)


crates/jazz-tools/src/row_histories/apply.rs
```diff
@@ -0,0 +1,784 @@
+//! apply_row_batch, patch_row_batch_state, and the visible-row computation helpers
+//! they pull in.
```
nrainhart: Love these comments on top of each file! I think we should aim to have them in most files, to make it clearer where each thing should go.

This one in particular feels a bit vague though. I'm not quite sure what an "apply algorithm" is, can we add a note about that here?

nrainhart (approved):

crates/jazz-tools/src/row_histories/apply.rs
antoniomuso: Done, I have split the `apply.rs` file into `resolution.rs` and `mutation.rs
`

PR #715, opened 2026-04-29
Reload browser on successful schema watch push

joeinnes:
## Summary
- Add an `onSchemaPush` callback to `ManagedDevRuntime.initialize`, fired after a schema watch push succeeds
- Wire the Vite and SvelteKit dev plugins to send a `full-reload` over the dev server's WebSocket from that callback, so the browser picks up the new schema without a manual refresh

## Test plan
- [ ] `pnpm --filter jazz-tools exec vitest run src/dev/sveltekit.test.ts src/dev/vite.test.ts` — new "sends a full-reload to the browser on a successful schema watch push" tests pass for both plugins
- [ ] In a Vite or SvelteKit example app, edit `schema.ts` while `pnpm dev` is running and confirm the browser reloads on its own

### DIFF
--- a/.changeset/reload-on-schema-push.md
+++ b/.changeset/reload-on-schema-push.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Trigger a Vite full-reload from the Jazz Vite and SvelteKit dev plugins
+whenever the schema watcher successfully pushes an updated schema, so the
+browser picks up the new schema without a manual refresh.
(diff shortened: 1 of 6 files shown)


gdorsi (approved):
Love this!

What about Next.js?

joeinnes:
> Love this!
> 
> What about Next.js?

Requires a fully different approach, followed up with https://github.com/garden-co/jazz2/pull/735

PR #724, opened 2026-04-29
feat: add `Db.transaction` & `Db.batch`

nrainhart:
## Description

Adds two new methods to Db & JazzClient: `transaction` and `batch`. These methods run a callback as part of a transaction/batch and commit it automatically once the callback finishes running:

```ts
const result = await db.transaction(async (tx) => {
  tx.insert(app.todos, { title: "Draft copy", done: false });
  const stagedDrafts = await tx.all(app.todos.where({ done: false }));
  tx.update(app.todos, todoId, { done: true });
});
await result.wait({ tier: "edge" });
```

This is now the recommended way of running batches and transactions.

This PR also:
- ensures changes made as part of a transaction are not visible until the transaction is committed (bug uncovered while testing this feature)
- stops requiring a table when creating transactions and batches. The schema is determined by the first CRUD operation performed inside the transaction/batch.
- renames `InsertHandle` to `WriteResult`, since these objects can now be returned by the insert API and also batches/transactions.


aeplay (approved):

packages/jazz-tools/src/runtime/db.transaction.test.ts
```diff
@@ -453,7 +614,168 @@ describe("Db transactions", () => {
...
+    expect(runtimeBatch.commit).toHaveBeenCalledTimes(1);
+    await expect(handle.wait({ tier: "edge" })).resolves.toEqual({
+      id: "todo-direct-callback",
+      title: "Callback batch",
+      done: false,
+    });
+    expect(client.waitForPersistedBatch).toHaveBeenCalledWith("batch-direct-callback", "edge");
+  });
+
+  it("does not commit a callback batch when the callback rejects", async () => {
```
gdorsi, line 670: Can we add a test for `db.batch(...)` failing after a write? The current rejection tests fail before any write, so they don’t lock down visibility/sealing semantics for partial direct batches.

packages/jazz-tools/tests/browser/db.transaction-reads.test.ts
```diff
@@ -65,8 +65,8 @@ describe("db transaction reads browser integration", () => {
       }),
     );
 
-    const aliceTx = db.beginTransaction(todos);
-    const bobTx = db.beginTransaction(todos);
+    const aliceTx = db.beginTransaction();
```
gdorsi, line 68: Could we add browser integration coverage for `db.transaction(...)` / `db.batch(...)`? The browser tests still only exercise `beginTransaction()`.

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -2310,46 +2382,70 @@ export class Db {
    * Use transactions when several writes should settle together after an authority validates them.
    *
    * Use {@link DbTransaction.commit} to commit the transaction.
+   *
+   * Prefer using {@link Db.transaction} when an explicit commit is not required.
    */
-  beginTransaction<T, Init>(table: TableProxy<T, Init>): DbTransaction {
-    const client = this.getClient(table._schema);
+  beginTransaction(): DbTransaction {
```
gdorsi, line 2392: Do we still need public `beginTransaction()` / `beginBatch()` now that the callback APIs are preferred? If yes, it may be worth documenting when to use the manual commit form.

gdorsi (approved):
LGTM

packages/jazz-tools/src/runtime/db.ts
nrainhart: Yes, there are some cases where this API is more convenient. E.g. in tests, you could begin a transaction as part of `beforeEach` to ensure the db changes made as part of a test are isolated

nrainhart: i'll add the docs as a fup, as the example i've got in mind (isolated writes in tests) makes more sense once we have a rollback operation

PR #725, opened 2026-04-29
Add explicit rollback cancelation for transactional batches

Takeno:
## Summary
- Add `rollback()` support on explicit transactions and wire it through client/db runtime APIs.
- Send an explicit `CancelBatch` sync message on rollback so servers discard still-pending transactional rows without creating a rejected settlement.
- Update sync handling, tracing, and transactional tests to cover rollback cleanup, upstream cancellation, and multi-client visibility.

## Testing
- Unit and integration coverage added for rollback behavior in Rust and TypeScript.
- Browser transaction-read integration validates that rolled-back staged rows disappear from ordinary reads and do not leak to other clients.
- Local verification passed for the targeted Rust, TS, browser, formatting, and type-check suites.


aeplay:
🤖 Closing because explicit rollback behavior and coverage are already present on the integration branch. The legacy `CancelBatch` protocol implementation belongs to the retired engine.

PR #730, opened 2026-04-29
Make transaction visibility strict by default

aeplay:
## What changed

- Removed the public/internal `strictTransactions` / `strict_transactions` query option across TS, WASM, NAPI, and Rust runtime surfaces.
- Made transaction visibility filtering unconditional in the query manager.
- Changed ordinary `Db` reads and `subscribeAll` to use deferred local overlays by default, so staged transactional rows do not leak into global reads.
- Kept explicit transaction-scoped reads (`tx.all` / local transaction overlay) able to see their own staged rows.
- Updated browser and runtime tests for Nico's repro: two open transactions see their own drafts, while `db.all(...)` returns `[]` before commit.
- Updated the transaction visibility spec language so it no longer describes an opt-in strict mode.

## Why

Nico found that open transactional inserts were globally visible before commit, even with `strictTransactions: true`. The strict option only controlled accepted transaction completeness; staged rows were still able to enter ordinary reads through the immediate local overlay path.

This makes strict transaction visibility the only behavior and removes the opt-out API entirely.

## Validation

- `pnpm lint`
- `pnpm build:ci`
- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.browser.ts tests/browser/db.transaction-reads.test.ts --reporter verbose`
- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.browser.ts tests/browser/db.transaction-reads.test.ts`
- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts src/runtime/client-tests/for-request.test.ts src/runtime/db.transaction.test.ts`
- `cargo test -p jazz-tools runtime_core::tests::query_subscription -- --nocapture`
- `cargo check -p jazz-tools -p jazz-wasm -p jazz-napi`
- `cargo fmt --all --check`
- `git diff --check`

Note: `pnpm lint` reports existing oxlint warnings in unrelated files, but exits successfully; clippy passes.


nrainhart:
Cherry-picked these changes into https://github.com/garden-co/jazz2/pull/724

PR #739, opened 2026-04-30
chat-react: migrate Playwright e2e to vitest browser; drop act()

joeinnes:
## Summary
- Move the auto-join race regression test from `e2e/invite-auto-join.spec.ts` to `tests/browser/auto-join-race.test.tsx`, so chat-react has a single test entry point (`pnpm test`) instead of two parallel stacks.
- Delete the e2e infrastructure that becomes redundant: `e2e/` directory, `playwright.config.ts`, and the `test:e2e` script. `@playwright/test` stays as a devDep — `walkthrough:shots` still uses it for the docs slideshow.
- Strip `act(...)` calls from every browser test. The suite relies on `waitFor()` polling, not React's flush guarantees, so `act()` was producing log noise (`The current testing environment is not configured to support act(...)`) without earning its keep.

## Test plan
- [ ] `cd examples/chat-react && pnpm test` — 8 files, 24 tests should pass with no `act(...)` warnings in the output.
- [ ] `pnpm walkthrough:shots --help` (or any Playwright invocation) — confirms `@playwright/test` is still wired up for screenshot capture.
- [ ] `pnpm test:e2e` — should now fail as "script not found", confirming the cleanup landed.


examples/chat-react/tests/browser/auto-join-race.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+      } catch {
+        /* best effort */
+      }
+      container.remove();
+    }
+    mounts.length = 0;
+    window.location.hash = "";
+    // Allow JazzProvider's async shutdown (worker termination, OPFS lock
+    // release) to complete before the next test starts.
+    await new Promise((r) => setTimeout(r, 1000));
```
nrainhart: Now that we expose the Jazz client via `window.__jazz`, could we use `await window.__jazz.shutdown()` here? Or something similar.

Not a blocker for merging this PR though

nrainhart (approved):
Thanks for cleaning this up!! ✨

PR #740, opened 2026-04-30
Update permissions docs for preflight helpers

Takeno:
## Summary

Adds local permission preflight APIs for write UI gating:

- `db.canInsert(...)` / `db.canUpdate(...)` returning `true | false | "unknown"`
- Runtime support through Rust core, WASM, NAPI, and React Native bindings
- Framework helpers for React/React Native, Vue, and Svelte
- Docs and chat examples showing permission-aware insert buttons with draft data


aeplay:
🤖 Closing as represented by the new-core permission-advice work, including #1349 and the current `db.canInsert` / `canInsertLocally` APIs. The former tri-state preflight design should not be ported.

PR #741, opened 2026-04-30
Add transaction rollback support and docs

Takeno:
## Summary
- Add `rollback()` to transaction handles and document the user-facing semantics
- Clarify that rollback closes the transaction handle without sealing the batch or removing staged pending rows
- Add a follow-up issue for cleaning up unsealed pending rows
- Include a `jazz-tools` changeset for the API addition

## Testing
- Not run (not requested)


packages/jazz-tools/src/runtime/db.ts
```diff
@@ -592,7 +590,6 @@ export class DbTransaction {
    * once it's committed with {@link DbTransaction.commit}.
    */
   insert<T, Init>(table: TableProxy<T, Init>, data: Init): T {
-    this.ensureActive();
```
nrainhart, line 602: This check is redundant because it's already made by the JazzClient, right?

packages/jazz-tools/src/runtime/client.test.ts
```diff
@@ -313,6 +313,38 @@ describe("JazzClient transactions", () => {
...
 
+  it("rolls back an open transaction without sealing the batch", () => {
+    const runtime = makeFakeRuntime();
+    const client = JazzClient.connectWithRuntime(runtime as any, makeContext());
+    const tx = client.beginTransaction();
+
+    tx.rollback();
+
+    expect(runtime.sealBatch).not.toHaveBeenCalled();
+    expect(() => tx.commit()).toThrow(/rolled back/i);
```
nrainhart, line 321: We need to integrate `tx.rollback()` with `db.transaction`/`client.transaction`:
- if an error is thrown inside the transaction callback, we should call `tx.rollback()` (this is mostly a nit, it doesn't make a difference in practice because the user can't get a hold of the transaction object anyway)
- you shouldn't be able to call `tx.rollback()` explicitly inside the callback

nrainhart (approved):
Looks good! Left only one comment I think we need to address

packages/jazz-tools/src/runtime/db.ts
Takeno: Yes, I moved all the transaction checks inside the runtime.

packages/jazz-tools/src/runtime/client.test.ts
Takeno: Implemented the tx.rollback call (mostly useless atm, but it already in place for future behaviors).

About the second point, since users have the `tx` object that can be passed around, isn't it an additional way to control the transaction flow instead of propagating errors and dealing with them?

nrainhart: hmmmm what i don't like is that you could rollback, and then keep doing stuff inside the callback. Which is very likely a footgun, because all db operations will fail.

Or maybe it's not that bad and we should just need to ensure folks use `return tx.rollback()` inside the callback 🤔. What do you think?

PR #745, opened 2026-04-30
Preserve row parent ordering when syncing to servers

aeplay:
## Summary
- queue unsent parent row batches before child batches when forwarding rows to servers
- use the parent-closure path for full sync, local writes, and server-relayed row batches
- add a regression test for a skipped upstream-confirmed parent followed by a local child
- add a patch changeset for jazz-tools

## Test
- cargo test -p jazz-tools server_sync

Note: the changeset amend bypassed lefthook because the fresh isolated worktree had no node_modules/oxfmt installed; the Rust test suite above passed.

### DIFF
--- a/crates/jazz-tools/src/sync_manager/sync_logic.rs
+++ b/crates/jazz-tools/src/sync_manager/sync_logic.rs
@@ -50,7 +50,23 @@ impl SyncManager {
         }
 
         for (object_id, metadata, row) in row_sync {
-            self.queue_row_to_server(server_id, object_id, metadata, row);
+            let Some(table) = metadata
+                .get(crate::metadata::MetadataKey::Table.as_str())
+                .cloned()
+            else {
+                self.queue_row_to_server(server_id, object_id, metadata, row);
+                continue;
+            };
+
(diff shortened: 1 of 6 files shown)


crates/jazz-tools/src/sync_manager/sync_logic.rs
```diff
@@ -50,7 +50,23 @@ impl SyncManager {
...
 
         for (object_id, metadata, row) in row_sync {
-            self.queue_row_to_server(server_id, object_id, metadata, row);
+            let Some(table) = metadata
+                .get(crate::metadata::MetadataKey::Table.as_str())
+                .cloned()
+            else {
+                self.queue_row_to_server(server_id, object_id, metadata, row);
+                continue;
+            };
```
nrainhart: Is there an actual case where a row may be missing its table metadata? If not, should we just throw an error?

nrainhart (approved):

crates/jazz-tools/src/sync_manager/sync_logic.rs
aeplay: addressed!

PR #747, opened 2026-04-30
Add batch transaction write API parity

aeplay:
## What changed

- Added custom insert id support to `DbTransaction.insert` and `DbDirectBatch.insert`.
- Added `upsert` to transaction and direct batch write scopes, wired through runtime batch contexts.
- Kept upsert behavior aligned with the simple write API, including partial updates for existing rows.

## Why

The simple write API supported custom ids and upsert, but batch and transaction scopes did not expose those capabilities. This made grouped writes less ergonomic and forced callers to special-case write paths.

## Validation

- `pnpm --filter jazz-wasm build`
- `pnpm --dir packages/jazz-tools build:runtime`
- `pnpm --dir packages/jazz-tools exec vitest run tests/browser/db.transaction-reads.test.ts --config vitest.config.browser.ts`
- `pnpm lint`

`oxlint` reports existing warnings, including an unused import in `packages/jazz-tools/src/runtime/client.ts`, but exits successfully; `cargo clippy --workspace -- -D warnings` passes.

### DIFF
--- a/packages/jazz-tools/src/runtime/client.ts
+++ b/packages/jazz-tools/src/runtime/client.ts
@@ -691,11 +691,6 @@ function isObjectAlreadyExistsError(error: unknown): boolean {
   return message.includes("object already exists") || message.includes("Create failed: Conflict");
 }
 
-function isInsertShapeValidationError(error: unknown): boolean {
-  const message = error instanceof Error ? error.message : String(error);
-  return message.includes("missing required field");
-}
-
 type BatchWriteContext = {
   batchMode: BatchMode;
   batchId: string;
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/runtime/client.ts
```diff
@@ -1112,7 +1144,7 @@ export class SessionClient {
       await this.create(table, values, options);
       return;
     } catch (error) {
-      if (!isObjectAlreadyExistsError(error)) {
+      if (!isObjectAlreadyExistsError(error) && !isInsertShapeValidationError(error)) {
```
nrainhart: ⚠️ Does it make sense to update the row after `isInsertShapeValidationError` is thrown?

nrainhart (approved):

PR #759, opened 2026-05-01
[codex] Budget remote sync ticks

tobiaslins:
## What

Caps how much remote sync work a single `RuntimeCore::batched_tick()` applies before yielding back to the scheduler.

- Limit transport inbox draining to `MAX_TRANSPORT_MESSAGES_PER_BATCHED_TICK`.
- Limit parked sync message application to `MAX_SYNC_MESSAGES_PER_BATCHED_TICK`.
- Preserve sequenced-message ordering and reschedule another batched tick when ready work remains.
- Store unsequenced parked messages in a `VecDeque` so bounded draining does not shift the whole buffer.
- Add a regression test and a Criterion benchmark for parked sync backlog cost per event-loop turn.

## Why

The browser traces showed long message handlers / runtime ticks after bulk `useAll` loads. The previous subscription fix removed the idle clean-subscription recompute freeze, but a large remote backlog can still be applied in one browser turn. This PR does not reduce total sync work; it budgets it across multiple scheduled ticks so the UI thread gets chances to breathe.

## Benchmarks

Benchmark: `cargo bench -p jazz-tools --bench batched_tick_budget_benchmark -- --quick`

Measures one `batched_tick()` call with a parked `QuerySettled` backlog. `origin/main` had the benchmark file temporarily added for comparison only.

| Backlog | main | this PR | per-tick improvement |
| ---: | ---: | ---: | ---: |
| 1,000 | 70.399 us | 10.844 us | 6.5x |
| 10,000 | 1.1503 ms | 15.298 us | 75.2x |
| 100,000 | 20.316 ms | 58.604 us | 346.7x |

## Validation

- `cargo test -p jazz-tools batched_tick_applies_parked_sync_messages_in_bounded_slices --lib`
- `cargo test -p jazz-tools runtime_core::tests --lib`
- `cargo fmt --check`
- `git diff --check`
- pre-commit hook: clippy + rustfmt

### DIFF
--- a/crates/jazz-tools/Cargo.toml
+++ b/crates/jazz-tools/Cargo.toml
@@ -148,6 +148,10 @@ harness = false
 name = "limit_offset_benchmark"
 harness = false
 
+[[bench]]
+name = "batched_tick_budget_benchmark"
+harness = false
+
 [[example]]
 name = "realistic_bench"
 required-features = ["client"]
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Closing because the old remote-tick budgeting implementation targets the retired runtime. Its responsiveness goal is tracked by the current interruptible/async direction, including #1661.

PR #761, opened 2026-05-01
[codex] Fix EXISTS policy sync context

### LINKED ISSUE #742: Client query permissions only resolve after querying dependency table first
We’re seeing a permissions-resolution issue with Jazz 2 alpha where a client query returns no rows unless we first query the table used inside the permission predicate.

### Setup

We have an admin-only dashboard with an `adminMembers` table. Admin access to other tables is derived from the current session user having an active row in `adminMembers`.

Relevant `permissions.ts`:

```ts
import { schema as s } from 'jazz-tools-v2'
import { app } from './schema'

export default s.definePermissions(app, ({ policy, session }) => {
  policy.adminMembers.allowRead.where({
    userId: session.user_id,
    active: true
  })

  const isAdmin = policy.adminMembers.exists.where({
    userId: session.user_id,
    active: true
  })

  policy.adminKnownOrgs.allowRead.where(isAdmin)
  policy.adminKnownOrgs.allowInsert.never()
  policy.adminKnownOrgs.allowUpdate.never()
  policy.adminKnownOrgs.allowDelete.never()

  policy.runHistory.allowRead.where(isAdmin)
  policy.runHistory.allowInsert.never()
  policy.runHistory.allowUpdate.never()
  policy.runHistory.allowDelete.never()
})
```

### What happens

This returns no rows on the client:

```ts
const recentOrgs = useAll(
  app.adminKnownOrgs.orderBy('createdAt', 'desc').limit(5)
)
```

But if we also query `adminMembers` first/in the same component:

```ts
const adminRows = useAll(app.adminMembers)
const recentOrgs = useAll(
  app.adminKnownOrgs.orderBy('createdAt', 'desc').limit(5)
)
```

then `recentOrgs` starts returning rows as expected.

The backend can read the same data fine with `context.asBackend()`, and the deployed permissions head targets the same structural schema hash as the client/backend (`836338d3fd5e`). The user id shown by the client also matches an active `adminMembers` row.

### Expected

`adminKnownOrgs` should be readable whenever the `exists` predicate on `adminMembers` matches the current session, without needing the frontend to explicitly query/load `adminMembers`.

### Actual

The `exists`-based permission appears not to resolve on the client until `adminMembers` has been queried directly.

tobiaslins:
## What

Fixes sync-backed client queries whose read policy depends on an `EXISTS` predicate over another table.

The server now includes the readable witness rows needed for the client to locally reproduce the policy decision, without requiring the client to explicitly query the predicate table first.

## Why

A client could receive no rows for a protected table when the policy depended on rows from a separate access/membership table, because those predicate rows were not part of the server query scope. The local client policy filter then failed closed.

## Details

- Collects successful policy witness scope from `EXISTS`, `EXISTS_REL`, and inherited policy evaluation.
- Merges only readable proof-chain rows into the sync scope.
- Withholds protected row bytes when the matching predicate rows are not readable.
- Recompiles/reevaluates subscriptions when upstream settled scope introduces branches needed for local evaluation.
- Adds sync-backed regression tests with generic `access_grants` / `protected_records` fixtures.

Resolves #742

## Checks

- `cargo test -p jazz-tools sync_backed_exists_policy -- --nocapture`
- `cargo test -p jazz-tools policy_filter -- --nocapture`
- `cargo test -p jazz-tools exists_policy -- --nocapture`
- `cargo test -p jazz-tools policy_context_tables -- --nocapture`
- pre-commit `clippy`
- pre-commit `rustfmt`
- `git diff --check`

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
+++ b/crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
@@ -1,6 +1,6 @@
 use std::collections::HashSet;
 
-use crate::object::ObjectId;
+use crate::object::{BranchName, ObjectId};
 use crate::query_manager::policy::{
     Operation, PolicyExpr, bind_outer_row_refs, bind_relation_refs,
     evaluate_expr_recursive_with_row_id, normalize_recursive_max_depth,
@@ -14,6 +14,14 @@ use crate::query_manager::types::{
 use crate::storage::Storage;
 
 use super::super::encoding::{column_is_null, decode_column};
+
+type WitnessScope = HashSet<(ObjectId, BranchName)>;
+
+#[derive(Default)]
+pub(crate) struct ScopedPolicyResult {
+    pub(crate) allowed: bool,
+    pub(crate) scope: WitnessScope,
+}
 
 pub(crate) struct PolicyContextEvaluator<'a> {
     schema: &'a Schema,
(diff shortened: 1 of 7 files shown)


nrainhart:
Already fixed, see https://github.com/garden-co/jazz/pull/1033

PR #774, opened 2026-05-02
Make batch settlements the durability source of truth

aeplay:
## What changed

This PR makes batch settlements the active durability and visibility model for Jazz sync.

Instead of treating every received row as its own durability unit, receive-side sync now treats a logical batch as the unit that settles, persists, replays, and releases waiters. `BatchSettlement` is the source of truth for whether a batch is durable, accepted, rejected, missing, or visible at a tier.

### Batch durability and visibility semantics

- Removed `RowBatchStateChanged` as an active sync protocol message.
- Removed the old received-row ack side channel.
- Made `BatchSettlement` resolve row-level persistence waiters and drive visibility recomputation.
- Moved tiered reads away from hot per-row `StoredRowBatch.confirmed_tier` mutation and toward authoritative batch settlement storage.
- Stopped patching every row history entry during receive-side settlement application.
- Added the batch-fate rule that one rejected write rejects the whole batch; independent fate should use separate batches.
- Kept subscription replay chunky: rows are delivered first, then settlements are emitted by batch.

### Replay and offline correctness

- Retain replayable direct sealed submissions synthesized from client settlements.
- Handle row/settlement replay arriving in either order without losing the eventual seal.
- Keep offline direct submissions around until server durability is authoritative enough, then prune them.
- Promote existing lower-tier direct settlements when an authority accepts the seal, so fresh edge reads see rows that live clients already received.
- Preserve idempotent replay behavior so exact row replays settle once per batch.

### Transport and scheduler work

- Fixed NAPI scheduling by replacing the unreliable scheduler path with a debounced Rust background thread that calls `core.batched_tick()`.
- Added `SyncUpdateBatch` transport frames so WebSocket sync updates can be sent in batches while preserving per-message sequencing.
- Updated native and WASM transport managers to unpack batched sync updates.

### Release note

- Added a patch changeset for `jazz-tools`, `jazz-wasm`, and `jazz-napi` describing the settlement-based durability/replay behavior.

## Why

The BoredM holder path exposed two related problems:

1. Receive-side durability was too row-shaped. A large incoming scope produced many row-level durability operations and repeated visible-row recomputation.
2. Replay/offline recovery needed to reason about complete batch fate rather than scattered row acknowledgements.

This PR moves the model closer to the product semantics: a batch is the durability unit. Rows carry the data; settlements carry fate and durability. That removes the worst per-row settlement hot path, gives us one protocol message for batch fate, and makes replay/offline recovery reason about complete batches.

Timing and payload-volume instrumentation has been split into stacked PR #799 so this one stays focused on the model and correctness changes.

## Validation

Local validation includes:

- `cargo test -p jazz-tools sync_manager::tests::transaction_sealing -- --nocapture`
- `cargo test -p jazz-tools direct_client_settlement -- --nocapture`
- `cargo test -p jazz-tools rc_transactional_insert_is_accepted_when_replayed_to_reconnected_upstream -- --nocapture`
- `cargo fmt --check --package jazz-tools`
- `git diff --check`

CI is green on the cleaned head: https://github.com/garden-co/jazz/actions/runs/25419123235

## Follow-ups

- Coalesce or reduce tier-only settlement persistence and ack volume further.
- Avoid per-member OPFS tier work where no local waiter or immediate cached durability read needs it.
- Revisit query result ordering so large scopes such as `dropdown_entry` do not delay unrelated holder release behind row/control traffic.


crates/jazz-tools/src/transport_manager.rs
```diff
@@ -670,6 +670,19 @@ impl<W: StreamAdapter + 'static, T: TickNotifier + 'static> TransportManager<W,
...
+                                            source: crate::sync_manager::types::Source::Server(self.server_id),
+                                            payload: update.payload,
+                                        };
+                                        let _ = self.inbound_tx.unbounded_send(TransportInbound::Sync {
+                                            entry: Box::new(entry),
+                                            sequence: update.seq,
+                                        });
+                                    }
+                                    self.tick.notify();
+                                }
```
gdorsi: Nit: this `SyncUpdateBatch` handler is identical to the one in the other dispatch loop. Worth extracting into a shared helper to avoid them drifting apart?

crates/jazz-tools/src/batch_fate.rs
```diff
@@ -92,6 +92,72 @@ impl BatchSettlement {
...
+                let mut visible_members = existing_members.clone();
+                for member in incoming_members {
+                    if !visible_members.contains(member) {
+                        visible_members.push(member.clone());
+                    }
+                }
+                Self::DurableDirect {
+                    batch_id: *batch_id,
+                    confirmed_tier: (*existing_tier).max(*incoming_tier),
+                    visible_members,
```
gdorsi, line 127: `merged_with` uses linear `.contains()` on a Vec of `VisibleBatchMember` for dedup. For large batches with many members this is O(n²).

crates/jazz-tools/src/sync_manager/inbox.rs
```diff
@@ -1256,6 +1296,24 @@ impl SyncManager {
...
+                        .iter()
+                        .map(|member| member.object_id)
+                        .collect::<Vec<_>>();
+                    let rows = self.transactional_batch_rows(storage, *batch_id, &object_ids);
+                    self.apply_transactional_batch_settlement_to_rows(
+                        storage,
+                        None,
+                        &settlement,
+                        &rows,
+                    );
```
gdorsi, line 1358: There doesn't seem to be a test covering the case where a `BatchSettlement` (especially `AcceptedTransaction`) arrives from the server before its member rows. If the settlement lands first, `transactional_batch_rows` finds nothing and the materialization is a no-op — but when the rows arrive later as `StagingPending`, there's no second attempt to apply the settlement.

Is that a possible scenario? If yes can we add an integration test to cover it?

gdorsi (approved):
LGTM

crates/jazz-tools/src/query_manager/rebac_tests.rs
```diff
@@ -199,7 +186,8 @@ fn client_write_rejection_reason(
         }
     }
 
-    settlement_reason.or_else(|| saw_rejected_state.then(|| "rejected".to_string()))
+    let _ = (row_id, branch);
```
nrainhart: Nit: we can just remove these params from this function

crates/jazz-tools/src/runtime_core/tests/write_batch/persisted.rs
```diff
@@ -194,20 +194,24 @@ fn rc_insert_persisted_does_not_touch_legacy_ack_storage() {
...
-            state: None,
-            confirmed_tier: Some(DurabilityTier::Local),
+        payload: SyncPayload::BatchSettlement {
+            settlement: crate::batch_fate::BatchSettlement::DurableDirect {
+                batch_id,
+                confirmed_tier: DurabilityTier::Local,
+                visible_members: vec![crate::batch_fate::VisibleBatchMember {
+                    object_id: row_id,
+                    branch_name,
+                    batch_id,
```
nrainhart, line 204: Do we need to send the batch id as part of each member, if it's already on the settlement?

crates/jazz-tools/src/transport_manager.rs
aeplay: Extracted!

crates/jazz-tools/src/batch_fate.rs
aeplay: uses a HashSet now

crates/jazz-tools/src/sync_manager/inbox.rs
aeplay: Added a test, and materialization is retried now

crates/jazz-tools/src/query_manager/rebac_tests.rs
aeplay: Done!

crates/jazz-tools/src/runtime_core/tests/write_batch/persisted.rs
aeplay: Cleaning this up would also affect the storage format, so deferring this for now

crates/jazz-tools/src/sync_manager/sync_tracer.rs
```diff
@@ -138,7 +132,16 @@ impl SyncMessage {
             SyncPayload::RowBatchCreated { row, .. } | SyncPayload::RowBatchNeeded { row, .. } => {
                 Some(row.row_id)
             }
-            SyncPayload::RowBatchStateChanged { row_id, .. } => Some(*row_id),
+            SyncPayload::BatchSettlement { settlement } => match settlement {
```
nrainhart, line 135: Don't we have a util function to extract visible members already?

PR #785, opened 2026-05-04
fix(jazz-tools): tighten null query checks in better-auth adapter

Gabrola:
### Motivation
- BetterAuth adapter should allow `ne` on `Timestamp` columns while preventing unsafe `null` comparisons on non-nullable fields.
- `null` filters must only be allowed for nullable schema columns and `ne null` should remain rejected for reference and `id` fields to avoid incorrect query semantics.
- Tests and a changeset are required to document and validate the tightened rules.

### Description
- Updated `isQuerySupported` to include `ne` in the supported operators for `Timestamp` columns by adding `"ne"` to the `Timestamp` case in `getSupportedOperators`.
- Tightened `null` handling in `isQuerySupported` to fetch the column and reject `null` comparisons against non-nullable columns and to explicitly reject `ne null` on columns with `references`.
- Adjusted tests in `packages/jazz-tools/src/better-auth-adapter/utils.test.ts` to add `session_expires_at` and `last_seen_at` timestamp columns, assert that `null` is only allowed on nullable columns, assert that `ne null` is rejected for `owner_id` and `id`, and assert that `ne` is supported for timestamp values.
- Added a changeset `.changeset/betterauth-nullability-query-rules.md` describing the patch-level change for `jazz-tools`.

### Testing
- Ran the focused test file with `pnpm --filter jazz-tools exec vitest run src/better-auth-adapter/utils.test.ts`, which passed (63 tests).
- Ran the package test runner with `pnpm test --filter jazz-tools`; the full package suite exercised many tests but shows unrelated failures in this environment (native binding / environment issues), so only the targeted adapter test was relied on for validation.

### DIFF
--- a/.changeset/betterauth-nullability-query-rules.md
+++ b/.changeset/betterauth-nullability-query-rules.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Updated BetterAuth adapter query-support rules: timestamp columns now support `ne`, `null` filters are only allowed for nullable columns, and `ne null` remains rejected for references and `id`.
--- a/packages/jazz-tools/src/better-auth-adapter/utils.test.ts
+++ b/packages/jazz-tools/src/better-auth-adapter/utils.test.ts
@@ -39,25 +39,22 @@ describe("isQuerySupported", () => {
         nullable: true,
       },
       {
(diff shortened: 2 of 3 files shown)


gdorsi (approved):
LGTM

PR #787, opened 2026-05-04
fix: retain row metadata on overlapping writes until server confirmation

nrainhart:
## Description

When performing write operations, the row metadata is only sent to the server if it hasn't previously been sent. This caused a race condition when doing multiple write operations on the same object (e.g. an insert followed by an update) before the server had processed the first operation: the later write could arrive without the row locator metadata the server needs to identify the row’s table and schema context. Even though messages are delivered in order, the first write may still be waiting on permission evaluation and not yet applied to storage, so the server cannot reliably recover that locator from its local state. In that window, policy evaluation for the later write can run with incomplete context and produce the wrong result.

This PR fixes the issue by keeping row metadata attached to server-bound writes until the object has at least one upstream-confirmed row. In other words, we no longer treat “metadata was sent once” as equivalent to “the server has definitely persisted enough state to infer it.” That makes overlapping writes self-describing until the server has fully caught up, which removes the race.

### DIFF
--- a/crates/jazz-tools/src/sync_manager/forwarding.rs
+++ b/crates/jazz-tools/src/sync_manager/forwarding.rs
@@ -7,6 +7,68 @@ use std::collections::HashSet;
 use uuid::Uuid;
 
 impl SyncManager {
+    fn object_has_upstream_confirmation<H: Storage>(
+        &self,
+        storage: &H,
+        table: &str,
+        branch_name: &BranchName,
+        object_id: ObjectId,
+    ) -> bool {
+        let Ok(Some(visible_entry)) =
+            storage.load_visible_region_entry(table, branch_name.as_str(), object_id)
+        else {
+            return false;
+        };
+        let local_tier = self.max_local_durability_tier();
+        match local_tier {
+            None => {
+                visible_entry.current_row.confirmed_tier.is_some()
+                    || visible_entry.worker_batch_id.is_some()
+                    || visible_entry.edge_batch_id.is_some()
+                    || visible_entry.global_batch_id.is_some()
+            }
+            Some(my_tier) => {
+                visible_entry
+                    .current_row
+                    .confirmed_tier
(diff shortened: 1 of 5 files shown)


crates/jazz-tools/src/sync_manager/forwarding.rs
```diff
@@ -7,6 +7,68 @@ use std::collections::HashSet;
...
+    ) {
+        let branch_name = BranchName::new(&row.branch);
+        let include_metadata = {
+            let Some(server) = self.servers.get(&server_id) else {
+                return;
+            };
+            let metadata_already_sent = server.sent_metadata.contains(&object_id);
+            let object_confirmed_upstream =
+                self.object_has_upstream_confirmation(storage, table, &branch_name, object_id);
+            !metadata_already_sent || !object_confirmed_upstream
```
gdorsi: If we end up checking object_has_upstream_confirmation anyway, we could either skip metadata_already_sent or use it to short circuit object_has_upstream_confirmation:

```rust
let include_metadata = if !metadata_already_sent {
    true
} else {
    !self.object_has_upstream_confirmation(storage, table, &branch_name, object_id)
};
```

gdorsi (approved):
Reviewed section by section. The metadata retention logic is sound — re-includes metadata until upstream confirmation arrives, closing the overlapping-writes gap. One minor nit left as an inline comment (unnecessary storage lookup when metadata_already_sent is false). Otherwise LGTM.

gdorsi (approved):
LGTM

The sync probably becomes a little more chatty, but we can't rely too much on optimistic data.

TBH, this makes me feel like the sync protocol currently is too much reliant on transmission ordering.

nrainhart:
> TBH, this makes me feel like the sync protocol currently is too much reliant on transmission ordering.

Yeah, and another problem is that we do not process sync messages synchronously when they arrive, so we can have problematic interleavings between consecutive messages

PR #790, opened 2026-05-05
chore: make jazz-rn optional peer dependencies

antoniomuso:
## Summary

- `jazz-rn` moves from `dependencies` to optional `peerDependencies` of `jazz-tools`. Web/Node consumers no longer install the React-Native native binary they don't use.
- Lazy-load + cache `jazz-rn` via a new `jazz-rn-loader` (`loadJazzRn` async, `getJazzRnSync` sync). The RN runtime module loads it once in `loadRuntime()`; the rest of the RN code path reads from the cache.
- Friendly install-hint error when the peer is missing — only real module-not-found error codes are rebranded; native-binding faults rethrow unchanged.
- `jazz-wasm` stays a regular dependency (web/Node consumers continue to get it transparently).
- Adds `jazz-rn` to the Expo example that was missing it.
- Docs note in `docs/install/client.mdx` flagging `jazz-rn` as required for Expo/RN.

## Why

The RN native binary was being pulled in for every consumer regardless of platform. Making it an optional peer keeps web/Node installs slim while a missing install on RN now shows a clear, actionable message instead of `Failed to resolve import`.

## Test plan

- [x] `pnpm exec tsc --noEmit` (jazz-tools) — clean.
- [x] `jazz-rn-loader.test.ts` — 7/7 pass; covers friendly message, cause preservation, non-resolution rethrow, cache hit, and accessed-before-loaded.
- [x] `jazz-rn-handoff.test.ts` — 4/4 pass; verifies `createJazzRnRuntime` succeeds after `createDb` primes the loader, throws otherwise, and exercises `Db#getLocalFirstIdentityProof` lazy-load.
- [ ] CI


packages/jazz-tools/src/runtime/peer-dep-error.ts
```diff
@@ -0,0 +1,14 @@
...
+// installed. Anything else (init crashes, syntax errors in wasm-bindgen glue,
+// native-binding faults) must rethrow so we don't misdiagnose real failures.
+const MODULE_NOT_FOUND_CODES: ReadonlySet<string> = new Set([
+  "ERR_MODULE_NOT_FOUND",
+  "MODULE_NOT_FOUND",
+  "ERR_PACKAGE_PATH_NOT_EXPORTED",
+]);
+
+export function isModuleNotFoundError(err: unknown): boolean {
+  if (!err || typeof err !== "object") return false;
```
gdorsi, line 18: Has `isModuleNotFoundError` been tested against the actual error shapes emitted by both Vite and Next.js (dev + prod)? Bundlers don't always set a `.code` property — Webpack, for example, tends to throw plain `Error` objects with only a message string. If the code-based check misses those, consumers would still see the raw bundler error instead of the friendly install hint.

antoniomuso: I removed the WASM part. Now only jazz-rn is an optional dep.

gdorsi (approved):
Non-blocking: How about flagging version mismatches in the metro plugins?

PR #794, opened 2026-05-05
chore: refactor db to inject the native runtime module

gdorsi:
Refactor db.ts to make the native runtime injected.

This will make it easier to split db in multiple "pluggable" modules as well as moving more stuff on Rust.

This PR also adds anonymous auth to the RN runtime.


nrainhart (approved):
💅

PR #795, opened 2026-05-05
fix: correctly resolve same-column where conditions

Gabrola:
## Summary

Fixes query planning for conjunctions that place multiple predicates on the same indexed column, and adds a changeset for the `jazz-tools` patch release.

## Motivation

The planner previously picked one index-scannable condition per disjunct and then considered all predicates on that same column to be covered by the index scan. That was incorrect for same-column conjunctions because a scan for only one bound could return rows that failed the other bound, while the residual filter had already been elided.

Examples that now return the correct rows:

```ts
// rows with values of [5, 10, 15, 20]
await db.all(
  app.table_with_defaults
    .where({ integer: { gt: 5, lt: 20 } })
    .select("integer")
    .orderBy("integer", "asc"),
);
// returns integer 10 and 15; excludes 5 and 20

// ----

// rows with values of [5, 10, 15, 20]
await db.all(
  app.table_with_defaults
    .where({ integer: { gt: 10, gte: 5 } })
    .select("integer")
    .orderBy("integer", "asc"),
);
// returns values greater than 10, not everything >= 5

// ----

await db.all(
  app.table_with_defaults
    .where({ nullableInteger: { lt: 10, ne: null } })
    .select("nullableInteger"),
);
// returns only non-null values below 10
```

## Planner behavior

Same-column scan conditions are now intersected before choosing the index scan. Redundant bounds collapse to the strictest scan, equality inside a range becomes an exact lookup, and contradictions become an empty scan.

Optimization examples:

```ts
await db.all(
  app.table_with_defaults.where({ integer: { eq: 15, gte: 5, lt: 20 } }),
);
// planned as an exact integer = 15 scan

await db.all(
  app.table_with_defaults.where({ integer: { eq: 10, lt: 10 } }),
);
// planned as ScanCondition::Empty and short-circuits to zero results

await db.all(
  app.table_with_defaults.where({ nullableInteger: { lt: 10, eq: null } }),
);
// impossible same-column predicate; returns []
```

For mixed-column predicates, the planner still keeps the residual filter when one index scan cannot fully cover the whole conjunction.

## Tests

- Added Rust query graph tests for redundant lower/upper bounds, equality within ranges, contradiction-to-empty scans, and mixed disjunct residual filtering.
- Added TS DSL integration coverage for integer, nullable integer, float, and timestamp same-column predicates.

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph/compile.rs
+++ b/crates/jazz-tools/src/query_manager/graph/compile.rs
@@ -1,10 +1,10 @@
 use std::collections::{HashMap, HashSet};
-use std::ops::Bound;
+use std::{cmp::Ordering, ops::Bound};
 
 use crate::object::BranchName;
 use crate::query_manager::types::{
     ColumnDescriptor, ColumnName, ColumnType, ComposedBranchName, RowDescriptor, RowPolicyMode,
-    Schema, SchemaHash, TableName, TupleDescriptor,
+    Schema, SchemaHash, TableName, TupleDescriptor, Value,
 };
 use crate::schema_manager::{
     SchemaContext, translate_column_for_index, translate_table_name_to_schema,
@@ -578,7 +578,7 @@ impl QueryGraph {
         // For multi-branch queries, we create scans for each branch and union them
         // Column names are translated for old schema branches
         let mut phase1_outputs: Vec<NodeId> = Vec::new();
-        let mut index_columns: Vec<String> = Vec::new();
+        let scan_plans: Vec<_> = plan.disjuncts.iter().map(index_scan_plan).collect();
 
         for branch in &branches {
             // Get schema hash for this branch to determine if column translation is needed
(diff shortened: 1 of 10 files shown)


crates/jazz-tools/src/query_manager/graph/compile.rs
```diff
@@ -591,18 +576,8 @@ impl QueryGraph {
...
-                // Find best index condition for this disjunct
-                let (scan_column, scan_condition) =
-                    if let Some(cond) = best_available_index_condition(disjunct, table_schema) {
-                        let column = cond.column().to_string();
-                        let scan_cond = condition_to_scan(cond);
-                        (column, scan_cond)
-                    } else {
-                        // No index condition, use "_id" for full scan
-                        ("_id".to_string(), ScanCondition::All)
-                    };
```
Gabrola, line 627: A simpler change to fix JUST the bug could have been implemented by preserving this logic and changing `best_available_index_condition` to detect lower bound + upper bound conditions for exactly two same-column predicates and collapse it into a ScanCondition::Range, and in all other cases of two "mismatched" conditions on the same column + >2 conditions fallback to `ScanCondition::All` and let filtering return the correct results, but I think the current proposed change is much better in terms of logical correctness while also allowing the usage of indexes is more cases when possible

gdorsi (approved):
LGTM, the fix makes sense to me. Thanks for the contribution

nrainhart (approved):
Amazing contribution, thanks @Gabrola!! 🥇

PR #796, opened 2026-05-05
Extract local-first auth from Db

gdorsi:
## Summary
- Move local-first auth config resolution, anonymous token minting, refresh scheduling, and identity proof minting into `db-auth.ts`.
- Keep `Db` responsible for coordinating auth state updates while removing token/timer details from `db.ts`.
- Add focused coverage for the extracted auth helper and refresh behavior.

## Test Plan
- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts src/runtime/db-auth.test.ts src/runtime/db.local-first-auth.test.ts src/runtime/db.anonymous.test.ts src/runtime/db.transport.test.ts src/react-native/db.test.ts`
- `pnpm --dir packages/jazz-tools build`

## Notes
- This is stacked on `refactor/db-cleanup-and-extract-cross-tab`.

### DIFF
--- a/packages/jazz-tools/src/runtime/db.ts
+++ b/packages/jazz-tools/src/runtime/db.ts
@@ -31,7 +31,7 @@ import {
   resolveEffectiveQueryExecutionOptions,
   runInBatch,
 } from "./client.js";
-import { type DbRuntimeModule, type RuntimeTokenOptions } from "./db-runtime-module.js";
+import { type DbRuntimeModule } from "./db-runtime-module.js";
 import { WasmRuntimeModule } from "./wasm-runtime-module.js";
 import { WorkerBridge, type PeerSyncBatch, type WorkerBridgeOptions } from "./worker-bridge.js";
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/src/runtime/db.ts
```diff
@@ -854,51 +854,17 @@ export class Db {
...
-  }
-
-  private mintLocalFirstToken(secret: string, audience: string, ttlSeconds: number): string {
-    if (!this.runtimeModule) {
-      throw new Error("Db runtime module is not initialized for this Db implementation");
-    }
-
-    return this.runtimeModule.mintLocalFirstToken({
+  /** @internal Attach local-first auth state and schedule token refresh. */
+  initLocalFirstAuth(secret: string, runtimeModule: AnyDbRuntimeModule): void {
```
nrainhart: Is the runtime module different from the one the Db was initialized with?

nrainhart (approved):

PR #800, opened 2026-05-06
Fix React Native UI freeze on db.all by making jazz-rn query async

joeinnes:
## Summary
- `jazz-rn`'s native `query` export ran `block_on` on the JS thread, which froze the UI for the duration of a query and could deadlock entirely: queries that need a later `batched_tick` to settle (e.g. server-sourced reads) cannot make progress while JS is blocked, since `batched_tick` is JS-driven.
- `query` is now `async fn`. uniffi-bindgen-react-native generates a Promise-returning JSI call and polls the future off the JS thread, so `batched_tick` callbacks keep firing while the query waits.
- The TS adapter now `await`s the binding's Promise; `JazzRnRuntimeBinding.query` is typed as `Promise<string>`. NAPI already shipped this shape.

## Test plan
- [ ] `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/react-native` — 30 tests pass, including the new `yields the JS event loop while a query is in flight` regression.
- [ ] `cargo test -p jazz-rn && cargo clippy -p jazz-rn --all-targets` — clean.
- [ ] Build the RN bindings (`pnpm --filter jazz-rn build`) and confirm the generated TypeScript surface for `query` becomes `Promise<string>`.

### DIFF
--- a/.changeset/rn-async-query.md
+++ b/.changeset/rn-async-query.md
@@ -1,0 +1,8 @@
+---
+"jazz-tools": patch
+"jazz-rn": patch
+---
+
+`jazz-rn`: `query` is now `async` and no longer blocks the React Native JS thread on one-shot reads.
+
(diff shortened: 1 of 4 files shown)


antoniomuso (approved):
Nice catch! LGTM!

PR #801, opened 2026-05-06
Add interactive lens diagram to migrations docs

joeinnes:
## Summary

- New `LensDiagram` MDX component that animates a row crossing the wire and being walked through the lens chain on the client. Pure path geometry lives in a sibling `lens-diagram-path.ts` so the React component stays thin.
- Embeds the diagram in `docs/content/docs/schemas/migrations.mdx` and links to it from `docs/content/docs/concepts/branches.mdx`.
- Reviews the migrations page: groups stub / customising-defaults / backwards-defaults under a single "The migration file" heading, tightens the post-diagram prose, and stamps `reviewedAt`.

## Test plan

- [ ] Load `/docs/schemas/migrations` and verify the diagram renders with the v1→v3 forward animation on first paint.
- [ ] Click each data card (v1, v2, v3) — arrow re-traces the correct path through schemas and lenses; tip dot tracks the line at constant velocity.
- [ ] Pick each schema version on the client device — backward direction (e.g. v3 data → v1 client) animates upward through the chain and flips lens labels.
- [ ] Idle (data and client on the same version) renders a single-card path, no double-animation.


gdorsi (approved):
Fantastric

antoniomuso (approved):

PR #802, opened 2026-05-06
Drop no-op Headers/Request/Response polyfills from jazz-tools/expo

joeinnes:
## Summary
- Removed three `polyfillGlobal` calls for `Headers`/`Request`/`Response` and the dead `react-native/Libraries/Network/fetch` imports they relied on.
- Those imports just re-export `global.*`, so each polyfill was either an identity assignment or — when this module evaluated before RN installed its networking globals — clobbered `globalThis.X` with `undefined`, breaking any consumer that touched `Headers.prototype` (e.g. Clerk's `new Headers(...)` surfacing as `TypeError: Cannot read property 'prototype' of undefined`).
- Kept `fetch` (streaming via `expo/fetch`) and `ReadableStream` (Hermes lacks WHATWG streams).

## Test plan
- [ ] `pnpm --filter jazz-tools test src/better-auth-adapter` still passes (URL/Request coercion in `fetchSpecCompliant` is unchanged).
- [ ] In an Expo app using both `jazz-tools/expo/polyfills` and Clerk, sign-in no longer throws `Cannot read property 'prototype' of undefined`.
- [ ] Existing Expo todo example still syncs end-to-end.

### DIFF
--- a/packages/jazz-tools/src/expo/polyfills.ts
+++ b/packages/jazz-tools/src/expo/polyfills.ts
@@ -1,11 +1,6 @@
 /// <reference path="./vendor.d.ts" />
 
 import { polyfillGlobal } from "react-native/Libraries/Utilities/PolyfillFunctions";
-import {
-  Headers as ReactNativeHeaders,
-  Request as ReactNativeRequest,
-  Response as ReactNativeResponse,
-} from "react-native/Libraries/Network/fetch";
 import { fetch as expoFetch } from "expo/fetch";
 
 import { ReadableStream as PonyfillReadableStream } from "web-streams-polyfill";
@@ -33,7 +28,4 @@ const fetchSpecCompliant: typeof globalThis.fetch = async (input, init) => {
 };
 
 polyfillGlobal("fetch", () => fetchSpecCompliant);
-polyfillGlobal("Headers", () => ReactNativeHeaders ?? globalThis.Headers);
-polyfillGlobal("Request", () => ReactNativeRequest ?? globalThis.Request);
-polyfillGlobal("Response", () => ReactNativeResponse ?? globalThis.Response);
 polyfillGlobal("ReadableStream", () => readableStreamCtor);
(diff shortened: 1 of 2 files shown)


packages/jazz-tools/src/expo/polyfills.ts
```diff
@@ -33,7 +28,4 @@ const fetchSpecCompliant: typeof globalThis.fetch = async (input, init) => {
 };
 
 polyfillGlobal("fetch", () => fetchSpecCompliant);
-polyfillGlobal("Headers", () => ReactNativeHeaders ?? globalThis.Headers);
-polyfillGlobal("Request", () => ReactNativeRequest ?? globalThis.Request);
-polyfillGlobal("Response", () => ReactNativeResponse ?? globalThis.Response);
 polyfillGlobal("ReadableStream", () => readableStreamCtor);
```
antoniomuso, line 31: I think we can remove this as well. And we can remove the dep `web-streams-polyfill`.

antoniomuso: Ok, I checked, and we use readableStream in file-storage.ts, so we can't remove it.

joeinnes: Yeah I think basically the only reason we need polyfills at all is because Hermes + RN don't support streaming response bodies, so we need to keep `fetch` and `ReadableStream`.

packages/jazz-tools/src/expo/polyfills.ts
```diff
@@ -33,7 +28,4 @@ const fetchSpecCompliant: typeof globalThis.fetch = async (input, init) => {
 };
 
 polyfillGlobal("fetch", () => fetchSpecCompliant);
```
antoniomuso, line 30: We can remove the fetch polyfill; we no longer have web-stream, and I don't see any call to fetch with ReadableStream. So I think we can keep the default fetch.

gdorsi (approved):

packages/jazz-tools/src/expo/polyfills.ts
joeinnes: Will do as a follow-up PR.

PR #806, opened 2026-05-06
fix: null field update is lost on sync

nrainhart:
## Description

A user reported an issue when updating optional fields to `null`. The update would be "lost", and the previous value preserved. If there were other intermediate updates, they would also be lost and the initial value would be visible instead.

I was able to repro this issue both in browser and napi integration tests, but not in single-DB setups (i.e. `update-api.test.ts` worked properly).

The problem was related to a combination of how row forwarding works across nodes, and the per-column merge strategy:
- for query-scope delivery we intentionally strip parents before sending visible rows to subscribers (see `scope_delivery_row`)
- the client then sees two parent-less visible rows (originally inserted row and updated row) with no common ancestor
- when there's no ancestor the "default" ancestor is a row with all fields set to `null` and, since the updated row also contains a null field, it's treated as if no change had been made.

The fix was to modify the merge strategy to consider the explicit `null` field as a new field.

### DIFF
--- a/crates/jazz-tools/src/row_histories/resolution.rs
+++ b/crates/jazz-tools/src/row_histories/resolution.rs
@@ -312,13 +312,14 @@ pub(super) fn build_computed_visible_preview(
     let mut merged_values = Vec::with_capacity(user_descriptor.columns.len());
     let mut contributing_rows: Vec<&StoredRowBatch> = Vec::new();
     let mut winner_batch_ids = Vec::with_capacity(user_descriptor.columns.len());
+    let null_ancestor = Value::Null;
 
     for column_index in 0..user_descriptor.columns.len() {
         let column = &user_descriptor.columns[column_index];
         let ancestor_value = ancestor_values
             .as_ref()
-            .map(|values| values[column_index].clone())
-            .unwrap_or(Value::Null);
+            .map(|values| &values[column_index])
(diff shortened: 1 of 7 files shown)


gdorsi (approved):
LGTM

crates/jazz-tools/src/row_histories/resolution.rs
```diff
@@ -319,23 +314,36 @@ pub(super) fn build_computed_visible_preview(
     let mut winner_batch_ids = Vec::with_capacity(user_descriptor.columns.len());
 
     for column_index in 0..user_descriptor.columns.len() {
-        let ancestor_value = ancestor_values[column_index].clone();
+        let column = &user_descriptor.columns[column_index];
+        let ancestor_value = ancestor_values
+            .as_ref()
+            .map(|values| values[column_index].clone())
```
gdorsi: Would it be worth refactoring this loop so `ancestor_value` is borrowed instead of cloned? A local `Value::Null` fallback could cover the no-ancestor case, and `merge_column_with_strategy` already accepts a borrowed value.

nrainhart: Good point! Done in [`f378c28` (this PR)](https://github.com/garden-co/jazz/pull/806/commits/f378c28419117c8cc03fb4a3ea162bcae08193ab)

PR #808, opened 2026-05-06
Support broader query membership filters

aeplay:
## What changed

- Widened typed query `where` inputs so `in` works on scalar and reference columns, not only row ids.
- Fixed query translation so scalar `in` lists are converted element-by-element instead of being treated as one scalar array literal.
- Added coverage for array-column `contains` with non-text element types: numbers, booleans, refs/UUIDs, enums, and timestamps.
- Added TS DSL coverage for text, boolean, required refs, nullable refs, array-column `in`, transformed numeric where inputs, and non-text array `contains` filters.

## Why

The query runtime already had general membership predicate shapes, but the typed TS query surface and tests only proved a narrow subset. This makes the public DSL match the relational behavior for easy scalar/reference `in` cases and confirms literal-in-array-column `contains` behavior across the scalar element types that the runtime already supports.

## Validation

- `pnpm lint`
- `pnpm --filter jazz-tools exec tsc --project tsconfig.tests.json --noEmit`
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts tests/ts-dsl/typed-app.test.ts tests/ts-dsl/query-api.test.ts src/runtime/query-adapter-tests/condition-translation.test.ts`
- `pnpm --filter jazz-wasm run build`

### DIFF
--- a/packages/jazz-tools/src/runtime/query-adapter.ts
+++ b/packages/jazz-tools/src/runtime/query-adapter.ts
@@ -363,7 +363,7 @@ function conditionToRelPredicate(
   }
   const valueTypeForCondition =
     cond.op === "contains" && columnType.type === "Array" ? columnType.element : columnType;
-  const rightLiteral =
+  const rightLiteral = () =>
     isFrontierRowIdToken(cond.value) && cond.op === "eq"
       ? { RowId: "Frontier" as const }
       : {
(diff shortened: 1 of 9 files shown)


packages/jazz-tools/src/runtime/query-adapter.ts
```diff
@@ -363,7 +363,7 @@ function conditionToRelPredicate(
   }
   const valueTypeForCondition =
     cond.op === "contains" && columnType.type === "Array" ? columnType.element : columnType;
-  const rightLiteral =
+  const rightLiteral = () =>
```
nrainhart, line 366: I don't get why this change was needed

nrainhart (approved):

aeplay:
Superseded by #1141, which implements broader query membership filters from scratch against the new relational core. This branch targets the retired pre-swap query engine, so nothing was ported — #1141 re-derives the behaviour (`in` over integer/float/boolean/uuid-reference columns, `contains` over non-text array elements), adds black-box coverage, and documents the supported matrix in SPEC 6.1.1.

PR #809, opened 2026-05-06
Replace batch settlements with batch fate

aeplay:
## What changed

- Renamed the replayable batch outcome model from settlement terminology to `BatchFate` / `BatchFateNeeded`.
- Removed `visible_members` / `VisibleBatchMember` from active Rust and TypeScript API expectations.
- Made successful batch fate apply by batch id to locally known rows, so tier checks no longer need per-row member scans or repeated giant settlement decodes.
- Kept the legacy storage column shape forward-compatible while new writers emit an empty compatibility member array.
- Updated status-quo specs and tests around whole-batch fate semantics.

## Why

BoredM cold subscription traces showed client time dominated by repeatedly decoding the same giant authoritative batch settlement row while checking row visibility. The design now makes batch fate singular and derives concrete row/query membership from local row delivery, sealed submissions, local batch records, and query scope.

## Validation

- `cargo check -p jazz-tools --no-default-features`
- `cargo test -p jazz-tools --no-default-features --no-run`
- `cargo test -p jazz-tools --no-default-features batch_fate`
- `cargo test -p jazz-tools --no-default-features seal_batch_accepts_all_staged_transactional_rows_as_one_settlement`
- `cargo test -p jazz-tools --no-default-features initial_query_sync_replays_current_direct_batch_fate`
- `pnpm lint`
- `git diff --check`


aeplay:
Closing this in favor of the split stacked PR series, which carries the work forward on top of current `main` and separates the large performance/correctness branch into reviewable slices:

1. #829 - Simplify batch seal lifecycle
2. #830 - Optimize query settlement replay
3. #831 - Coalesce sync replay transport
4. #832 - Optimize row storage hot paths
5. #833 - Add selective table index metadata
6. #834 - Deliver subscription rows in native format
7. #835 - Optimize sync replay storage hot paths
8. #836 - Fast path initial output delivery

A couple of notes from the split: the older OPFS EOF fix is intentionally not part of the stack because `main` already has the more principled #816 fix, and the duplicate batch-fate conversion work from this PR was skipped where `main` already had the parallel implementation.

PR #811, opened 2026-05-07
Simplify batch seal lifecycle

aeplay:
## Summary
- stop rebuilding replayable local batch records from every synced row
- persist sealed batch submissions and authoritative BatchFate as the post-seal source of truth
- make waiters and mutation-error replay read BatchFate directly via WASM/NAPI bindings
- update write-batch regression expectations for the new sealed-submission/fate lifecycle

## Why
Heavy inserts were pathologically slow because local batch membership was being decoded and rewritten repeatedly. This moves membership to a one-time sealed manifest and keeps fate separate for settlement/wait paths.

## Validation
- cargo check -p jazz-tools -p jazz-wasm -p jazz-napi
- cargo test -p jazz-tools write_batch -- --nocapture
- cargo clippy --workspace -- -D warnings
- pnpm dlx oxlint . --ignore-path .oxlintignore (completed with existing warnings, 0 errors)

Note: `pnpm lint` could not run directly because this worktree has no node_modules; `oxlint` and `oxfmt` were run via pnpm dlx instead.


aeplay:
Closing this mega PR in favor of the split stacked PR series, which carries the work forward on top of current `main` and separates the large performance/correctness branch into reviewable slices:

1. #829 - Simplify batch seal lifecycle
2. #830 - Optimize query settlement replay
3. #831 - Coalesce sync replay transport
4. #832 - Optimize row storage hot paths
5. #833 - Add selective table index metadata
6. #834 - Deliver subscription rows in native format
7. #835 - Optimize sync replay storage hot paths
8. #836 - Fast path initial output delivery

A couple of notes from the split: the older OPFS EOF fix is intentionally not part of the stack because `main` already has the more principled #816 fix, and duplicate batch-fate conversion work was skipped where `main` already had the parallel implementation from #809.

PR #812, opened 2026-05-07
Drop fetch polyfill from jazz-tools/expo

joeinnes:
Follow-up to #802.

## Summary
- Removed `polyfillGlobal("fetch", ...)`, the `fetchSpecCompliant` URL/Request coercion, and the `expo/fetch` import. RN's default fetch (`whatwg-fetch` under the hood) already accepts URL and Request inputs per the WHATWG spec, so better-auth's URL-input path keeps working without the wrapper.
- `expo/fetch` was originally swapped in for streamable response bodies, but nothing in jazz-tools consumes `response.body` as a stream — sync runs over the Rust WebSocket transport, and every `fetch()` call site uses buffered methods (`.json()`, `.text()`).
- Cleaned up the now-unused `expo/fetch` and `react-native/Libraries/Network/fetch` ambients in `vendor.d.ts`.
- `ReadableStream` polyfill stays — `runtime/file-storage.ts` constructs `new ReadableStream(...)` for the chunked-file API, which Hermes can't service.

## Test plan
- [ ] `pnpm --filter jazz-tools test src/better-auth-adapter` still passes.
- [ ] In an Expo app, a better-auth round-trip (URL input) completes without "The 2nd argument cannot be cast to type URL" or any other native-bridge error — i.e. RN's default fetch genuinely handles URL inputs in Hermes.
- [ ] In an Expo app, the file-storage API (`db.fileFromStream` / `db.fileToStream`) still works — `ReadableStream` global is intact.
- [ ] Existing Expo todo example still syncs end-to-end.

### DIFF
--- a/.changeset/drop-expo-fetch-polyfill.md
+++ b/.changeset/drop-expo-fetch-polyfill.md
@@ -1,0 +1,9 @@
+---
+"jazz-tools": patch
+---
+
+Drop the `fetch` polyfill from `jazz-tools/expo`.
+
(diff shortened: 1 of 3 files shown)


antoniomuso (approved):
LGTM

PR #813, opened 2026-05-07
feat(jazz-wasm): Rust-owned worker bridge

antoniomuso:
## Summary

Move the browser dedicated-worker bridge orchestration out of TypeScript and into the `jazz-wasm` Rust crate. The TS `WorkerBridge` survives as a thin adapter, so `db.ts` doesn't need to change. The public `Db` API and the on-the-wire `postMessage` envelope shape are unchanged. Diff: **+4.6k / −2.7k**.

## Why

`jazz-wasm` was the only `SyncSender` binding that exposed itself as a **JS callback** (`onSyncMessageToSend`) — NAPI doesn't, Rust's `TransportManager` owns the wire there. The only reason `JsSyncSender` existed was to let `worker-bridge.ts` and `jazz-worker.ts` intercept outbox messages and re-route them via `postMessage`. That JS-callback indirection was the symptom of a deeper split: the worker bridge protocol was **half in TypeScript** (envelope, init handshake, peer routing, lifecycle hints, buffering, shutdown) **and half in Rust** (sync routing, persistence, transport). The costs:

- `unsafe impl Send for JsSyncSender` to satisfy the multi-thread Tokio backend's `Send` bound — sound under WASM, but a smell.
- **Two protocol definitions** (`worker-protocol.ts` and the Rust runtime emitter) kept in sync by hand; drift was invisible at compile time.
- Tests for message-protocol edges fragmented across both layers.

This PR replaces the JS-callback `SyncSender` with `RustOutboxSender` that calls `worker.post_message` directly. The trait stays; the JS escape hatch is gone.

## Architecture (after)

```mermaid
graph LR
    subgraph Main["Main thread"]
        Db["Db<br/>(jazz-tools/runtime/db.ts)"]
        WB["WorkerBridge.ts<br/>~250-LOC TS adapter"]
        WWB["WasmWorkerBridge (Rust)<br/>crates/jazz-wasm/src/worker_bridge.rs"]
        MR["WasmRuntime<br/>(main client)"]
        Db --> WB --> WWB --> MR
    end
    subgraph Worker["Dedicated Worker"]
        JW["jazz-worker.ts<br/>~150-LOC bootstrap shim"]
        WH["worker_host (Rust)<br/>crates/jazz-wasm/src/worker_host.rs"]
        WR["WasmRuntime<br/>(worker client + OPFS)"]
        WSU["WebSocket upstream<br/>(Rust transport, unchanged)"]
        JW -->|"runAsWorker(init, pending)"| WH
        WH --> WR --> WSU
    end
    WWB <-->|"postcard binary<br/>over postMessage"| WH
```

`WorkerBridge.ts` is now a delegating adapter (same surface: `new WorkerBridge(worker, runtime)` + listeners before `init(options)`). `jazz-worker.ts` only does the bootstrap-handoff dance — post `{type:"ready"}`, buffer init + pending while WASM loads, hand off to `wasmModule.runAsWorker`. After that, **Rust owns `self.onmessage` / `self.postMessage`**. Both sides encode as postcard-binary `Uint8Array`s; `worker_protocol.rs` is the single source of truth. JS-object envelopes survive only for `{type:"ready"}` and pre-WASM-load error frames.

## What changed

- **Stage 1 — outbox.** `JsSyncSender` + `onSyncMessageToSend` deleted. `RustOutboxSender` posts to `Worker` (main) or `DedicatedWorkerGlobalScope` (worker). Encoding rules + sequence assignment preserved. `+ Send` bound on `RuntimeCore::sync_sender` `cfg`-gated to non-wasm32 — no more `unsafe impl Send`.
- **Stage 2 — worker host.** `runAsWorker(init, pending)` takes over `self.onmessage`. Three `thread_local!`s (`HOST` / `RUNTIME` / `PEER_ROUTING`) split borrowing so outbox callbacks don't re-enter the host borrow.
- **Stage 3 — main bridge.** `WasmWorkerBridge` owns the state machine, init/shutdown handshakes (`INIT_RESPONSE_TIMEOUT_MS=12_000`, `SHUTDOWN_ACK_TIMEOUT_MS=5_000`), peer channels, forwarder pass-through, and best-effort `Drop` cleanup.

## Key invariants

- **Worker-side init order**: open runtime → register clients → attach outbox target → bootstrap catalogue → connect upstream **before** draining → sync retained local batches → flip `Ready` → drain pending → bail if a `Shutdown` was drained → post `InitOk`. Connecting upstream before draining keeps server-bound traffic from drained writes routing through the Rust transport rather than into the (closed) bootstrap forwarder.
- **Pre-init outbox** is buffered inside `RustOutboxSender::pending_sync_entries`; `attach()` closes the init-gate, `init()` opens it on `init-ok`.
- **Shutdown** is synchronous up to the await: `batched_tick` + `flush_now` to drain, `install_noop_sync_sender` + `set_server_payload_forwarder(None)` + `remove_server` to detach, **then** post `Shutdown`. The shutdown future clears `worker.onmessage` (Closure drop alone does not).
- **`Drop` for `WasmWorkerBridge`** runs the same teardown on exception paths.

## Perf — Rust bridge vs TS bridge

Same bench (`packages/jazz-tools/tests/browser/worker-bridge.bench.test.ts`, uncommitted), same machine (Apple Silicon, Chromium headless), one run each.

| Scenario | Metric | Rust (this PR) | TS (main) | Δ |
|---|---|---:|---:|---:|
| `local-wait-rt` (30 iters) | p50 / p95 ms | 2002 / 2009 | 2005 / 2010 | ~0 |
| `burst-then-wait` (2000 inserts) | avg ms / insert | 10.04 | 10.04 | 0 |
| `local-query-rt` (500 iters) | p50 / p95 ms | 0.70 / 0.84 | 0.72 / 0.99 | **−5% / −15%** |
| `local-query-rt` | ops/sec | 1399 | 1331 | **+5%** |

`local-wait-rt` and `burst-then-wait` aren't bridge-bound (the first pins on `LOCAL_SYNC_ACK_TIMEOUT_MS`, the second on OPFS WAL writes), so the move is neutral there as expected. `local-query-rt` exercises the inbound dispatch path and the Rust bridge is ~5% faster on avg, ~15% better on p95. No regression detected; the wasm-bindgen boundary crossing added by the move is offset by postcard + transferable `ArrayBuffer` savings.

## Tests

- **`crates/jazz-wasm/tests/worker_bridge.rs`** — 27 wasm-bindgen-test cases against a synthetic `Worker` shim. Covers init/error/shutdown handshakes, pre-init buffering + init-gate flush, forwarder routing, drop-after-shutdown idempotency, and the `Shutdown`-during-init race.
- **`crates/jazz-wasm/tests/opfs.rs`** — OPFS persistence under `run_in_dedicated_worker`.
- **`packages/jazz-tools/tests/browser/worker-bridge.test.ts`** — existing browser integration tests still green.
- `todo-client-localfirst-react` dev server boots, 16/16 example tests pass in Chromium.
- `cargo check -p jazz-wasm --target wasm32-unknown-unknown` and `cargo check --workspace --tests` clean.

## Preserved / out of scope

- Public `Db` API, `postMessage` envelope, WebSocket transport, leader election, BroadcastChannel tab-sync, NAPI/RN bindings: untouched.
- Deferred: moving `db.ts` to talk to the Rust bridge directly and deleting the TS adapter.


crates/jazz-wasm/src/worker_bridge.rs
```diff
@@ -0,0 +1,1032 @@
...
+#[wasm_bindgen]
+pub struct WasmWorkerBridge {
+    inner: Rc<BridgeInner>,
+}
+
+#[wasm_bindgen]
+impl WasmWorkerBridge {
+    /// Attach a Rust bridge to an externally-constructed Worker.
+    ///
+    /// Per spec, options are parsed at attach time. `init()` is parameter-less.
```
gdorsi: Which spec is this referring to? 😄

crates/jazz-wasm/src/worker_bridge.rs
```diff
@@ -0,0 +1,1032 @@
...
+    #[wasm_bindgen(js_name = waitForLocalSyncFlush)]
+    pub fn wait_for_local_sync_flush(&self, batch_id: Option<String>) -> js_sys::Promise {
+        let inner = Rc::clone(&self.inner);
+        wasm_bindgen_futures::future_to_promise(async move {
+            if inner.is_disposed_like() {
+                return Ok(JsValue::UNDEFINED);
+            }
+            let init_promise = inner.init_promise.borrow().clone();
+            if let Some(promise) = init_promise {
+                let _ = JsFuture::from(promise).await;
```
gdorsi, line 325: When init fails (worker sends an error), the state becomes `Failed` but `is_disposed_like()` only checks for `Disposed` or `ShuttingDown`. So after the rejected init promise is silently discarded by `let _ =` on line 325, the loop keeps running — it tries to talk to a dead worker and waits the full 2 seconds for a reply that will never come.

A quick fix: bail early if the init promise was rejected, or include `Failed` in `is_disposed_like()`.

gdorsi:
Code LGTM, going to test it manually later

crates/jazz-wasm/src/worker_bridge.rs
antoniomuso: I removed the spec; let me clean it.

crates/jazz-wasm/src/worker_bridge.rs
antoniomuso: Done!

gdorsi (approved):

PR #820, opened 2026-05-07
feat: batches can read their own writes and be rolled back

nrainhart:
## Description

Now that batched writes are not immediately visible globally until the batch is committed, it makes sense for batches to allow reading their own writes, and also be rolled back.

This makes the API of transactions and batches identical, so we can simplify the code a lot as a bonus.

### DIFF
--- a/.changeset/batch-rollback-and-reads.md
+++ b/.changeset/batch-rollback-and-reads.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Add rollback and scoped read support to explicit batches. `DirectBatch`/`DbDirectBatch` now expose `rollback()`, and batch handles can read their own local writes before commit through `DirectBatch.query(...)` and `DbDirectBatch.all(...)`/`one(...)`.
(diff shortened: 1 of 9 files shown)


gdorsi:
if a batch only reads and never writes, the code may still try to commit an empty batch that was never really created in storage.

Could you double-check if this is a real bug?

Repro flow:

- Use a real runtime Db, not the mocked tests.
- Call db.batch(async (batch) => ...).
- Inside the callback, call await batch.all(...) or await batch.one(...).
- Based on that result, return without calling insert, upsert, update, or delete.
- runInBatch auto-calls commit() after the callback resolves.
- commit() calls sealBatch(...).
- Rust seal_batch looks for a local batch record, but reads do not create one.
- It fails with a missing local batch record error instead of returning the callback result.

gdorsi (approved):
LGTM

nrainhart:
> Could you double-check if this is a real bug?

It is, thanks for catching this. Fixed in [`43d0992` (this PR)](https://github.com/garden-co/jazz/pull/820/commits/43d09924c5c9b95d9b9f22150298c223d4f07b51)

PR #828, opened 2026-05-09
Fix invite write durability waits

gdorsi:
## What

Fixes the invite flow getting stuck while waiting for writes to reach the configured durability tier.

## Broken message flow before

1. The chat app creates the initial chat with an `edge` durability wait.
2. The main browser runtime records the new write batch as durable only at `local`.
3. The worker runtime syncs that same batch to the server.
4. The server confirms the batch and the worker observes a stronger settlement, for example `global` durability.
5. The worker hydrates its local batch records back into the main runtime.
6. The main runtime already has a batch record for the same batch id, so `localBatchRecord(batchId)` kept returning the main runtime's older `local` settlement.
7. `waitForPersistedBatch(batchId, "edge")` looked at that older `local` settlement, decided it had not reached `edge`, and stayed pending.
8. In the invite test, that meant `CreateChatRedirect` stayed on `Creating chat...`; the user never reached the chat view, so the invite flow could not continue.

## Why it works now

When worker-hydrated batch records arrive, the main client now compares the worker settlement with the runtime settlement for the same batch id.

If the worker record has a stronger durable settlement, such as `global` over `local`, the hydrated worker record is allowed to override the main runtime record for wait/reconciliation checks. So the same flow now becomes:

1. Main runtime still starts with a `local` batch record.
2. Worker still receives the stronger server-confirmed settlement.
3. Worker hydrates that stronger settlement back to main.
4. Main chooses the stronger worker-hydrated record.
5. `waitForPersistedBatch(batchId, "edge")` sees `global >= edge` and resolves.
6. The chat creation flow navigates to the chat, and the invite flow can continue.

Rejected worker-hydrated records still keep their previous behavior: they override non-rejected runtime records so mutation errors are surfaced.

## Changes

- Let worker-hydrated batch records override main runtime records when they carry a stronger durable settlement.
- Add a regression test for the worker hydration case.

## Validation

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/client.test.ts -t "worker batch hydration"`
- `cargo test -p jazz-tools rc_query_settled_before_data_should_not_drop_upstream_rows --lib --features test`
- `pnpm --filter chat-react exec vitest run --config vitest.config.browser.ts tests/browser/invite.test.tsx -t "allows a user to join a private chat via invite link"`

### DIFF
--- a/.changeset/fix-worker-hydrated-batch-records.md
+++ b/.changeset/fix-worker-hydrated-batch-records.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Use stronger worker-hydrated batch settlements when resolving durability waits in browser runtimes.
--- a/packages/jazz-tools/src/runtime/client.test.ts
+++ b/packages/jazz-tools/src/runtime/client.test.ts
@@ -3,6 +3,7 @@ import {
   WriteResult,
   JazzClient,
   resolveDefaultDurabilityTier,
+  type LocalBatchRecord,
   type MutationErrorEvent,
   type Runtime,
   WriteHandle,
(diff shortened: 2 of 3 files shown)


nrainhart:
Fixed by https://github.com/garden-co/jazz/pull/884

PR #840, opened 2026-05-11
fix: correctly resolve magic columns in subqueries

### LINKED ISSUE #825: Selecting magic columns in included relation returns null result
## Summary

When querying a parent table with an included reverse relation, adding Jazz magic columns like `$createdAt` and `$updatedAt` to the included relation's `.select(...)` causes the included relation to come back missing/null/empty. Selecting only normal table columns with `.select('*')` returns the expected related row.

## Environment

- `jazz-tools`: `2.0.0-alpha.46`
- Runtime: Node backend service
- Query is executed through `jazzService.asBackend().one(...)`

## Schema shape

The included table has a normal forward ref back to the parent table:

```ts
export const child = s.table({
  parentId: s.refs.parent(),
  name: s.string(),
});
```

## Query that unexpectedly fails

This query returns the parent row, but the reverse relation does not contain the expected child row when the included child relation selects magic columns:

```ts
const parent = await jazzService.asBackend().one(
  app.parent
    .where({ id: parentId })
    .include({
      childViaParent: app.child
        .limit(1)
        .select('*', '$createdAt', '$updatedAt'),
    })
    .requireIncludes(),
);
```

## Workaround

Removing the magic columns from the included relation makes the include resolve correctly:

```ts
childViaParent: app.child
  .limit(1)
  .select('*')
```

With that workaround, `parent.childViaParent[0]` is present as expected.

## Expected behavior

Selecting magic columns on an included relation should not change whether the related row is included. The row should still be returned, with `$createdAt` and `$updatedAt` populated on the included entity.

## Actual behavior

Selecting `'*'`, `'$createdAt'`, and `'$updatedAt'` on the included reverse relation causes the include to resolve as missing/null/empty, even though the same include works when selecting only `'*'`.

Gabrola:
## Summary

- Fix magic-column projection inside array subqueries and nested/reverse relation includes.
- Ensure selected magic columns like `$createdAt` and `$updatedAt` are added to subquery row descriptors even though they are not physical table columns.
- Centralize magic-column descriptor creation and add Rust + TS DSL coverage for timestamp magic columns in subquery/include projections.

## Motivation / Problem

Issue #825 reported that including a reverse relation worked with `.select("*")`, but returned a missing/null/empty included row when the relation selected Jazz magic columns such as `$createdAt` and `$updatedAt`.

The expected behavior is that selecting magic columns should only add those fields to the included entity, not change whether the related row is returned.

## Testing
- Added Rust query-manager coverage for array subqueries selecting `$createdAt` and `$updatedAt`.
- Added TS DSL tests for reverse relation includes and nested array relation includes projecting magic timestamp columns.

## Follow-up
This change also surfaced a separate issue around permission-introspection magic columns like `$canRead`, `$canEdit`, and `$canDelete`.
Those columns are more complex than provenance timestamps because they are relational: their values depend on policy evaluation, the active session/auth context, and related tables used by the permission graph. Supporting them correctly in subqueries would likely require forwarding the parent query session into subgraph compilation and subscribing relation/subquery nodes to the relevant policy and magic-column dependency tables, not just the directly included relation table.

Closes #825

### DIFF
--- a/crates/jazz-tools/src/query_manager/magic_columns.rs
+++ b/crates/jazz-tools/src/query_manager/magic_columns.rs
@@ -1,3 +1,5 @@
+use crate::query_manager::types::{ColumnDescriptor, ColumnType};
+
 pub const RESERVED_MAGIC_COLUMN_PREFIX: char = '$';
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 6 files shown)


crates/jazz-tools/src/query_manager/magic_columns.rs
```diff
@@ -38,6 +40,28 @@ pub fn magic_column_kind(name: &str) -> Option<MagicColumnKind> {
     }
 }
 
+pub(crate) fn magic_column_descriptor(kind: MagicColumnKind) -> ColumnDescriptor {
```
nrainhart, line 43: ✨

nrainhart (approved):
Thanks for the fix and the code cleanup @Gabrola!! Added the issue with `$canRead`, `$canEdit`, and `$canDelete` & subqueries to our backlog

PR #849, opened 2026-05-11
fix: update indexes on tx/batch commit

nrainhart:
## Description

Batched/transaction writes used to modify indexes, which are shared with "regular" queries. That means e.g. that deleting a row in a tx would also make it disappear from db.one/all queries.

This PR modifies batched/transaction writes to not modify indexes until the batch/tx is committed, and wires batch/tx queries to retrieve these rows from the overlay instead

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph_nodes/index_scan.rs
+++ b/crates/jazz-tools/src/query_manager/graph_nodes/index_scan.rs
@@ -113,7 +113,9 @@ impl IndexScanNode {
             ) else {
                 continue;
             };
-            if row.is_soft_deleted() || row.is_hard_deleted() {
+            if self.column.as_str() == "_id_deleted" && row.is_soft_deleted() {
+                new_ids.insert(row_id);
+            } else if row.is_soft_deleted() || row.is_hard_deleted() {
                 new_ids.remove(&row_id);
             } else if self.overlay_value_matches_condition(row_id, &row.data) {
                 new_ids.insert(row_id);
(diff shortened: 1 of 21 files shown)


crates/jazz-tools/src/query_manager/graph_nodes/index_scan.rs
```diff
@@ -65,11 +69,114 @@ impl IndexScanNode {
...
         &self.output_descriptor
     }
+
+    fn overlay_value_matches_condition(&self, row_id: ObjectId, data: &[u8]) -> bool {
+        let Some(value) = self.overlay_index_value(row_id, data) else {
+            return false;
+        };
+        match &self.condition {
+            ScanCondition::All => true,
+            ScanCondition::Eq(expected) => value == *expected || array_contains(&value, expected),
```
aeplay, line 79: don't we have a more appropriate operator for contains?

aeplay (changes requested):
One issue brought up by codex:

Staged deletes disappear from include_deleted queries" body="When the overlay row is a staged soft delete, `_id_deleted` scans currently remove the row from `new_ids` instead of adding it. `include_deleted` queries are compiled as a union of `_id` and `_id_deleted` scans, so inside a transaction/batch a deleted row is removed from the `_id` scan and never appears in the `_id_deleted` scan. That makes `include_deleted` return less data in the writer's own overlay than the same query would after commit. The `_id_deleted` overlay path should mirror the normal soft-delete index mutation by inserting the staged deleted row for `_id_deleted` scans.

crates/jazz-tools/src/query_manager/graph_nodes/index_scan.rs
aeplay: Not unique to this PR - will address in a follow up

aeplay (approved):

PR #850, opened 2026-05-11
Add repro for worker hydration durability wait

gdorsi:
## What

Adds an expected-failing regression test for the invite durability wait issue described in #828.

The test models the browser runtime case where the main runtime has an older `local` settlement for a batch, then the worker hydrates a stronger `global` settlement for the same batch. The expected behavior is that `waitForBatch(batchId, "edge")` can resolve from the stronger worker-hydrated record.

## Repro steps from #828

1. The chat app creates the initial chat with an `edge` durability wait.
2. The main browser runtime records the new write batch as durable only at `local`.
3. The worker runtime syncs that same batch to the server.
4. The server confirms the batch and the worker observes a stronger settlement, for example `global` durability.
5. The worker hydrates its local batch records back into the main runtime.
6. The main runtime already has a batch record for the same batch id, so `localBatchRecord(batchId)` keeps returning the main runtime's older `local` settlement.
7. `waitForBatch(batchId, "edge")` looks at that older `local` settlement, decides it has not reached `edge`, and stays pending.
8. In the invite test, that means `CreateChatRedirect` stays on `Creating chat...`; the user never reaches the chat view, so the invite flow cannot continue.

## Validation

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/client.test.ts -t "worker batch hydration"`
  - Result: 1 expected fail, 31 skipped.

### DIFF
--- a/packages/jazz-tools/src/runtime/client.test.ts
+++ b/packages/jazz-tools/src/runtime/client.test.ts
@@ -252,6 +252,7 @@ describe("JazzClient worker batch hydration", () => {
       }),
     ]);
 
+    // Current result: this still returns the runtime's older "local" settlement.
     expect(client.localBatchRecord(batchId)?.latestSettlement).toMatchObject({
       confirmedTier: "global",
     });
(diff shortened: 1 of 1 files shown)


nrainhart:
Fixed by https://github.com/garden-co/jazz/pull/884

PR #851, opened 2026-05-12
Make contains index scans explicit

aeplay:
## What changed

- Add an explicit `ScanCondition::Contains` for index scans.
- Lower `Condition::Contains` / `filter_contains` to that scan condition instead of scanning all rows.
- Keep equality overlay evaluation as strict equality only.
- Route referenced UUID-array contains scans through the existing expanded index lookup path while keeping that storage detail named as contains behavior.
- Add a runtime overlay regression test proving `filter_contains` sees a staged array-ref row while `filter_eq` does not match array membership.

## Why

PR #849 made local overlay evaluation treat equality as `equals OR array contains`, leaking the physical reference-array index expansion into query semantics. Query semantics already have an explicit contains operator, so equality should not broaden to arbitrary membership.

## Validation

- `cargo fmt --check`
- `pnpm lint`
- `cargo test -p jazz-tools rc_query_local_transaction_overlay_uses_contains_not_eq_for_array_refs -- --nocapture`
- `cargo test -p jazz-tools runtime_core::tests::query_subscription -- --nocapture`
- `cargo test -p jazz-tools query_manager::manager_tests::array_subqueries -- --nocapture`
- `cargo test -p jazz-tools graph_nodes::index_scan -- --nocapture`

## Notes

Attempted `pnpm --filter jazz-tools test:browser tests/browser/db.transaction-reads.test.ts`, but this worktree cannot load the browser Vitest config because `jazz-napi` is unresolved. A broader `pnpm --filter jazz-tools test ...` attempt similarly failed on unresolved `jazz-napi` / `jazz-wasm` and missing built `dist` artifacts.

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph/compile.rs
+++ b/crates/jazz-tools/src/query_manager/graph/compile.rs
@@ -2103,6 +2103,7 @@ fn apply_condition_to_builder(mut builder: QueryBuilder, condition: &Condition) 
 fn condition_to_scan(cond: &Condition) -> ScanCondition {
     match cond {
         Condition::Eq { value, .. } => ScanCondition::Eq(value.clone()),
+        Condition::Contains { value, .. } => ScanCondition::Contains(value.clone()),
         Condition::Lt { value, .. } => ScanCondition::Range {
             min: Bound::Unbounded,
             max: Bound::Excluded(value.clone()),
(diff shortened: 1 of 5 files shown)


nrainhart (approved):

aeplay:
Superseded by the new relational core (#1094). The old `query_manager` index-scan implementation this makes explicit is gone; `contains` semantics now live in `crates/groove`/`crates/jazz`. Any remaining requirement should be expressed as new-core tests rather than landed from this branch.

PR #852, opened 2026-05-12
Cache policy evaluation during server settlement

aeplay:
## Summary

- add a settlement-scoped `PolicyEvalCache` shared across a downstream client's active server-side query settlement generation
- reuse row authorization, inherited reference authorization, and recursive relation recompute results while settling authorized server query scopes
- clear client cache state on dirty recompute/client removal and add regression coverage for stale cache poisoning after initial settlement
- include a patch changeset for `jazz-tools`

## Validation

- `pnpm lint`
- `cargo fmt --check`
- `cargo check -p jazz-tools`
- `cargo test -p jazz-tools server_subscription_policy_eval_cache_is_cleared_after_initial_settlement_and_dirty_recompute -- --nocapture`

### DIFF
--- a/.changeset/policy-eval-settlement-cache.md
+++ b/.changeset/policy-eval-settlement-cache.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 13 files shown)


aeplay:
Closing in favor of #853

PR #853, opened 2026-05-12
Cache policy subexpressions during scope evaluation

aeplay:
## Summary

This keeps server-side authorization work scoped to the places where it is actually needed:

- skip resettling and re-authorizing server subscriptions whose graph is already clean and whose scope has already settled
- reuse the previous `last_scope` for clean subscriptions, while still emitting `QuerySettled` if durability tier progress requires it
- add a short-lived policy subexpression cache for repeated referenced-row authorization work inside a single authorization/scope evaluation pass
- add pass-local recursive relation result reuse keyed by stable relation semantics plus the full seed/input fingerprint

The cache intentionally does not store whole row-access decisions across subscription passes. After avoiding clean subscription recomputation, that broader row cache was no longer useful for the large initial-load path and carried a larger correctness surface.

The remaining cache namespaces are narrower:

- referenced-row authorization (`allowed.toRead(ref)` shape), keyed by branch, referenced row, operation, and evaluation depth
- recursive relation results, keyed by a stable semantic fingerprint for the compiled subgraph/template plus correlation/hop/depth metadata and a full seed fingerprint including tuple content, batch ids, row provenance, and tuple provenance

Both cache namespaces are scoped to the current settlement/evaluation pass. They do not survive into later ticks, so they avoid dependency-tracking and cross-subscription invalidation complexity.

The env-gated policy counters remain so we can keep measuring cache hit/miss behavior while tuning this path.

## Load timing

Using a large downstream app fixture with 39 simultaneous subscriptions and ~20k rows in the largest subscribed table, fresh server per run:

- cache narrowed to one evaluation pass, before clean-scope skip: `26384ms`, `25071ms`
- clean-scope skip with row cache still present: `7474ms`, `7246ms`
- simplified design with no row-access cache: `7094ms`, `7049ms`
- after removing unsafe recursive relation result caching: `14427ms`, `13832ms`
- final pass-local semantic recursive relation cache: `8449ms`, `8411ms`

Final counter shape with semantic recursive relation reuse:

- largest table row access stays to one pass: `19894`
- ref subexpr cache: `21` hits / `30` misses, branch-scoped
- recursive relation cache: `58` hits / `2` misses
- recursive relation evals: `2` actual evaluations per load

## Validation

- `cargo fmt --check`
- `cargo test -p jazz-tools recursive_cache_site_includes_step_semantics`
- `cargo check -p jazz-tools`
- `cargo build -p jazz-tools --bin jazz-tools --features cli`
- `pnpm turbo run build --filter=jazz-wasm --filter=jazz-tools --only`
- pre-commit lefthook clippy/rustfmt


crates/jazz-tools/src/row_format.rs
```diff
@@ -1323,6 +1324,9 @@ pub fn encode_value(value: &Value) -> Vec<u8> {
 /// Encode a Value to binary bytes with type information (needed for Row values).
 pub fn encode_value_with_type(value: &Value, col_type: &ColumnType) -> Vec<u8> {
     match (value, col_type) {
+        (Value::Text(raw), ColumnType::Uuid) => Uuid::parse_str(raw)
+            .map(|uuid| ObjectId::from_uuid(uuid).uuid().as_bytes().to_vec())
+            .unwrap_or_else(|_| encode_value(value)),
```
gdorsi: When `Uuid::parse_str` fails on a `Value::Text`, the fallback calls `encode_value(value)` which writes variable-length UTF-8 bytes into a column expecting exactly 16 UUID bytes. This will produce wrong-length encoded data. 

Should this return an error or encode a sentinel value instead?

crates/jazz-tools/src/query_manager/graph/compile.rs
```diff
@@ -2386,28 +2401,52 @@ fn apply_condition_to_builder(mut builder: QueryBuilder, condition: &Condition)
...
+    table_schema
+        .columns
+        .columns
+        .iter()
+        .find(|descriptor| descriptor.name == column)
+        .map(|descriptor| &descriptor.column_type)
+}
+
+fn normalize_scan_value(value: &Value, column_type: Option<&ColumnType>) -> Value {
+    match (value, column_type) {
```
gdorsi, line 2446: The text-to-UUID normalization in `column_scan_plan` has no dedicated test. The policy side has `test_session_user_id_matches_text_and_uuid_owner_columns`, but nothing covers the analogous index-scanning path.

Could you add one?

crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
```diff
@@ -240,14 +270,7 @@ impl<'a> PolicyContextEvaluator<'a> {
...
             PolicyExpr::ExistsRel { rel } => self.evaluate_exists_rel_with_context(
-                rel,
-                operation == Operation::Select,
-                row,
-                descriptor,
-                table_name,
-                io,
-                row_loader,
-                depth,
+                rel, true, row, descriptor, table_name, io, row_loader, depth,
```
gdorsi: ExistsRel` now hardcodes `structural_scans: true` instead of the previous `operation == Operation::Select`. This means write operations (Insert/Update/Delete) will now use structural scans when they previously did not.

Is this intentional?

crates/jazz-tools/src/query_manager/settlement_eval_cache.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub(crate) fn ref_access_get(&self, key: &RefAccessSubexprKey) -> Option<bool> {
+        self.ref_access.get(key).copied()
+    }
+
+    pub(crate) fn ref_access_insert(&mut self, key: RefAccessSubexprKey, result: bool) {
+        self.ref_access.insert(key, result);
+    }
+
+    pub(crate) fn relation_result_get(&self, key: &RelationSubexprKey) -> Option<AHashSet<Tuple>> {
+        self.relation_results.get(key).cloned()
```
gdorsi, line 45: Shall we use an `Rc` here to wrap the relation result, or is it ok to clone it on each cache hit?

crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
```diff
@@ -343,6 +366,47 @@ impl<'a> PolicyContextEvaluator<'a> {
             _ => return false,
         };
 
+        let cache_key = if depth == 0 && visited.is_empty() {
+            Some(RefAccessSubexprKey {
```
gdorsi, line 365: Is it intentional that the ref-access cache only activates at `depth == 0 && visited.is_empty()`? Since the guard ensures `depth` is always 0 here, `parent_eval_depth` in the cache key is always 1 — so the depth field in the key is effectively unused. Deeper hops (which tend to be the expensive ones in long inheritance chains) are never cached.

crates/jazz-tools/src/query_manager/policy_counters.rs
```diff
@@ -0,0 +1,100 @@
...
+
+static COUNTERS: OnceLock<Mutex<PolicyCounters>> = OnceLock::new();
+
+fn enabled() -> bool {
+    std::env::var_os("JAZZ_POLICY_COUNTERS").is_some()
+}
+
+pub(crate) fn increment(bucket: &'static str, key: impl Into<String>) {
+    if !enabled() {
+        return;
```
gdorsi, line 46: I made the "local telemetry" system to make this kind of debugging easier to setup and more effective.

Advantages:
- batching is already in place
- doesn't print, so it's not as slow
- output is stored in files that can be queried

See https://github.com/garden-co/jazz/tree/main/dev/local-telemetry

If you have some troubles on setting it up, feel free to ping me so we can improve the system.

gdorsi (approved):

crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
aeplay: removed the unused depth field

crates/jazz-tools/src/query_manager/settlement_eval_cache.rs
aeplay: ok to clone until it shows up in profiles. Many clones get optimized away during inlining anyways if short-lived

PR #855, opened 2026-05-12
refactor: simplify `QueryManager`'s write paths

nrainhart:
## Description

This is just a cleanup PR to reduce duplication in the implementation of `QueryManager`'s write methods. Behaviour remains unchanged

### DIFF
--- a/crates/jazz-tools/src/query_manager/writes.rs
+++ b/crates/jazz-tools/src/query_manager/writes.rs
@@ -986,193 +986,16 @@ impl QueryManager {
         values: &[Value],
         write_context: Option<&WriteContext>,
     ) -> Result<InsertResult, QueryError> {
-        self.insert_with_write_context_and_id(storage, table, values, None, write_context)
-    }
-
-    pub fn insert_with_write_context_and_id<H: Storage>(
-        &mut self,
-        storage: &mut H,
-        table: &str,
-        values: &[Value],
-        external_object_id: Option<ObjectId>,
-        write_context: Option<&WriteContext>,
-    ) -> Result<InsertResult, QueryError> {
         let _span = tracing::debug_span!("QM::insert", table).entered();
         let current_branch = self.current_branch().as_str().to_string();
-        let table_name = TableName::new(table);
-        let write_schema = self.schema.clone();
-        let table_write = self.write_table_cache_entry_for_schema(
-            &current_branch,
-            table_name,
-            write_schema.as_ref(),
-        )?;
-        let descriptor = table_write.descriptor.as_ref();
-        let insert_policy = table_write.insert_policy.as_deref();
-
-        if values.len() != descriptor.columns.len() {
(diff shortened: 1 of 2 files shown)


crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -1400,6 +1064,7 @@ impl QueryManager {
             None,
             write_schema,
             write_context,
+            false,
         )
```
gdorsi, line 1068: This consolidation makes the SchemaManager insert path stricter than before.

Before: when SchemaManager inserted a row and the table had no explicit insert policy, the old code just asked "is the table in the schema? OK, allowed." It didn't care about strict mode.

After: the same path now also checks "are we in strict mode?" If yes and the table has no insert policy, the insert gets blocked.

This matters more than it looks: `row_policy_mode` is set to `Enforcing` whenever *any* table in the schema has *any* policy (even just a SELECT policy on a completely different table). So a schema like:
- `users` table with a SELECT policy
- `posts` table with no INSERT policy

...would cause `posts` inserts through SchemaManager to be denied when a session is present. The old code allowed this.

Same concern applies to the `else` branch (line 1079) and to the switch from `None` to real `object_id` in the policy evaluator, which changes outcomes for `InheritsReferencing` policies.

gdorsi: I guess this is ok-ish, since we evaluate permissions there.

But I think it should be covered by an integration test if it's not.

crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -1227,160 +1034,17 @@ impl QueryManager {
...
-            let _ = self.apply_local_row_batch(storage, visibility_change)?;
-        }
-
-        Ok(InsertResult {
-            row_id: object_id,
-            batch_id: row_batch_id,
-            row_values: values.to_vec(),
-        })
+            true,
+        )
```
gdorsi, line 1047: Nice side-effect of the consolidation: `insert_on_branch_with_write_context_and_id` (and the rest of the `insert_on_branch` family) now rejects anonymous writes via `deny_anonymous_writes: true`. Before this PR, the branch-aware insert path had no anonymous check at all — an anonymous session could slip through to policy evaluation without hitting the guard that `insert_with_session` had.

Good fix! Could you add a test for `insert_on_branch_with_session` with an anonymous session to lock it in? The existing `anonymous_insert_is_denied_before_policy_eval` test only covers the user-facing `insert_with_session` path.

gdorsi (approved):
Always love to see LOC go down.

crates/jazz-tools/src/query_manager/writes.rs
nrainhart: Actually, after some further cleanup (https://github.com/garden-co/jazz/pull/897) I found most of these methods are only used only on tests 🙃

PR #862, opened 2026-05-13
fix: permission magic columns in array subqueries

Gabrola:
Follow-up on #840 

## Summary
- Preserve the parent session and row policy mode when compiling array and recursive subqueries.
- Fix reverse and nested include queries that select permission magic columns like `$canRead`, `$canEdit`, and `$canDelete`.
- Add regression coverage plus a patch changeset for `jazz-tools`.

## Testing
Added coverage:
- Rust regression test for array subgraphs inheriting the parent session when resolving permission magic columns.
- TS Query API regression tests for `$canRead`, `$canEdit`, and `$canDelete` on reverse relation includes and nested array relation includes.

Verified:
- `cargo test -p jazz-tools --lib` passed: 1193 passed, 0 failed.
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts tests/ts-dsl/query-api.test.ts` passed: 39 passed, 0 failed.

### DIFF
--- a/.changeset/fix-subquery-permission-magic-columns.md
+++ b/.changeset/fix-subquery-permission-magic-columns.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Fix queries with reverse relations that select permission magic columns such as `$canRead`, `$canEdit`, and `$canDelete`.
+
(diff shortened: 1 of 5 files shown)


gdorsi (approved):
Thanks for the contribution!

Added some tests to increase the coverage, the fix LGTM!

PR #864, opened 2026-05-13
spec: discriminable typescript errors

Gabrola:
## Summary

The TypeScript API currently throws too many plain, message-only `Error`s, so app code cannot reliably distinguish
failures with any reliable mechanism such as `instanceof` or typed `Error.name` values.


specs/todo/issues/discriminable-typescript-errors.md
```diff
@@ -0,0 +1,31 @@
...
+The TypeScript API currently throws too many plain, message-only `Error`s, so app code cannot reliably distinguish
+failures with any reliable mechanism such as `instanceof` or typed `Error.name` values.
+
+## Priority
+
+medium
+
+## Notes
+
+- Use `Error.name` as the stable discriminator. Do not add a separate `code` field.
```
nrainhart, line 14: Sounds good. I do think we should add a `JazzError` class to allow catch-all `instanceof` checks, but it'd be great to have a typed `Error.name` that we can use for exhaustiveness checking

specs/todo/issues/discriminable-typescript-errors.md
```diff
@@ -0,0 +1,31 @@
...
+- Keep names in the existing PascalCase style, e.g. `RecoveryPhraseError`, `PasskeyBackupError`, `FileNotFoundError`,
+  and `JazzAnonymousWriteDeniedError`.
+- Group errors by the surface area that can throw them. For example, database query/write errors should have their own
```
nrainhart, line 17: What do you mean by "grouping" exactly? Having separate error classes? Namespacing `Error.name` values?

Gabrola: Yes, exactly, but they will still all extend `JazzError` ultimately. It's just to make error handling a bit more sensible. For example a DB query/write could not possibly ever throw a `PasskeyBackupError` so we could expose something like `JazzQueryError` that would not include `PasskeyBackupError` in the union of possible `Error.name`s so exhaustiveness checking is a bit easier.

I'll need to double check the feasibility of this while implementing because it might not be really feasible to have clean boundaries between the different error types and which services/APIs can possibly throw them

aeplay:
🤖 Closing this obsolete spec-only branch. Its old tracking tree no longer exists and several proposed typed public errors have since landed. The remaining cross-Rust, WASM, NAPI, and TypeScript structured-error contract is tracked in #2061, including a security review of which policy-denial details are safe to expose.

PR #865, opened 2026-05-13
Edge nodes forward catalogue commands to core node

Takeno:
Merge #867 first

## Summary
- Remove the explicit `catalogueAuthority*` configuration and derive edge catalogue behavior from `upstreamUrl`.
- In edge mode, validate the admin secret locally, then forward catalogue HTTP reads and writes to the upstream core.
- Update the CLI, NAPI, TypeScript dev-server surfaces, route handling, and status-quo specs for the new model.
- Add real core + edge `TestingServer` coverage for schema and permissions catalogue forwarding.

## Testing
- `cargo test -p jazz-tools --features test --test catalogue_sync_integration edge_catalogue_http_reads_and_writes_forward_to_real_core`
- `cargo fmt --check`
- `git diff --check` on the touched files


gdorsi (commented):
x

crates/jazz-tools/tests/catalogue_sync_integration.rs
```diff
@@ -74,6 +100,141 @@ async fn seed_schema_catalogue(server: &TestingServer, schema: &jazz_tools::Sche
...
+//          | forwards after local admin-secret validation
+//          v
+//   core TestingServer
+//
+// The assertions verify that writes sent to the edge are persisted by the real
+// core, and reads sent to the edge return the core catalogue state.
+#[tokio::test]
+async fn edge_catalogue_http_reads_and_writes_forward_to_real_core() {
+    let app_id = TestingServer::default_app_id();
+    let core = TestingServer::builder()
```
gdorsi, line 120: Nit: both core and edge use the same default admin secret here, so if `forward_catalogue_request` accidentally hardcoded the wrong credential (or an empty string), this test would still pass. Consider giving the core a distinct admin secret (e.g. `with_admin_secret("core-admin-secret")`) so the test actually validates that the edge relays the right value.

gdorsi (approved):
LGTM

PR #867, opened 2026-05-13
Pull core catalogue on edge reconnect and keep edge state in sync

Takeno:
## Summary
- Make edge reconnects verify their catalogue hash against core and pull missing catalogue entries on reconnect
- Keep core as the source of truth for catalogue data while preventing direct catalogue writes from edges
- Ensure catalogue updates accepted by core propagate to connected edges
- Add coverage for reconnect replay, edge-forwarded migrations, WS upstream URLs, and the catalogue hash refresh path
- Update docs for embedded server `adminSecret` behavior and edge `--upstream-url` requirements

## Testing
- Added and ran integration coverage for edge/core migration forwarding, WS upstream edge sync, and stale-edge reconnect replay
- Added runtime regression tests for catalogue hash refresh behavior on sync
- Ran formatting and lint checks; all passed


docs/content/docs/getting-started/client-setup.mdx
```diff
@@ -268,7 +268,7 @@ All plugins accept the same base options:
 | `server`      | `true` (default) starts an embedded local server. `false` disables the plugin. A URL string connects to an existing server. An object configures the embedded server (see below). |
 | `schemaDir`   | Directory containing `schema.ts` and `permissions.ts`. Defaults to the project root, or `src/lib/` for SvelteKit.                                                                 |
 | `appId`       | Override the app ID. Defaults to the value in `.env`, otherwise a generated UUID persisted on first run.                                                                          |
-| `adminSecret` | Required when `server` is a URL. Ignored for the embedded server (a random secret is generated).                                                                                  |
+| `adminSecret` | Required when `server` is a URL. For the embedded server, used as its admin secret unless `server.adminSecret` is set; otherwise a random secret is generated.                    |
```
gdorsi, line 236: 🤖 This docs change also needs the packaged docs index regenerated. `packages/jazz-tools/bin/docs-index.txt` still says root `adminSecret` is ignored for embedded servers, so packaged docs readers will keep seeing the old rule.

docs/content/docs/getting-started/server-setup.mdx
```diff
@@ -31,8 +31,11 @@ Your app ID namespaces your data for storage and sync. Click below to generate o
 | `--allow-local-first-auth`          | Allow local-first auth (`Authorization: Bearer <self-signed Jazz JWT>`)                                                                  | `JAZZ_ALLOW_LOCAL_FIRST_AUTH` | see `NODE_ENV` note below |
 | `--backend-secret <BACKEND_SECRET>` | Enable backend session impersonation                                                                                                     | `JAZZ_BACKEND_SECRET`         | unset                     |
 | `--admin-secret <ADMIN_SECRET>`     | Required for `deploy`, `migrations push`, and schema catalogue reads. In development mode, structural schema auto-sync works without it. | `JAZZ_ADMIN_SECRET`           | unset                     |
+| `--upstream-url <UPSTREAM_URL>`     | Run as an edge server connected to the upstream core server. Requires `--peer-secret` and `--admin-secret`.                              | `JAZZ_UPSTREAM_URL`           | unset                     |
```
gdorsi, line 34: 🤖 Please regenerate the packaged docs index too. `packages/jazz-tools/bin/docs-index.txt` does not include `--upstream-url` or `--peer-secret`, so packaged docs readers will not see how to start edge mode.

gdorsi (commented):
Two docs index follow-ups.

crates/jazz-tools/src/runtime_core/sync.rs
```diff
@@ -289,13 +295,34 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
...
-            sm.add_client_with_storage(&self.storage, client_id);
-            sm.set_client_role(client_id, ClientRole::Peer);
+        let client_existed = sm.get_client(client_id).is_some();
+
+        if !client_existed {
+            sm.add_client(client_id);
+        }
+        sm.set_client_role(client_id, ClientRole::Peer);
+
+        let queued_catalogue_replay = sm.queue_catalogue_sync_to_client_if_hash_mismatch(
```
gdorsi, line 371: 🤖 This hash-based skip is only wired into the peer path. Backend/admin client setup still uses `add_client_with_storage`, so a first connection or a reconnect after the client state has been reaped queues the full catalogue replay even when the handshake already sent a matching catalogue hash. That can add avoidable load for deployments with many catalogue entries.

gdorsi (approved):
LGTM

crates/jazz-tools/src/sync_manager/inbox.rs
```diff
@@ -1287,6 +1279,14 @@ impl SyncManager {
...
+                        self.apply_payload_from_client(
+                            storage,
+                            client_id,
+                            payload,
+                            AuthoritativeFateRecording::Skip,
+                        );
+                    }
                     ClientRole::User => {
                         let Some(_session) = &client.session else {
                             self.outbox.push(OutboxEntry {
```
gdorsi, line 1309: This PR blocks peers from writing catalogue entries via `CatalogueEntryUpdated` (line 1273), but the same data can still slip through here as a `RowBatchCreated`. The `Peer` arm doesn't check `is_catalogue()` before accepting the write — `Backend` and `User` both do.

Split `Peer` out of `Peer | Admin` and add the `is_catalogue()` guard.

crates/jazz-tools/src/transport_manager.rs
```diff
@@ -199,7 +199,7 @@ pub struct AuthConfig {
 
 impl AuthConfig {
     pub fn can_publish_catalogue(&self) -> bool {
-        self.admin_secret.is_some() || self.peer_secret.is_some()
+        false
```
gdorsi: `can_publish_catalogue()` always returns `false` now, so the `can_publish_catalogue` field on `TransportHandle`, its accessor, the gating check in `flush_runtime_outbox`, and the test asserting both secrets don't grant it are all dead code — no path ever reaches `true`.

Either remove the whole indirection, or add a comment here explaining the plan to re-enable it per-transport later.

PR #879, opened 2026-05-14
Add reactive LocalFirstAuth for Svelte and Vue

joeinnes:
## Summary
- New `LocalFirstAuth` reactive class in `jazz-tools/svelte` and `useLocalFirstAuth` composable in `jazz-tools/vue`, mirroring the existing React hook. Both read the secret store only on the client and notify sibling instances on `login`/`signOut` so a single sign-out invalidates the whole tree.
- `BrowserAuthSecretStore` throws a clearer error when used outside a browser and resolves `localStorage` lazily, so server-side imports of the module singleton don't blow up.
- `jazzSvelteKit` plugin restarts the dev server on warm starts when `PUBLIC_JAZZ_SERVER_URL` is missing from the captured env — previously only the first-ever cold start triggered a restart, leaving warm runs unable to reach the dev server.

## Test plan
- [ ] `pnpm test` in `packages/jazz-tools` (vue + svelte + sveltekit configs).
- [ ] Scaffold a SvelteKit app using the docs snippet, log in / sign out / refresh — verify the provider remounts cleanly each time and `$env/dynamic/public.PUBLIC_JAZZ_SERVER_URL` resolves on warm starts.
- [ ] Same for a Vue + Vite app via the docs snippet.

### DIFF
--- a/.changeset/reactive-local-first-auth.md
+++ b/.changeset/reactive-local-first-auth.md
@@ -1,0 +1,12 @@
+---
+"jazz-tools": patch
+---
+
+Added reactive local-first auth helpers for Svelte and Vue, matching `useLocalFirstAuth` in `jazz-tools/react`:
+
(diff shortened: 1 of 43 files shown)


gdorsi (approved):
LGTM

PR #880, opened 2026-05-14
fix: mark local batch rows rejected only for replayble

Takeno:
## Summary

Fix server startup hanging on large databases when stored authoritative `Rejected` batch fates have no local replay metadata.

## Cause

Startup replay was scanning all stored authoritative `Rejected` fates and treating each one as a local rejected batch that might need optimistic rows retracted.

For rejected fates that no longer had a local batch record or sealed batch submission, `local_batch_rows()` fell back to a global history scan across all row locators. On the reported database this meant roughly:

- 3,015 rejected fates without local replay metadata
- 18,891 row locators
- ~57M worst-case row-history scans before the server could bind its port

These fates are server-side authoritative history, not necessarily local optimistic writes. Replaying them at startup was both semantically too broad and very expensive.

## Fix

Startup now only replays stored rejected fates when there is local replay metadata proving the batch may own optimistic local rows:

- persisted sealed batch submission
- persisted local batch record

The startup replay path also uses an indexed/member-based lookup only, avoiding the unindexed global row-history scan during `RuntimeCore::new`.

The global fallback remains available for the recovery/reconciliation paths that still need it.

## Why This Fix

The original startup replay check existed to cover a real crash-recovery case: if the runtime crashes after persisting a rejection settlement but before retracting the optimistic visible row, restart must apply the rejection before queries can observe stale data.

That case still has local replay metadata, so it continues to be handled.

Orphan authoritative rejected fates without local replay metadata cannot reliably be attributed to local optimistic rows. Treating them as startup replay work causes pathological startup cost on server databases that retain many historical rejected fates.

## Consequences

Startup no longer surfaces orphan server-side rejected fates as local mutation errors.

If a database contains a corrupted local optimistic row whose rejected batch has lost both its sealed submission and local batch record, startup will not repair it through the global scan anymore. That state is not reliably distinguishable from ordinary server history, so repairing it should be handled by an explicit migration/diagnostic path rather than blocking every startup.

Expected effect: large server databases with retained rejected fates should reach the listening state without the rejected-fate × row-locator scan explosion.

## Tests

Added regression coverage for orphan rejected fates with row locators but no local replay metadata.

Verified:

```bash
cargo test -p jazz-tools --features test --lib
```

Result: `1398 passed, 0 failed`.

### DIFF
--- a/crates/jazz-tools/src/runtime_core/mod.rs
+++ b/crates/jazz-tools/src/runtime_core/mod.rs
@@ -367,8 +367,14 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
             .scan_acknowledged_rejected_batch_fates()
             .map(|batch_ids| batch_ids.into_iter().collect())
             .unwrap_or_default();
+        let replayable_rejected_batch_fates = rejected_batch_fates
+            .into_iter()
+            .filter(|fate| {
+                Self::has_startup_rejected_batch_replay_metadata(&storage, fate.batch_id())
+            })
+            .collect::<Vec<_>>();
         let pending_mutation_error_events: BTreeMap<BatchId, MutationErrorEvent> =
-            rejected_batch_fates
+            replayable_rejected_batch_fates
                 .iter()
                 .filter_map(|fate| {
                     let crate::batch_fate::BatchFate::Rejected {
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Closing because this startup-replay implementation targets the deleted `RuntimeCore` fate-record path. The bounded-startup performance question may be retained as a fresh benchmark if it reproduces on the new core.

PR #884, opened 2026-05-14
refactor: simplify `.wait({ tier })` and `onMutationError` logic

nrainhart:
## Description

- In browsers, the in-memory DB was creating local batch acks as soon as a batch was sealed, when it should actually be waiting for worker persistence: https://github.com/garden-co/jazz/commit/a67393ec904074ee695649a1875cb87539135665
- Greatly simplifies how batch acks/rejections are propagated between the worker and the main thread. We used to have a whole protocol parallel to sync payloads to share this data, now the main thread's runtime relies only on sync payloads for this.
	- The only case where we still forward acks/rejections from the worker to the main thread is on startup, since the worker may have pending/rejected batches that the main thread still needs to keep track of. This is addressed by a hydration process that runs only on startup.
- Reduced code duplication between `Db` and `ClientBackedDb`: https://github.com/garden-co/jazz/commit/ac9204048e132e743d63234a0eb6cdad9046e88e


crates/jazz-tools/src/runtime_core/mod.rs
```diff
@@ -450,6 +485,53 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
...
+    }
+
+    pub fn set_rejected_batch_acknowledged_callback(
+        &mut self,
+        callback: Option<RejectedBatchAcknowledgedCallback>,
+    ) {
+        self.rejected_batch_acknowledged_callback = callback;
+    }
+
+    pub fn pending_mutation_error_delivery(
```
gdorsi, line 509: `pending_mutation_error_delivery` drains events via `drain_mutation_error_events`. If the callback returns `false` for some events (indicating "not handled"), those events are lost — they were already removed from the tracker. Should unhandled events be re-queued?

crates/jazz-tools/src/runtime_core/writes.rs
```diff
@@ -886,15 +891,18 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
...
+                    confirmed_tier,
+                };
+                record.apply_fate(settlement.clone());
+                self.storage
+                    .upsert_authoritative_batch_fate(&settlement)
+                    .map_err(|err| {
+                        RuntimeError::WriteError(format!("persist batch fate: {err}"))
+                    })?;
+            }
             self.publish_direct_batch_rows(&record)?;
```
gdorsi, line 906: When `synthesize_direct_write_fate` is false and the batch is `Direct`, no fate is written. But `publish_direct_batch_rows` is still called unconditionally after the `if let`. The batch record never gets `apply_fate` called — are downstream consumers okay receiving rows from a batch with no fate? Could cause the batch to stay in a limbo state if no remote peer ever sends a fate.

crates/jazz-wasm/src/worker_bridge.rs
```diff
@@ -670,6 +582,71 @@ impl BridgeInner {
...
+            }
+        }
+    }
+
+    fn hydrate_worker_local_batch_record(&self, record: LocalBatchRecord) {
+        let should_reconcile = local_batch_record_needs_fate_reconciliation(&record);
+        let batch_id = record.batch_id;
+        let mut core = self.runtime.core.borrow_mut();
+        if let Some(BatchFate::Rejected { code, reason, .. }) = record.latest_fate.clone() {
+            if let Err(error) = core.replay_batch_rejection(record.batch_id, &code, &reason) {
```
gdorsi, line 611: Might be worth checking: `replay_batch_rejection` is called before `hydrate_local_batch_record`, so the `local_batch_record(batch_id)` lookup inside `replay_batch_rejection` would miss — the record isn't in `MemoryStorage` yet. That means the `MutationErrorEvent` would get a synthetic stub instead of the real worker record. Would swapping the order (hydrate first, then replay) give the error event richer data, or is there a reason for doing it this way?

crates/jazz-napi/src/lib.rs
```diff
@@ -857,31 +836,22 @@ impl NapiRuntime {
...
     pub fn on_mutation_error(&self, callback: MutationErrorTsfn) -> napi::Result<()> {
-        {
-            let mut slot = self
-                .mutation_error_callback
-                .lock()
-                .map_err(|_| napi::Error::from_reason("lock"))?;
-            *slot = Some(callback);
-        }
-        emit_pending_mutation_errors(&self.core, &self.mutation_error_callback);
+        let callback: MutationErrorCallback = Arc::new(move |event| {
```
gdorsi, line 838: The callback is dispatched with `NonBlocking` (fire-and-forget) but always returns `true`, so `acknowledge_handled_rejected_batch` fires immediately — before the JS callback has actually executed.

Worth flagging with a comment if this behavior is intentional

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -2334,8 +2271,12 @@ class ClientBackedDb extends Db {
   }
 
   override onMutationError(listener: (event: MutationErrorEvent) => void): () => void {
```
gdorsi, line 2273: Why does this return a no-op unsubscribe?

gdorsi (approved):
LGTM

crates/jazz-wasm/src/worker_bridge.rs
```diff
@@ -670,6 +582,71 @@ impl BridgeInner {
...
+    fn hydrate_worker_local_batch_records(&self, encoded_records: Vec<serde_bytes::ByteBuf>) {
+        for row in encoded_records {
+            match LocalBatchRecord::decode_storage_row(row.as_ref()) {
+                Ok(record) => self.hydrate_worker_local_batch_record(record),
+                Err(error) => tracing::warn!("decode worker local batch record: {error:?}"),
+            }
+        }
+    }
+
+    fn hydrate_worker_local_batch_record(&self, record: LocalBatchRecord) {
```
gdorsi, line 606: Does this replace the logic we commented out in https://github.com/garden-co/jazz/pull/882?

crates/jazz-tools/src/runtime_core/mod.rs
nrainhart: So false currently means “delivered but not acknowledged,” not “not delivered”. I think this is unnecessary complexity. `onMutationError` is a notification surface, not a transactional ack protocol. I’d lean toward simplicity here and just mark a rejection as delivered/ack-ed at the runtime core when the callback is invoked

crates/jazz-napi/src/lib.rs
nrainhart: See https://github.com/garden-co/jazz/pull/884#discussion_r3248847655

packages/jazz-tools/src/runtime/db.ts
nrainhart: `ClientBackedDb` is used only for tests, I just didn't want to keep adding unnecessary complexity here

crates/jazz-wasm/src/worker_bridge.rs
nrainhart: It should. I'll double-check and clean this up

nrainhart: It turned out to be necessary, but I simplified it and configured it to run only on client DBs: [`82e1fc0` (this PR)](https://github.com/garden-co/jazz/pull/884/changes/82e1fc060ea8fc0e091ae612ddb1fd11b09dc07e)

crates/jazz-tools/src/runtime_core/mod.rs
nrainhart: [`300c178` (this PR)](https://github.com/garden-co/jazz/pull/884/commits/300c178ec7fdf6c8111f2ee345c76aaddc1d95b4)

crates/jazz-wasm/src/worker_bridge.rs
nrainhart: `hydrate_local_batch_record(record)` calls `upsert_local_batch_record(&record)`, and storage also persists `record.latest_fate` into the authoritative fate table. So if we hydrate first, then call `replay_batch_rejection()`, the batch rejection will be considered as already delivered, and thus `onMutationError` won't be triggered.

crates/jazz-tools/src/runtime_core/writes.rs
nrainhart: This "limbo state" is intentional for the browser main-thread runtime. We're relying on having a the worker runtime eventually answer with the record's fate

PR #892, opened 2026-05-15
Drop Node.js 20 support; require Node 22.12 (Jod LTS)

joeinnes:
## Summary
- Raise `engines.node` to `>=22.12` (Jod LTS) on the root and published `jazz-tools` / `create-jazz` / `inspector`; drop Node.js 20. Changeset included (patch).
- Bump `@types/node` to `^22` across the examples and prune the lockfile accordingly.
- Remove Node-20-era workarounds: the `import.meta.dirname ?? __dirname` fallback (19 files) and a redundant `?? {}` spread guard in the realistic benchmark runner.
- De-version-pin the MCP backend-selection comments/warning (it's a runtime capability check, not Node-version-gated) and fix the `nextjs-csr-ssr` example so it builds (sync URL + demo backend secret).

## Why now
Node.js 20 has reached end-of-life and Node.js 22 is now the oldest supported LTS line, so 22.12 (Jod LTS) is a safe new minimum. See the [Node.js release/EOL schedule](https://nodejs.org/en/about/eol).

## Test plan
- [ ] `pnpm build` — full workspace builds clean on Node 22.12+ (verified locally)
- [ ] `pnpm test`
- [ ] Confirm `pnpm install` on Node 20 emits the `EBADENGINE` warning

### DIFF
--- a/examples/auth-simple-chat/package.json
+++ b/examples/auth-simple-chat/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "@types/express": "^4.17.21",
-    "@types/node": "^22.0.0",
+    "@types/node": "catalog:default",
     "@types/react": "^19.0.0",
     "@types/react-dom": "^19.0.0",
     "@vitejs/plugin-react": "catalog:default",
--- a/examples/auth-simple-chat/package.json
+++ b/examples/auth-simple-chat/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "@types/express": "^4.17.21",
-    "@types/node": "^20.0.0",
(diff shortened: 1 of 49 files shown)


examples/auth-simple-chat/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "@types/express": "^4.17.21",
-    "@types/node": "^20.0.0",
+    "@types/node": "^22.0.0",
```
nrainhart: Nit: should we move the node version into the pnpm catalog?

nrainhart (approved):

PR #898, opened 2026-05-17
[codex] Add composite index query acceleration

tobiaslins:
## What changed

This PR adds maintained composite indexes for hot scoped pagination queries, plus the operational plumbing needed to build and use them safely:

- Adds `CompositeIndex`, `CompositeIndexColumn`, and `IndexDirection` schema support.
- Encodes composite index keys as ordered byte tuples, including descending components.
- Plans `scope = value ORDER BY column LIMIT n` queries onto ready composite indexes.
- Plans global `ORDER BY column LIMIT n` queries onto one-column composite indexes.
- Auto-maintains system timestamp indexes for `$createdAt` and `$updatedAt`.
- Plans global `ORDER BY $createdAt/$updatedAt LIMIT n` onto those system indexes.
- Adds forward and reverse limited index scans, so one ordered physical index can satisfy both ASC and DESC reads.
- Adds limited index-range scans for memory, OPFS BTree, and RocksDB storage backends.
- Adds online composite-index backfill with persisted build state and readiness tracking.
- Adds budgeted backfill APIs so large index builds can be chunked instead of completed in one tick.
- Keeps composite indexes out of `SchemaHash`, so adding/changing an index backfills an existing branch instead of creating a new branch identity.
- Fixes write-table cache invalidation by including a composite-index signature in the cache key.
- Avoids copying all existing pending local rows into every synced subscription.
- Adds schema validation for composite index names/components and namespace collisions.
- Adds scoped pagination and write-amplification benchmarks.

## Why

The target use case is a very large table, e.g. 10M rows across many users/orgs, where each user owns roughly 100-10k rows and the hot query is:

```sql
WHERE owner_id = current_user
ORDER BY updated_at DESC
LIMIT 50
```

A single-column `owner_id` index avoids a full table scan, but it still has to load all rows for that owner, sort by `updated_at`, and then apply the limit. A composite `owner_id ASC, updated_at DESC` index lets storage scan exactly the user's ordered slice and stop after the page size.

The same machinery now also supports admin-style global queries:

```sql
ORDER BY updated_at DESC
LIMIT 50
ORDER BY $updatedAt DESC
LIMIT 50
```

A one-column composite index like `updated_at DESC` can serve user-declared timestamps directly. Internal `$createdAt`/`$updatedAt` are maintained automatically as system indexes. The reverse scan path means one physical ordered index can serve both ASC and DESC.

## Performance notes

The 10M-row RocksDB scoped pagination benchmark showed the composite limited scan bringing the hot query into the sub-millisecond range after the subscription pending-row cloning issue was fixed.

The write-amplification benchmark shows the expected tradeoff:

- Default all-column indexing now reports 11 logical insert index entries: `_id`, 8 user columns, `$createdAt`, `$updatedAt`.
- Minimal intentional single-column indexing reports 6 logical insert index entries: `_id`, 3 user columns, `$createdAt`, `$updatedAt`.
- Adding one or two composite indexes adds predictable write cost, but buys the fast scoped ordered reads.
- Title-only updates now still mutate `$updatedAt`, so they carry 2 system timestamp index mutations even when no user column index changes.

The last encoder micro-optimization was intentionally small: it removed per-component allocation for composite key construction by appending into a shared buffer. Criterion showed only tiny/noisy improvements, so the important optimization remains intentional indexing plus composite limited scans.

## Developer impact

Composite indexes are currently schema-declared and only planned once their backfill state is ready. Adding or changing an index on an existing table does not change branch identity; it rebuilds index entries in place and keeps the planner off the index until ready.

System timestamp indexes are automatic and schema-free. They cost two index entries on insert, and `$updatedAt` changes add two index mutations on update. That feels worth it for admin/debug/feed-style latest-row queries.

Since each Jazz Cloud tenant gets isolated deployment/storage, index builds do not need cross-tenant fairness machinery in this layer. The useful operational follow-up is per-deployment visibility and controls: backfill progress, throttling/pause, cancellation, and stuck-job detection.

## Validation

Ran:

- `cargo fmt`
- `cargo test -p jazz-tools magic_ordered_limit --lib`
- `cargo test -p jazz-tools system_timestamp --lib`
- `cargo test -p jazz-tools system_updated_at --lib`
- `cargo test -p jazz-tools global_ordered_limit --lib`
- `cargo test -p jazz-tools index_range_limited_reverse --lib`
- `cargo test -p jazz-tools opfs_btree_index_ops --lib`
- `cargo test -p jazz-tools composite --lib`
- `cargo test -p jazz-tools --features rocksdb index_range_limited_reverse --lib`
- `cargo test -p jazz-tools --features rocksdb composite --lib`
- `cargo bench -p jazz-tools --bench write_amplification_benchmark --no-run`
- `cargo bench -p jazz-tools --bench write_amplification_benchmark`

Also used the scoped pagination benchmark/example during development to validate the 10M-row RocksDB path.


aeplay:
🤖 Closing because the new Groove substrate already provides variant-aware live-index admission and backfill. The legacy planner/storage implementation should not be ported; any remaining user-facing ordered-index feature should start as a fresh PR.

PR #899, opened 2026-05-18
Harden graceful server shutdown lifecycle

Takeno:
## Summary
- Adds an internal HTTP shutdown path gated by admin secret
- Coordinates graceful shutdown across websocket drain, runtime flush, and storage close
- Propagates shutdown failures through server, client, WASM, NAPI, and RN bindings
- Adds coverage for shutdown lifecycle, websocket gating, and storage flush failure handling

## Testing
- `cargo test -p jazz-tools --features test --lib server::`
- `cargo test -p jazz-tools --features test --lib storage_flush`
- `cargo test -p jazz-tools --features test --lib runtime_tokio::tests::flush`
- `cargo test -p jazz-tools --features test --bin jazz-tools`
- `git diff --check`

### DIFF
--- a/crates/jazz-tools/src/commands/server.rs
+++ b/crates/jazz-tools/src/commands/server.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 
 use jazz_tools::middleware::AuthConfig;
 use jazz_tools::schema_manager::AppId;
-use jazz_tools::server::{ServerBuilder, StorageBackend};
+use jazz_tools::server::{ServerBuilder, ShutdownPhase, StorageBackend};
 use tracing::info;
 
 const STANDALONE_INSPECTOR_URL: &str = "https://jazz2-inspector.vercel.app/";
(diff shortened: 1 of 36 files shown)


crates/jazz-tools/src/server/shutdown.rs
```diff
@@ -0,0 +1,351 @@
...
+            return;
+        }
+
+        let mut rx = self.subscribe();
+        while rx.borrow().is_running() {
+            if rx.changed().await.is_err() {
+                break;
+            }
+        }
+    }
```
gdorsi, line 108: Possible race between `try_enter_app_request` / `try_enter_websocket` and `wait_for_*_drain`:

The drain waiter can return `true` ("all drained") while a `try_enter_*` call is still mid-rollback. Here's a concrete interleaving:

```
Thread A (try_enter_app_request)          Thread B (request_shutdown)         Drain Waiter
─────────────────────────────────         ───────────────────────────         ────────────
1. is_shutting_down() → false
   (passes first check)
                                          2. requested.swap(true)
                                          3. set_phase(ShuttingDown)
                                          4. drain_notify.notify_waiters()
                                                                              5. wakes up
                                                                              6. active_app_requests() → 0
                                                                              7. returns true  ← "all drained!"
8. fetch_add(1) → count is now 1
9. is_shutting_down() → true
10. fetch_sub(1) → count back to 0
11. drain_notify.notify_waiters()
    (nobody listening, waiter already exited)
12. returns None
```

The drain waiter sees count == 0 at step 6 and exits, but Thread A is still executing inside `try_enter_app_request` (steps 8–11), touching `self.inner` atomics and calling `notify_waiters`. The request *is* ultimately rejected, so no guard leaks — but the caller of `wait_for_app_request_drain` may start tearing down resources (closing storage, dropping the runtime) while Thread A's rollback is still in flight.

Likely low-risk in practice (the rollback is just two atomic ops + a notify call, and the `Arc` keeps memory alive), but the contract "drain returned `true` → all request processing is finished" is technically violated.

crates/jazz-tools/src/server/routes/websocket.rs
```diff
@@ -424,6 +473,13 @@ async fn handle_ws_connection(
     heartbeat.tick().await; // consume the immediate first tick
     loop {
         tokio::select! {
+            biased;
+            changed = shutdown_rx.changed() => {
+                if changed.is_ok() && state.shutdown.is_shutting_down() {
+                    close_ws_for_shutdown(&mut socket).await;
+                    break;
+                }
+            }
```
gdorsi, line 482: This arm has a busy-spin bug when the shutdown sender gets dropped.

Here's what happens step by step:

1. This `select!` uses `biased;`, which means Tokio always checks the arms **top to bottom** — so `shutdown_rx.changed()` is checked first on every loop iteration.
2. When the sender side of a `watch` channel is dropped, `changed()` doesn't hang — it resolves **immediately** with `Err`.
3. When that happens, `changed.is_ok()` is `false`, so we skip the `if` block and do… nothing.
4. The loop continues, we're back at step 1, and `changed()` resolves immediately again.

Since there's no `await` that actually waits, this loop spins as fast as the CPU can go — 100% core usage doing nothing useful.

The handshake `select!` blocks (lines 320 and 386) handle this correctly — they close the socket and return on both `Ok` and `Err`. This main loop should do the same: `break` when the sender is dropped.

Something like:

```rust
changed = shutdown_rx.changed() => match changed {
    Ok(()) if state.shutdown.is_shutting_down() => {
        close_ws_for_shutdown(&mut socket).await;
        break;
    }
    Ok(()) => {} // changed but not shutting down yet — keep looping
    Err(_) => break, // sender dropped — no one to signal us, stop gracefully
}
```

crates/jazz-tools/src/server/hosted.rs
```diff
@@ -66,36 +73,32 @@ impl HostedServer {
...
+            && tokio::time::timeout(shutdown_budget, &mut shutdown_task)
+                .await
+                .is_ok()
+        {
+            finalization_completed = true;
+        }
 
-        if let Some(tx) = self.shutdown_tx.take() {
-            let _ = tx.send(());
+        if !finalization_completed {
```
gdorsi, line 90: When the shutdown_task times out, `shutdown()` returns early without aborting or awaiting the axum serve task. The serve task keeps accepting connections until `Drop` fires.

Consider cleaning up the serve task on this path too (abort + bounded await), same as the normal-exit branch below.

crates/jazz-tools/src/server/hosted.rs
```diff
@@ -66,36 +73,32 @@ impl HostedServer {
...
-            .flush()
-            .await
-            .expect("flush server runtime");
+        self.state.shutdown.request_shutdown();
+        let shutdown_budget = self.state.shutdown.timeout() * 2 + Duration::from_secs(5);
+
+        let mut finalization_completed = false;
+        if let Some(mut shutdown_task) = self.shutdown_task.take()
+            && tokio::time::timeout(shutdown_budget, &mut shutdown_task)
+                .await
```
gdorsi, line 84: The `is_ok()` check on the `JoinHandle` doesn't distinguish a panic from successful completion — `JoinHandle` yields `Result<(), JoinError>`, so a panic resolves as `Ok(Err(JoinError))` within the timeout and `finalization_completed` would be set `true` even if finalization panicked.

Consider matching on the inner `Result` to catch this.

crates/jazz-tools/src/server/hosted.rs
```diff
@@ -38,11 +38,18 @@ impl HostedServer {
             .expect("bind server listener");
         let port = listener.local_addr().expect("local addr").port();
 
-        let (shutdown_tx, shutdown_rx) = oneshot::channel();
+        let (serve_shutdown_tx, serve_shutdown_rx) = oneshot::channel();
+        let shutdown_state = built.state.clone();
+        let shutdown_task = tokio::spawn(async move {
+            shutdown_state.shutdown.wait_requested().await;
+            tokio::time::sleep(Duration::from_millis(50)).await;
```
gdorsi, line 45: This 50ms sleep has no comment explaining its purpose. If it's meant as a grace period for in-flight requests to observe the ShuttingDown phase, a fixed sleep doesn't guarantee that — and it adds latency to every shutdown.

Worth adding a comment explaining the intent, or replacing with a more deterministic mechanism.

gdorsi (approved):

crates/jazz-tools/src/commands/server.rs
```diff
@@ -60,7 +64,82 @@ pub async fn run(
 
     info!("Listening on http://{}", bound_addr);
     info!("Open the inspector: {}", inspector_link);
-    axum::serve(listener, built.app).await?;
+
+    let state = built.state.clone();
```
gdorsi, line 68: Right now `run()` hands control to `axum::serve` without installing a signal handler, so Ctrl+C / `kill <pid>` just drops the process immediately. That means `--shutdown-timeout-secs` is only reachable through the admin HTTP endpoint, which is a fairly narrow path — most operators stop servers via signals.

How would you feel about wiring a `tokio::signal` handler (SIGTERM + SIGINT) that feeds into the same shutdown controller? Something like using `axum::serve`'s `with_graceful_shutdown` to hook the signal, then running the drain/finalization sequence from there. That way the timeout flag works for the common case too.

PR #901, opened 2026-05-19
Use jemalloc for Rust and RocksDB allocators

antoniomuso:
## Summary

This switches the production allocator setup to jemalloc in both places that matter for the RocksDB-backed server path:

- Rust global allocator: `tikv-jemallocator`
- RocksDB allocator: `rust-rocksdb` with its `jemalloc` feature enabled

The goal is to avoid the split allocator setup where Rust used mimalloc while RocksDB stayed on the system allocator. The benchmark results point to the unified jemalloc setup as the best real production configuration.

## Benchmark Rationale

Benchmarked 36 cells across 4 allocator configurations, with 3 repeats and median results.

Configuration meanings:

- `unified`: jemalloc for both Rust and RocksDB
- `hybrid`: mimalloc for Rust, jemalloc for RocksDB
- `system`: no custom allocator
- `current`: the previous shipping setup

Throughput vs `current`:

| scenario | unified | hybrid | system |
| --- | ---: | ---: | ---: |
| write_heavy | +16.4% | -6.3% | +13.1% |
| read_heavy | +19.8% | +15.4% | +20.5% |
| balanced | +1.8% | +2.3% | +2.5% |

Server CPU/op vs `current`:

- `unified`: -18% / -20.5% / -0.6%

Server peak RSS vs `current`:

- `unified`: -3.8% / +7.8% / -10.2%, effectively neutral overall

Key interpretation:

- `unified` wins the important read/write-heavy cases, with roughly +16-20% throughput and about 20% lower server CPU/op.
- `hybrid` is worse than unified and loses write-heavy throughput, which suggests mixing allocator runtimes costs memory and does not buy enough back.
- `system` can beat the previous current setup, but the unified jemalloc setup is the recommendation because it performs well without the Rust/RocksDB allocator split.

## Changes

- Enables `jemalloc` on direct `rust-rocksdb` consumers.
- Uses `tikv-jemallocator` as the Rust global allocator for `jazz-tools`.
- Uses `tikv-jemallocator` in the Rust todo-server example.
- Updates `Cargo.lock` so Rust and RocksDB allocator paths converge through `tikv-jemalloc-sys`.

`jazz-napi` intentionally stays on `mimalloc-safe`; that crate documents a separate Node/NAPI allocator boundary.

## Validation

- `cargo tree -p jazz-tools --features cli -i tikv-jemalloc-sys`
- `cargo tree -p jazz-tools --features rocksdb -e features -i rust-rocksdb`
- `cargo tree -p opfs-btree --features rocksdb -e features -i rust-rocksdb`
- `cargo check -p jazz-tools --features cli`
- `cargo check -p todo-server`
- `cargo fmt --check`
- `git diff --check`
- pre-commit hook: `clippy` and `rustfmt`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1804,7 +1804,6 @@ dependencies = [
  "jsonschema",
  "jsonwebtoken",
  "lz4_flex",
- "mimalloc",
  "opentelemetry",
  "opentelemetry-otlp",
  "opentelemetry-stdout",
@@ -1827,6 +1826,7 @@ dependencies = [
  "smolset",
  "tempfile",
  "thiserror 1.0.69",
+ "tikv-jemallocator",
  "tokio",
  "tokio-tungstenite",
  "tower",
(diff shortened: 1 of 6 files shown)


gdorsi (approved):
Nice!

I guess we gotta generate new prebuilt artifacts to make this change effective, right?

PR #908, opened 2026-05-19
Mirror requested CORS headers for Jazz server preflights

grp06:
## Summary

- Mirror `Access-Control-Request-Headers` in the shared Jazz server CORS layer.
- Add a regression test proving auth-related preflight headers receive a concrete `Access-Control-Allow-Headers` value instead of `*`.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo test -p jazz-tools --features test server::routes::tests::cors_preflight_mirrors_requested_auth_headers`
- `cargo test -p jazz-tools --features test server::routes::tests`
- `cargo test -p jazz-tools --features test`

## Not run

- `pnpm format:check`: local checkout is missing `node_modules`, so `oxfmt` is unavailable. Rust formatting was checked separately with `cargo fmt --check`.

### DIFF
--- a/.changeset/mirror-cors-preflight-headers.md
+++ b/.changeset/mirror-cors-preflight-headers.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Mirror requested CORS preflight headers from the shared Jazz server.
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Superseded by #1955, which ports the fix and regression coverage to the current jazz-server Axum router.

PR #909, opened 2026-05-20
[codex] Spec global epoch snapshot isolation

aeplay:
## What changed

Adds a motivated design spec for global-epoch snapshot isolation in Jazz, including storage shape, transaction dependency validation, predicate/index concerns, sync protocol changes, permission semantics, examples, diagrams, and self-review concerns.

## Why

Jazz needs true snapshot isolation for transactions and branches without version vectors over potentially millions of local-first clients. The spec proposes global authority-assigned epochs for committed history plus a device-local optimistic overlay.

## Validation

Docs-only change. Pre-commit formatting ran during commit; no test suite run.

### DIFF
--- a/specs/a_mvp/global_epoch_snapshot_isolation.md
+++ b/specs/a_mvp/global_epoch_snapshot_isolation.md
@@ -1,0 +1,696 @@
+# Global Epoch Snapshot Isolation
+
+## Why This Exists
+
+Jazz needs true snapshot isolation for two features:
+
+- transactions
+- branches
+
+The current row-history system already has durable row versions, branch frontiers, visible-row
+caches, and replayable batch fate. That gives us history, but not a clean global read cut. Reads
+mostly mean "current visible state at a durability tier", and durability tier is not the same thing
+as snapshot membership.
+
+The tempting distributed-systems answer is a version vector. That does not fit Jazz. A Jazz app can
+have millions of local-first clients, and a snapshot token must not grow with the number of possible
+writers.
+
+The proposed answer is deliberately smaller:
+
+> Snapshot isolation is defined over globally committed history using one monotonically increasing
+> `global_epoch` per branch authority. Device-local writes form a private overlay and never become
+> globally meaningful until the global authority accepts them.
+
+This is not "timestamp per row" in the naive distributed sense. A row's `global_epoch` is not minted
+by arbitrary clients and is not compared across independent local clocks. It is assigned only by the
+global authority that serializes accepted writes for a branch.
+
+## Core Claim
+
+We can get away with:
+
+```text
+globally committed history:
+  ordered by global_epoch
+
+device-local optimistic history:
+  ordered by local_epoch, visible only on that device/runtime path
+```
+
+No edge epoch, worker epoch, or client vector is needed for MVCC truth.
+
+Workers and edge servers can still cache, persist, relay, preflight permissions, and answer queries
+from their latest known global state. They do not assign canonical snapshot order.
+
+## What Needs Snapshot Isolation
+
+### Transactions
+
+A transaction starts from a known global snapshot:
+
+```text
+base_global_epoch = latest global epoch known to this runtime
+```
+
+It may run while offline. That is allowed. Offline transactions are highly optimistic: if the base
+epoch is stale, the transaction may fail at the global authority. On cold data, it can still succeed.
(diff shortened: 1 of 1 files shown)


specs/a_mvp/global_epoch_snapshot_isolation.md
```diff
@@ -0,0 +1,696 @@
...
+### Transactions
+
+A transaction starts from a known global snapshot:
+
+```text
+base_global_epoch = latest global epoch known to this runtime
+```
+
+It may run while offline. That is allowed. Offline transactions are highly optimistic: if the base
+epoch is stale, the transaction may fail at the global authority. On cold data, it can still succeed.
```
nrainhart, line 57: Does this mean transactions won't see optimistic local state? Or that they assume that state will be accepted by the authority before it accepts the tx (and will otherwise fail)? I think with transactions it's important to do the latter (even more so than with branches)

specs/a_mvp/global_epoch_snapshot_isolation.md
```diff
@@ -0,0 +1,696 @@
...
+
+A branch is created from a global snapshot:
+
+```text
+branch_base = { source_branch, base_global_epoch }
+```
+
+The branch may intentionally be stale. That is a feature, not a bug. Branching from "the last global
+state this device knew" is a clear semantic boundary and avoids embedding local speculative writes
+into durable branch identity.
```
nrainhart, line 69: Doesn't this mean that if an app never connects to a server, branches will always start up from an "empty" db?

specs/a_mvp/global_epoch_snapshot_isolation.md
```diff
@@ -0,0 +1,696 @@
...
+
+### Batches
+
+Batches do not need snapshot isolation. They group row-version production and replayable submission.
+They may be direct or transactional, but the batch itself is not the read cut.
```
nrainhart, line 74: Agree

specs/a_mvp/global_epoch_snapshot_isolation.md
```diff
@@ -0,0 +1,696 @@
...
+
+History rows keep all durable row versions.
+
+```text
+history row key:
+  (row_id, branch, batch_id)
+
+history row payload:
+  parents
+  status                  // local_pending | global_committed | rejected
```
nrainhart, line 152: Nit: this could be derived from a global_epoch/local_epoch/rejected_epoch timestamp

specs/a_mvp/global_epoch_snapshot_isolation.md
```diff
@@ -0,0 +1,696 @@
...
+  global_epoch            // nullable
+  local_epoch             // nullable, device-local only
+  local_base_global_epoch  // nullable, device-local only
+  tx_id                   // nullable
```
nrainhart, line 156: Is this `tx_id` different form the batchId?

specs/a_mvp/global_epoch_snapshot_isolation.md
nrainhart: Oh, I see you propose having a separate "private local branch" feature in the future. I'd like to discuss this a bit more, if implementing a [dotted-version-vector](https://gsd.di.uminho.pt/members/vff/dotted-version-vectors-2012.pdf) like overlay is possible, it'd be great to not have two types of branches

specs/a_mvp/global_epoch_snapshot_isolation.md
```diff
@@ -0,0 +1,696 @@
...
+  lower_bound,
+  upper_bound,
+  predicate_fingerprint,
+  observed_max_global_epoch,
+}
+```
+
+Row dependencies cover point reads and conditional writes.
+
+Index range dependencies cover predicates and absence checks. They prevent phantoms.
```
nrainhart, line 289: This is not needed to achieve a snapshot isolation level, right? It would give us serializable isolation (or something close to it)

nrainhart (commented):
Leaving some thoughts as comments. It probably makes more sense to discuss this live tomorrow.

aeplay:
Implemented by the new relational core (#1094): snapshot/transaction isolation semantics are specified in `crates/jazz/SPEC/` rather than the retired `specs/a_mvp/` tree this draft targeted. Closing.

PR #913, opened 2026-05-20
Fix local write policy relation evaluation

aeplay:
## Summary
- evaluate local row write policies through the contextual policy evaluator so complex boolean expressions preserve semantics
- bind the outer row id when extracting `ExistsRel` complex clauses for local writes
- allow structural relation scans for local write `ExistsRel` policy graphs and cover the regression with a focused test
- make local writes fail closed until the current-permissions authorization schema is loaded, then evaluate insert policies from that authorization schema even when the write schema is structural

## Why
The local write path could flatten complex `OR` relation policy clauses into separate graph checks and require all of them to pass. That denied valid writes for policies where access can be granted through one of multiple relation paths.

For current-permissions clients, the write schema may be structural while policies live in the authorization schema. Local writes now have an explicit contract: if permissions are required but the authorization schema is not available yet, the write is denied locally; once the authorization schema is loaded, local policy evaluation uses those policies instead of treating the structural schema as permissive.

## Validation
- `cargo check -p jazz-tools --lib`
- `cargo test -p jazz-tools local_update_with_exists_rel_policy_can_bind_outer_row_id --lib`
- `cargo test -p jazz-tools --lib query_manager::rebac_tests::insert_policies::local_insert -- --nocapture`
- `cargo test -p jazz-tools --lib`


gdorsi (commented):
.

crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -1384,6 +1384,59 @@ impl QueryManager {
         if depth > crate::query_manager::policy::RECURSIVE_POLICY_MAX_DEPTH_HARD_CAP {
             return false;
         }
+        if let Some(current_row_id) = row_id {
```
gdorsi: If I understand things correctly here, `_inner` is only ever called from `_for_row` which always passes `Some(row_id)`, which means that the new guard at line 1387 (`if let Some(current_row_id) = row_id`) always fires.

If this is true it means that the entire old evaluation path from line 1389 onward is now dead code.

Is this intentional?

aeplay: thanks for calling this out, this led me to discover a bunch of weird edge cases

crates/jazz-tools/src/query_manager/policy_graph.rs
```diff
@@ -262,8 +258,10 @@ impl PolicyGraph {
     ) -> Option<Self> {
         let use_structural_rows = structural_scans
             || current_table
-                .and_then(|table| {
-                    Self::exists_rel_output_table(rel, branch).map(|output| output == *table)
+                .map(|table| {
+                    let output_is_current =
+                        Self::exists_rel_output_table(rel, branch) == Some(*table);
+                    output_is_current || Self::relation_references_table(rel, table)
```
gdorsi, line 264: If I understand this path correctly, any EXISTS_REL relation that merely mentions the current table is compiled structurally, and structural mode clears policies for every table in the relation. That looks like a new bug: a write policy could join the current table into an otherwise protected relation and stop enforcing SELECT policies on the other scanned tables. Should this only make the current-table scan structural instead of the whole compiled schema?

crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -1472,237 +1378,30 @@ impl QueryManager {
...
+            provenance.clone(),
+        );
+        let mut evaluator =
+            crate::query_manager::graph_nodes::policy_eval::PolicyContextEvaluator::new(
+                self.schema.as_ref(),
+                session,
+                branch,
+                self.row_policy_mode,
+            );
+        evaluator.evaluate_row_access(
```
gdorsi, line 1394: I may be missing a guard elsewhere, but this looks like a possible behavior regression in the write-authorization path.

**What seems to have changed**
Local write checks now delegate the whole policy to `PolicyContextEvaluator`. From my read, that evaluator keeps one visited set for inherited parent rows while it evaluates sibling policy clauses.

**Why this could be a problem for us**
For a composite policy that requires two inherited checks through the same parent row, the first inherited check can succeed and mark that parent as visited. The second inherited check can then see the same parent as already visited and treat it like a cycle, causing the write to be denied.

A concrete shape would be a rule like: the user may update a row only if they can both read and update its related project. Those are two separate policy questions, but they can follow the same foreign key parent.

If that reading is right, this can deny legitimate inserts or updates locally, and can also affect server-side permission checks, for app schemas that combine multiple inherited requirements in one policy.

gdorsi (commented):
Review comments.

gdorsi (approved):
LGTM

aeplay:

PR #914, opened 2026-05-20
feat(docs): WriteTierDiagram — pick a tier, watch the write propagate

joeinnes:
An interactive diagram for the durability-tiers section of `writing-data.mdx` that makes the wait-tier story concrete: pick a tier, hit "New colour", and watch the write light up the phone, race up the trunk to the chosen tier, fan out from edge to the global core and to another device, and resolve at the tier you picked. Aeroplane mode queues every non-local write on the device; flipping it off releases the queue, staggered so each propagation wave reads as its own event rather than a chord.

Stacks on `joe/mobile-diagrams` — review only the last four commits here.

## What's in the diff

- `feat(docs): expose roundedPath/Point and a disabled NodeAction` — engine surface gains a public `roundedPath` (so consumer diagrams can build bespoke elbow connectors without re-deriving the corner-arc maths) and a `disabled` prop on `NodeAction` (dimmed + not-allowed via existing CSS, for in-flight affordances).
- `refactor(docs): lift the colour-write policy out of tier-sync` — moves `pickNextColor` and the hue-gap helpers from `tier-sync-diagram.tsx` to a shared `components/mdx/colour.ts`. Pure relocation; both propagation diagrams want the same "consecutive writes are visibly distinct" guarantee.
- `refactor(docs): extract reusable PhoneChrome from lens-diagram` — pulls the lens diagram's phone bezel (status bar, themed screen inside a near-black shell) into `diagram/phone-chrome.tsx`. Obeys engine portability rules (no Tailwind, no `fd-*`, no `lucide-react`); the screen follows `--diagram-*` tokens, the bezel/notch stay near-black either way. Optional `offline` prop swaps wifi/signal for an iOS-style aeroplane glyph. Lens output is byte-equivalent.
- `feat(docs): WriteTierDiagram — pick a tier, watch the write propagate` — the new component plus a one-line caption in `writing-data.mdx` tying it to the offline behaviour described above.

## Domain policy lives in the definition

The engine just measures, lays out, and animates. The diagram itself owns:

- last-write-wins per node
- the "local-only" counter that only clears once a write reaches the edge
- the writing → settled lifecycle on the in-flight pill
- queue release ordering when aeroplane mode flips off

The trunk and branch connectors are drawn from measured anchors via the newly-exported `roundedPath`. The in-flight "Write" affordance uses `NodeAction`'s new `disabled` state.

## Testing

Pure functions (geometry, BFS waves, colour gap) covered by the existing `vitest` suite. Interactive behaviour and look — desktop and mobile widths, light and dark, aeroplane on/off, every tier — checked by eye.

### DIFF
--- a/docs/components/mdx/colour.ts
+++ b/docs/components/mdx/colour.ts
@@ -1,0 +1,71 @@
+// Colour-propagation payload policy, shared by the animated sync diagrams
+// (tier-sync & write-tier). The diagram engine owns the propagation
+// *mechanism*; choosing the next colour to write is diagram *policy*, so it
+// lives here alongside the definitions rather than inside the engine.
+
+const MIN_HUE_GAP = 40;
+export const INITIAL_COLOR = "#146aff";
+
+function hexToHue(hex: string): number | null {
+  const m = /^#([0-9a-f]{6})$/i.exec(hex);
+  if (!m) return null;
+  const n = parseInt(m[1], 16);
+  const r = ((n >> 16) & 0xff) / 255;
+  const g = ((n >> 8) & 0xff) / 255;
+  const b = (n & 0xff) / 255;
+  const max = Math.max(r, g, b);
+  const min = Math.min(r, g, b);
+  const d = max - min;
+  if (d === 0) return 0;
+  let h: number;
+  if (max === r) h = ((g - b) / d) % 6;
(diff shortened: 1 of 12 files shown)


nrainhart (approved):
Beautiful!! ✨

PR #919, opened 2026-05-21
feat: transaction visibility

nrainhart:
## Description

This PR extends transactions to allow configuring when they become visible to other transactions. The goal is to provide an abstraction that unifies transactions and (direct) batches, so that users only think in terms of transactions.

Transactions now support two `visiblity` modes:
- `immediate`: transaction becomes visible on commit (replaces existing batches)
- `deferred`: transaction becomes visible on authority approval
	- for DBs connected to a sync server, the authority is the _core_ server (⚠️ this is a change with respect to existing transactions, that just wait for local/edge approval depending on the setup)
	- for local DBs, the authority is the _local_ server

By default, a transaction becomes visible when the relevant authority accepts it:

```ts
db.transaction((tx) => {});
// This is equivalent to
db.transaction((tx) => {}, { visibility: "deferred" });
```

Manual transactions use the same option:

```ts
const tx = db.beginTransaction({ visibility: "immediate" });
tx.insert(todos, { title: "Immediate" });
tx.commit();
```

Existing direct writes/direct batches are now transactions with an `immediate` visibility. We could even remove the `beginBatch`/`batch` API.

Internally, we allow defining when a transaction becomes visible based on any durability tier. A transaction's writes are visible by reads outside the transaction only when: `visibility == immediate or confirmed_tier >= visibility.durabilityTier`.


specs/transaction_visibility.md
```diff
@@ -0,0 +1,312 @@
...
+db.transaction(async (tx) => {
+  tx.insert(todos, { title: "Ship it" });
+});
+```
+
+By default, a transaction becomes visible when the highest relevant authority accepts it:
+
+```ts
+// With server authority: default visibleAt is "global"
+db.transaction(async (tx) => {}, { visibleAt: "global" });
```
joeinnes, line 17: `deferTo` as an alternative for `visibleAt`? To me `visibleAt` implies *only* visible at, when the truth is that once it's accepted it's visible everywhere. Alternatively, what about just `tier`, and supplement with clear docs?

specs/transaction_visibility.md
```diff
@@ -0,0 +1,312 @@
...
+
+Manual transactions use the same option:
+
+```ts
+const tx = db.beginTransaction({ visibleAt: "global" });
+tx.insert(todos, { title: "Carefully" });
+await tx.commit();
+```
+
+Existing direct writes/direct batches become implementation details: internally they are transactions visible at `immediate` acceptance. Public `batch` wording should be deprecated or aliased, but not kept as a separate semantic concept.
```
joeinnes, line 45: Proposed user-facing taxonomy:

* `transaction` - a group of changes committed together
* `optimistic transaction` - a group of changes committed together with `immediate` or `local` visibility
* `durable transaction` - only if needed for distinguishing - a group of changes with a visibility higher than `local`. Technically there's an argument that `local` is also a durability level, but I think this slight imprecision in wording saves a lot of complicated linguistic gymnastics, and allows us to reuse the 'durable' wording we have elsewhere.

specs/transaction_visibility.md
nrainhart: Yeah, I'm not convinced about `visibleAt` either. Not a fan of `deferTo` either (doesn't seem immediately obvious to me). `tier` is vague, but it's definitely my plan B. Will keep thinking of alternatives

nrainhart: After discussing it offline with Joe, we decided it doesn't make sense to expose this level of granularity into the user-facing API. We now expose just a `visibility` of `'immediate' | 'deferred'`, and obtain the actual transaction visibility based on the DB's configuration:
```ts
db.transaction(async (tx) => {}, { visiblity: "immediate" })
```

packages/jazz-tools/src/runtime/db.transaction.test.ts
nrainhart, line 1: Removed a ton of mostly-mocked tests. `db.transaction-reads.test.ts` actually validates how transactions works

crates/jazz-tools/src/storage/mod.rs
```diff
@@ -1624,6 +1625,49 @@ fn exact_visible_row_table_locator_for_delete<H: Storage + ?Sized>(
 }
 
 fn sealed_batch_submission_storage_descriptor_with_branch_ords() -> RowDescriptor {
+    RowDescriptor::new(vec![
```
nrainhart, line 1628: The changes to storage are backwards-compatible. Sync protocol changes are not, so updating to the latest version of Jazz will still be necessary to connect to our servers.

nrainhart:
Superseded by https://github.com/garden-co/jazz/pull/1052

PR #922, opened 2026-05-22
Reuse materialized query scope rows during sync delivery

aeplay:
## Stack
This PR stacks on #923 and should be reviewed after the set-level resource-policy evaluator.

## Summary
- threads materialized query scope rows through sync delivery so rows already produced by the query graph do not need to be point-read again during scope installation
- preserves the existing storage-backed fallback for scope entries that were not materialized by the graph
- limits materialized scope replay to rows from the subscription's current schema hash so schema-mismatched rows keep using the existing storage path
- keeps the policy evaluation behavior unchanged; earlier row-load caching and set-level inherited-authorization experiments were removed from this PR after they exposed write-policy correctness risks

## Why
The subscription pipeline already materializes result rows while computing query scope. Before this change, scope installation collapsed those rows back to object IDs and then asked sync delivery to fetch the same rows from storage one by one before replaying them to the client.

This PR lets the query graph return the rows in scope alongside the scope IDs. Sync delivery can replay those already-loaded rows directly, while still installing the same scope and still falling back to storage when a row is unavailable or schema context does not match.

## Measurements
Representative local measurements from the permissioned load harness while this optimization was isolated:
- row-aware scope delivery reduced the largest scope installation from roughly 1.5-2.0s to roughly 300ms
- end-to-end permissioned load remained in the same overall ballpark as the prior optimized stack, with this PR addressing the scope-installation point-read cost specifically

## Validation
- cargo fmt --check
- cargo test -p jazz-tools --features test --test policies_integration -- --nocapture
- pnpm --filter jazz-tools test:browser
- pre-commit clippy/rustfmt hook passed while committing

## Notes
This intentionally does not include the more ambitious set-level policy evaluator experiment. That path looked promising for redundant inherited-policy work, but in this branch it changed write-policy behavior under rejected optimistic mutations, so it should be redesigned separately rather than bundled into this sync delivery optimization.

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph/execute.rs
+++ b/crates/jazz-tools/src/query_manager/graph/execute.rs
@@ -3,7 +3,7 @@ use smallvec::SmallVec;
 use std::collections::HashMap;
 
 use crate::object::ObjectId;
-use crate::query_manager::settlement_eval_cache::SettlementEvalCache;
+use crate::query_manager::settlement_eval_cache::{RowLoadKey, SettlementEvalCache};
 use crate::query_manager::types::{LoadedRow, RowDelta, TableName, Tuple, TupleDelta};
 use crate::storage::Storage;
 use crate::sync_manager::RowBatchKey;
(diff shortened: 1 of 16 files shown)


aeplay:

PR #923, opened 2026-05-22
Evaluate supported resource policies at set level

aeplay:
## Stack
This PR stacks on #913 and should be reviewed after the local write policy evaluation fixes.

## Summary
- evaluates supported resource-policy relation shapes at set level instead of repeating the same correlated policy graph for every candidate row
- keeps the optimization behind shape checks and falls back to the existing per-row evaluation path when a policy cannot be decorrelated safely
- reduces policy graph compile overhead by reusing schema context handles instead of cloning the full context through hot compile paths
- makes browser wasm sync entrypoints resilient to reentrant delivery by retrying parked sync messages/ticks instead of panicking on `RefCell` reborrows
- uses wasm-safe timing in shared query/sync code and allows structural schema catalogue bootstrap in non-production/testing runtimes

## Why
Large permissioned subscriptions can produce many child rows that inherit access through a much smaller parent resource set. The old path evaluated the parent policy graph separately for each child or parent candidate, which made repeated authorization work dominate load time even when storage access itself was reasonably fast.

This PR teaches the policy graph to recognize the reusable relation-set shape and evaluate it once for a candidate set. It preserves the scalar/per-row fallback for unsupported policy expressions.

The latest fixes also cover the browser persistent-worker path exercised by `test-ts`: direct wasm sync delivery can re-enter while the runtime is already borrowed, and dev/test schema bootstrap needs to accept structural schema catalogue payloads from peer-local runtimes without opening broader catalogue writes.

## Validation
- cargo fmt --check
- cargo test -p jazz-tools structural_catalogue -- --nocapture
- pnpm --filter jazz-wasm build:profiling
- pnpm --filter jazz-napi build
- pnpm --filter jazz-tools test:browser
- pnpm exec vitest run --config vitest.config.browser.ts tests/browser/todo-app.test.tsx (from examples/todo-client-localfirst-react)
- pnpm test --filter=!moon-lander-react --filter=!@jazz/rust --filter=!jazz-rn --concurrency=2 (local Node v20 cannot run the full suite: repo requires Node >=22.12; failures were from missing `node:sqlite` and wasm init under Node v20)

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph/compile.rs
+++ b/crates/jazz-tools/src/query_manager/graph/compile.rs
@@ -1006,7 +1006,6 @@ impl QueryGraph {
         })?;
 
         validate_execution_plan(&plan, schema)?;
-
         Self::compile_execution_plan_with_schema_context(
             &plan,
             schema,
@@ -2145,7 +2144,7 @@ fn index_scan_plan(
     for condition in disjunct
         .conditions
         .iter()
-        .filter(|c| c.is_index_scannable() && table_schema.is_indexed_column(c.column()))
+        .filter(|c| c.is_index_scannable() && scan_index_exists(table_schema, c.column()))
     {
         if !column_plans
             .iter()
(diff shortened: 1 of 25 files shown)


aeplay:

PR #926, opened 2026-05-22
Interactive group-permissions demo

joeinnes:
## Summary
- Adds a live demo to the group-permissions recipe (under the role table): an onboarding gate where the reader picks reader/contributor/writer/admin, which flips into a Drive-style file list where the available actions react to the chosen role — swappable any time from a role chip in the header.
- Routes the engine primitives `DiagramFrame`, `DiagramStyles` and `PhoneChrome` through the `./diagram` barrel and migrates the lens & write-tier diagrams off their deep imports (no behavioural change).

## Test plan
- [ ] Run the docs dev server, open `/docs/recipes/access-control/group-permissions`, pick each role; confirm Members only appears for admin, New greys out for reader, and the per-row edit pencil follows the edit-own (contributor) / edit-any (writer, admin) rules.
- [ ] `pnpm --filter docs types:check` passes.

### DIFF
--- a/docs/components/mdx/diagram/index.ts
+++ b/docs/components/mdx/diagram/index.ts
@@ -21,6 +21,10 @@ export type { SequenceProps, Participant, SequenceStep } from "./sequence";
 
 export { NodeShell, NodeIcon, NodeTitle, NodeSubtitle, NodeFooter, NodeAction } from "./kit";
 
+export { DiagramFrame } from "./frame";
+export { DiagramStyles } from "./styles";
+export { PhoneChrome } from "./phone-chrome";
+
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Closing this as stale rather than reconstructing it. Its four role affordances still match the recipe policy, but the Drive-style mock is separate from the executable docs snippets and duplicates the role table while omitting other material policy behavior such as delete, leave, and bootstrap. The current docs direction is to keep examples runnable and source-backed. If we add a real group-workspace example, we can revisit an interaction derived from that shared policy; until then this mock would be a second drift-prone implementation.

PR #927, opened 2026-05-22
Add invite-links access-control recipe

joeinnes:
## Summary
- Add an "Invite links" recipe to the access-control docs: a resource owner shares a URL carrying a resource ID plus a secret, and an invitee redeems it to get a membership row.
- Primary pattern uses a small server route to validate the code and insert the membership (with backend privileges); secondary pattern does the check inside a permission rule via a session claim, for local-first auth with no server.
- Redeem route is idempotent (no duplicate membership on re-redeem); post-redeem navigation targets corrected; note added that the short join code should be widened when it isn't single-use.
- Worked, type-checked snippets in `todo-server-ts` and `todo-client-localfirst-react`.

## Test plan
- [ ] `pnpm build` and confirm the recipe page renders with every `<include>` snippet resolving
- [ ] Open the recipe and sanity-check the primary vs secondary pattern guidance reads correctly

### DIFF
--- a/docs/content/docs/recipes/access-control/invite-links.mdx
+++ b/docs/content/docs/recipes/access-control/invite-links.mdx
@@ -1,0 +1,113 @@
+---
+title: "Invite links"
+description: Let users share a URL that grants another user access to a private resource without knowing their identity in advance.
+reviewedAt: "9712a744f"
+---
+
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Superseded by #1953, which replaces the subscription-only claim pattern with atomic, server-backed invite redemption.

PR #928, opened 2026-05-22
feat(docs): ConcurrentWritesDiagram — watch concurrent writes converge

joeinnes:
## Summary

Adds an auto-playing diagram to the **Local-First Data Model** docs page (`concepts/local-first-data-model.mdx`), replacing the static `<Graph converge>` version-history example.

It tells the deterministic-merge story concretely: starting from a baseline row, Alice and Bob edit it concurrently — Alice renames it, Bob marks it complete. They touch *different fields*, so both writes survive, and every peer converges on the same `m3`. The Callout below now grounds the "beware of unexpected results" warning in this exact example (the merge is mechanically correct but semantically wrong — Bob may have bought cow's milk).

- Each node is a card showing the full row state (`title`, `done`) so the diff between branches reads at a glance; a2/b2 carry Alice/Bob author tints.
- Builds itself once on mount: v1 → a2/b2 → m3, with a trace + node pulse as each write lands.
- A blue pulse re-flows root → leaves every 5s to keep it alive, with no controls.
- Built on the existing diagram engine (`Graph` + `useDiagramTraces` + `playWaves` + node kit). Adds two engine colour tokens (`--diagram-author-a` / `--diagram-author-b`) with sensible defaults.

No changeset: `docs` is a private package.

## Test plan

- [ ] `pnpm --filter docs types:check` passes
- [ ] `pnpm --filter docs test` passes
- [ ] Visit `/docs/concepts/local-first-data-model`: diagram builds on load, columns/arrows are symmetric, node pulses are rounded, both branches (incl. b2→m3) pulse on the recurring flow, hover/tap shows nothing extraneous
- [ ] Check light + dark mode and a narrow (mobile) width

### DIFF
--- a/docs/components/mdx/concurrent-writes-diagram.tsx
+++ b/docs/components/mdx/concurrent-writes-diagram.tsx
@@ -1,0 +1,346 @@
+"use client";
+
+import { type ReactNode, useCallback, useEffect, useRef, useState } from "react";
+
+import { cn } from "@/lib/cn";
+import {
+  Graph,
+  type GraphEdge,
+  type GraphNode,
+  type GraphOverlayCtx,
+  type Hop,
+  hopEdgeId,
+  NodeShell,
+  NodeSubtitle,
+  NodeTitle,
+  playWaves,
+  useDiagramTraces,
+} from "./diagram";
+
+type NodeId = "v1" | "a2" | "b2" | "m3";
+
+type VersionVisual = {
+  id: NodeId;
+  rank: number;
+  order?: number;
+  author?: "a" | "b";
+  title: string;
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Closing this as superseded. The concrete different-field merge example and semantic-conflict warning are already present in the current local-first data-model guide. The remaining change is a large auto-playing wrapper around a fixed version-history DAG; our current diagram contract intentionally keeps fixed data-only DAGs inline as Graph data, reserving bespoke components for genuine interaction or domain policy. Keeping the current inline graph avoids another timer and trace implementation with no corresponding runnable example.

PR #929, opened 2026-05-22
feat(otel): export logs via OTLP and drop JAZZ_OTEL gate

Elfo404:
## Summary

- Wire `opentelemetry-appender-tracing` so every `tracing::` event becomes an OTel `LogRecord` with the active trace context attached. Logs + traces now correlate on `(trace_id, span_id)` in the backend.
- Single knob: `OTEL_EXPORTER_OTLP_ENDPOINT`. When set, the CLI initializes both the tracer and logger providers and ships traces + logs over OTLP/HTTP. When unset, neither is initialized.
- Stdout continues to emit human-readable text logs unchanged — durable, simple, no format flag.
- Drop the redundant `JAZZ_OTEL=1` opt-in and the stdout span-exporter fallback inside `init_tracer_provider`.
- Wrap the synchronous startup logs in a `server_startup` span so they carry trace_id/span_id when exported.

## Why

Before: traces and stdout text were two unrelated streams; correlating a log to its span required eyeball + timestamp guessing. Adding the OTel logs appender — the official `tracing` → OTel-logs bridge in the Rust SDK — gives us proper correlation without rolling a custom JSON formatter, and keeps stdout as the human-readable channel.

The `JAZZ_OTEL` gate predated the SDK honoring `OTEL_EXPORTER_OTLP_ENDPOINT`; with the stdout-fallback exporter removed, the endpoint env var is the natural on/off signal.

## Verification

Ran the CLI against a local OTLP collector with `OTEL_EXPORTER_OTLP_ENDPOINT` set:

- `otel_traces` shows the `server_startup` span.
- `otel_logs` shows the two in-span startup messages carrying the same `trace_id` + `span_id`.
- Logs emitted outside the span correctly have empty trace context.
- Builds clean with `--features otel` and with `--features cli` (no otel).

## Test plan

- [ ] CI passes
- [ ] `cargo build -p jazz-tools --features otel` succeeds
- [ ] Run `jazz-tools server` with `OTEL_EXPORTER_OTLP_ENDPOINT` unset → only stdout text logs, no OTLP traffic
- [ ] Run with the endpoint set → traces + logs land in collector, in-span logs carry trace context

### DIFF
--- a/.changeset/otel-log-export.md
+++ b/.changeset/otel-log-export.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Export `tracing` events as OpenTelemetry logs over OTLP, correlated with traces via the active span's `trace_id`/`span_id`. The CLI now keys both trace and log export off `OTEL_EXPORTER_OTLP_ENDPOINT`; the `JAZZ_OTEL=1` opt-in and the stdout span-exporter fallback are removed. Stdout continues to emit human-readable text logs.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1806,6 +1806,7 @@ dependencies = [
  "lz4_flex",
  "mimalloc",
  "opentelemetry",
+ "opentelemetry-appender-tracing",
  "opentelemetry-otlp",
  "opentelemetry-stdout",
  "opentelemetry_sdk",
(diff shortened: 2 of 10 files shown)


Takeno (approved):

gdorsi (approved):
Nice job!

PR #931, opened 2026-05-22
Fix policy relation review followups

aeplay:
Follow-up to Guido's review comments on #913.

## What changed
- Keep same-table EXISTS_REL scans structural only for the current table, while preserving SELECT policy enforcement on other joined tables.
- Scope inherited-policy cycle tracking to the active traversal path so sibling inherited checks through the same parent row can both evaluate.
- Add ReBAC regression coverage for both cases.

## Tests
- cargo test -p jazz-tools query_manager::rebac_tests
- pre-commit hook: clippy + rustfmt


aeplay:

PR #933, opened 2026-05-23
Optimize output auth, OPFS locality, and worker startup

aeplay:
## Summary

This stack targets the Boredm cold-start bottleneck from the storage, authorization, query-settlement, and worker-startup sides:

- groups ref-shaped policy authorization so parent-set output auth can fast-path correlated child loads
- avoids main-thread all-ready stalls from repeated initial settle behavior
- improves OPFS B-tree locality/cache behavior for the app-shaped cold-start workload
- caches structural authorization schemas and row-format layout offsets on hot paths
- fast-paths flat row-history tail projection during replay/decode
- sends worker startup progress logs for schema/OPFS/rehydrate/bootstrap phases and OPFS open/retry behavior
- uses a catalogue-only bootstrap for worker startup, then skips edge-reconciliation replay in the local-only worker sync path
- reuses materialized query scope rows for initial/reactive client query scope delivery instead of re-reading scope rows from storage
- replaces/drops query scopes incrementally so updates avoid rebuilding the full union of all client scopes twice
- instruments local batch row resolution and amortizes legacy broad-scan fallback results across batch ids until storage changes
- adds browser harness diagnostics for W4 reopen breakdown and lets scenario-selected runs skip heavyweight diagnostic-only tests

## Latest Trace / Log Read

The latest app logs show `DataEntry children` holder processing is cheap once the rows arrive: ~19,894 child changes are nested in about 40-45ms. The long pole is before that callback, in worker query/sync replay and scope delivery.

The new `local_batch_rows` instrumentation showed the remaining hot path clearly: every logged lookup was falling through to `broad_row_locator_scan`. In one run there were 126 broad fallback calls over only 17 batch ids, totaling ~9.6s of instrumented resolution time and ~868k row-locator/history scans. The worst repeats scanned ~20k rows to answer tiny batches of 4/68/126 rows.

The newest commit changes that fallback from per-request rediscovery into a per-storage-generation cache: if we have to scan all row histories, we index rows for every batch id discovered and reuse those results until storage changes. This should specifically reduce repeated worker-side scans in the app trace while preserving correctness during sync replay by clearing the cache on storage writes.

## Local Browser Benchmarks

After rebuilding `crates/jazz-wasm`:

- W4 cold reopen: 25 cycles, wall ~1058ms, avg ~42.3ms/cycle
  - `create_db` avg ~9.5ms
  - local query avg ~32.0ms
  - shutdown avg ~0.7ms
- B3 cold reopen/sync shape: 12 cycles, wall ~587ms, avg ~48.8ms/cycle
- B1 mutation mix: 12,288 ops, wall ~767ms, ~16k ops/sec
- B6 permission-heavy read mix: 6,000 ops, wall ~4691ms, ~1.28k ops/sec

## Verification

- `cargo fmt --check && git diff --check`
- `cargo test -p jazz-tools sync_manager::tests::query_scope --lib`
- `cargo test -p jazz-tools server_subscriptions --lib`
- `cargo test -p jazz-tools runtime_core::tests::write_batch --lib`
- `cargo test -p jazz-wasm --lib`
- `cargo test -p opfs-btree --lib`
- `cargo test -p jazz-tools opfs_btree --lib`
- `pnpm --dir crates/jazz-wasm build`
- browser realistic benchmark spot checks for W4, B3, B1, and B6

## Known Test Gap

`cargo test -p jazz-tools --lib` still has three failing sync-backed policy tests:

- `sync_backed_joined_exists_rel_session_subscription_keeps_local_rows_when_server_scope_is_empty`
- `sync_backed_exists_session_subscription_keeps_local_rows_when_server_scope_is_empty`
- `sync_backed_exists_rel_session_subscription_keeps_local_rows_when_server_scope_is_empty`

I verified the same failure exists at the remote stack head before the newest worker/query-scope commits (`bb13c6a16`), so this appears inherited rather than introduced here.

## App Smoke Note

The wasm package has been rebuilt for the app to consume. The PR remains labeled `preview-build` so the preview package should build from the pushed branch.


aeplay:

PR #934, opened 2026-05-23
Add SolidJS Support

FelipeEmos:
Hi!
This PR is now ready with Solid
-  API adapters
- updates to docs
- updates to examples

## Design overview

I used the React and Vue adapters as reference but this is far from a 1:1 copy of them.

Whenever there was a more "Solid-like" way of doing things, that's how I went so we can keep the strengths and familiarities of the framework. So, to sum it up, I heavily used:
- Input Objects as Accessors
- Stores
- Output objects with signals as getters

I find these patters very wise and they are heavily used in TanstackQuery, so I took a good amount of inspiration from there. They allow great fine grained reactivity and protects you from small footguns.

For example, some libs accept both accessor or plain objects, like:
```tsx
useMyLib({ config: "foo" }) // plain input object
useMyLib(() => ({ config: "foo" })) // accessor to input object
```

That works, but if you are new to solid you may forget at some point that the `plain input object` option is **not reactive**.

I think Tanstack Query's choice of **only allowing** `accessor to input object` is very wise, because it cuts down complexity (by having only one way to do the thing) and it's always reactive.

So that's how I designed the api...

```tsx
// Input is an Accessor<DbConfig>
const client = createSolidJazzClient(() => ({
  account,
  sync: { peer: "wss://cloud.jazz.tools/?key=..." },
}));

<JazzProvider client={client}>
  <Todos />
</JazzProvider>;

function Todos() {
  const db = useDb<MyAppDb>();               // Accessor<T>
  const session = useSession();              // Accessor<Session | null>
  const todos = useAll(() => ({ query: db().todos }));

  createEffect(() => {
    // This works because todos.isLoading is a signal and triggers reactivity
    if (todos.isLoading) console.log("loading..."); 
  });

  return <pre>{JSON.stringify(todos.data)}</pre>;
}
```

#### Infra

- [x] Added Solid-specific build/test setup for jazz-tools (including Solid JSX/typecheck path).
- [x] Refactored web client creation into an agnostic base used by Solid bindings.
  - @todo [another PR] : make other frameworks ( react/vue/svelte ) use this file also

#### API & Runtime Integration

- [x] Implemented Solid client factories (createSolidJazzClient, createSolidExtensionJazzClient).
- [x] Implemented Solid provider/context and core hooks (useDb, useSession, useAuthState).
- [x] Implemented Solid query subscription API (useAll) with cache + delta-driven updates.
- [x] Implemented Solid local-first auth API (useLocalFirstAuth) with shared-store synchronization semantics.
- [x] Added coverage for Solid lifecycle/auth/query behavior and validated framework parity.
 
#### Examples & Docs

( More inspired by Vue )
- [x] Added todo-client-localfirst-solid example with parity to existing framework examples.
- [x] Added docs Solid example app/snippets for auth, provider usage, queries, and runtime config.
- [x] Updated docs/README references so Solid appears as a first-class framework option.

## How I used AI
I really do believe it's important to be transparent about this subject. For all of us DEVs, so let's go...

I used AI heavily to
- Study the existing codebase
- Help me make commit messages
- Help catch bugs
- Update docs and examples ( the last 3 commits 99% AI, but of course steered and tested by me )
- Design tests

The tests were very close to 1:1 parity with other frameworks so I felt more confident about them being AI, but I read and tweaked in some small amount all of them. Note that a couple more tests were added on top of the "1:1 framework parity" strategy because I was catching some lifecycle bugs.

To develop the actual APIs... these were practically 100% oldschool. 
I tried using AI but it was awful and I didn't feel like becoming a mess cleaner.

AI helped tweaking a minor detail here and there  about the API but that's it.


packages/jazz-tools/src/web/create-jazz-client.ts
```diff
@@ -0,0 +1,84 @@
...
+import { registerWindowJazzStorageClient } from "../window-client-storage.js";
+
+/**
+ * @todo Make VUE, React and SVELTE consume from this agnostic client
+ *
+ * The following files are practically 1:1 to this one:
+ * ./packages/jazz-tools/src/react/create-jazz-client.ts
+ * ./packages/jazz-tools/src/vue/create-jazz-client.ts
+ * ./packages/jazz-tools/src/svelte/create-jazz-client.ts
+ **/
```
FelipeEmos, line 15: I wonder how opened you guys are for this suggestion, so I'm pointing it out here.

This file is indeed very framework agnostic and is being repeated in React, Vue and Svelte, so what about extracting it here, at `./src/framework-agnostic/create-agnostic-jazz-client.ts` ?

joeinnes: I think @gdorsi already mentioned on Discord that this would be a good improvement for a follow-up PR. 

What do you think about `./src/browser/create-jazz-client.ts` to keep things simple?

FelipeEmos: @joeinnes 
Nice! I 100% agree this belongs to another PR

About the file name, I think this code is not a *browser* thing. With SSR enabled, these frameworks may run this file on the server somewhat frequently.

I suggested the folder `./framework-agnostic` to encourage a place where the adapters could share code that would normally exist inside their framework specific folder, a.k.a. `solid`, `vue`, `svelte`, `react` ... that was my line of thought

That being said, I'm a big fan of keeping things simple! I just wonder `browser` may not be the best name here 🤔 , let me know what you think

joeinnes: Good point. The main reason I'm not fully comfortable with "framework-agnostic" is that React Native still needs special treatment.

What about "web"?

FelipeEmos: Well, you were a few steps ahead of me here.
I’m currently hitting SSR issues where framework hydration tries to serialize this client object.

What should be the intended SSR contract for Jazz framework adapters? Are they currently SSR ready?

joeinnes: So for now, the framework client doesn't run on the server, the SSR pattern is to run two separate Jazz instances: one on the server, one on the client (pattern here: https://github.com/garden-co/jazz/tree/main/examples/nextjs-csr-ssr).

Likely we won't support hydrating the client literally, but we might expose helpers for data serialisation etc.

FelipeEmos: Thanks, man!
I examined and poked this example it's now clear to me the scope of the SSR behavior.

The `JazzProvider` is not resolved during SSR, so its children are not rendered.
- In other words, `<TodoForm />` and `<TodoList />` don't appear when we CURL the page

https://github.com/garden-co/jazz/blob/ac2f9f3f55fdf0522e09fcbca5f356e36a88c945/examples/nextjs-csr-ssr/app/ClientTodo.tsx#L8-L30

FelipeEmos: I'll move on, cus this PR already has the same behavior present in the example!

But I'm wondering if that's the best DX jazz can have 😅
I mean, it's gonna be very tempting to wrap your whole app in a `JazzProvider` ( I would, it's part of my auth and data context... ) and people may not have a pleasent surprise when their whole app practically "turns off SSR" -> more precisely their entire app renders to a blank tree on the server

Of course we should not hydrate the client, but maybe `<JazzProvider />` could provide a "server specific client" on the server? I don't know... just food for thought, maybe that's too much juggling

packages/jazz-tools/src/solid/create-solid-jazz-client-internal.ts
```diff
@@ -0,0 +1,89 @@
...
+  let disposed = false;
+  onCleanup(() => {
+    disposed = true;
+  });
+
+  const [activeRunId, setActiveRunId] = createSignal(0);
+  const [connectedRunId, setConnectedRunId] = createSignal<number | undefined>(undefined);
+
+  const [res, { mutate, refetch }] = createResource(
+    config,
```
joeinnes: `createResource` will check for referential equality on the `config`, so if a signal updates the `config`, we'll tear down the whole client, even if there's no change in the actual values. We stringify here for the same reason: 

https://github.com/garden-co/jazz/blob/ac2f9f3f55fdf0522e09fcbca5f356e36a88c945/packages/jazz-tools/src/react-core/provider.tsx#L218

We should probably do something similar here, and add a similar test to:

https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/react-core/provider.stable-deps.test.tsx

docs/content/docs/reading/queries.mdx
```diff
@@ -144,12 +144,13 @@ Each framework has a reactive binding that re-renders when query results change.
 | React/Expo | `useAll(query)`                | Returns `T[] \| undefined`  |
 | Vue        | `useAll(query)`                | Returns a `Ref`             |
 | Svelte     | `new QuerySubscription(query)` | Exposes reactive `.current` |
+| Solid      | `useAll(() => ({ query }))`    | Returns `T[] \| undefined`  |
```
joeinnes:
```suggestion
| Solid      | `useAll(() => ({ query }))`    | Returns `{ data, loading, error }` |
```

Looks like the actual return type is `{ data: T[] | undefined, loading: boolean, error: Error | unknown }`

docs/content/docs/reading/queries.mdx
```diff
@@ -175,21 +181,22 @@ Each framework has a reactive binding that re-renders when query results change.
...
+Vue's `useAll`, Svelte's `QuerySubscription`, and Solid's `useAll` reconcile new query results into the existing reactive array in place rather than swapping the reference. When an upstream change touches a single row, only that row's affected fields write into the reactive proxy, so `$effect` (Svelte), Solid computations/effects, and `watch` / template bindings (Vue) only re-fire for components that depend on the fields that actually changed.
 
 In practice this means:
 
 - Adding, removing, or reordering rows updates the array structure only — untouched row objects keep their identity, so `{#each items as item (item.id)}` and `<TransitionGroup>` keyed renders are stable.
 - Editing a single field on one row writes only that field — sibling rows do not re-render, and per-row components that read other fields stay quiet.
 - Vue's `useAll` returns a deep `Ref` (not a `shallowRef`), so per-field reactivity composes with the rest of your component tree without manual unwrapping.
+- Solid's `useAll` returns an accessor function, so Solid's fine-grained dependency tracking only re-runs computations that read changed fields.
```
joeinnes:
```suggestion
- Solid's `useAll` returns a store, so Solid's fine-grained dependency tracking only re-runs computations that read changed fields.
```

packages/jazz-tools/src/solid/use-local-first-auth.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+  return listeners;
+}
+
+export type UseLocalFirstAuth = ReturnType<typeof useLocalFirstAuth>;
+
+export function useLocalFirstAuth(store: Accessor<AuthSecretStore> = () => browserAuthSecretStore) {
+  const [secret, setSecret] = createSignal<string | null>(null);
+  const [isLoading, setIsLoading] = createSignal(true);
+  const [error, setError] = createSignal<unknown>(undefined);
```
joeinnes: This is nice, we should probably add this to the other framework adapters.

Suggestion: you've already got an error normalisation function in `use-all.ts`, could we make that reusable so we could type this as `Error | null`?

examples/docs/todo-client-localfirst-solid/src/AuthAnonymous.tsx
```diff
@@ -0,0 +1,9 @@
+import { JazzProvider, createSolidJazzClient } from "jazz-tools/solid";
+
+export function AuthAnonymous(props: { children?: unknown }) {
```
joeinnes:
```suggestion
export function AuthAnonymous(props: ParentProps) {
```

examples/docs/todo-client-localfirst-solid/src/AuthAnonymous.tsx
```diff
@@ -0,0 +1,9 @@
...
+  const client = createSolidJazzClient(() => ({
+    appId: "my-app",
+    serverUrl: "http://127.0.0.1:4200",
+  }));
+  return <JazzProvider client={client}>{props.children as any}</JazzProvider>;
```
joeinnes:
```suggestion
  return <JazzProvider client={client}>{props.children}</JazzProvider>;
```

examples/docs/todo-client-localfirst-solid/src/AuthAnonymous.tsx
```diff
@@ -0,0 +1,9 @@
+import { JazzProvider, createSolidJazzClient } from "jazz-tools/solid";
```
joeinnes, line 2:
```suggestion
import { type ParentProps } from "solid-js";
import { JazzProvider, createSolidJazzClient } from "jazz-tools/solid";
```

examples/docs/todo-client-localfirst-solid/src/AuthExternal.tsx
joeinnes, line 1: Same props typing

packages/jazz-tools/vite.config.solid.ts
```diff
@@ -0,0 +1,77 @@
+import path from "node:path";
+import { defineConfig } from "vite";
+import solid from "vite-plugin-solid";
+
+const solidRoot = path.resolve(__dirname, "src/solid");
+const entryFile = path.resolve(solidRoot, "index.ts");
+
+function shouldExternalizeModule(moduleId: string, parentModuleId: string | undefined): boolean {
+  if (moduleId === "solid-js" || moduleId === "solid-js/web") {
```
joeinnes:
```suggestion
  if (moduleId === "solid-js" || moduleId.startsWith("solid-js/")) {
```
Need to externalise at least `store` and `web`, but probably we should cover any subpath.

joeinnes (changes requested):
This is cool! Thanks for building this out. 

A couple of small points (there's one about typing the props which applies in a number of places but only commented twice) and stability of the config prop, but looks good!

After that, we'll need to exclude `src/solid/**` from tests globally (root `vitest.config.ts`) and add a changeset.

packages/jazz-tools/src/solid/use-local-first-auth.ts
FelipeEmos: I agree, I'm happy you liked it!

I'll use `normalizeError` here then, but since it's appearing only at 2 places and it's a 1 line helper function, I'll copy paste it if it's okay for you.

About the other framework adapters, I'll asssume that's out of scope for the PR.
If we use it in more places, like other framework adapters, then it's worth extracting to `./packages/jazz-tools/src/errors.ts` in my opinion

packages/jazz-tools/vite.config.solid.ts
FelipeEmos: You're right!

packages/jazz-tools/src/solid/create-solid-jazz-client-internal.ts
FelipeEmos: Good call!

I added a test to solid and then fixed with a `stableConfig` that uses `createMemo` but checking equality via the string, as you mentioned from the react version.

About adding tests to other frameworks, I guess that's best to be done in another PR?

FelipeEmos:
Thanks for the review, @joeinnes !
I fixed everything you pointed out on new commits so it's easier to "re-review"

One small question: is the `ChangesetBot` supposed to re-run in this PR now?
Its comment here is outdated since we now have a proper changeset file
- https://github.com/garden-co/jazz/pull/934#issuecomment-4526670411

packages/jazz-tools/src/solid/use-local-first-auth.ts
joeinnes: The scope of the change I'm requesting is really just switching from typing the error as `unknown` to the `Error | null` pattern we use elsewhere.

We'll probably ship a better error discrimination helper that can be used everywhere, so copy-pasting is also fine. The 'we should probably add this to the other framework adapters' wasn't a suggestion for you to add it here, just an acknowledgement that this is neat, and a note for myself on adding it elsewhere.

FelipeEmos:
okay then :)

Let me know if there's something more to change in the PR or if it's good to go

joeinnes (approved):

PR #940, opened 2026-05-26
Fix stack overflow forwarding rows with deep parent history

joeinnes:
## Summary
- Forwarding a row's parent batches to a server walked the history recursively, so a row with a deep history chain overflowed the stack — surfacing as "memory access out of bounds" on the WASM client worker and as OOM on the server. ~300 ancestors is enough to crash a default browser worker (~1 MiB shadow stack).
- Replaced the recursion with an iterative post-order walk over an explicit worklist, guarded by a persistent `visited` set, so each parent batch is loaded and queued at most once regardless of graph shape (deep chain or merged DAG). Queue ordering (parents before children) and cross-forward dedup are preserved.
- Removed an O(n²) clone exposed by the fix: `queue_row_to_server_with_metadata` and `queue_row_to_client` cloned the entire per-object sent-batch set just to test membership of a single batch; it's now checked by borrow. A 2,000,000-deep chain dropped from ~18 min to ~13 s.

## Test plan
- [ ] `cargo test -p jazz-tools --lib sync_manager::` — 75 tests pass, including the new `forwarding_a_deep_parent_chain_does_not_overflow_the_stack`
- [ ] Confirm the regression test is genuinely red on the pre-fix recursion (aborts with a stack overflow) and green with the fix

### DIFF
--- a/crates/jazz-tools/src/sync_manager/tests/forwarding_recursion.rs
+++ b/crates/jazz-tools/src/sync_manager/tests/forwarding_recursion.rs
@@ -84,22 +84,8 @@ fn forwarding_a_deep_parent_chain_does_not_overflow_the_stack() {
     );
 }
 
-/// Scaling guard (a ratio, not a wall-clock budget): forwarding a second batch
-/// of updates against a larger already-sent set must not take materially longer
-/// than the first. Cloning the sent set per forward made it ~3x slower.
-#[test]
-fn forwarding_hot_row_updates_do_not_scale_with_synced_history() {
-    use std::time::{Duration, Instant};
-    const PER_PHASE: usize = 100_000;
-
-    let io = MemoryStorage::new();
-    let mut sm = SyncManager::new();
-    let server_id = ServerId::new();
-    sm.add_server_with_storage(server_id, false, &io);
-    let row_id = ObjectId::new();
-
-    // Build the tips up front so the timed phases measure only forwarding.
-    let tips: Vec<_> = (0..2 * PER_PHASE)
+fn hot_row_tips(row_id: ObjectId, count: usize) -> Vec<StoredRowBatch> {
+    (0..count)
         .map(|k| {
             visible_row(
                 row_id,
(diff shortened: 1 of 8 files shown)


crates/jazz-tools/src/sync_manager/tests/forwarding_recursion.rs
```diff
@@ -0,0 +1,139 @@
...
+        depth + 1,
+        "every batch in the chain should be queued exactly once"
+    );
+}
+
+/// Scaling guard (a ratio, not a wall-clock budget): forwarding a second batch
+/// of updates against a larger already-sent set must not take materially longer
+/// than the first. Cloning the sent set per forward made it ~3x slower.
+#[test]
+fn forwarding_hot_row_updates_do_not_scale_with_synced_history() {
```
gdorsi: This guard only exercises the server forwarding path, but the same sent-batch membership change was made for client queueing too. Could we add a client-scoped hot-row case as well, so the client send path cannot regress back to cloning the full sent set without this test catching it?

crates/jazz-tools/src/sync_manager/tests/forwarding_recursion.rs
```diff
@@ -0,0 +1,139 @@
...
+            );
+        }
+        start.elapsed()
+    };
+
+    // First phase grows the sent-batch set 0 -> PER_PHASE; second PER_PHASE -> 2*PER_PHASE.
+    let first = forward_range(&mut sm, 0..PER_PHASE);
+    let second = forward_range(&mut sm, PER_PHASE..2 * PER_PHASE);
+
+    let ratio = second.as_secs_f64() / first.as_secs_f64();
```
gdorsi: This assertion depends on a wall-clock ratio, so a loaded CI runner could make the second phase look more than 2x slower even when the code is fine. Can we make the regression guard deterministic, or otherwise avoid failing purely on scheduler noise?

gdorsi (commented):
Review comments.

gdorsi (approved):
Looks good to me.

joeinnes:
Switched to deterministic for the regression test and added the client path case.

PR #942, opened 2026-05-26
SQLite core spike

aeplay:
## What

This PR captures the SQLite core spike for rebuilding Jazz bottom-up on top of SQLite.

It includes:

- a new `crates/mini-jazz-sqlite` prototype crate
- an RFC-style `SPEC.md` plus status-quo comparison and research notes
- archived Attempt 1 and Attempt 2 reference implementations
- a continuous `DECISION_LOG.md` for the overnight implementation pass
- a broad whole-system test harness covering sync, durable query descriptors, branches, policies, lenses, transactions, conflict handling, subscriptions, and recursive queries

## Why

The goal is to de-risk whether Jazz can keep its high-level local-first/product semantics while delegating storage, indexing, projection rebuilds, and ACID boundaries to SQLite.

The spike is intentionally ambitious and exploratory. It is not framed as a final production implementation, but it gives us executable evidence for the architecture and highlights remaining gaps.

## Validation

- `cargo test -p mini-jazz-sqlite`
- Latest run: 264 whole-system tests passing

## Notes

- The untracked local `scratch/` directory contains generated wasm-size experiment artifacts and is intentionally not part of this PR.
- Open questions and contradictions are captured in `crates/mini-jazz-sqlite/SPEC.md`, `DECISION_LOG.md`, and `CURRENT_JAZZ_RESEARCH.md`.


crates/mini-jazz-sqlite/src/runtime.rs
```diff
@@ -0,0 +1,4922 @@
...
+                serde_json::to_string(&query_read.value)
+                    .map_err(|err| crate::Error::new(err.to_string()))?,
+                now_ms()
+            ],
+        )?;
+        Ok(())
+    }
+
+    pub fn apply_untrusted_bundle(&mut self, bundle: &Bundle) -> Result<()> {
+        let stale_exclusive_tx_ids =
```
gdorsi, line 1414: Can we make the stale read-set check use the same policy-filtered view as the user who made the read? Otherwise, a row that user could not see might look like a conflict and cause a valid transaction to be rejected.

gdorsi (commented):
Could we add a small COUNT aggregation primitive to this spike? The useful validation is whether we can use SQLite's native aggregation while still keeping the versioned table layout needed for migrations.

crates/mini-jazz-sqlite/src/runtime.rs
aeplay: See [`SPEC.md` authority validation](https://github.com/garden-co/jazz/blob/d788a1676/crates/mini-jazz-sqlite/SPEC.md#L1411-L1438), especially the paragraphs around current authority policy and writer-principal policy-filtered stale comparison.

aeplay:
> Could we add a small COUNT aggregation primitive to this spike? The useful validation is whether we can use SQLite's native aggregation while still keeping the versioned table layout needed for migrations.

See [`SPEC.md` aggregate queries](https://github.com/garden-co/jazz/blob/d788a1676/crates/mini-jazz-sqlite/SPEC.md#L1165-L1198). The benchmark/research matrix also calls out aggregate `COUNT` and account-aggregator-shaped workloads in [`SPEC.md` performance research](https://github.com/garden-co/jazz/blob/d788a1676/crates/mini-jazz-sqlite/SPEC.md#L2158-L2258).

aeplay:
Superseded by the new relational core (#1094). The `crates/mini-jazz-sqlite*` spike crates never landed; the core that shipped is `crates/jazz` + `crates/groove` with its own SPEC. Closing as part of a pre-new-core PR audit — the spike stays in history if we want to revisit the SQLite direction.

PR #945, opened 2026-05-26
spec(jazz-sqlite): Clarify transaction isolation level

nrainhart:
## Description

Made the transaction isolation model more explicit in the spec. The code still feels pretty POC-y in general, so I focused mostly on the spec and the tests.

### DIFF
--- a/crates/mini-jazz-sqlite/spec/08-transactions.md
+++ b/crates/mini-jazz-sqlite/spec/08-transactions.md
@@ -61,6 +61,22 @@ The local write path:
 7. commits the embedded database transaction
 8. publishes local subscription diffs
 
+Transaction isolation:
+
+- A transaction reads from a consistent semantic snapshot of the whole database
+  captured when the transaction starts.
+- Reads inside the transaction include the transaction's own staged writes,
+  layered over the start snapshot.
+- Reads inside the transaction do not include staged writes from any other
+  transaction.
+- Writes committed by other transactions after the transaction starts are not
+  visible inside the transaction; starting a new transaction is the way to read
+  a later current view.
+- Sealing publishes all staged mutations as one transaction. The implementation
+  may lower staged state through SQLite transactions, in-memory overlays, or
+  another embedded-database mechanism, but the public semantics are the
+  start-snapshot plus own-staged-writes view.
+
 Patch updates preserve omitted fields from the effective visible row. The
 effective base may be a current branch row, a row inherited from branch sources,
 or a pinned historical base snapshot. Unknown user fields fail closed before
--- a/crates/mini-jazz-sqlite/spec/18-invariants.md
+++ b/crates/mini-jazz-sqlite/spec/18-invariants.md
@@ -24,6 +24,11 @@ feature exists.
 - One simple write creates one sealed transaction.
 - One explicit transaction may contain multiple row mutations and still seals as
   one transaction.
+- Transactions read from the semantic database snapshot captured when the
+  transaction started.
+- Transaction reads include the transaction's own staged writes.
+- Transaction reads exclude other transactions' staged writes and later
+  committed writes.
 - An explicit transaction with no staged mutations is a no-op and creates no
   transaction record.
 - Multiple staged mutations to the same row in one explicit transaction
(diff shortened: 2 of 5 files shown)


aeplay:
🤖

Findings

[P1] Commit still materializes patches against current state, not the transaction start snapshot.
In [runtime.rs](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/src/runtime.rs#L3629), TransactionBuilder::commit() sends row mutations through insert_row_in_tx(), whose effective_write_values() reads the current effective row via [effective::row_values](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/src/runtime.rs#L2471-L2493). Deletes also snapshot through current runtime.read_rows() at [L3577-L3583](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/src/runtime.rs#L3577-L3583).
That means tx.read_rows() has start-snapshot semantics, but tx.update_row(...).commit() can preserve fields from a later committed row. Example: tx starts with {body:"A", pinned:false}, another tx commits {pinned:true}, then first tx patches only {body:"B"}. The committed history likely becomes {body:"B", pinned:true}, not the start-snapshot {body:"B", pinned:false}. This contradicts the new spec text at [SPEC.md#L742-L756](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/SPEC.md#L742-L756). I’d add this as a failing test before merging.

[P1] Transaction read reconstruction can silently erase branch conflict candidates.
[collapse_transaction_rows()](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/src/runtime.rs#L3488-L3523) keeps all rows at the minimum source depth, but then [read_rows() collects them into BTreeMap<id, row>](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/src/runtime.rs#L3218-L3222). If two equal-depth branch sources expose different candidates for the same row id, one candidate overwrites the other by iteration order. Current effective-row logic treats multiple source candidates as ambiguous/erroring, so this new transaction path is not semantically equivalent. This matters because transactions are exactly where unresolved branch conflicts and write preconditions will become important.

[P2] Snapshot visibility is rebuilt with a simpler “latest physical tx_num” rule.
In [read_history_rows_at_start()](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/src/runtime.rs#L3350-L3368), visibility is “in dotted vector, non-rejected, no newer tx_num.” That is not obviously the same reducer as ordinary semantic visibility once outcomes, pending/exclusive transactions, global epochs, merge resolution, branch candidates, and policy visibility are all in play. For this PR’s simple tests it works, but I would not want this to become a second, subtly different projection engine. Ideally transaction snapshot reads should call into one shared visibility reducer, parameterized by “as of vector/base,” rather than fork the rules here.

[P3] The PR reintroduces DECISION_LOG.md.
[DECISION_LOG.md](https://github.com/garden-co/jazz/blob/2dda6f36c2e646f8413422e5586acb4398dcef82/crates/mini-jazz-sqlite/DECISION_LOG.md) is useful context, but we’ve been folding logs into the spec and deleting them from the branch. I’d either fold the note into SPEC.md or leave it out of the PR.

Open Question

The decision log asks whether commit-time read-set versions should be forced from the start snapshot. My read: yes. Otherwise the transaction’s validation/read-set story diverges from its read API, which is a particularly sharp footgun for exclusive transactions and policy-dependent writes. **Anselm: I agree**

Summary

The spec clarification and the four new tests are directionally excellent: “start snapshot plus own staged writes” is the right contract. The main issue is that only transaction reads got the snapshot treatment; commit-time patch materialization, delete snapshots, policy/write checks, and recorded read sets still appear to consult current state. That should be fixed before merging, or the PR will document stronger isolation than the runtime actually provides.

PR #953, opened 2026-05-27
Add mini-sqlite WASM runtime, query API, and todo benchmark

gdorsi:
## Summary

This PR makes `mini-jazz-sqlite` usable from a browser, exposes a Rust/WASM query surface based on `BuiltQuery`, and adds a browser todo benchmark that exercises paginated query speed over a large local dataset with realistic query controls.

The main direction is:

```text
specialized runtime helpers     BuiltQuery JSON
read_rows_where_*          ->   query / one / subscribe
ad-hoc browser smoke       ->   examples/mini-sqlite-todo
runtime-side page scans    ->   SQL-lowered predicates/order/window
```

## Browser/WASM Runtime

The WASM wrapper exposes the embedded Rust runtime through `wasm-bindgen`, with both memory and OPFS-backed storage:

```text
browser UI
   |
   v
examples/mini-sqlite-todo/src/db-worker.js
   |
   v
wasm-bindgen MiniJazzRuntime
   |
   +-- openMemory(nodeId, user)
   +-- openOpfs(dbName, nodeId, user)
   +-- insertRow(table, id, values)
   +-- updateRow(table, id, values)
   +-- deleteRow(table, id)
   +-- query(builtQuery)
   +-- one(builtQuery)
   +-- subscribe(builtQuery, callback)
   +-- unsubscribe(handle)
   v
mini_jazz_sqlite::Runtime
   |
   v
SQLite WASM VFS / OPFS
```

Example browser usage:

```js
import init, { MiniJazzRuntime } from "./generated/mini-jazz-sqlite-wasm/mini_jazz_sqlite_wasm.js";

await init();

const db = await MiniJazzRuntime.openOpfs(
  "mini-sqlite-todo",
  "browser-tab-1",
  "alice",
);

db.insertRow("todos", `todo-${crypto.randomUUID()}`, {
  title: "Measure pagination",
  done: false,
  project: "todo-list",
});

const rows = db.query({
  table: "todos",
  conditions: [{ column: "done", op: "eq", value: false }],
  orderBy: [["$createdAt", "desc"]],
  limit: 10,
});
```

## Query Pipeline

The Rust query API accepts the same JSON-shaped query descriptor at the WASM boundary and lowers supported operations into SQLite instead of routing through prebuilt runtime methods.

```text
BuiltQuery JSON
   |
   v
parse + validate bounded fields
   |
   v
resolve schema fields and system fields
   |
   v
compile predicates / order / limit / offset
   |
   v
SQLite over current projections
   |
   v
semantic RowView values
```

Supported query operations now go through one path:

```js
const PAGE_QUERY = {
  table: "todos",
  conditions: [
    { column: "done", op: "eq", value: false },
  ],
  orderBy: [["$createdAt", "desc"]],
  limit: 10,
};

const latestOpenTodos = db.query(PAGE_QUERY);
const firstOpenTodo = db.one(PAGE_QUERY);
```

The intended SQL shape is now the ordinary database shape:

```sql
SELECT current.*
FROM current_todos AS current
WHERE current.done = ?
  AND <branch visibility>
  AND <read policy visibility>
ORDER BY current.j_created_at DESC, <stable tie breaker>
LIMIT 10 OFFSET 0;
```

That matters for the 100k todo benchmark: pagination should be an indexed query over the current projection, not a runtime scan followed by in-memory filtering and slicing.

## Query-Scoped Sync And Repair

Query exports remain history-based, not authoritative result snapshots. A receiver applies history and observed facts, then reruns its local query.

```text
upstream query export
   |
   +-- current result rows
   +-- repair rows for rows that left the scope
   +-- predicate / page / policy observed facts
   +-- branch and snapshot context when needed
   v
sync bundle
   |
   v
receiver applies history
   |
   v
receiver reruns query locally
```

This PR tightens the built-query refresh path so SQL-lowered queries can repair rows that enter or leave the visible result set through:

```text
field update
row delete
timestamp/order change
policy dependency change
branch/source change
page boundary displacement
```

The generic query repair tests cover combinations of supported operators, ordering, pagination, policy filtering, and incremental updates.

## Subscriptions

Subscriptions now use `BuiltQuery` for the query-shaped API and emit callback deltas from the WASM wrapper.

```js
const handle = db.subscribe(PAGE_QUERY, ({ all, delta }) => {
  for (const change of delta) {
    if (change.kind === 0) {
      // added: change.row is present
    } else if (change.kind === 1) {
      // removed: change.id and change.index identify the removal
    } else if (change.kind === 2) {
      // updated or moved: change.row is present for value updates
    }
  }

  render(all);
});

db.unsubscribe(handle);
```

The WASM notification loop keeps subscription state consistent when callbacks write rows or throw. One failing callback no longer prevents later callbacks from seeing the write that already happened.

## Removed Specialized Surface

The public direction is query language first. This PR removes or avoids public APIs that bypass that path:

```text
read_rows_where_eq_top_created_at_desc   removed from the public browser path
read_rows_where_eq_top_field_desc        replaced by BuiltQuery orderBy + limit
export_table_history                     not exposed on the WASM API
apply_bundle                             not exposed on the WASM API
```

The remaining browser-facing shape is smaller:

```text
writes:        insertRow / updateRow / deleteRow
reads:         readRows / query / one
live reads:    subscribe / unsubscribe
storage info:  storageStats / storageFormatVersion
```

## Todo Benchmark Example

The example is intentionally small and named `mini-sqlite-todo`, but now exercises a more realistic local app shape:

```text
examples/mini-sqlite-todo
   |
   +-- browser app
   +-- db worker
   +-- generated WASM package
```

The worker stores todos plus a many-to-many label model:

```text
todos
labels
todo_labels
```

Labels are maintained as an aggregate app-level subscription rather than hydrated row-by-row for each visible todo:

```text
subscribe({ table: "labels", orderBy: [["name", "asc"]] })
   |
   v
labelsById cache
   |
   +-- global label filter chips
   +-- filter validation
```

The paged todo benchmark path is kept separate from that aggregate label catalogue:

```text
generate 100000 todos with deterministic labels
   |
   v
store todos and label edges through MiniJazzRuntime.openOpfs(...)
   |
   v
keep labels available through the labels subscription
   |
   v
compose BuiltQuery JSON from UI controls
   |
   +-- title contains search text
   +-- optional label filter using todo_labels to find matching todo ids
   +-- order by title or $createdAt
   +-- asc or desc direction
   v
return only the first 10 matching todos
```

The UI supports:

```text
manual todo creation with comma-separated labels
full text-style title search through the query API contains operator
global label filter chips backed by the labels subscription
ordering controls for name/date in both directions
top-10 todo query timing
100k row generation timing
```

The worker reports:

```js
{
  type: "state",
  todos,
  labels,
  filters,
  queryMs,
  currentRows,
  generateMs,
}
```

Per-row label chips were removed from the visible todo list. That keeps the unfiltered timing focused on the paged todo query instead of including N extra `todo_labels` queries to decorate the 10 visible rows.

This keeps the example focused on the performance question while making the benchmark closer to an actual app query: how fast can the browser runtime answer a paged query after a large local seed, once search, label filters, and user-controlled ordering are involved?

## Correctness Notes

The query implementation is meant to preserve these contracts:

```text
same logical data + same branch + same policy + same query
   => same semantic result rows
   => same semantic order
   => same page window
```

Local storage ids are implementation details. Query results, ordering, and pagination should be based on semantic values and public ids rather than SQLite allocation order or internal row numbers.

Read visibility is also part of query planning, not a post-filter. Query, export, and repair paths need to use the same branch and policy visibility rules before selecting result rows or page windows.

## Validation

```text
cargo fmt --check
cargo test -p mini-jazz-sqlite
cargo clippy -p mini-jazz-sqlite --all-targets -- -D warnings
CC_wasm32_unknown_unknown=/opt/homebrew/Cellar/llvm@21/21.1.8/bin/clang cargo check -p mini-jazz-sqlite-wasm --target wasm32-unknown-unknown
pnpm --filter mini-sqlite-todo build
pnpm --dir examples/mini-sqlite-todo build
```

Browser smoke covered:

```text
memory runtime: generate 100k todos, query latest 10
OPFS runtime:   generate todos through real worker path, query latest 10
subscriptions: query, subscribe, callback delta, unsubscribe
API surface:    removed specialized helpers are absent from the browser path
todo example:   labels aggregate through subscription, add labeled todo, filter by label, search title text, change name/date sort order
```

### DIFF
--- a/crates/mini-jazz-sqlite/src/read_set.rs
+++ b/crates/mini-jazz-sqlite/src/read_set.rs
@@ -64,7 +64,7 @@ pub(crate) fn record_tx_read_num_with_observed(
     reason: i64,
     observed_tx_num: Option<i64>,
 ) -> Result<()> {
-    let mut stmt = conn.prepare_cached(
+    let mut stmt = conn.prepare(
         "INSERT OR REPLACE INTO jazz_tx_read
          (tx_num, table_num, row_num, reason, observed_tx_num)
          VALUES (?, ?, ?, ?, ?)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2311,6 +2311,7 @@ dependencies = [
  "sqlite-wasm-vfs",
  "wasm-bindgen",
  "wasm-bindgen-futures",
+ "wasm-bindgen-test",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2302,6 +2302,7 @@ name = "mini-jazz-sqlite-wasm"
 version = "0.1.0"
 dependencies = [
  "console_error_panic_hook",
+ "js-sys",
  "mini-jazz-sqlite",
  "serde",
  "serde-wasm-bindgen 0.6.5",
--- a/crates/mini-jazz-sqlite-wasm/Cargo.toml
+++ b/crates/mini-jazz-sqlite-wasm/Cargo.toml
@@ -9,6 +9,7 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 console_error_panic_hook = "0.1"
 mini-jazz-sqlite = { path = "../mini-jazz-sqlite" }
+js-sys = "0.3"
 serde = { version = "1", features = ["derive"] }
 serde-wasm-bindgen = "0.6"
 serde_json = "1"
--- a/crates/mini-jazz-sqlite-wasm/Cargo.toml
+++ b/crates/mini-jazz-sqlite-wasm/Cargo.toml
@@ -19,3 +19,6 @@ wasm-bindgen-futures = "0.4"
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 sqlite-wasm-rs = "0.5"
 sqlite-wasm-vfs = "0.2"
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3"
--- a/crates/mini-jazz-sqlite-wasm/src/lib.rs
+++ b/crates/mini-jazz-sqlite-wasm/src/lib.rs
@@ -165,22 +165,6 @@ impl MiniJazzRuntime {
         )
     }
 
-    #[wasm_bindgen(js_name = readRowsWhereEqTopCreatedAtDesc)]
-    pub fn read_rows_where_eq_top_created_at_desc(
-        &self,
-        table_name: &str,
-        field_name: &str,
-        value: JsValue,
-        limit: usize,
-    ) -> Result<JsValue, JsValue> {
-        let value = parse_json_value(value)?;
-        to_js_value(
-            self.runtime
-                .read_rows_where_eq_top_created_at_desc(table_name, field_name, value, limit)
-                .map_err(to_js_error)?,
-        )
-    }
-
     #[wasm_bindgen(js_name = exportTableHistory)]
     pub fn export_table_history(&self, table_name: &str) -> Result<JsValue, JsValue> {
         to_js_value(
--- a/crates/mini-jazz-sqlite-wasm/src/lib.rs
+++ b/crates/mini-jazz-sqlite-wasm/src/lib.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeMap;
 
-use mini_jazz_sqlite::{BuiltQuery, RowsSubscription, Runtime, Storage, SubscriptionDelta};
+use mini_jazz_sqlite::{BuiltQuery, RowsSubscription, Runtime, Storage};
 use serde::Serialize;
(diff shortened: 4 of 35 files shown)


crates/mini-jazz-sqlite/src/read_set.rs
```diff
@@ -64,7 +64,7 @@ pub(crate) fn record_tx_read_num_with_observed(
     reason: i64,
     observed_tx_num: Option<i64>,
 ) -> Result<()> {
-    let mut stmt = conn.prepare_cached(
+    let mut stmt = conn.prepare(
```
gdorsi, line 67: This change is for WASM compat

PR #956, opened 2026-05-27
Fix batched_tick deadlock with deferred query subscriptions

antoniomuso:
## Summary

- `runtime_core::ticks::batched_tick` could strand a `CatalogueEntryUpdated` in `parked_sync_messages` forever when a server-side query subscription was deferred, because the early-return at `ticks.rs:724` jumped past `handle_sync_messages` — the only path that drains the parked queue.
- The same early-return (and its twin at line 741) unconditionally re-armed the scheduler, so the runtime kept rescheduling itself with no forward progress. On React Native, `RnScheduler::schedule_batched_tick` fired its JS callback synchronously with no debounce, turning that into a microtask hot loop that pegged the JS thread and starved `setInterval`/render.
- Drain parked messages before deciding to reschedule, only reschedule when progress was made, and defer the RN scheduler callback through a 1ms-sleeping background thread to break synchronous re-entry.

## Changes

**`crates/jazz-tools/src/runtime_core/ticks.rs`**
- `handle_sync_messages` now returns `bool` (true iff anything was drained).
- `batched_tick` always calls `handle_sync_messages` and reschedules only when it drained something AND a subscription is still pending. Progressless ticks now exit cleanly; the next inbound message will schedule via `park_sync_message` on its own.

**`crates/jazz-rn/rust/src/lib.rs`**
- `RnScheduler::schedule_batched_tick` defers `request_batched_tick` through `thread::spawn` + `sleep(1ms)`, mirroring `schedule_mutation_error_delivery` and `NapiScheduler::schedule_batched_tick`. Coalesces bursts and breaks synchronous JS re-entry.

**`crates/jazz-tools/src/runtime_core/tests/batched_tick_parked_drain.rs`** (new)
Four regression tests:
1. Parked messages must be drained even when a sub is deferred.
2. `batched_tick` must not hot-spin on a progressless tick.
3. End-to-end: a parked `CatalogueEntryUpdated` unblocks the deferred sub after a bounded number of ticks.
4. Control: the no-deferred-sub path still drains parked messages.

## Test plan

- [x] `cargo test -p jazz-tools --lib runtime_core::tests::batched_tick_parked_drain` — 4/4 pass
- [x] `cargo test -p jazz-tools --lib` — 1227/1227 pass, no regressions
- [x] `cargo test -p jazz-rn` — 3/3 pass
- [x] `cargo check --workspace` — clean
- [ ] Manual verification on a React Native client: deferred-sub scenario no longer pegs CPU or starves `setInterval`


crates/jazz-tools/src/runtime_core/tests/batched_tick_parked_drain.rs
```diff
@@ -0,0 +1,287 @@
...
+    assert_eq!(core.parked_sync_messages.len(), 0);
+
+    let ticks = 10;
+    let baseline = scheduler.schedule_count();
+    for _ in 0..ticks {
+        core.batched_tick();
+    }
+    let reschedules = scheduler.schedule_count() - baseline;
+
+    assert!(
```
gdorsi: This still allows a mostly-hot loop: if 9 out of 10 progressless ticks rescheduled, `reschedules < ticks` would pass while the scheduler is still burning CPU. Since the setup has no parked messages, no outbox, and no storage flush, can this assert `reschedules == 0` so progressless ticks never re-arm the scheduler?

.changeset/batched-tick-parked-drain.md
```diff
@@ -0,0 +1,9 @@
+---
+"jazz-tools": patch
+"jazz-rn": patch
+---
+
+Fix a `batched_tick` deadlock that left React Native apps spinning at 100% CPU when a server-side query subscription was deferred waiting for its catalogue/schema.
+
+- `jazz-tools`: `RuntimeCore::batched_tick` now drains parked sync messages before deciding whether to reschedule, so a `CatalogueEntryUpdated` that arrives while a subscription is parked actually unblocks it instead of sitting in the parked queue forever. Progressless ticks no longer re-arm the scheduler, breaking the reschedule hot loop.
+- `jazz-rn`: `request_batched_tick` is now deferred off the JS thread (mirroring `schedule_mutation_error_delivery` and `NapiScheduler`) so the JS callback can't synchronously re-enter `batched_tick` and starve `setInterval` / rendering.
```
gdorsi, line 9: This release note says the React Native scheduler change is part of the fix, but the new regression coverage only exercises `runtime_core`. Could we add a small `jazz-rn` scheduler test so the synchronous JS re-entry loop is covered if it regresses?

gdorsi (approved):
Approved with two small coverage nits.

.changeset/batched-tick-parked-drain.md
antoniomuso: We don't have a way to test this now. I am going to add an example app that works like an integration test in another PR.

PR #965, opened 2026-05-28
Cap WS connections per client_id and time out idle handshakes

joeinnes:
## Summary
Closes the reconnect-storm amplification path on the Jazz server's WebSocket route described in ticket jaz0-a803. Two mitigations:

- **Per-`client_id` connection cap (4) with evict-oldest** in `ConnectionEventHub::register_connection`. A single `client_id` cannot pin unbounded server-side fan-out memory by opening many concurrent sockets, and a reconnecting client always wins against its own stale sockets. Evicted clients receive a `ServerEvent::Error { code: RateLimited }` followed by a close with code 1008 (Policy) so the eviction is observable application-side, not just at the TCP layer.
- **10s handshake-read timeout** on the initial `socket.recv()` in `handle_ws_connection`. A WS upgrade that never sends an `AuthHandshake` frame is closed with a policy close instead of pinning a Tokio task, an fd, and a recv buffer indefinitely.

The wire protocol's `ErrorCode::RateLimited` was previously defined but never emitted — it's now sent on the eviction path with the post-handshake binary encoding (the existing `send_ws_error_with_code` was JSON-only and only correct pre-handshake, so a separate `send_ws_error_binary` helper was added).

Out of scope for this PR (per ticket scoping discussion): bounded outbound channels, shorter heartbeats, global connection caps, per-IP rate limits, TCP keepalives. Each can be filed as a follow-up if telemetry justifies.

## Test plan
- [x] `cargo test -p jazz-tools --lib --features test reconnect_storm_tests` — 7 tests pass in ~12s
  - 2 exploit-primitive demonstrations at 100-socket scale (concurrent-storm + idle-upgrade slowloris)
  - 5 fix-shape contracts (evict-oldest; rate-limited error frame; cross-client_id isolation; steady-state eviction; handshake-timeout × shutdown race)
- [x] Full `cargo test -p jazz-tools --tests --features test` suite — 1461 lib + ~200 integration tests, no regressions
- [x] Tests observe the server only through the public WS protocol plus the live `ServerState::connections` map, so they remain valid under any implementation that satisfies the behavioural contract
- [x] Changeset added for jazz-tools patch bump

### DIFF
--- a/.changeset/reconnect-storm-connection-cap.md
+++ b/.changeset/reconnect-storm-connection-cap.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Bound the reconnect-storm amplification on the server's WebSocket path: enforce a per-`client_id` connection cap (4) with evict-oldest semantics so a single client_id cannot pin unbounded fan-out memory, and time out pre-handshake sockets after 10s so an unauthenticated peer cannot park resources without sending the `AuthHandshake` frame. Evicted clients receive a `RateLimited` error frame followed by a policy close.
--- a/crates/jazz-tools/src/server/mod.rs
+++ b/crates/jazz-tools/src/server/mod.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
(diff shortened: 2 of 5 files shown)


gdorsi (approved):
Reviewed the release-note section. No concerns.

PR #966, opened 2026-05-28
docs: lens diagram — typed field pills instead of fake checkbox

joeinnes:
## Summary
- The client phone in the lens diagram rendered the projected row as a to-do with a non-functional `<input type="checkbox">`. The checkbox looked interactive but didn't toggle anything, which surprised reviewers who tried it.
- Replace the simulated to-do UI with a column of typed field pills: one chip per field, strings in the brand accent (`#146aff`), booleans as green/red status chips with a coloured dot.
- No simulated affordances any more — the only interactive controls are the data cards on the left and the v1/v2/v3 segmented picker.

## Test plan
- [ ] `pnpm --filter=docs dev`, open `/docs/schemas/migrations`, scroll to the lens diagram.
- [ ] Click each data card on the left; confirm pills morph and JSON updates.
- [ ] Toggle v1 / v2 / v3 on the phone's segmented control; watch pills slide in after the trace dot, JSON cascades behind.
- [ ] Switch v2 ↔ v3 on the client; the `done`/`completed` pill label crossfades in place (no re-mount).
- [ ] Compare light vs dark mode — booleans + string chips should remain legible in both.

### DIFF
--- a/docs/components/mdx/lens-diagram.tsx
+++ b/docs/components/mdx/lens-diagram.tsx
@@ -212,27 +212,108 @@ function LensCard({
 }
 
 const PROJECTION_CSS = `
-@keyframes lens-projection-in {
+@keyframes lens-pill-in {
   from { opacity: 0; transform: translateX(-4px); }
-  to { opacity: 1; transform: translateY(0); }
+  to { opacity: 1; transform: translateX(0); }
 }
-.projection-row { animation: lens-projection-in 200ms ease-out 350ms both; }
+@keyframes lens-pill-key-flip {
+  0%   { opacity: 0; transform: translateY(-3px); }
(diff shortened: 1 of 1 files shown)


nrainhart:
Screenshot 2026-05-28 at 11 37 24 AM

@joeinnes I found this UI a bit confusing. I first thought that these were 2 separate tasks (instead of 2 fields of the same task). I think both the phone diagram and the "My To Dos" title led me in that direction. Alternatively, what do you think about greying out the checkbox (and maybe change the cursor on hover) to make it more explicit that it's not clickable?

aeplay:
🤖 Closing this historical documentation branch because it describes an obsolete lens presentation. Any current lens illustration should be regenerated from the maintained API and terminology.

PR #967, opened 2026-05-28
Migrate inspector to TanStack Router 

cleminso:
Migrates the inspector to TanStack Router with URL-driven connection, branch, schema, table, and table-search state.

Key changes:
- Adds typed route constants, route resolution helpers, and file-based routes.
- Makes full URL params authoritative for standalone runtime setup.
- Keeps stored connections as local credential/default storage.
- Adds typed table search param validation.
- Updates tests and removes React Router usage from inspector.


aeplay:
🤖 Closing as obsolete after a current-main review. The historical implementation is based on the removed inspector extension/bridge architecture and bundles unrelated old commits, so a rebase would be a fresh migration rather than a safe conflict resolution. Its TanStack Router direction can be reconsidered as new work against the current standalone/embedded inspector if still desired.

PR #975, opened 2026-05-29
Fix permissions head-before-bundle race causing local write denials

joeinnes:
## Summary
- `process_catalogue_permissions_head` called `require_authorization_schema()` unconditionally, flipping `row_policy_mode` to `Enforcing` the moment a permissions head arrived. When the matching bundle hadn't yet been received, `authorization_schema` stayed `None`, and every local insert/update in that window was denied by the missing-explicit-policy check — even writes against tables with wide-open policies.
- `apply_permissions_head` already calls `set_authorization_schema` (which sets Enforcing + the required flag) when the bundle is known, so the earlier call was redundant on success and incorrect during the gap. Dropping it lets the client stay `PermissiveLocal` until the bundle merges in.
- Resolves the world-tour browser flake where `useAll resolves .include() relations and reflects later updates` intermittently failed with `policy denied INSERT on table bands` under CI load (CI runs 26412268975, 26460534994, 26489581001, 26511724023, 26579926475).

## Test plan
- [ ] `cargo test -p jazz-tools` — new `permissions_head_without_bundle_does_not_deny_local_writes` unit test exercises the gap; fails at the red commit, passes at the fix.
- [ ] `pnpm --filter world-tour test`

### DIFF
--- a/crates/jazz-tools/src/schema_manager/manager.rs
+++ b/crates/jazz-tools/src/schema_manager/manager.rs
@@ -1356,13 +1356,16 @@ impl SchemaManager {
             parent_bundle_object_id,
             bundle_object_id,
         };
-        self.query_manager.require_authorization_schema();
         if let Some(current_head) = self.current_permissions_head
             && current_head.version > head.version
         {
             return Ok(());
         }
         self.current_permissions_head = Some(head);
+        // Defer flipping row_policy_mode to Enforcing until apply succeeds —
+        // apply_permissions_head calls set_authorization_schema which sets it.
+        // Flipping earlier denies writes against tables whose explicit policy
+        // lives in the not-yet-arrived bundle.
         if self.apply_permissions_head(head) {
             self.pending_permissions_head = None;
         } else {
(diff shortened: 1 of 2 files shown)


crates/jazz-tools/src/schema_manager/manager.rs
```diff
@@ -2590,6 +2593,61 @@ mod tests {
...
+        );
+
+        let write_context = WriteContext::from_session(Session::new("test-user"));
+        let values = HashMap::from([
+            ("id".to_string(), Value::Uuid(ObjectId::new())),
+            ("name".to_string(), Value::Text("Alice".into())),
+            ("email".to_string(), Value::Text("alice@example.com".into())),
+        ]);
+
+        manager
```
gdorsi, line 2647: nit: the new test proves that writes work while the permissions bundle is missing, but it stops before the bundle arrives. 

Shall we also cover that the permissions are applied correctly once the bundle is applied?

gdorsi (approved):
Approved with one inline note.

PR #976, opened 2026-05-29
feat: Safari SharedWorker leader mode

antoniomuso:
## Summary

Implements the **Safari leader path**: a SharedWorker that holds `LOCK_NAME`, opens OPFS, runs the Jazz runtime, and answers tab connections with a per-tab `MessagePort`. On Safari, tabs become pure followers and no longer spawn a dedicated Worker. Chrome/Firefox are unchanged — the capability probe falls through to the existing BroadcastChannel + dedicated-Worker election path.

Implements `docs/superpowers/plans/2026-05-28-shared-worker-leader-safari.md`.

### How it works
- **Capability probe** (`capability.ts`) runs *inside* a SharedWorker to detect sync-OPFS-in-SharedWorker (true on WebKit, false on Chromium today).
- **Single-message handshake**: tab → `CONNECT {tabId, scope, version, appId, dbName, schemaJson, …}` → leader → `PEER_PORT {port, generation}` or `LEADER_FAULT {reason}`, plus a cheap `CHECK_CAPABILITY` / `CAPABILITY_RESULT` probe pair.
- **`Db` integration**: eager `CHECK_CAPABILITY` probe in `createWithWorker`; on support, skips `TabLeaderElection` + `spawnWorker`. The actual `CONNECT` (carrying schema) is deferred to the first `getClient(schema)`, which installs a `MessagePortRuntimeTransport` on the follower runtime.
- **Rust `worker_host`** is widened to run in either `DedicatedWorkerGlobalScope` or `SharedWorkerGlobalScope` (via `globalThis`/`Reflect`), accepts a transferred `MessagePort` (`attach-follower-port`), and posts leader-bound sync directly to attached ports. An `on_init_ok` callback replaces the previous postcard-tag sniff.

### Naming / scope
- Narrow, honest names (`SharedWorkerLeader`, `LEADER_PROTOCOL_VERSION`, `CONNECT`/`PEER_PORT`) — in this mode the SharedWorker *is* the leader, not a broker.
- **Out of scope**: tab-hosted SharedWorker mode (Chrome/Firefox election state machine) — a separate future plan.

## Verification (Chromium + units)
- Node unit tests: 16 pass (`protocol` 12, `url` 4, incl. a `JAZZ_PACKAGE_VERSION`↔`package.json` drift guard).
- Browser tests (Chromium): `capability` (3), `spike` (1), `attach-follower-port` end-to-end in a dedicated Worker (1), `client` against a stub SharedWorker (3) — all pass.
- Rust: `worker_to_main_round_trips` extended for the new ack variants; `pnpm build:core` (jazz-wasm) compiles clean; clippy/rustfmt clean.
- **Regression**: `worker-bridge` 48/48 pass; the Chromium `createDb` path falls through to the dedicated-Worker election correctly (probe returns `false` fast — no timeout).

## Deferred / needs follow-up
- **WebKit end-to-end verification is pending.** The 4 E2E tests (`cold-start`, `two-tab`, `tab-close`, `restart`) are written and gated with `describe.skipIf(!leaderSupported)` — they **skip on Chromium** (visible) and only exercise the leader path on WebKit. The repo's `vitest.config.browser.ts` only configures Chromium, so they have **not** been run against WebKit yet. Needs a WebKit-configured runner (or manual Safari Technology Preview run) to confirm the full data plane.
- **Storage reset** is unavailable on the leader path (the `StorageResetCoordinator` rides BroadcastChannel, which the leader path does not open).
- **Bundle size**: the leader entry bundles the full WASM runtime unconditionally; a capability-gated dynamic import is a follow-up.
- **`generation`** is always 1 in v1 (reserved field; becomes load-bearing only in the tab-hosted follow-up).
- On a supported-but-failed CONNECT, the design **throws** (`SharedWorkerLeaderConnectError`) rather than falling back — by explicit decision.

## Test Plan
- [ ] Run the 4 `shared-worker-leader-*` E2E tests against WebKit / Safari Technology Preview and confirm pass.
- [ ] Confirm Chromium/Firefox behaviour is unchanged in a real app (dedicated-Worker path).


packages/jazz-tools/src/runtime/shared-worker-leader/message-port-runtime-transport.ts
```diff
@@ -0,0 +1,73 @@
...
+      }
+    });
+    this.runtime.replayFollowerServerEdge();
+
+    this.port.onmessage = (event: MessageEvent) => {
+      const data = event.data as { type?: string; payload?: unknown };
+      if (data?.type !== "leader-sync") return;
+      const payload = data.payload;
+      if (!Array.isArray(payload)) return;
+      for (const entry of payload) {
```
antoniomuso, line 54: [high] After applying each batch of incoming `leader-sync` payloads, the transport never calls `runtime.batchedTick()`. The dedicated-Worker bridge does this (see `worker_bridge.rs:851`), and without it the parked sync messages are never processed — queries won't settle and subscriptions won't fire until some unrelated event happens to trigger a tick.

antoniomuso (commented):
Review of the SharedWorker-leader transport layer.

packages/jazz-tools/src/runtime/shared-worker-leader/message-port-runtime-transport.ts
antoniomuso: Confirmed. `apply_incoming_follower_payload` only calls `on_sync_message_received` and never ticks, whereas WorkerBridge calls `batched_tick()` after each worker→main sync batch (worker_bridge.rs:851). Added a `batchedTick()` after each applied `leader-sync` batch (guarded so empty batches don't tick), plus a Chromium transport unit test (`shared-worker-leader-transport.test.ts`) that asserts the tick fires exactly once per non-empty batch — it covers the data-plane apply path that previously had no non-WebKit coverage. Fixed in cf12083dc.

PR #979, opened 2026-05-29
Type-check docs example snippets in the build

joeinnes:
## Summary
- Type-check every `examples/docs/*` snippet package (`tsc` / `svelte-check` / `vue-tsc`) and wire all six into the docs site's turbo `build`, so a broken snippet now fails the build instead of shipping silently.
- Fix the type errors this surfaced across the React/Svelte/Vue/TS/Expo snippets (e.g. `s.enum` arity, destructuring `db.insert(...).value`, the Svelte `JazzSvelteProvider` import), and align the Rust `where` snippet to TypeScript — the `$createdAt` magic column instead of a non-existent `created_at`, plus a `priority` column added to the `todo-server-rs` schema and migration.
- Export `RowRefValue` / `SessionRefValue` from `jazz-tools` so helpers inside `definePermissions(...)` callbacks can be typed against row and session references.
- Rename the quickstart snippet files to `AddTodo` / `TodoItem` (and point Expo's docs at its real components) so filenames match the titles shown to readers; add pre-commit hooks covering the snippet and docs-site type-checks.

## Test plan
- [ ] `pnpm build:core`, then `pnpm turbo run build --filter='./examples/docs/*' --only` — all six docs example packages type-check.
- [ ] `pnpm --filter docs types:check` passes.


examples/docs/todo-client-localfirst-react/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "node ../../../packages/jazz-tools/dist/cli.js validate && vite build",
+    "build": "node ../../../packages/jazz-tools/dist/cli.js validate && tsc --noEmit && vite build",
```
nrainhart: Nit: can we extract `tsc --noEmit` into a `check` command?

turbo.json
```diff
@@ -10,7 +10,13 @@
...
         "todo-client-localfirst-react#build",
         "todo-client-localfirst-ts#build",
-        "todo-server-ts#build"
+        "todo-server-ts#build",
+        "todo-client-localfirst-react-docs#build",
+        "todo-client-localfirst-ts-docs#build",
+        "todo-client-localfirst-svelte-docs#build",
+        "todo-client-localfirst-vue-docs#build",
+        "todo-client-localfirst-expo-docs#build",
+        "todo-server-ts-docs#build"
```
nrainhart, line 19: This seems error prone if we keep adding examples for docs. Is there a way to say "all examples in `examples/docs` should build"?

nrainhart (approved):
This is great, thanks!!!

turbo.json
joeinnes: I don't think `dependsOn` can glob, but it's belt and braces because of this:
https://github.com/garden-co/jazz/pull/979/changes/BASE..8d01dafa799af5402107a24aaa011e34054cd53d#diff-ad6a01e589b8b1b214ca310dbb8d2e4314f6c612b921050c73c97455de43884dR29

PR #980, opened 2026-05-29
fix: permissions that reference row IDs in `exists` policies

nrainhart:
## Description

Fixes update permission checks that reference `id` inside `exists` policies.

The policy evaluator now carries the current row object ID through these policies, so checks like `chatMembers.chatId == chat.id` and self-checks like `chats.id == chat.id` evaluate correctly.

Also added an example of how policies can be used to enforce only specific columns in a table can be updated.

### DIFF
--- a/.changeset/fix-policy-row-id-exists-updates.md
+++ b/.changeset/fix-policy-row-id-exists-updates.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix update permission policies that compare `id` inside `EXISTS` checks.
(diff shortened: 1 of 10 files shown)


gdorsi (approved):
Thanks for the fix.

Do you think that would be doable to add an integration test for this?

Rust, or Browser. Even better if both.

nrainhart:
> Do you think that would be doable to add an integration test for this?
Rust, or Browser. Even better if both.

Added a Rust integration test. I wouldn't add a browser test for this, as the browser architecture shouldn't really affect policy enforcement. We don't have TS integration tests for policies either, and I think that's the right call. Policies are pretty complex and require thorough testing, so I'd prefer to centralize them in Rust.

PR #982, opened 2026-06-01
chore(jazz-tools): enable OTEL in published server versions

Takeno:
Publish jazz-tools with OTEL feature enabled

### DIFF
--- a/.changeset/slimy-lines-nail.md
+++ b/.changeset/slimy-lines-nail.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
-jazz-tools server now has OTEL feature enabled by default
+jazz-tools server now has built-in support for OTEL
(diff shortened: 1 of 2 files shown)


Elfo404 (approved):

.changeset/slimy-lines-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+jazz-tools server now has OTEL feature enabled by default
```
gdorsi: nit: this note says OTEL is enabled by default, but the server still only starts OTEL when `OTEL_EXPORTER_OTLP_ENDPOINT` is set. Users may read this as telemetry sending automatically.

gdorsi (approved):
Approved with one minor release-note wording comment.

.changeset/slimy-lines-nail.md
Takeno: You're right. Fixing

PR #983, opened 2026-06-01
Implement transaction upload protocol

gdorsi:
This PR adds a client-to-server upload path to `mini-jazz-sqlite`.

It also changes locally generated `tx_id` values from `tx-{node_id}-{local_epoch}` strings to UUIDv7 strings. Upload handling still treats `tx_id` as an opaque transaction identifier, so the protocol does not derive behavior from the id format.

Upload row data is now row-image-shaped: inserts and updates carry the effective field values for the row, while deletes carry empty values. The durable upload registry only stores retry metadata; upload data is reconstructed from transaction write metadata and committed history rows.

## Client Transaction Upload Protocol

The unit of sync is one transaction. A transaction can contain multiple row changes across multiple tables, but it is uploaded as one protocol message. Batching is handled by the transport layer, not by stuffing multiple transactions into one upload message.

The client sends transaction data, not history. The receiving server reconstructs and validates authoritative storage/history from its local state, trust context, policies, and any upstream state it needs to fetch. A client cannot forge receipts, rejection state, system fields, or server history.

Updated the Rust example app to use a sync server: https://github.com/garden-co/jazz/pull/984

The client changes are uploaded correctly, but the subscription updates are not propagated currently.
The auth is also not implemented yet (going to be the next thing I'm going to do).

Full spec [here](https://github.com/garden-co/jazz/blob/guido/transaction-upload-proto/docs/superpowers/specs/2026-06-01-mini-jazz-sqlite-client-data-protocol.md)

## Goals

- Let clients send local committed transactions upstream.
- Reconcile unsynced local transactions on startup and reconnect, even with no active subscriptions.
- Keep upload retry state durable on the client.
- Separate message receipt from transaction acceptance.
- Replace unsubscribe-by-`Replay`-difference with an explicit unsubscribe message.

## Wire Messages

Client upload:

```rust
ClientMessage::UploadTx {
    tx: ClientTx,
    data: Vec<ClientDataRecord>,
    reads: Vec<ReadRecord>,
}
```

Server receipt ACK:

```rust
ServerMessage::UploadAck {
    tx_id: String,
}
```

Server transaction fate:

```rust
ServerMessage::TxStatus {
    tx_id: String,
    status: TxStatusKind,
}
```

Unsubscribe:

```rust
ClientMessage::Unsubscribe {
    subscription_id: SubscriptionId,
}
```

Protocol version is bumped to v2 and `ProtocolCapabilities` now includes `tx_upload`.

## Upload Flow

`UploadAck` is flow control only. It means the server received the upload message on this connection and the client can free one in-flight upload slot. It does not mean the transaction was accepted, persisted as authoritative history, edge-filed, globally accepted, or rejected.

Transaction outcome is reported separately with `TxStatus`.

```text
client durable upload queue                     server
---------------------------                     ------
active txs ordered by
(created_at, sync_seq)

[tx A] [tx B] [tx C]
   |      |      |
   |      |      +---- UploadTx(C) -----------> receive message
   |      +----------- UploadTx(B) -----------> receive message
   +------------------ UploadTx(A) -----------> receive message

<---------------------- UploadAck(A) --------- frees one in-flight slot
<---------------------- UploadAck(B) --------- frees one in-flight slot

...later, when fate is known...

<---------------------- TxStatus(A, EdgeAccepted | GlobalAccepted | Rejected)
                         completes or keeps the retry entry according to tx mode
```

The default max in-flight upload count is `1000`. The client sends transactions in order, but it does not wait for `tx A` to be acked before sending `tx B`. WebSocket and `postMessage` already provide ordered delivery for the intended transports.

On reconnect, connection-local in-flight state is cleared. The client scans the durable queue again and replays active transactions that have not completed.

## Transaction Shape

`ClientTx` carries the transaction header:

```rust
struct ClientTx {
    tx_id: String,
    branch_id: Option<String>,
    conflict_mode: TxConflictMode,
    created_at: i64,
    author: Option<String>,
}
```

`branch_id = Some(id)` applies to that branch. `branch_id = None` uses the authenticated session/default branch.

For untrusted client connections, the server derives the author/provenance from the authenticated session and ignores forged client authors. For trusted peer/server connections, the peer may be authoritative for author/provenance, subject to the connection role.

`ClientDataRecord` carries the actual transaction data:

```rust
struct ClientDataRecord {
    table: String,
    row_id: String,
    op: DataOp,
    values: BTreeMap<String, JsonValue>,
}
```

`values` is row-image-shaped:

- `Insert`: effective fields for the new row.
- `Update`: effective fields after the update.
- `Delete`: empty values.

System fields such as `j_created_at`, `j_updated_at`, `j_created_by`, and `j_updated_by` are not accepted from client data. The server derives them.

One upload message must contain exactly one transaction. That transaction may contain many row changes.

## Reads And Conflict Mode

Mergeable transactions may send `reads: []`.

Exclusive transactions need the read facts required for validation. Missing or stale reads reject the transaction rather than closing the connection.

This PR keeps the existing `ReadRecord` shape; read-set redesign is intentionally outside this change.

## Server Handling

The server receives upload messages in transport order and processes them in that order.

```text
UploadTx
   |
   v
handshake + auth checks
   |
   +-- failed envelope/auth check ----> protocol error / close
   |
   v
send UploadAck(tx_id)
   |
   v
validate shape, branch, tx id, reads, policy, current rows
   |
   +-- invalid or denied ------------> TxStatus(tx_id, Rejected)
   |
   v
expand client row data into authoritative local storage/history
   |
   v
record edge/global/rejection fate when known
   |
   v
TxStatus(tx_id, EdgeAccepted | GlobalAccepted | Rejected)
```

Duplicate upload of an already known transaction is idempotent. The server still sends `UploadAck`, and if the transaction already has terminal/known fate, it can send the current `TxStatus` so the client can finish reconciliation quickly.

If the server lacks authoritative state needed to validate an update/delete, it should fetch or wait for trusted upstream state instead of trusting client-provided history. Missing local state is not automatically proof that the transaction is invalid.

## Client Upload Queue

Every normal committed local transaction is inserted into a durable upload registry atomically with the transaction commit. Registry insertion failure fails the commit.

Registry schema:

```sql
CREATE TABLE IF NOT EXISTS jazz_tx_upload_queue (
  sync_seq INTEGER PRIMARY KEY AUTOINCREMENT,
  tx_num INTEGER NOT NULL UNIQUE,
  status INTEGER NOT NULL,
  created_at INTEGER NOT NULL,
  branch_id TEXT,
  author TEXT,
  completed_at INTEGER,
  last_upload_attempt_at INTEGER,
  last_ack_at INTEGER,
  attempt_count INTEGER NOT NULL DEFAULT 0
);

CREATE INDEX IF NOT EXISTS jazz_tx_upload_queue_active_idx
ON jazz_tx_upload_queue(status, created_at, sync_seq)
WHERE status = 1;
```

`jazz_tx_upload_queue` owns durable retry/completion state. Upload data is reconstructed from `jazz_tx_write` and the committed history rows for the transaction.

Queue rows are ordered by:

```sql
ORDER BY created_at, sync_seq
```

`created_at` is the transaction timestamp. `sync_seq` is a local monotonic tie-breaker and is never sent over the wire.

`TxStatus` is non-cumulative:

- `EdgeAccepted` means edge acceptance happened.
- `GlobalAccepted` means global acceptance happened and also satisfies edge-level waiting/retry needs.
- `Rejected` is terminal for upload retry.

Completion rules:

```text
mergeable tx:
  complete when local fate has edge receipt OR global receipt OR rejection

exclusive tx:
  complete when local fate has global receipt OR rejection
```

For registry/retry purposes, an edge receipt is enough to complete mergeable uploads. Exclusive uploads keep waiting/replaying until global acceptance or rejection. Unknown or downgrade-like status messages are ignored.

The queue is never completed by `UploadAck`; it is completed only from authoritative local transaction fate.

## Cleanup

Cleanup deletes completed upload registry metadata only. It never deletes transaction records, row data, history, receipts, or rejection detail.

Active rows are never deleted by cleanup, regardless of age.

MVP cleanup policy:

```text
retention_age = 7 days
max_completed_rows = 10_000
delete_batch_size = 500
min_cleanup_interval = 60 seconds
```

## Unsubscribe

`ClientMessage::Unsubscribe { subscription_id }` removes exactly that subscription. It also drops pending download messages and last acknowledged cursor state for that subscription.

It does not touch upload queue state and does not rely on replaying the remaining subscription set.

## Validation

- `cargo test -p mini-jazz-sqlite --test whole_system`
- `cargo test -p mini-jazz-sqlite`
- `cargo fmt --check -p mini-jazz-sqlite`
- `cargo clippy -p mini-jazz-sqlite -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/mini-jazz-sqlite/src/runtime.rs
+++ b/crates/mini-jazz-sqlite/src/runtime.rs
@@ -1,3 +1,4 @@
+use crate::protocol::{ClientDataRecord, ClientTx, DataOp, TxConflictMode, TxStatusKind};
 use crate::query_api::{predicate_query, BuiltQuery, QueryCondition, QueryConditionOp};
 use crate::read_visibility::ReadVisibility;
 use crate::rows::{ensure_row_id, ensure_row_id_with_status, public_row_id, row_num};
@@ -29,6 +30,13 @@ pub struct Runtime {
     auth: RuntimeAuth,
     node_num: i64,
     branch_num: i64,
+}
+
+#[derive(Clone, Debug)]
+pub struct UploadQueueEntry {
+    pub tx: ClientTx,
+    pub data: Vec<ClientDataRecord>,
+    pub reads: Vec<ReadRecord>,
 }
 
 #[cfg(not(target_arch = "wasm32"))]
@@ -129,6 +137,33 @@ impl QueryScopeOptions<'_> {
 }
 
 pub const ADMIN_SYSTEM_USER: &str = "@system/admin";
+const UPLOAD_STATUS_ACTIVE: i64 = 1;
+const UPLOAD_STATUS_COMPLETED: i64 = 2;
+
+#[derive(Clone, Copy)]
+enum UploadCompletion {
+    EdgeAccepted,
+    GlobalAccepted,
+    Rejected,
+}
+
+struct UploadCompletionFate {
+    tx_num: i64,
+    conflict_mode: i64,
+    rejected: bool,
+    has_edge_receipt: bool,
+    has_global_receipt: bool,
+}
+
+impl From<&crate::protocol::TxStatusKind> for UploadCompletion {
+    fn from(status: &crate::protocol::TxStatusKind) -> Self {
+        match status {
+            crate::protocol::TxStatusKind::EdgeAccepted => Self::EdgeAccepted,
+            crate::protocol::TxStatusKind::GlobalAccepted { .. } => Self::GlobalAccepted,
+            crate::protocol::TxStatusKind::Rejected { .. } => Self::Rejected,
+        }
+    }
+}
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum RuntimeAuth {
@@ -365,6 +400,12 @@ impl Runtime {
         let table = self.schema.table_def(table_name)?.clone();
         let user = self.attribution_user().to_owned();
         let bypass_policy = self.bypasses_policy();
+        let upload_records = vec![ClientDataRecord {
+            table: table_name.to_owned(),
+            row_id: id.to_owned(),
+            op: data_op_from_storage(op)?,
+            values: values.clone(),
+        }];
         let db = self.conn.transaction()?;
         let now = now_ms();
         let (tx_num, tx_id) = tx::create_tx(&db, self.node_num, &self.node_id, now)?;
@@ -390,6 +431,17 @@ impl Runtime {
                     crate::schema::current_table(&table.name)
                 ),
                 params![row_num, self.branch_num, tx_num],
+            )?;
+        }
+        if allowed {
+            let branch_id = branch_id_for_num(&db, self.branch_num)?;
+            enqueue_upload_tx(
+                &db,
+                tx_num,
+                now,
+                Some(&branch_id),
+                Some(&user),
+                &upload_records,
             )?;
         }
         db.commit()?;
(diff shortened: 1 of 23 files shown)


crates/mini-jazz-sqlite/src/runtime.rs
```diff
@@ -5273,6 +5753,270 @@ fn record_tx_write_num(
...
+            if let Some(row_num) = existing_row_num {
+                if row_has_current_branch_value(conn, &record.table, row_num, branch_num)? {
+                    return Err(crate::Error::new(format!(
+                        "row id {} already exists in table {}",
+                        record.row_id, record.table
+                    )));
+                }
+            }
+            existing_row_num.unwrap_or(0)
+        } else {
```
gdorsi, line 5959: Should we also keep track of the server ingestion time?

Could be useful to offer conflict resolution based on authoritative timestamps

PR #985, opened 2026-06-01
feat(jazz-sqlite): align permissions with mainline Jazz

nrainhart:
## Description

Added a policy implementation that aligns more closely with Jazz's, and used it to add the concept of groups to the todo example.

Things I added:
- separate insert/update/delete policies (instead of a single write policy)
- separate using & with check policies
- some of the policies supported by Jazz: Cmp, And/Or/Not, Exists, Inherits

## Takeaways

Even in large tables (I tested a table with 1M todos) with recursive groups, queries completed in <10ms. There's probably room for improvement, but for a super naive implementation I'm pleasantly surprised with the results.

## Discussion

Permissions are only partially implemented. I focused on the policies that allowed testing recursive groups. I don't think we need to focus on until we decide whether we want to cleanup and "productize" the current POC or start from scratch.

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph_nodes/policy_filter.rs
+++ b/crates/jazz-tools/src/query_manager/graph_nodes/policy_filter.rs
@@ -12,7 +12,7 @@ use crate::query_manager::graph_nodes::policy_eval::{
     PolicyContextEvaluator, collect_policy_dependency_tables,
 };
 use crate::query_manager::policy::{
-    Operation, PolicyExpr, evaluate_expr_recursive, normalize_recursive_max_depth,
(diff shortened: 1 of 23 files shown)


crates/jazz-tools/src/query_manager/graph_nodes/policy_filter.rs
nrainhart, line 1: This is cherry-picked from https://github.com/garden-co/jazz/pull/980

crates/jazz-tools/src/query_manager/policy.rs
nrainhart, line 1: Same thing here

PR #986, opened 2026-06-02
feat(svelte, vue): reactive session handles for getSession and useSession

joeinnes:
## Summary
- `getSession()` (Svelte) now returns `{ current: Session | null }` — a getter handle that always reads from the underlying `$state`, so session changes propagate to any consumer holding the handle without destroying the provider.
- `useSession()` (Vue) now returns `ComputedRef<Session | null>` derived from the injected `clientRef` shallowRef, so `triggerRef()` calls from `onAuthChanged` update all template and computed consumers automatically.

## Test plan
- [ ] Run `pnpm exec vitest run --config vitest.config.svelte.ts` — all Svelte tests pass
- [ ] Run `pnpm exec vitest run --config vitest.config.ts src/vue/` — all Vue tests pass

### DIFF
--- a/packages/jazz-tools/src/vue/provider.test.ts
+++ b/packages/jazz-tools/src/vue/provider.test.ts
@@ -1,0 +1,95 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { isRef, shallowRef, triggerRef } from "vue";
+import type { Session } from "../runtime/context.js";
+
+// Override inject so tests can control what the provider key resolves to without
+// mounting a real Vue component tree.  computed/shallowRef/triggerRef are kept real.
+const mockClientRef = shallowRef<any>(null);
+
+vi.mock("vue", async (importOriginal) => {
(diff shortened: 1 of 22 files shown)


packages/jazz-tools/src/vue/provider.test.ts
```diff
@@ -0,0 +1,95 @@
...
+function makeSession(userId: string): Session {
+  return { user_id: userId, claims: {}, authMode: "local-first" };
+}
+
+describe("vue/useSession", () => {
+  beforeEach(() => {
+    mockClientRef.value = null;
+  });
+
+  it("VU-SES-01: throws when no JazzProvider context is present", () => {
```
nrainhart: Nit: I don't see the point in having these codes on each test, and they're annoying when you move tests around
```suggestion
  it("throws when no JazzProvider context is present", () => {
```

nrainhart (approved):

PR #988, opened 2026-06-03
docs: spec for operator hard delete and compliance erasure

joeinnes:
## Summary
- Adds a design spec (`specs/todo/projects/hard-delete/spec.md`) for exposing the engine's existing — built, tested, but unexposed — hard-delete primitive as an **operator-only** erasure capability, aimed primarily at compliance ("right to be forgotten") erasure.
- Key decisions: exposed **only on the backend handle** (not client bindings); erasure spans all schema-versions and user-branches **within one env** (local cross-schema erasure + server-side cross-prefix fan-out); honest-client propagation via **reconcile-on-settle** (a per-row pull — the single new protocol message); transaction support with a fail-loud conflict rule; `includeDeleted` on the backend query path for subject-erasure and a retention-GC job; audit left to the application.
- Adds a deferred-idea note (`specs/todo/ideas/3_later/reconcile-on-reconnect.md`) for narrowing the offline-settle residual later.

This is a design spec opened as a draft for comment — no code changes.

### DIFF
--- a/specs/todo/ideas/3_later/reconcile-on-reconnect.md
+++ b/specs/todo/ideas/3_later/reconcile-on-reconnect.md
@@ -1,0 +1,22 @@
+# Reconcile-on-Reconnect for Hard-Delete Erasure
+
+## What
+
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Closing this historical old-engine design branch. Operator erasure remains an acknowledged product/design topic, but it needs a fresh specification against the new core rather than a port of this protocol plan.

PR #989, opened 2026-06-03
fix: useAll / QuerySubscription correctness and parity across bindings and the orchestrator

joeinnes:
## Summary

Fixes a cluster of `useAll` / `QuerySubscription` correctness and parity bugs across the React, Svelte and Vue bindings and the shared subscriptions orchestrator discovered while preparing hydrated snapshots for SSR. Reads are render-safe, a session change drops the previous session's rows, and JWT refresh is deduped and timeout-guarded.

## Changes by area

**Orchestrator**
- New render-safe `computeKey()` / `peekState()` reads, backing the React rewrite.
- Subscription-setup failures no longer surface as unhandled promise rejections for callback-only consumers.
- On a session change, a settled entry resets to `pending` and listeners are told to clear, so a previous session's rows are dropped on logout/login rather than served until the new subscription's first delta.

**React**
- `useAll` rewritten on `useSyncExternalStore` with a `getServerSnapshot` path. Reads are render-safe (no subscription opened during render); an inline `app.todos.where(...)` no longer does render-phase work; SSR reads a seeded snapshot synchronously without a layout-effect warning; a pending suspense query suspends on its real entry promise (opened during render) rather than one a suspended effect could never resolve, and a not-yet-supplied query suspends so the boundary shows its fallback.
- JWT refresh is deduped at the client level — a second provider or a remount can't double-fire it — and the refresh latch times out, so a hung `onJWTExpired` can no longer wedge auth and silently drop every later expiry.

**Svelte**
- `QuerySubscription` returns its unsubscribe directly from the effect (no shared mutable field to drop), drops `onDestroy` so it works inside `$effect.root` / `.svelte.ts`, and always starts `current` as `undefined`.

**Vue — BREAKING**
- `useAll` now returns `{ data, error, loading }` refs instead of a bare `Ref<T[] | undefined>`, so a failed query is distinguishable from loading or empty. Adds a Suspense-compatible `useAllSuspense`.
- Migration: `const todos = useAll(...)` → `const { data: todos } = useAll(...)`. All in-repo Vue consumers, docs and examples are migrated in this PR.

**Docs**
- The `subscribeAll` JSDoc example uses `change.item` and the now-exported `RowChangeKind.Added` (the old `change.row` / `change.kind === 0` yielded `undefined`); `SubscriptionDelta.all`'s freshly-allocated-per-delta contract is documented with a pointer to `applyDelta` / `reconcileArray`.

## Test plan

- [ ] `pnpm --filter jazz-tools test` — node/React/Svelte suites green (1193 + 19 + 46; 1 expected-fail xfail).
- [ ] `pnpm --filter jazz-tools test:browser` — jazz-tools browser suite green.
- [ ] `pnpm --filter world-tour test` and `pnpm --filter todo-client-localfirst-vue test` — Vue example browser tests green.
- [ ] Typecheck clean: `tsc --noEmit` against `tsconfig.json` and `tsconfig.tests.json`.

Changeset: `.changeset/useall-querysubscription-correctness.md` (`jazz-tools` patch).


docs/content/docs/reading/queries.mdx
```diff
@@ -139,15 +139,15 @@ magic columns.
...
 Each framework has a reactive binding that re-renders when query results change. See [Framework Patterns](/docs/reference/framework-patterns#query-subscriptions) for side-by-side examples.
 
-| Framework  | API                            | Notes                       |
-| ---------- | ------------------------------ | --------------------------- |
-| React/Expo | `useAll(query)`                | Returns `T[] \| undefined`  |
-| Vue        | `useAll(query)`                | Returns a `Ref`             |
-| Svelte     | `new QuerySubscription(query)` | Exposes reactive `.current` |
+| Framework  | API                            | Notes                                               |
+| ---------- | ------------------------------ | --------------------------------------------------- |
+| React/Expo | `useAll(query)`                | Returns `T[] \| undefined`                          |
```
nrainhart, line 144: What do you think about migrating the React API to `{ data, error, loading }` as well? (not necessarily as part of this PR, but as a fup)

nrainhart: Now that I think about it... we use `WriteResult.value` to return inserted values. If we're doing breaking changes, maybe we should align on `WriteResult.data`?

packages/jazz-tools/src/vue/use-all.ts
```diff
@@ -43,31 +78,65 @@ function subscribeToEntry<T extends { id: string }>(
...
+ *
+ * @param query - the database query (e.g. `app.todos.where({ done: false })`)
+ * @param options - optional query execution options
+ *
+ * @returns the same reactive `{ data, error, loading }` as {@link useAll}
+ */
+export async function useAllSuspense<T extends { id: string }>(
+  query: QueryBuilder<T>,
+  options?: QueryOptions,
+): Promise<UseAllResult<T>> {
```
nrainhart: I'd omit the `loading` field from the suspense API, since it'll always be true anyway

packages/jazz-tools/src/react-core/use-all.ts
```diff
@@ -16,58 +20,88 @@ function useAllBase<T extends { id: string }>(
...
-    }
-    return undefined;
-  }
+  const getSnapshot = useCallback(
+    (): UseAllState<T> | null => (key === null ? null : manager.peekState<T>(key)),
+    [manager, key],
+  );
 
-  const state = entry.state;
+  const state = useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
```
nrainhart, line 62: @gdorsi I seem to recall you saying `useSyncExternalStore` caused issues with things like transitions. Am I right or totally off-base?

gdorsi: Yes, it disables concurrent UI features.

But keeping concurrent UI support and correctness is very hard, so it's ok to opt-out for them.

docs/content/docs/reading/queries.mdx
joeinnes: Agree on both points :)

packages/jazz-tools/src/vue/use-all.ts
joeinnes: Makes sense

nrainhart (approved):
Thanks for all the fixes! Regarding the use of `useSyncExternalStorage`, Guido's comment makes sense: let's focus on correctness for now, and then we'll worry about concurrent ui

PR #990, opened 2026-06-03
Add on-device integration test runner for React Native (Maestro)

antoniomuso:
## What

Adds a standalone Expo app — `dev/integration-tests-expo` — that is itself an on-device integration-test runner for the React Native client (`jazz-rn`), plus a dedicated Maestro flow and CI workflow.

Today CI exercises RN through a single happy-path Maestro walkthrough of the todo example. The browser/JS client has a rich integration suite (bulk writes, queries, subscriptions, relations, convergence) that RN has no equivalent of. This fills that gap, and gives us a place to host RN regression scenarios going forward.

## How it works

- On launch the app runs every suite in `tests/_registry.ts` sequentially and renders each test as a row (`pending → running → PASS/FAIL`).
- Each test gets a fresh, **local-only** `Db` (`createDb({ appId })`, no server) and shuts it down afterwards. Default-allow permissions mean no server or policy setup is needed.
- When every test passes, a `suite-passed` element renders. The Maestro flow waits for it; a red test **or a freeze** means it never appears, the wait times out, and the flow fails. The failure screenshot shows which test died.

## Covered (Phase 1, local-only)

bulk writes (100 rows), CRUD round-trip, query conditions, order + limit, `subscribeAll` add/update deltas, forward-ref `include`, local-tier `wait`. Adding a test = one `tests/*.suite.ts` file + one line in `_registry.ts`. Assertions use a small Hermes-safe Jest-style `expect` shim (`runner/expect.ts`).

image

## CI

`expo-android-integration-e2e.yml` builds a release APK and runs the flow on Maestro Cloud (Android). It mirrors the existing todo e2e workflow but drops the Vercel-sandbox / jazz-tools-server / deploy steps — Phase 1 is local-only, so there are no secrets beyond the Maestro key, which also avoids the known WS-upgrade flakiness.

## Verification

- `tsc --noEmit` clean; runner unit tests 16/16 (Node); oxlint clean.
- The actual suite bodies were run through the runner against the Node/WASM `createDb` (same Rust core + `Db` API as native) — all green — to validate the DSL semantics before the on-device run.
- The APK build + on-device Hermes run + Maestro gate are validated by the new workflow.

## Notes

- The runner (`runner/`) is decoupled from the RN registry so its pure-TS logic is Node-unit-testable.
- Lives under `dev/` (not `examples/`) because it is internal QA tooling, not a user-facing showcase.
- Design + plan: `specs/rn-integration-test-runner-{design,plan}.md`.
- **Phase 2 (not in this PR):** multi-peer sync tests (server-backed or an in-process relay) — the registry/ctx are designed to accommodate a second `Db`.

### DIFF
--- a/.changeset/expo-crypto-random-values.md
+++ b/.changeset/expo-crypto-random-values.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+`jazz-tools/expo/polyfills` now installs `globalThis.crypto.getRandomValues` from `expo-crypto` when React Native/Hermes does not provide `globalThis.crypto`, fixing anonymous/local-first `createDb({ appId })` startup in Expo apps.
--- a/.changeset/rn-batched-tick-macrotask.md
+++ b/.changeset/rn-batched-tick-macrotask.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Schedule React Native `batchedTick` callbacks on a macrotask so repeated native tick requests do not starve timers, React rendering, or runner timeouts.
--- a/.github/workflows/expo-android-integration-e2e.yml
+++ b/.github/workflows/expo-android-integration-e2e.yml
@@ -1,0 +1,133 @@
+name: Expo Android Integration Tests (Maestro Cloud)
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - changeset-release/*
+  pull_request:
+    branches: [main]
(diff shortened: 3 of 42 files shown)


gdorsi (approved):

aeplay:
🤖 Superseded by the current React Native direction in #1832 and roadmap issue #1756. The replacement deliberately uses native/Expo validation without retaining Maestro as a separate vendor runner.

PR #992, opened 2026-06-04
fix(server): bound upstream sync retries with connect deadlines

Takeno:
## Summary

Fix edge upstream reconnect behavior when the upstream Jazz server is slow to become available, such as after scaling from zero.

The existing transport retry loop only retried after the current WebSocket connect or auth handshake attempt returned. In production, those attempts could hang on network/LB timeouts before Jazz got a chance to apply its backoff, causing long delays before the edge connected to the core.

This change adds optional per-attempt transport deadlines and enables them for edge upstream sync.

## Changes

- Add `TransportRetryConfig` with optional timeouts for:
  - WebSocket connect attempts
  - Jazz auth handshake attempts
- Add `connect_with_retry_config`/`install_transport_with_retry_config` paths while preserving existing default behavior.
- Configure edge upstream sync with 5s connect and 5s auth-handshake deadlines.
- Keep auth failures unchanged: unauthorized handshakes still suspend retries until auth is updated.
- Add regression tests for pending connect and pending auth handshake attempts retrying after timeout.

## Testing

- `cargo test -p jazz-tools --features test transport_manager::tests::`
- `cargo test -p jazz-tools --features test edge_connects_to_core_with_admin_secret_only`
- `cargo test -p jazz-tools --features test builder_allows_edge_mode_with_admin_secret_only`

### DIFF
--- a/crates/jazz-tools/src/transport_manager.rs
+++ b/crates/jazz-tools/src/transport_manager.rs
@@ -93,6 +93,10 @@ pub struct TransportHandle {
     pub(crate) declared_schema_hash: std::sync::Arc<std::sync::Mutex<Option<String>>>,
 }
 
+/// Per-attempt retry deadlines for native Tokio transports.
+///
+/// Non-Tokio transports currently ignore these fields and keep their existing
+/// connection and authentication timing behavior.
 #[derive(Clone, Copy, Debug, Default)]
 pub struct TransportRetryConfig {
     pub connect_attempt_timeout: Option<Duration>,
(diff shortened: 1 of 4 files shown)


crates/jazz-tools/src/transport_manager.rs
```diff
@@ -92,6 +93,12 @@ pub struct TransportHandle {
     pub(crate) declared_schema_hash: std::sync::Arc<std::sync::Mutex<Option<String>>>,
 }
 
+#[derive(Clone, Copy, Debug, Default)]
+pub struct TransportRetryConfig {
+    pub connect_attempt_timeout: Option<Duration>,
```
gdorsi, line 102: This setting is handled only by Tokyo runtimes.

I guess that’s intentional, but should be flagged here because it’s not immediately recognizable from here.

gdorsi (approved):
Approved with one inline note.

PR #995, opened 2026-06-04
feat: SSR with data hydration

joeinnes:
## SSR with data hydration

Server-render Jazz queries and hydrate them on the client, so the first paint shows real rows instead of a loading state — then hand over to live sync with no blank frame in between.

The seam is the **hook, not the provider**: prefetch a query on the server (in a Server Component or a route loader) and pass its snapshot to the hook that reads it. Hydration is co-located with the query and composable, rather than every app query hoisted to provider level.

```tsx
// server — prefetch against a backend Db, dehydrate to an opaque snapshot
builder.prefetch(db, app.todos.where({ done: false })); // db = context.asBackend() / forSession()
const snapshot = builder.dehydrate();

// client — the same query, seeded for the first paint
<JazzProvider ssr={!!snapshot}>
const todos = useAll(app.todos.where({ done: false }), { snapshot });
```

### How it works

- **Two-form snapshot envelope.** `createSnapshotBuilder(...).dehydrate()` produces an opaque `DehydratedSnapshot` carrying both the rendered rows (for a synchronous first paint) and the server's own versioned CRDT **sync bundle**. Jazz does no serialisation of its own — it's plain JSON the host framework serialises.
- **Single-orchestrator attach.** The provider's `ssr` mode renders synchronously from a db-less seed orchestrator that each hook's snapshot seeds. When the live client connects it attaches to that *same* orchestrator — draining the bundle into the store and re-subscribing — instead of swapping a seed store for a live one. Seeded rows persist through attach and reconcile without churn, so there's no flash.
- **Hook-level seed.** `useAll(query, { snapshot })` and Svelte's `new QuerySubscription(query, { snapshot })` apply the snapshot once and read the seeded entry synchronously (Svelte does it in the constructor, before any effect, so the SSR HTML already has the rows). `useAllSuspense` takes the same option and renders the seeded rows synchronously instead of suspending — which it can't usefully do during SSR anyway. The snapshot never affects the query key.
- **Schema-fingerprint guard.** The client computes the fingerprint from its own query's schema and discards a snapshot built against a divergent schema — the client's code is the source of truth, never the server's.
- **Principal scoping.** The snapshot's principal is derived from the `Db` you prefetch with (`asBackend()` → public/`null`; `forSession`/`forRequest` → that user). Scoping is the server's job; on the client, seeding a user-stamped snapshot into a different live session throws rather than render another user's rows, as a backstop.
- **Tier stays the user's choice.** SSR doesn't force a global/edge tier; the query runs with whatever tier the caller threads.

### React + Svelte

Both `JazzProvider` and `JazzSvelteProvider` gain an `ssr` mode (the Svelte provider's `client` is now optional — the server has none). A suspense hook that renders during the seed phase degrades to an empty result instead of suspending forever (the server can't resolve it, and suspending on the client would stall the connect); errors still surface. Svelte's `QuerySubscription` keeps its options bag reactive and treats `snapshot` as one-shot (a later, different snapshot is ignored, with a development warning).

### Examples & docs

- `examples/nextjs-csr-ssr` and `examples/sveltekit-csr-ssr` — mirrored server-prefetch + client-hydrate panels, each with an e2e round-trip through a shared backend.
- Guide: **SSR with data hydration** — route-level prefetch, the security/scoping boundary, and caching caveats; framework tabs with tested `<include>` snippets.

### Test plan

- Red/green TDD throughout — the snapshot envelope and orchestrator bundle queue, the seed-phase suspense latch, the schema-fingerprint guard, and the flash-on-attach fix (a seeded entry must not reset to `pending` when the live db attaches).
- React, Svelte and orchestrator suites green; `cargo check` (WASM + NAPI) clean; the Rust sync-bundle tests cover cold seed, reconcile-without-churn, permission filtering, wire round-trip and version-gate rejection. CI green on the head commit.

> The earlier "bundled runtime fix" (multi-client teardown OOB) has since landed on `main` and dropped out of this branch on rebase, so this PR is now just the SSR feature.


aeplay:
🤖 Closing this pre-new-core implementation rather than attempting a restack. #989 retained render-safe seeded-query primitives, and #1909 established the current SSR-safe provider lifecycle, but this branch public dehydration API and Rust sync-bundle transport no longer exist in the current architecture. The current nextjs-csr-ssr example remains the supported Server Component path. Fresh design work is tracked in #2059, including principal and schema isolation plus first-paint and reconnect receipts.

PR #997, opened 2026-06-04
refactor: remove unused methods from `Runtime` JS API

nrainhart:
## Description

I think the `Runtime` API exposed to JS captures fairly well what a future public Rust API should expose. While reviewing it I found several methods that are no longer used, so I removed them. I also unified insert/update/delete/restore with their `*WithSession` counterparts.


gdorsi (approved):
Approved.

PR #998, opened 2026-06-05
Shared worker broker leader election

antoniomuso:
## Summary

This PR replaces the BroadcastChannel-based tab leader election with a **SharedWorker broker**: one broker per `{appId, dbName}` coordination domain that owns cross-tab membership and leader election, while sync data flows over dedicated `MessagePort`s straight between follower runtimes and the single leader-owned OPFS worker.

Exactly one tab (the leader) owns the dedicated persistence worker. Every other tab (followers) runs only a main-thread runtime and reads/writes durably through the leader's worker — no fallback workers, no per-tab OPFS namespaces, no broadcast fan-out. The public `createDb` / framework API is unchanged.

### Topology

```
                 ┌────────────────────────────┐
                 │  SharedWorker broker        │   control plane only:
   hello/elect   │  (jazz-broker-worker.ts)    │   membership, election,
  ┌─────────────►│  leadershipId, fencing,     │   liveness, port assignment,
  │  ┌──────────►│  storage-reset coordination │   storage reset
  │  │           └────────────┬───────────────┘
  │  │                        │ assigns MessageChannel per follower
┌─┴──┴────┐          ┌────────▼─────────┐
│ follower │  data   │ leader tab        │
│ tabs     │◄───────►│  └─ dedicated     │──► OPFS (single namespace)
│ (main-   │  ports  │     worker (wasm) │
│  thread  │         └───────────────────┘
│  runtime)│   postcard envelopes, transferred ArrayBuffers,
└──────────┘   worker posts directly to follower ports
```

The broker never touches sync payloads. It assigns ports and arbitrates leadership; Jazz data flows worker ⇄ follower directly.

### Attach and election

- A tab connects with a `hello` carrying a deterministic **fingerprint** (app/db, persistent driver namespace, env, user branch, server URL, auth class, runtime-source identity, **control-protocol version** `jazz-browser-broker-v2`, storage-format version). Mismatched tabs get an explicit `unsupported` reply at attach — mixed deploy versions fail loud instead of misbehaving silently.
- The broker ranks eligible candidates (visible tabs preferred, schema-fingerprint compatible) and sends `become-leader` with a **monotonically increasing `leadershipId`**.
- The promoted tab acquires the leader **Web Locks**, spawns the dedicated worker (which takes the worker lock as a preflight before opening OPFS), and reports `leader-ready`. The broker then broadcasts the leader to all tabs and starts **lock-release monitors** on the leader's locks — the crash detector.

### Follower data path

- Per follower, the broker creates a `MessageChannel`: one end goes to the leader tab (`attach-follower-port`, forwarded into the Rust worker host with the port transferred), the other to the follower (`use-follower-port`).
- Follower durable startup gates on the leader worker acknowledging the port attach.
- Sync payloads are postcard envelopes with transferred `ArrayBuffer`s; the worker-side outbox routes follower-bound batches **directly to the follower's port** (the leader's main thread is not on the hot path), and ingest on both sides decodes bytes natively in wasm without a JS round-trip.

### Failure handling and fencing

- **Leader crash/close** → the browser releases its Web Locks → the broker's lock monitors fire → leader cleared, replacement election, followers re-attach under the new `leadershipId`. Writes in flight survive via outbox buffering while no sync sender is attached, and batch reconciliation handles a leader that flushed its WAL but never delivered the batch fate.
- **Broker crash** (SharedWorker restart) → tabs detect via ping/pong liveness and reconnect: demote locally, re-hello the new instance, re-report visibility and schema. Every tab→broker message is stamped with the `brokerInstanceId` it was composed for; a fresh broker drops (and warns once about) messages addressed to a dead instance, so nothing composed pre-restart can corrupt the new instance's state.
- **Leadership is only ever granted via `become-leader`.** A leader tab that re-hellos a live broker (spurious liveness reconnect, frozen tab resuming) has its stale leadership cleared and goes through a fresh election instead of being resurrected by a `leader-ready` echo.
- **Stuck demotion** → forced takeover steals the Web Locks after a timeout.
- **Zombie workers are impossible at the terminal layer**: the dedicated worker refuses to open OPFS without holding the worker lock, and the Rust worker host rejects follower-port attaches fenced to a different `leadershipId`.
- **Attach failures fail fast**: a SharedWorker `error` event (e.g. script URL mismatch during a deploy) rejects the connection with the real reason instead of a generic hello timeout.

### Storage reset

`wipeData` is coordinated by the broker in the same control plane: leadership is paused, every tab prepares, one tab performs the reset, and finished outcomes are redelivered to tabs that rejoin mid-reset so nobody waits out a timeout.

## Advantages over the old leader election

The old stack was three BroadcastChannel protocols glued together: `tab-leader-election.ts` (lock-lease probing + heartbeats + term fencing), `tab-sync-protocol.ts` (follower/leader sync messages broadcast to every tab), and `storage-reset-coordinator.ts` — plus per-follower fallback workers writing to private `${dbName}__fallback__${tabId}` OPFS namespaces.

| | Old (BroadcastChannel) | New (SharedWorker broker) |
|---|---|---|
| **Election authority** | Each tab probes/heartbeats and converges by gossip; lease timeouts race re-acquisition | One broker owns membership and election; single writer of `leadershipId` |
| **Durable storage** | Leader namespace **plus** one fallback OPFS namespace per follower tab, reconciled later | One OPFS namespace, owned by one worker — nothing to reconcile, no duplicated quota |
| **Follower writes** | Into the follower's own fallback worker, merged eventually | Directly into the leader worker over a dedicated port — durable on first ack |
| **Data plane** | Broadcast: every payload structured-cloned to every tab, then filtered by addressee; leader main thread relays worker traffic | Targeted per-follower ports with transferred buffers; worker posts straight to the follower port |
| **Crash detection** | Heartbeat timeouts over BroadcastChannel | Platform-enforced Web Lock release (fires even on hard tab kill) + ping/pong liveness |
| **Fencing depth** | `term` checked in main-thread TS | `leadershipId` enforced in TS **and** in the Rust worker host; worker-lock preflight blocks zombie OPFS writers; broker-instance stamping fences across broker restarts |
| **Version skew** | Undetected — old/new tabs shared channels | Protocol + storage versions embedded in the attach fingerprint; mismatches rejected explicitly |
| **Election quality** | First lock winner | Visibility-ranked (prefers a tab the user can see), schema-compatible candidates only |
| **Storage reset** | Separate broadcast coordinator | Same broker control plane, with outcome redelivery to rejoining tabs |

Net effect: fewer moving protocols (one control plane instead of three broadcast ones), fewer split-brain windows, less storage churn, and a data path with no broadcast amplification and no leader main-thread hop.

## Behavioral changes

- Persistent browser mode is **stricter**: missing `SharedWorker`, `MessageChannel`, or Web Locks now fails with an explicit error instead of silently falling back to the old peer-election/fallback-worker path. Memory-mode browser clients and non-browser runtimes are unchanged.
- New `runtimeSources.brokerWorkerUrl` (and `baseUrl` derivation) lets apps host the broker SharedWorker script explicitly, e.g. on a CDN.
- Removed: `tab-leader-election.ts`, `tab-sync-protocol.ts`, `storage-reset-coordinator.ts`, and all `__fallback__` namespace usage.

## Testing

- **Browser (Playwright/Chromium)**: broker attach/fingerprint rejection, election and monotonic leadership ids, single-worker ownership across tabs, follower startup without fallback workers, follower reads/writes through the leader worker, leader/follower failover (including in-flight follower writes and local-wait reconciliation after a bridged leader crash), broker-restart fencing, same-tab leader re-hello re-promotion, storage reset through the broker (including mid-reset rejoins).
- **Unit**: protocol fingerprints/versioning, leader-candidate ranking, Web Locks acquire/monitor/steal, broker-client reconnect semantics (instance-stamped sends, no replay of reconnect-time messages, visibility replay, storage resets deferred across reconnects, error causes preserved), broker worker URL resolution.
- **Rust/wasm**: worker protocol round-trips, follower-port attach/detach and peer routing, outbox buffering with no sync sender.

## Review guide

1. `specs/todo/a_mvp/shared_worker_broker_leader_election.md` — target behavior and non-goals.
2. `packages/jazz-tools/src/runtime/browser-broker-protocol.ts` + `leader-lock.ts` — protocol, fingerprinting, Web Locks primitives.
3. `packages/jazz-tools/src/worker/jazz-broker-worker.ts` + `packages/jazz-tools/src/runtime/browser-broker-client.ts` — broker control plane and tab-side lifecycle.
4. `packages/jazz-tools/src/runtime/db.ts` — persistent browser `createDb` integration (promotion, demotion, follower durable path).
5. `packages/jazz-tools/src/runtime/worker-bridge.ts`, `packages/jazz-tools/src/worker/jazz-worker.ts`, `crates/jazz-wasm/*` — follower data-port path into the leader worker.
6. Browser/wasm test suites for the behavioral contract.


gdorsi (commented):
> A tab connects with a hello carrying a deterministic fingerprint (app/db, persistent driver namespace, env, user branch, server URL, auth class, runtime-source identity, control-protocol version jazz-browser-broker-v2, storage-format version). 

Why does this needs to be this complicated?
Can't we inline the Jazz version and check that?

> Mismatched tabs get an explicit unsupported reply at attach — mixed deploy versions fail loud instead of misbehaving silently

So this means that, if I have multiple tabs open, my app can't upgrade until I close all the tabs?

If yes, this needs to be documented somewhere, along with a way to provide info to the user on how do complete the upgrade (e.g. a callback to trigger a "Please close the other tabs" modal)

docs/superpowers/plans/2026-06-05-shared-worker-broker-leader-election.md
```diff
@@ -0,0 +1,347 @@
+# SharedWorker Broker Leader Election Implementation Plan
```
gdorsi: Do we want to keep this?

packages/jazz-tools/src/runtime/browser-broker-client.ts
```diff
@@ -0,0 +1,754 @@
+import {
```
gdorsi, line 1: Can't we move this stuff to Rust as well?

It would make it possible to:
- Simulate the full browser environment using only Rust (we should also inject the platform deps)
- Build web apps using Rust frameworks more easily (e.g. Yew)

specs/todo/a_mvp/shared_worker_broker_leader_election.md
```diff
@@ -0,0 +1,593 @@
+# SharedWorker Broker Leader Election — TODO (MVP)
```
gdorsi: This also should be removed right?

gdorsi (approved):

.changeset/shared-worker-broker.md
```diff
@@ -0,0 +1,10 @@
...
+"jazz-wasm": patch
+---
+
+Route persistent browser runtimes through a SharedWorker broker so tabs for the same Jazz app share one OPFS-backed leader runtime instead of each opening independent storage handles. The broker coordinates leader promotion, follower message ports, schema compatibility, visibility hints, storage resets, and failover after tab or worker crashes, preserving pending local writes while the durable path reconnects.
+
+**Breaking change — browser support:** persistent browser mode now requires `SharedWorker`, `MessageChannel`, and Web Locks support. Browsers or embedded webviews missing those capabilities will reject `createDb()`/`createJazzClient()` startup for persistent storage instead of using the previous BroadcastChannel tab-election path. Use a supported browser runtime for persistent local storage, or switch to the memory driver with a `serverUrl` in unsupported environments.
+
+This replaces the BroadcastChannel tab-election and fallback-namespace path with broker-managed routing, adds the broker worker to the published `jazz-tools` worker assets, and extends the WASM worker bridge so leader and follower ports can exchange sync batches directly.
```
nrainhart: This part feels too implementation-specific
```suggestion
```

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -803,26 +836,41 @@ export type BatchScope = Scoped<DbDirectBatch>;
...
+  private followerReadyResolved = false;
+  private followerPortReadyLeadershipId: number | null = null;
+  private followerLeaderReadyLeadershipId: number | null = null;
+  private durablePathError: Error | null = null;
+  private brokerSchemaFingerprint: string | null = null;
+  private brokerResetSchema: WasmSchema | null = null;
+  private readonly pendingLeaderFollowerPorts = new Map<
+    string,
+    { followerTabId: string; leadershipId: number; port: MessagePort }
+  >();
```
nrainhart, line 874: The Db facade is already pretty complex. Do you think we could encapsulate the tab election state into a separate class to keep the Db more manageable?

No need to do this in this PR, it could be a fup

nrainhart: Now a Db can _either_ be connected to:
- a worker, if it's the leader
- another tab, if it's a follower

right? Maybe that's one thing we can abstract away, and we avoid redundant calls like:
```ts
this.workerBridge?.updateAuth({
  jwtToken,
});
this.followerPortBridge?.updateAuth({
  jwtToken,
});
```

antoniomuso: Not really, the `Db` is connected to its own dedicated worker when its tab is the leader. If its tab is a follower, the `Db` connects to the current leader’s dedicated worker through a transferred `MessagePort`; the follower tab did not spawn or own that worker.



```text
leader tab Db
  -> WorkerBridge
  -> its own dedicated worker

follower tab Db
  -> MessagePortRuntimeBridge
  -> leader tab's dedicated worker
```

antoniomuso: For the rest, yes. This could be abstracted. And we can try to encapsulate the leader election in a separated class.

specs/todo/a_mvp/shared_worker_broker_leader_election.md
antoniomuso: yes :+1:

packages/jazz-tools/src/runtime/browser-broker-client.ts
antoniomuso: Yeah, I will implement it as a follow up.

docs/superpowers/plans/2026-06-05-shared-worker-broker-leader-election.md
antoniomuso: No, I am going to remove it.

packages/jazz-tools/src/runtime/db.ts
nrainhart: > Not really, the Db is connected to its own dedicated worker when its tab is the leader. If its tab is a follower, the Db connects to the current leader’s dedicated worker through a transferred MessagePort

Sorry, that's what I meant to say. I think my point still stands. Right now we're doing a bunch of "is this connected to its own worker" and "is this connected to another tab's worker" checks, what if instead we do Db->Bridge, where Bridge can either be a worker bridge or a message port bridge?

antoniomuso: I am gonna do it as a follow up.

PR #1000, opened 2026-06-05
refactor: simplify TS JazzClient

nrainhart:
## Description

Removes boilerplate from the TS JazzClient and moves core logic over to the Rust side, with the goal of having a simpler and more principled interface between the Rust core and the TS bindings.

While working on this I also:
- fixed `updateAt` override not working for inserts
- added `includeDeleted()` to TS query builders for reading soft-deleted rows (necessary to test the above fix, but also generally useful to expose)

_Note for reviewers:_ each change is isolated in its own commit, in case you find the whole diff hard to review.


crates/jazz-tools/src/runtime_core/writes.rs
```diff
@@ -9,6 +9,158 @@ use crate::row_histories::{BatchId, RowState, patch_row_batch_state};
...
+                    RuntimeBatchStatus::Committed,
+                ))
+            }
+            Some((_, RuntimeBatchStatus::Active)) | None => Ok(()),
+        }
+    }
+
+    fn finish_batch(&mut self, batch_id: BatchId, status: RuntimeBatchStatus) {
+        if let Some(context) = self.batch_contexts.get_mut(&batch_id) {
+            context.status = status;
```
gdorsi, line 73: Do we have a reason for keeping finished batches contexts in memory, or shall we evict them?

gdorsi (approved):
Nice job!

crates/jazz-tools/src/runtime_core/writes.rs
nrainhart: Yes, it is needed to make stale handle operations fail after `commit_batch` / `rollback_batch`. Are you worried about memory usage?

nrainhart: Merging the PR now, but we can adjust this as a fup if needed

gdorsi: Not a big deal, these are small objects, but we add them for each write right?

Can't we make them fail if the batch_id is missing?

nrainhart: We can, at the expense of making the error less useful (i.e. we can't distinguish between "this batch/tx is already closed" and "this batch was never created"). It sounds like a fair tradeoff, I'll make the change.

nrainhart: https://github.com/garden-co/jazz/pull/1006

PR #1001, opened 2026-06-08
Improve opfs-btree browser storage performance

gdorsi:
Jazz persists browser data in a B-tree stored in OPFS. Profiling the todo stress test (50k items) showed the storage worker CPU-bound in this backend — first on reads, then (after fixing those) on writes. The changes, ordered by impact:

### 1. Durability flushes append to a WAL instead of checkpointing
`flush_wal` used to be a full checkpoint: every dirty page written in place, freelist rebuilt, superblock swapped. It is now a pure append of dirty pages + a commit marker + one `flush()`. Recovery replays the WAL tail on open; checkpointing remains the rare compaction path. Crash safety: the WAL never overwrites the checkpointed region, and torn tails fail their checksums and are ignored.

### 2. Writes no longer thrash the page cache
Every `put` ran a full O(cached-pages) eviction scan — and once dirty/WAL-pinned pages held the cache over budget, the scan re-fired on every call without ever reaching its target, evicting warm pages that the next put re-read from OPFS. Eviction now runs only where pages are actually inserted into the cache, and its trigger accounts for pinned pages: one scan per ~`max/4` insertions instead of per put.

### 3. Tree descents work on raw pages
Lookups *and* writes used to fully decode each internal page (one `Vec` per key, hundreds per page) at every level. Both now binary-search the raw page bytes directly, with bounds validated once per page instead of per probe; writes decode only when a split actually happens. A last-leaf hint lets sequential reads skip the descent entirely.

### 4. Page checksums once per flush, not per mutation
In-place upserts no longer recompute the xxh3 checksum over the full 16 KiB page on every write; checksums refresh once per dirty page at flush/checkpoint.

### 5. Batch writes sorted by key (jazz-tools layer)
Consecutive inserts land in the same or adjacent leaves instead of dirtying random pages across the tree.

Also: WAL replay reads page runs instead of single pages on wasm32 (kept per-page natively — measured faster there), reusable staging buffers in `range`, and `persisted_pages` tracking so reads need no `file.len()` call.

## Numbers

Browser traces from `dev/stress-tests/todo-react`, focused on the dedicated worker CPU profiles. Load captures include two worker profiles, so the rows below combine both instead of picking one.

Load action:

| worker metric | main | optimized | delta |
|---|---:|---:|---:|
| Worker active sampled CPU | 6.183 s | 4.203 s | −32% |
| Time in `opfs_btree` stacks | 2.548 s | 1.261 s | −51% |
| `OpfsBTree::get` stacks | 1.550 s | 0.631 s | −59% |
| `OpfsBTree::range` stacks | 1.107 s | 0.519 s | −53% |
| Page checksum stacks | 0.270 s | 0.025 s | −91% |

Generate action:

| worker metric | main | optimized | delta |
|---|---:|---:|---:|
| Worker active sampled CPU | 8.680 s | 5.302 s | −39% |
| Time in `opfs_btree` stacks | 4.821 s | 2.145 s | −56% |
| `OpfsBTree::put` stacks | 3.147 s | 0.727 s | −77% |
| `decode_page` stacks | 1.082 s | 0.039 s | −96% |
| Page checksum stacks | 0.512 s | 0.076 s | −85% |


crates/jazz-tools/src/storage/opfs_btree/mod.rs
```diff
@@ -266,21 +278,38 @@ impl Storage for OpfsBTreeStorage {
...
+
+            fn flush_pending_puts(
+                tree: &mut OpfsBTree<AnyFile>,
+                pending_puts: &mut Vec<(String, &[u8])>,
+            ) -> Result<(), StorageError> {
+                if pending_puts.is_empty() {
+                    return Ok(());
+                }
+
+                pending_puts.sort_by(|left, right| left.0.cmp(&right.0));
```
gdorsi, line 291: Sorting a batch before inserting is one of the bigger wins here.

Random-order inserts descend a different tree path each time, dirtying unrelated leaves and triggering more splits, while sorted inserts land in the same or adjacent leaves so the descent path stays cached and far fewer pages are dirtied. 

A `Delete` flushes the pending batch first, so apply order is preserved wherever it could matter.

crates/opfs-btree/src/db.rs
```diff
@@ -354,7 +375,8 @@ impl<F: SyncFile> OpfsBTree<F> {
             total_pages = self.total_pages
         )
         .entered();
-        let mut dirty_page_ids: Vec<PageId> = self.dirty_pages.iter().copied().collect();
+        let mut dirty_page_ids: Vec<PageId> =
+            self.dirty_pages.union(&self.wal_pages).copied().collect();
```
gdorsi, line 380: Checkpoint is now the only path that rewrites home locations: it writes both dirty and WAL-resident pages to their home offsets, swaps superblocks, then truncates the WAL tail off the file. 

The union here is what guarantees pages living only in the WAL make it home before the tail is dropped.

crates/opfs-btree/src/db.rs
```diff
@@ -93,14 +102,21 @@ pub struct OpfsBTree<F: SyncFile> {
     active: Superblock,
     root_page_id: Option<PageId>,
     total_pages: u64,
+    // Pages >= active.total_pages are WAL tail pages, not home locations.
+    // Page ids in wal_pages have newer bytes only in the cache/WAL and must
+    // not be evicted until a checkpoint writes them to their home locations.
+    persisted_pages: u64,
```
gdorsi, line 109: `persisted_pages` tracks how much of the file is actually on disk. 

It serves three roles: 
1. tells the WAL where to start appending 
2. bounds page reads without a `file.len()` call per read
3. and detects when the file needs extending before a checkpoint.

crates/opfs-btree/src/db.rs
```diff
@@ -1469,6 +1719,7 @@ impl<F: SyncFile> OpfsBTree<F> {
             if Some(*page_id) == protected_page
                 || Some(*page_id) == root_page_id
                 || self.dirty_pages.contains(page_id)
+                || self.wal_pages.contains(page_id)
```
gdorsi, line 1732: Safety-critical line: a page whose newest bytes exist only in the cache/WAL must not be evicted, otherwise a later cache miss would read the stale home-location copy from disk. WAL-resident pages stay pinned until the next checkpoint writes them home.

crates/opfs-btree/src/page.rs
```diff
@@ -215,43 +215,72 @@ pub(crate) fn raw_page_kind(raw: &[u8], expected_page_size: usize) -> Result<Pag
     Ok(header.kind)
 }
 
-pub(crate) fn raw_internal_child_for_key(
-    raw: &[u8],
-    expected_page_size: usize,
+// Hot path: bounds for the whole slot directory are validated once up front
+// so each binary-search probe is two direct loads plus one key bounds check.
+fn internal_child_for_key_parsed(
```
gdorsi, line 220: Hot read fix: 

the old per-probe path re-validated slot bounds with checked arithmetic and routed every read through `Result`-returning helpers. 

Bounds for the whole slot directory are now validated once before the loop, so each binary-search probe is two direct little-endian loads plus one key bounds check. 

`raw_leaf_find_value` below gets the same treatment.

crates/opfs-btree/src/page.rs
```diff
@@ -215,43 +215,72 @@ pub(crate) fn raw_page_kind(raw: &[u8], expected_page_size: usize) -> Result<Pag
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum RawDescendStep {
+    Leaf,
+    Child(PageId),
+    Other(PageKind),
+}
+
+/// One tree-descent step on a raw page: parses the header once and, for
+/// internal pages, finds the child for `key` in the same pass.
+pub(crate) fn raw_descend_step(
```
gdorsi, line 270: Combines page-kind dispatch and child selection into a single header parse. 

Before, each descent level parsed the header twice (`raw_page_kind`, then the child lookup) with extra cache hashmap lookups in between.

crates/opfs-btree/src/wal.rs
```diff
@@ -0,0 +1,682 @@
...
+        write_page(file, page_size, cursor + 1, frame.raw)?;
+        cursor += 2;
+    }
+
+    let commit = encode_wal_commit_page(page_size, header.generation, frame_count)?;
+    write_page(file, page_size, cursor, &commit)?;
+    Ok(wal_pages)
+}
+
+pub(crate) fn read_commit<F: SyncFile>(
```
gdorsi, line 112: Recovery entry point: replay only adopts a commit if the header, every frame meta page, every frame's data checksum, and the trailing commit marker (matching generation + frame count) all validate. Any mismatch truncates the tail and falls back to the last durable state, so torn or partial flushes can never be misread as committed.

crates/opfs-btree/src/wal.rs
```diff
@@ -0,0 +1,682 @@
...
+    // batched replay to win there.
+    false
+}
+
+const READ_RUN_PAGES: u64 = 64;
+
+/// Serves WAL page reads from a buffered run of up to READ_RUN_PAGES pages,
+/// so sequential replay does one backing-file read per run instead of one
+/// (freshly allocated) read per page.
+struct PageRunReader<'a, F: SyncFile> {
```
gdorsi, line 260: Run-buffered replay reads: one backing-file read per 64 pages instead of one zeroed allocation + one read call per page. 

Gated to wasm32 (`batch_wal_replay_reads` above) because each OPFS sync-access-handle read is a costly JS call, while on native/MemoryFile the run buffer's extra frame copy measured ~25% slower in the bench, hence both modes staying behind one `Cow`-returning interface.

gdorsi (commented):
Inline notes on the load-bearing changes, for easier review.

crates/opfs-btree/src/db.rs
```diff
@@ -364,11 +392,83 @@ impl<F: SyncFile> OpfsBTree<F> {
...
+        self.truncate_wal_tail()?;
         self.dirty_pages.clear();
+        self.wal_pages.clear();
+        self.freelist_dirty = false;
         self.evict_pages_if_needed(None);
         Ok(())
     }
 
+    pub fn flush_wal(&mut self) -> Result<(), BTreeError> {
+        let _span = tracing::debug_span!(
```
gdorsi, line 399: This is the core durability change: flush_wal used to be a full checkpoint (every dirty page written in place, freelist rebuilt, superblock swap).

Now it only appends dirty pages as WAL frames past the checkpointed region plus a commit marker, then one flush().

Because nothing inside the checkpointed region is touched, a crash mid-flush just means recovery falls back to the previous superblock.

crates/opfs-btree/src/db.rs
```diff
@@ -555,26 +646,22 @@ impl<F: SyncFile> OpfsBTree<F> {
...
-                "insert reached freelist page {}",
-                page_id
+            RawDescendStep::Other(kind) => Err(BTreeError::Corrupt(format!(
+                "insert reached {:?} page {}",
+                kind, page_id
             ))),
         }
     }
 
     fn delete_recursive(&mut self, page_id: PageId, key: &[u8]) -> Result<bool, BTreeError> {
```
gdorsi, line 659: Changed from “decode every internal page” to “use raw page bytes to choose the next child.”

crates/opfs-btree/src/db.rs
```diff
@@ -1101,6 +1171,8 @@ impl<F: SyncFile> OpfsBTree<F> {
         dirty_page_ids: &[PageId],
         freelist_pages: &[(PageId, Page)],
     ) -> Result<(), BTreeError> {
+        self.refresh_dirty_page_checksums(dirty_page_ids)?;
```
gdorsi, line 1179: Three optimizations live in this checkpoint write path:

1. **Deferred checksums get settled here.** In-place leaf mutations no longer hash the page on every write; this call recomputes xxh3 once per dirty page just before it hits disk. It only touches ids still in `dirty_pages` — WAL-resident pages in the batch were already refreshed by the flush that wrote them, so a checkpoint doesn't re-hash them.

2. **Contiguous pages are written in runs.** `collect_pages_for_write` returns the batch sorted by page id; the loop below groups consecutive ids and issues one `write_all_at` per run instead of one per page. Each OPFS sync-access-handle write is a JS call, so this matters most in the browser. (`alloc_page_near` is what makes runs common: split pages get ids adjacent to their siblings.)

3. **No page cloning.** The collected writes borrow the cached page buffers (`Cow::Borrowed`); only the freshly encoded freelist pages are owned. The run buffer is the single copy made on the way out.

crates/opfs-btree/src/db.rs
nrainhart: Sounds like a good comment to have in the code

nrainhart (approved):
All changes look reasonable to me, but I'm admittedly out of my element here

antoniomuso (approved):

PR #1007, opened 2026-06-09
refactor: simplify Rust & TS JazzClients

nrainhart:
## Description

This is a collection of multiple cleanups (see details in commits). The most notable change is the removal of the schema alignment logic that reordered returned inserted/queried values, both from Rust and TS. I couldn't find any real use cases that justified having it.


gdorsi (approved):
Wow, nice cleanup.

PR #1009, opened 2026-06-10
Batch settlement lifecycle

gdorsi:
## Problem

On the `todo-react` stress dataset, reloading the browser app took seconds (or hung) while the worker thread sat at 100% CPU. Deleting browser storage and re-syncing from the server made reload instant — a strong hint that the problem was not data volume but *locally retained batch bookkeeping*.

The root cause: the codebase had **four divergent answers to "is this batch settled?"**, and for a local-tier browser worker none of them was ever satisfiable:

1. `sealed_batch_still_needs_edge_reconciliation` (runtime core) — hard-coded `confirmed_tier < EdgeServer`, so a `DurableDirect { Local }` fate *always* "still needs reconciliation", with or without a server.
2. `local_batch_record_needs_fate_reconciliation` (wasm worker bridge) — the same hard-coded gate, duplicated on the main thread.
3. `retained_batch_terminal_tier` — dynamic, but flipped to `GlobalServer` whenever any server was *registered*, socket health and configuration validity irrelevant.
4. Submission pruning in the sync inbox — only on `Rejected` or `confirmed_tier >= GlobalServer`.

The consequences compounded:

- **The pending set was immortal.** Every direct batch ever committed stayed "pending" forever, so the worker→main replay on reload grew monotonically with total writes.
- **Membership recovery was quadratic.** When the worker settled a batch it deleted the sealed submission but kept the fate at `Local` — which gate 1 said was non-terminal. On the next reload, rebuilding "which rows belong to this batch" for each fate-only batch fell back to a full row-locator/history scan: O(batches × total rows). The two halves of the lifecycle actively sabotaged each other — settlement deleted the membership artifact that the never-terminating reconciliation path then had to reconstruct forensically.
- **`wait({ tier: "local" })` could hang forever.** A non-durable client with no tiered peer had no fate producer at all, and nothing guarded against registering a structurally unresolvable waiter.

We first tried treating the symptoms — a dedicated batch index, limiting replay to query-relevant rows. Both helped (reload dropped to ~3s) but couldn't fix it, because they reduced the cost *per pending batch* while the pending set itself never shrank.

## Why this solution

Instead of making the scans faster, this PR makes the pending set **self-bounding** by giving every runtime a single, configuration-derived **settlement target**:

> A batch is settled when its fate reaches this runtime's settlement target: `GlobalServer` when an upstream is configured, otherwise the runtime's own strongest durability tier.

Everything else falls out of that one definition:

- **Local-only runtimes settle at commit.** No upstream → target is `Local` → the self-confirmed fate settles the batch immediately and its bookkeeping retires. A serverless reload replays *nothing*, regardless of dataset size.
- **Submissions are retained exactly as long as they're needed.** An authority below its target keeps the sealed submission — which already carries the member list — so pending-batch membership is a point lookup, never a row-store scan. When the fate reaches the target, submission + local batch record + cache entry retire together; the fate row stays behind as the idempotency tombstone.
- **The reload replay is the pending set.** Worker→main sync now carries only genuinely in-flight batches: empty on a healthy local-only reload, proportional to unsynced work after an offline session — which is real replay work that must happen anyway.
- **The reconnect reconciliation path no longer treats every visible row as normal pending-batch state.** Since committed batches durably carry a fate or a retained submission, the submission/fate scans derive the normal pending set. The remaining full-scan fallback is kept as a deferred inconsistency-repair path rather than removed in this PR.
- **`wait(tier)` is total.** It resolves through the same producer model (self-confirmation at commit for durable runtimes, peer fates otherwise) and returns an error immediately when the runtime has no configured path to the requested tier, instead of hanging.
- **Late-server promotion still works.** Batches retired at `Local` are re-offered when an upstream appears: the existing `add_server` reconciliation regenerates seals from row history, and authorities already promote lower-tier direct fates.

The predicate itself lives once, on `SyncManager` (`fate_settled_at` / `fate_needs_settlement_at`), with the runtime core, both inbox gates, the tick-path retirement, and the worker bridge consuming it. The bridge now delegates its fate-reconciliation decision through `LocalBatchRecord` instead of carrying a wasm-only duplicate, so the four-way divergence can't silently come back.

## Client -> Server Reconciliation

The client-to-server handoff is still deliberately conservative:

- A local-only direct batch may retire its sealed submission once it settles at `Local`, but its row history and fate remain enough to regenerate a server seal later.
- When an upstream is added, reconciliation re-offers locally settled direct batches that are below the new settlement target.
- If a retained submission exists, the client replays the stored seal and exact member list.
- If the local runtime only has the lower-tier fate plus row history, reconciliation regenerates the seal from the committed rows.
- Stored `Missing` fates remain pending across restart, so a client that had retained a seal for server reconciliation will re-offer both the seal and rows instead of incorrectly treating the batch as settled.

This keeps the serverless fast path fast without losing the ability to promote local work to a server later.

## Validation

- Profile A/B on the stress dataset: the retained-batch replay and the reconnect row sweep no longer appear in the worker profile; main-thread time is idle/waiting rather than CPU-bound.
- New black-box integration tests cover the lifecycle invariants: serverless commits settle and replay nothing, authorities retain submissions below the target and prune at it, settled fates retire bookkeeping (leaving the fate tombstone), stored `Missing` fates replay retained seals and rows after restart, regenerated offline seals preserve their full contents, and `wait` on an unattainable tier errors instead of hanging.
- `LocalBatchRecord` now owns the bridge-facing fate-reconciliation gate, with a matrix test covering target/fate combinations.
- Full local verification:
  - `cargo test -p jazz-tools --lib`
  - `cargo check --workspace`
  - `cargo check -p jazz-wasm --target wasm32-unknown-unknown`
  - `pnpm --filter jazz-tools build`
  - `pnpm format:check`
  - `pnpm --filter jazz-tools test`
  - `pnpm --filter jazz-tools test:browser`
- Remote CI on `bac0a3050`: `lint`, `test-rust`, and `test-ts` passed.

## Behaviour changes to review

1. A serverless commit no longer persists a sealed submission (settled at `Local` immediately); replay to a later-added server regenerates the seal from row history.
2. An authority with an upstream now *retains* the submission it settles at `Local` until the upstream confirms (previously deleted immediately — this was the source of the quadratic membership recovery).
3. `wait_for_batch` returns an error for tiers no configured producer can reach.
4. Post-rollback TypeScript batch/transaction handles preserve the expected API error shape across both direct Node use and browser worker-backed use.

## Deferred

- Full removal of the remaining reconciliation full-scan fallback.
- Promotion watermark to skip the one-shot reconciliation scan when the settlement target hasn't changed across restarts.
- Config validation for stale `serverUrl` (a registered-but-unreachable upstream keeps batches legitimately pending).
- Per-row visible-region loads during query materialization — now the dominant remaining reload cost, independent of batching.
- Explicit error to the origin client when an untiered peer receives a seal (currently a warning).


packages/jazz-tools/src/runtime/db.ts
nrainhart, line 1: I don't understand why the changes in this file are needed

crates/jazz-tools/src/runtime_core/sync.rs
```diff
@@ -1,105 +1,120 @@
...
+            .sync_manager()
+            .batch_needs_settlement(fate)
     }
+
     fn pending_batch_ids_needing_reconciliation(&self) -> Vec<crate::row_histories::BatchId> {
-        let terminal_tier = self.retained_batch_terminal_tier();
+        let target = self.settlement_target();
+        let fates_by_batch: HashMap<crate::row_histories::BatchId, BatchFate> = self
+            .storage
+            .scan_authoritative_batch_fates()
```
nrainhart, line 28: Can you remind me what `scan_authoritative_batch_fates` and `scan_sealed_batch_submissions` return? May be worth adding a comment on those methods to make this clearer in the future as well

nrainhart (commented):
Core changes look reasonable, left only one question around reconciliation and one on the TS db changes

packages/jazz-tools/src/runtime/db.ts
gdorsi: They went through after my review 🤦 

Looks like CI is failing on main, and Codex does this instead of removing the stale tests.

gdorsi: Removed the commit

crates/jazz-tools/src/runtime_core/sync.rs
gdorsi: https://github.com/garden-co/jazz/pull/1009/commits/374a8d684bc9b70ab0e039a53df990e695d4ffa5

Added comments!

nrainhart: Thanks! For `scan_authoritative_batch_fates`, are we keeping the fate of every write ever done persisted forever?

gdorsi: That's correct.

It also means that this operation grows unbounded over time, which is far from ideal.

We probably need some more follow-up fixes to keep the situation in check

nrainhart (approved):

PR #1010, opened 2026-06-10
feat: add Rust permissions DSL

nrainhart:
## Description

This PR adds a permissions DSL to Rust that mirrors the TS DSL as closely as possible. For now, it's not meant to be a user-facing API, but rather just a utility for writing more readable tests. It's not as feature-complete as the Typescript DSL, the goal is to add the remaining parts as we need them.

To test the new DSL, I migrated all ReBAC permission tests to use the new permission DSL along with the existing schema/table builders, making them way easier to follow.

My goal is to migrate important Rust tests to use just schema/table builders, the permission DSL and the Rust JazzClient (instead of internals).

### DIFF
--- a/crates/jazz-tools/src/query_manager/types/policy.rs
+++ b/crates/jazz-tools/src/query_manager/types/policy.rs
@@ -1,4 +1,9 @@
 use super::*;
+use crate::object::ObjectId;
+use crate::query_manager::policy::{CmpOp, Operation, PolicyValue};
+use crate::query_manager::relation_ir::{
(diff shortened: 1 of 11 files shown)


crates/jazz-tools/src/query_manager/types/policy.rs
nrainhart, line 1: This is shamelessly vibecoded. I focused on getting a DSL similar to the one we have in Typescript

gdorsi (approved):

PR #1013, opened 2026-06-10
test: update rollback error expectations

gdorsi:
## Summary
- Update rollback error assertions in db transaction tests to match current batch lifecycle errors.
- Keep implementation unchanged.

## Verification
- pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts src/runtime/db.transaction.test.ts
- pnpm --filter jazz-tools test


nrainhart (approved):
Thanks for the fix! And sorry for the CI noise 😅

PR #1015, opened 2026-06-10
Fix realistic benchmark native build

gdorsi:
## Summary
- update the realistic benchmark example for the current synchronous Rust write API
- remove stale awaits from insert/update calls
- account for the BatchId now returned by insert where the benchmark keeps the row id

## Root cause
The native benchmark example still used the older async write-call shape after the Rust client write API became synchronous. CI failed while building the benchmark example with E0277 because Result is not a Future.

## Verification
- cargo build --release -p jazz-tools --features client,rocksdb --example realistic_bench
- cargo build --release -p jazz-tools --features client,sqlite --example realistic_bench
- pre-commit hook: clippy, rustfmt


nrainhart (approved):
Thanks for fixing this!

PR #1016, opened 2026-06-11
Support in-operator where queries

gdorsi:
## Summary

- Fixes runtime lowering for `in` conditions so array operands are handled before scalar literal coercion.
- Extends `in` support across id, enum, string, and reference query paths, including Better Auth pushdown and inspector operator selection.
- Documents the expanded operator support and adds browser integration coverage for `db.all`, `db.subscribeAll`, `useAll`, and `useAllSuspense`.

## Root Cause

The runtime query adapter eagerly coerced every condition value as a scalar before switching on the operator. Array values for `in` therefore failed before reaching the `in` lowering branch for scalar columns.

## Validation

- `pnpm build:runtime && pnpm build:test` from `packages/jazz-tools`
- `npx vitest run --config vitest.config.ts tests/ts-dsl/query-api.test.ts src/better-auth-adapter`
- `npx vitest run --config vitest.config.browser.ts tests/browser/db.all.test.ts tests/browser/db.subscribeAll.test.ts tests/browser/useAll.test.tsx tests/browser/useAllSuspense.test.tsx`
- `npx vitest run --config vitest.config.ts` from `packages/inspector`
- `npx tsc -b` from `packages/inspector`


crates/jazz-tools/src/query_manager/relation_ir_query_plan.rs
```diff
@@ -205,7 +205,15 @@ fn relation_predicate_to_disjuncts(predicate: &PredicateExpr) -> Option<Vec<Conj
...
-                return None;
+                let column = to_scoped_runtime_column(left);
+                return Some(vec![Conjunction {
+                    conditions: vec![
+                        Condition::IsNull {
+                            column: column.clone(),
+                        },
+                        Condition::IsNotNull { column },
+                    ],
+                }]);
```
gdorsi, line 219: This is a little hack to make `in: []` return an empty set.

Going to add a comment.

packages/jazz-tools/tests/ts-dsl/query-api.test.ts
```diff
@@ -118,6 +138,51 @@ describe("TS Query API", () => {
       expect(results.map((todo) => todo.id)).toEqual([todoWithoutOwner.id, todoWithOwner.id]);
     });
 
+    it("filters enum columns with in", async () => {
```
nrainhart: Nit: we could have separate describe blocks for different operators

packages/jazz-tools/src/typed-app.ts
```diff
@@ -237,25 +237,23 @@ type TimestampWhere<TOptional extends boolean> = WhereEqNe<
...
   }
 >;
-type UuidWhere<TOptional extends boolean, TRef extends string | undefined> = TRef extends string
-  ? WhereEqNe<string, TOptional, TOptional extends true ? { isNull?: boolean } : {}>
-  : WhereEqNe<
-      string,
-      TOptional,
-      TOptional extends true ? { in?: string[]; isNull?: boolean } : { in?: string[] }
-    >;
+type UuidWhere<TOptional extends boolean> = WhereEqNe<
```
nrainhart, line 241: Shouldn't we support `in` everywhere we support `eq`/`ne`? At least for non-nullable columns

packages/jazz-tools/src/runtime/query-adapter.ts
```diff
@@ -366,6 +366,19 @@ function conditionToRelPredicate(
   if (!columnType) {
     throw new Error(`Unknown column "${column}" in table "${table}"`);
   }
+  if (cond.op === "in") {
```
nrainhart, line 369: not sure why we need to pull this up, but tbh this whole function looks pretty messy

packages/inspector/src/utility/where-operators.ts
```diff
@@ -26,7 +26,7 @@ export function getSupportedWhereOperatorsForColumn(column: WhereOperatorColumn)
 
   switch (column.columnType.type) {
     case "Text":
-      return ["eq", "ne", "contains"];
+      return ["eq", "ne", "contains", "in"];
```
nrainhart: This logic is duplicated across the inspector, better auth and jazz-tools packages. We could probably centralize it.

packages/jazz-tools/tests/browser/useAllSuspense.test.tsx
```diff
@@ -366,6 +419,37 @@ describe("useAllSuspense browser integration", () => {
     });
   }
 
+  it("does not include rows for an empty in list", async () => {
```
nrainhart: Do we really need all of these browser tests? How's this different from testing it outside of the browser?

nrainhart (approved):
The overall change looks good! Left only minor comments

packages/jazz-tools/src/runtime/query-adapter.ts
gdorsi: Had to pull this up because the `toWasmValue` throws when passing arrays for scalar columns.

Yeah, pretty messy.

packages/jazz-tools/src/typed-app.ts
gdorsi: Yes, we can.

packages/jazz-tools/tests/browser/useAllSuspense.test.tsx
gdorsi: I just do it as rule of thumb, to cover main -> worker -> server sync.

Sometimes it helps to discover unexpected bugs.

There is probably no need to repete them for the different hooks, going to fix it.

PR #1018, opened 2026-06-11
Improve opfs-btree hint reuse and WAL write batching

gdorsi:
Stacked on #1001. After that round, the remaining opfs-btree cost was structural leaf-hint misses (every `put` cleared the hint; append keys and range starts could never validate against one), one OPFS write call per WAL page, and split candidates being encoded just to size-check them. This PR addresses those.

## What changed

1. **Leaf hints survive mutations and multiplied (1 → 4 slots).** Hints are validated against the page's *current* bytes on every use (`raw_leaf_covers_key`), which makes clearing them in `put`/`delete` unnecessary — the only real invalidation points are `remove_page` (the page-free chokepoint) and WAL replay. Each slot caches its leaf's key span so misses prefilter with a couple of byte comparisons, and slots reuse their buffers instead of allocating per descent.
2. **The rightmost leaf owns `[first_key, +∞)`.** Leaves are never merged and `next` pointers never cleared, so a cached leaf with no successor is provably the tail of the current tree. Append-pattern workloads — every newly created row — now hit hints on reads and writes instead of descending; the put path keeps the new tail hinted across splits.
3. **Inline-sized `put`s upsert through the hints with no descent.** A covers-hit identifies the unique leaf responsible for the key, so the fast path writes in place; splits, misses, and overflow values fall through to the recursive insert unchanged.
4. **Range starts resolve through hints too.** A start falling in the empty gap between a hinted leaf and its cached successor jumps straight to the successor (this is what the query layer's synthesized prefix bounds hit), and every scan remembers the leaf it *ended* in — which is exactly where a paginated caller's next start lands.
5. **WAL commits write in 64-page runs on wasm/OPFS** — one JS call per run instead of two per frame — keeping the direct per-page path natively, where batching measured slower. Mirrors the replay-read batching from #1001.
6. **`page_fits` computes the encoded size arithmetically** instead of building the full payload, and split candidates are moved rather than cloned; a test pins the size computation to `encode_fields`.

## Benchmarks

Compared browser worker CPU profiles from `dev/stress-tests/todo-react`. `main` and `1st round` are the values from #1001; `this PR` is from the latest captures after the tail/range hint work.

Load action:

| worker metric | main | 1st round | this PR |
|---|---:|---:|---:|
| Worker active sampled CPU | 6.183 s | 4.203 s | 1.959 s |
| Time in `opfs_btree` stacks | 2.548 s | 1.261 s | 0.457 s |

Generate action:

| worker metric | main | 1st round | this PR |
|---|---:|---:|---:|
| Worker active sampled CPU | 8.680 s | 5.302 s | 4.846 s |
| Time in `opfs_btree` stacks | 4.821 s | 2.145 s | 1.305 s |

The latest captures show the tail/range hints taking another chunk out of tree descents on top of the first round: `raw_descend_step` is down to 163.9 ms in generate and 38.8 ms in load. The latest load capture has a long idle tail, so active worker CPU is the meaningful comparison there.

## Verification
- `cargo test -p opfs-btree`
- `cargo clippy -p opfs-btree --all-targets -- -D warnings`
- `cargo check -p opfs-btree --target wasm32-unknown-unknown`
- `pnpm build:core`
- `cargo bench -p opfs-btree --bench hot_paths`


crates/opfs-btree/src/page.rs
```diff
@@ -332,26 +337,36 @@ pub(crate) fn raw_leaf_find_value<'a>(
...
+/// its last entry still belong to it.
 pub(crate) fn raw_leaf_covers_key(
     raw: &[u8],
     expected_page_size: usize,
     key: &[u8],
 ) -> Result<bool, BTreeError> {
+    let Some(span) = raw_leaf_span(raw, expected_page_size)? else {
+        return Ok(false);
+    };
+    Ok(span.first_key <= key && (key <= span.last_key || span.next_page_id.is_none()))
```
gdorsi, line 351: This function is used to validate leaf hints, and before we were missing some "opportunities" of skipping a full pages traversal by not considering that rightmost leaves top limit should be considered unbounded.

This helps alot when doing append-only inserts.

crates/opfs-btree/src/page.rs
```diff
@@ -96,8 +102,7 @@ pub(crate) fn freelist_ids_per_page(page_size: usize) -> Result<usize, BTreeErro
 }
 
 pub(crate) fn page_fits(page: &Page, page_size: usize) -> Result<bool, BTreeError> {
-    let encoded = encode_fields(page)?;
-    Ok(encoded.payload.len() <= page_payload_capacity(page_size)?)
+    Ok(encoded_payload_len(page)? <= page_payload_capacity(page_size)?)
```
gdorsi, line 105: `page_fits` used to fully encode the candidate page just to compare its length, and the split paths then encoded the winner *again* in `set_dirty_page`. 

The encoded size is fully determined by entry lengths, so this computes it arithmetically.

crates/opfs-btree/src/wal.rs
```diff
@@ -253,6 +251,81 @@ fn batch_wal_replay_reads() -> bool {
 }
 
 const READ_RUN_PAGES: u64 = 64;
+#[cfg(target_arch = "wasm32")]
+const WRITE_RUN_PAGES: usize = 64;
+
+/// Accumulates consecutive WAL pages and writes them to the backing file in
+/// runs of up to WRITE_RUN_PAGES, so a commit costs one write call per run
+/// instead of one per page.
+struct PageRunWriter<'a, F: SyncFile> {
```
gdorsi, line 260: Optimizing also the writes, to batch subsequent wal paages writes into one like we have started doing with reads in the previous PR.

This approach is effective with WALs when we do write-intensive workloads.

gdorsi (commented):
Inline notes on the load-bearing changes, for easier review.

crates/opfs-btree/src/db.rs
```diff
@@ -258,6 +305,40 @@ impl<F: SyncFile> OpfsBTree<F> {
...
+            match upsert {
+                RawLeafUpsertResult::Inserted => {
+                    self.mark_dirty_loaded_page(page_id);
+                    self.remember_leaf_hint(page_id)?;
+                    return Ok(());
+                }
+                RawLeafUpsertResult::Updated { old_overflow } => {
+                    self.mark_dirty_loaded_page(page_id);
+                    self.remember_leaf_hint(page_id)?;
+                    if let Some(old_overflow) = old_overflow {
```
gdorsi, line 283: In this B-tree, small values are stored inline in the leaf. Large values are stored in overflow pages, and the leaf stores a pointer like:

```rs
OverflowRef {
    head_page_id,
    total_len,
}
```

When updating a slot that was pointing to an overflow page, we need to free it to avoid leaking overflow pages.

crates/opfs-btree/src/db.rs
```diff
@@ -359,13 +393,18 @@ impl<F: SyncFile> OpfsBTree<F> {
...
                 }
             }
 
+            last_scanned = Some(page_id);
             current = next;
         }
 
+        if let Some(page_id) = last_scanned {
+            self.remember_leaf_hint_if_not_mru(page_id)?;
+        }
```
gdorsi, line 407: This if just checks that we have scanned at least one page

crates/opfs-btree/src/db.rs
```diff
@@ -1019,25 +1060,129 @@ impl<F: SyncFile> OpfsBTree<F> {
...
+        for idx in 0..LEAF_HINT_SLOTS {
+            let Some(page_id) = self.leaf_hints.covering_candidate(idx, key, prefilter_mru) else {
+                continue;
+            };
             let covers = match self.pages.get(&page_id) {
                 Some(raw) => raw_leaf_covers_key(raw, self.options.page_size, key)?,
                 None => false,
             };
             if covers {
+                self.leaf_hints.promote(idx);
```
gdorsi, line 1087: Just markds the leaf as the most recently used

nrainhart (approved):
Nice performance improvements! :shipit:

PR #1019, opened 2026-06-12
test: refactor ReBAC tests to use JazzClient

nrainhart:
## Description

Migrates ReBAC unit tests that previously accessed Jazz internals directly to use the higher-level JazzClient and TestingServer. This makes them easier to read as well as more resilient to future internal refactorings.

The most important part in this PR are the utils I added for higher-level tests:
- `JazzClient::test_client` simplifies single-node tests
- `JazzClient::connect_with_row_policy_mode` allows creating permissive clients (to simulate untrusted clients)
- moved `wait_for_query` to the test_support crate, so that it can be used in more places
- added a migration guide. I'll use this guide to migrate more existing tests, and them I'll modify it to aid LLMs in creating new tests 

Note: I plan to move the migrated rebac tests to `/crates/jazz-tools/tests/policies_integration/`. I didn't move them yet to make the diff view nicer for review, I'll do it once the PR is approved.


crates/jazz-tools/tests/policies_integration/declared_fk_inheritance.rs
```diff
@@ -1,158 +1,126 @@
...
+
     assert!(
         visible_ids.contains(&file_id),
         "alice should see file via allowedTo.readReferencing(policy.todos, \"image\")"
     );
 }
 
-#[test]
-fn rebac_declared_fk_inheritance_grants_update_access() {
+#[tokio::test]
```
nrainhart, line 81: Since the `JazzClient` is async, tests using it need to be Tokio tests. If this leads to slower test runs in the long run, we could maybe switch to using a sync JazzClient just for tests, but I think this is fine for now

crates/jazz-tools/src/query_manager/rebac_tests.rs
```diff
@@ -745,13 +620,18 @@ fn declared_file_inheritance_schema(array_edge: bool) -> Schema {
         .build()
 }
 
+#[cfg(feature = "test-utils")]
```
nrainhart: These cfg attributes are needed in all utils and tests that use the JazzClient (since the client is only compiled with the `client` flag, which `test-utils` includes).

This should become less noisy once I move these tests over to `/crates/jazz-tools/tests/policies_integration/`

crates/jazz-tools/TEST_MIGRATION
```diff
@@ -0,0 +1,32 @@
+# Test Migration Guide
```
gdorsi, line 1: Shall we make this one a skill?

gdorsi (approved):
LGTM

I think that would be valuable to also add some docs comment to the tests, because in the current state it's hard for me to understand what the test actually covers.

For example while reviewing, the agent reported:
> `rebac_declared_fk_inheritance_reacts_to_fk_updates` no longer tests subscription reactivity. The old test used a single subscription and checked its results before and after the FK change. The new test calls `query_ids_as` (a one-shot query) twice instead, so it only validates point-in-time correctness

Now, my guess is that the subscription is outside of the test scope, but a doc comment on the test would have helped.

crates/jazz-tools/TEST_MIGRATION
nrainhart: Yes! My plan is to convert this into a "writing Rust tests for Jazz" skill after I'm done with the migration

PR #1020, opened 2026-06-12
docs: add E2EE shared key management and algorithms design spec

gdorsi:
In short:

Per-column end-to-end encryption for Jazz2: developers mark columns as encrypted in the schema, values are encrypted client-side before sync, and the server only ever sees ciphertext. 

This document covers shared key management (DX + persistence) and encryption algorithms. 

ACL remains the job of row policies; E2EE protects data from the service provider, not from other authorized members.

Schema:
```ts
const schema = {
  users: s.table({
    name: s.string(),
    e2eeKey: s.string().optional(),      // app-managed directory, ordinary column
  }),
  projects: s.table({
    name: s.string(),
  }).encryptionSpace(),                   // rows of this table can own a shared key
  todos: s.table({
    title: s.string().encrypted("projectId"),
    done: s.boolean(),                    // plaintext, queryable
    projectId: s.ref("projects"),
  }),
};
```

Usage:

```ts
// publish my key (app-managed directory)
await db.users.update(me, { e2eeKey: db.e2ee.publicKey() });

// invite: app fetches the recipient's key from its own table, then
await db.projects.shareKey(projectId, recipient.e2eeKey);

// revoke (v1 = policy-only): delete their sealed copy; app updates its own ACL rows
await db.projects.unshareKey(projectId, recipient.e2eeKey);

// member listing for UI: [{ recipient_user_id, recipient_public_key }]
await db.projects.keyHolders(projectId);

// writes & reads: transparent
await db.todos.insert({ title: "secret", projectId });  // encrypts under projectId's key
const todo = await db.todos.get(id);                     // decrypts; todo.title is string
```

### DIFF
--- a/docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
+++ b/docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
@@ -1,0 +1,325 @@
+# E2EE Shared Key Management & Encryption Algorithms — Design
+
+Status: approved design, pre-implementation.
+
+## Goal
+
+Per-column end-to-end encryption for Jazz2: developers mark columns as encrypted in
+the schema, values are encrypted client-side before sync, and the server only ever
+sees ciphertext. This document covers shared key management (DX + persistence) and
+encryption algorithms. ACL remains the
+job of row policies; E2EE protects data from the service provider, not from other
+authorized members.
+
+Scope for v1: LocalFirst Auth users only.
+
+## Decisions at a glance
+
+| Question | Decision |
+|---|---|
+| Key scope | Per **space row** — a row of a table marked `.encryptionSpace()` owns one shared key |
+| Key handoff | Sealed at invite time via an explicit `shareKey` call |
+| Revocation | Policy-only in v1 (no rotation); envelope carries `key_id` so epochs can be added later |
+| Public-key directory | App-managed column in the app's own schema; framework never queries it |
+| Schema DX | `s.string().encrypted("refColumn")` — each encrypted column names the ref to its space |
+| Sealed-key persistence | Framework-managed companion table `<table>$keys` per space table |
+| Crypto venue | Rust core (WASM / NAPI / RN) — one implementation, synchronous in-engine |
+| Symmetric cipher | XChaCha20-Poly1305 |
+| Sealing | HPKE (RFC 9180), DHKEM(X25519) + HKDF-SHA256 + ChaCha20-Poly1305 |
+| Post-quantum | Not in v1; algorithm-id byte in every envelope/sealed blob keeps the upgrade additive |
+| LoFi auth key algorithm | Unchanged (ed25519 already ships in the WASM engine) |
+| Rust API | Same core feature surfaced on `JazzClient` (`share_key` / `unshare_key` / `key_holders`); schema via `TableSchemaBuilder` |
+
+## 1. Identity & private keys
+
+No new secret material and no new storage. The 32-byte LocalFirst Auth secret is the
+single root. Alongside the existing `jazz-auth-sign-v1` ed25519 derivation
+(`crates/jazz-tools/src/identity.rs`), add one domain-separated derivation:
+
+- `jazz-e2ee-seal-v1` → X25519 keypair (the user's encryption identity).
+
+Consequences:
+
+- Everything is re-derivable from the auth secret, so existing backup/recovery paths
+  (passphrase, passkey backup) automatically cover E2EE keys.
+- Recovery = auth recovery. Lose the secret, lose the keys.
+- `x25519-dalek` reuses the `curve25519-dalek` backend already pulled in by
+  `ed25519-dalek`, so the bundle delta is minimal.
+
+The public encryption key is exposed to apps as `db.e2ee.publicKey()`. The **app**
+publishes it into its own users/profiles table (app-managed directory). The framework
+never reads app tables for keys: `shareKey` takes the recipient's public key as an
+argument, and the app looks it up with an ordinary query. Directory keys are trusted
+on first use in v1; key transparency / verification is out of scope.
+
+## 2. Schema DX
+
+```ts
+const schema = {
+  users: s.table({
+    name: s.string(),
+    e2eeKey: s.string().optional(),      // app-managed directory, ordinary column
+  }),
+  projects: s.table({
+    name: s.string(),
+  }).encryptionSpace(),                   // rows of this table can own a shared key
+  todos: s.table({
+    title: s.string().encrypted("projectId"),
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
```diff
@@ -0,0 +1,329 @@
...
+
+- `.encrypted(refColumn)` is valid only when `refColumn` is a **non-nullable** ref to
+  an `.encryptionSpace()` table; anything else is a schema-build error.
+- TS types are unchanged for reads/writes: `todo.title` is `string` (modulo the
+  `Locked` state, §6). Under the hood `encrypted()` composes with the existing
+  `ColumnTransform` mechanism in `dsl.ts`: serialize per declared type → encrypt →
+  store as BYTEA envelope.
+- Encrypted columns are excluded from the query/where DSL, indexes, and policy
+  expressions — enforced both at the type level and at schema-build time. Filtering
+  on encrypted values happens client-side; this trade-off is accepted.
```
nrainhart, line 86: Couldn't equality comparisons still work? If they do, it'd be nice to keep them

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
```diff
@@ -0,0 +1,329 @@
...
+  app-level trust decision, not a protocol property.)
+
+## 6. Locked values
+
+If a client holds a row but not its key (key row not yet synced, or never shared),
+the plaintext cannot exist. Encrypted columns are typed `T | Locked`, where `Locked`
+is an exported sentinel with an `isLocked(v)` guard. This forces app code to handle
+the state honestly instead of receiving an ambiguous `null`. In practice the state is
+rare — row policies usually prevent non-members from receiving rows at all — but
+"key hasn't synced yet" is a genuine local-first state and the types should say so.
```
nrainhart, line 229: This brings back jazz1 loaded-state-handling vibes. I get the feeling that I'd rather clients don't see rows if they cannot decrypt them

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
joeinnes: My 2c: keep it all client side, at least for now. To keep equality comparisons, you'd either need deterministic encryption (nonce is currently random, we'd need to derive it) or a [blind index](https://paragonie.com/blog/2017/05/building-searchable-encrypted-databases-with-php-and-sql#solution-literal-search).

Deterministic encryption is probably a weakening we don't want to encourage, but developing blind indexes might be worth a follow-up.

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
```diff
@@ -0,0 +1,329 @@
...
+
+**Who can share.** Anyone who already holds the space key — sealing requires the
+plaintext key, so only current key holders can produce a valid share, and a
+non-member's `shareKey` fails locally with `E2EEKeyUnavailable`. This gate is
+cryptographic and cannot be bypassed via the server (which never has the key). In
+v1 there is no admin tier: possession = full re-share rights, and "only owners can
+invite" is an app-level convention enforced by the app's own policies on row
+delivery, not by the key layer. Damage from a rogue share is bounded because the
+key alone reads nothing — the recipient still needs the rows, which the app's
+ordinary policies control.
```
joeinnes, line 163: Note on this: I think it's worth documenting this very carefully. On first read, I came to the conclusion that this is a regression compared to classic Jazz, but after spending some time on it, I've changed my opinion, and I think this is just a formal documentation of a threat which is fundamentally the same category as shoulder-surfing or screenshotting, and I think our posture is in fact stronger than it was with classic Jazz (or it will be once rotation lands).

Classic Jazz model: access to ciphertext is security-through-obscurity (need the CoValue ID).
Jazz 2 model: access to ciphertext is governed by ReBAC policies.

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
```diff
@@ -0,0 +1,329 @@
...
+| Public-key directory    | App-managed column in the app's own schema; framework never queries it                                                    |
+| Schema DX               | `s.string().encrypted("refColumn")` — each encrypted column names the ref to its space                                    |
+| Sealed-key persistence  | Framework-managed companion table `<table>$keys` per space table                                                          |
+| Crypto venue            | Rust core (WASM / NAPI / RN) — one implementation, synchronous in-engine                                                  |
+| Symmetric cipher        | XChaCha20-Poly1305                                                                                                        |
+| Sealing                 | HPKE (RFC 9180), DHKEM(X25519) + HKDF-SHA256 + ChaCha20-Poly1305                                                          |
+| Post-quantum            | Not in v1; algorithm-id byte in every envelope/sealed blob keeps the upgrade additive                                     |
```
joeinnes, line 29: I think we should try to keep this as pluggable as possible, XChaCha20-Poly1305 is not NIST-standard, so anyone wanting to use Jazz in a regulated context might have issues. 

Also, PQ is likely to become a bigger ask than it is currently sooner rather than later: https://blog.google/innovation-and-ai/technology/safety-security/cryptography-migration-timeline/

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
gdorsi: Agree with Joe.

If you can do equality it means that you can do some permission checks, but also that a third party can "reverse engineer" the values.

I think it's better to make encryption more strict and make people pick only columns that are important for privacy in their app.

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
gdorsi: Agree.

Got in the spec because it's a question I've asked to Claude that wans't totally clear from the rest of the spec.

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
gdorsi: Thanks for this feedback!

> XChaCha20-Poly1305 is not NIST-standard

Ok, better to replace it then

> PQ is likely to become a bigger ask than it is currently sooner rather than later

Going to give it a look. Last time I checked the problem wasn't even performance but bundle size.

docs/superpowers/specs/2026-06-12-e2ee-shared-keys-design.md
gdorsi: Good point, going to fix it.

PR #1025, opened 2026-06-15
chat-react: scoped updates, sender deletes, idempotent reactions

joeinnes:
Follow-up to #980, which landed the permissions-engine fix for `id` comparisons inside `exists`. This builds the worked example, tests and polish on top of that fix.

## Summary
- Restrict chat updates to non-protected fields — `createdBy` and `isPublic` can no longer be changed — demonstrating granular field-level update permissions in the chat-react example.
- Let a message's sender delete it and cascade-delete its attachments, files and reactions (including reactions left by others); deny everyone else.
- Make the reaction picker idempotent: picking the same preset twice no longer creates a duplicate reaction.
- Polish the example's docs, error handling and copy (ErrorBoundary, ChatImage, ChatFile, ChatView).
- Add a spec note proposing an unchanged-columns permission helper to replace hand-rolled per-field `exists` checks.

## Test plan
- [ ] `cd examples/chat-react && pnpm run test:unit` — 7 permission tests pass (incl. sender delete + cascade)
- [ ] `cd examples/chat-react && pnpm run test:browser` — 25 browser tests pass (incl. reaction-picker idempotency)

### DIFF
--- a/examples/chat-react/README.md
+++ b/examples/chat-react/README.md
@@ -6,8 +6,12 @@ Real-time, permission-aware chat app. Public rooms, private chats with invite li
 
 ```bash
 pnpm install
-pnpm dev        # starts the Jazz server, pushes the schema, and opens Vite
+pnpm dev        # starts Vite — the app runs local-first by default
 ```
+
+The app is local-first: without configuration it persists everything in the browser. To
+sync across devices, point it at a Jazz server with `VITE_JAZZ_SERVER_URL` (and
+`VITE_JAZZ_APP_ID`). `pnpm dev` does not start a server or push the schema.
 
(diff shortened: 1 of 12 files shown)


aeplay:
🤖 Closing this pre-new-core chat-react branch. Its sender-delete and profile-ownership plus attachment-delete policy fixes are already present in main; #1505 preserved the required correlated EXISTS semantics, #1911 added authenticated message-attribution coverage, and #1635 fixed subscription removal. Per #1949, new authenticated chat behavior belongs in BandChat. The remaining reaction-deduplication requirement is tracked in #2063 rather than transplanted from this conflicted branch.

PR #1026, opened 2026-06-15
test(integration): improve schema migration coverage

nrainhart:
## Description

FUP to https://github.com/garden-co/jazz/pull/1019. Migrates unit tests on schema migration to integration tests that use the Rust JazzClient.

While adding these tests I found two issues, which I fixed:
- Permissions were not being enforced if a table was renamed while the permissions still referred to the old table name
- Writes on a newer schema didn't update subscriptions for the previous schema (even when there was a lens between both table versions)

### DIFF
--- a/crates/jazz-tools/src/sync_manager/inbox.rs
+++ b/crates/jazz-tools/src/sync_manager/inbox.rs
@@ -664,10 +664,14 @@ impl SyncManager {
         let context =
             crate::storage::resolve_history_row_write_context(storage, table, &row).ok()?;
         let branch_name = BranchName::new(&row.branch);
-        let row_locator = RowLocator {
-            table: table.to_string().into(),
-            origin_schema_hash: Some(context.history_row_raw_table_id().schema_hash),
-        };
+        let row_locator = storage
+            .load_row_locator(row.row_id)
+            .ok()
+            .flatten()
+            .unwrap_or_else(|| RowLocator {
+                table: table.to_string().into(),
+                origin_schema_hash: Some(context.history_row_raw_table_id().schema_hash),
+            });
 
         apply_row_batch_with_context(
             storage,
--- a/crates/jazz-tools/src/sync_manager/inbox.rs
+++ b/crates/jazz-tools/src/sync_manager/inbox.rs
@@ -536,10 +536,31 @@ impl SyncManager {
         &self,
         storage: &H,
         batch_id: crate::row_histories::BatchId,
+        target_branch_name: Option<&BranchName>,
         object_ids: &[ObjectId],
     ) -> Vec<(String, StoredRowBatch)> {
         let mut rows = Vec::new();
         for row_id in object_ids {
+            if let Some(target_branch_name) = target_branch_name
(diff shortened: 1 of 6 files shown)


crates/jazz-tools/tests/catalogue_sync_integration.rs
```diff
@@ -901,6 +1255,1338 @@ async fn catalogue_sync_e2e_schema_evolution_through_sync_manager() {
...
+        ]
+    );
+
+    alice.shutdown().await.expect("shutdown alice");
+    bob.shutdown().await.expect("shutdown bob");
+    server.shutdown().await;
+}
+
+#[tokio::test]
+async fn draft_lens_does_make_rows_from_old_schema_visible() {
```
gdorsi: nit: is this test covering the opposite of the test name?

Shall we rename it into `draft_lens_does_not_make_rows_from_old_schema_visible`?

crates/jazz-tools/src/sync_manager/inbox.rs
```diff
@@ -583,14 +604,88 @@ impl SyncManager {
...
+        storage: &mut H,
+        table: &str,
+        row: StoredRowBatch,
+    ) -> Option<crate::row_histories::ApplyRowBatchResult> {
+        let context =
+            crate::storage::resolve_history_row_write_context(storage, table, &row).ok()?;
+        let branch_name = BranchName::new(&row.branch);
+        let row_locator = RowLocator {
+            table: table.to_string().into(),
+            origin_schema_hash: Some(context.history_row_raw_table_id().schema_hash),
```
gdorsi: 🤖  `apply_row_batch_for_table` constructs `RowLocator` using `origin_schema_hash` from the write context (`context.history_row_raw_table_id().schema_hash`) rather than preserving the row's actual origin schema hash. 

The old `apply_row_batch` read the existing global row locator from storage, which kept the origin schema hash set during the initial write. 

When a row authored under an older schema version is applied through a lens to a newer schema's table, the constructed `origin_schema_hash` can reflect the target/write schema, not the original one. This hash ends up in `RowVisibilityChange.row_locator` and is later used by `prepare_row_update_with_origin` to resolve table names through schema migrations, potentially causing incorrect table name mapping during deferred visibility change processing.

----

Is this concern valid?

gdorsi (approved):

crates/jazz-tools/tests/catalogue_sync_integration.rs
nrainhart: Ha, good catch. Fixed.

crates/jazz-tools/src/sync_manager/inbox.rs
nrainhart: Tbh I'm not sure. The current implementation is pretty hard to follow. I'm assuming this will not survive the rewrite, but wanted the test cases with the correct behavior to already be passing when we do the rewrite.

I just modified the current implementation to be more similar to what we previously had, while still keeping all tests green.

PR #1028, opened 2026-06-17
otel: capture Rust panics through the OTLP log pipeline

Elfo404:
## Problem

The OTLP log exporter buffers records in-process and flushes on an interval. When the process **panics**, that buffer dies with it — so the crash is invisible to structured telemetry and survives only as the raw stderr backtrace scraped off disk by the node collector's filelog pipeline.

## Change

Add `otel::install_panic_hook`, installed from `init_tracing` when OTLP export is active. It:

1. Turns the panic into a structured, trace-correlated `error!` event (`panic.message`, `panic.location`, `panic.backtrace`) — so the OTLP appender bridge converts it to a `LogRecord`.
2. **Force-flushes** the logger provider synchronously, so the record actually ships before the process unwinds.
3. Preserves and re-invokes the previously installed hook, so the stderr backtrace still prints and the **filelog safety net stays intact**.

## Why `force_flush` is safe here

The logger uses the **dedicated-thread** `BatchLogProcessor` with the blocking HTTP client (`with_batch_exporter`), not the async-runtime variant. The flush runs on the processor's own thread — independent of the unwinding Tokio runtime — and its force-flush is internally bounded (~5s), so a missing collector (e.g. a freshly provisioned node) can't stall the crash path.

## Scope / limitations

Covers Rust `panic!` (the workspace builds with the default `panic = unwind`, so the hook fires with a clean unwind window). Does **not** cover OOM-kill / SIGKILL / `abort` — the kernel terminates the process with no chance to run the hook; those need a metrics-side signal (`kube_pod_container_status_last_terminated_reason`), tracked separately.

## Testing

New `tests/otel_panic_hook.rs` uses an **in-memory exporter**, so it runs in CI without a collector. It asserts a panic produces a flushed `process panicked` record carrying the message and source location, and that the previously installed hook still runs.

- `cargo test -p jazz-tools --features otel --test otel_panic_hook` — passes
- `cargo check -p jazz-tools` (with and without `otel`) — clean
- `cargo clippy --features otel` — no new warnings
- `cargo fmt --check` — clean


Elfo404:
Good catch — `tests/AGENTS.md` is explicit that everything there must be a no-mock, end-to-end `TestingServer` test, and this one uses an `InMemoryLogExporter` and calls the hook directly, so it doesn't fit.

Making it a `TestingServer` e2e test would be contrived (you'd have to make a real running server panic on demand to exercise a hook whose logic is entirely module-local), so I moved it to a **unit test** in `src/otel.rs`'s existing `#[cfg(test)] mod tests` — alongside `active_websockets_gauge_reports_the_count`, which already uses the in-memory-exporter pattern. The `tests/` file is deleted. The unit test also now restores the prior global panic hook so it doesn't leak into the rest of the lib test binary.

`cargo test -p jazz-tools --features "otel,test" --lib otel::tests::panic_hook` passes; fmt + clippy clean.

Elfo404:
Reverted the unit-test relocation in a follow-up commit (`0c6cad777`) — and to be precise about *why*, since the real reason is the one you raised:

The test installs a **process-global** panic hook (`std::panic::set_hook`). The lib unit-test binary runs its tests **in parallel in one process**, so even with save/restore there's a window where another test panicking would route through our hook — the global mutation can influence unrelated tests. As an integration test it compiles to its **own test binary / process** where it's the only test, so the global hook swap is fully isolated and can't affect anything else. That isolation — not just "crash telemetry belongs in integration" — is the motivation.

It also now carries the `///` doc comment AGENTS.md asks for (contract + actor + ASCII flow) and keeps the local `Arc` sentinel + save/restore so it's tidy within its own process. `cargo test --features otel --test otel_panic_hook` passes; fmt clean.

Happy to capture this "why it's an integration test" note in the test's module doc too, so it's answered in-code for the next reader — say the word.

antoniomuso (approved):

gdorsi (approved):

PR #1029, opened 2026-06-17
Expose framework-agnostic binding utilities via jazz-tools/shared

joeinnes:
## Summary
- Add a `jazz-tools/shared` entry point re-exporting the framework-agnostic utilities
  the React/Svelte/Vue bindings use to turn a live query into a reactive,
  in-place-updated result set: `applyDelta`, `reconcileArray`, `RowChangeKind`, plus the
  `SubscriptionDelta` / `RowDelta` / `QueryBuilder` / `QueryOptions` types and the
  orchestrator's `SubscriptionsOrchestrator` / `CacheEntryHandle` / `UseAllState` shapes.
- Refactor the React, Svelte and Vue `useAll` bindings to consume that public surface
  instead of deep relative imports — a pure import swap, no behaviour change. Dogfooding
  is the proof the surface is complete.
- Surfaced so an external author can build their own binding (e.g. a signals-based
  `useAllSignal`) on the same utilities. Documented as an advanced, use-at-your-own-risk
  surface — not semver-covered.
- Wired through the package `exports` map only; `tsc` already emits `dist/shared` JS + `.d.ts`.

## Test plan
- [ ] `pnpm --filter jazz-tools test` — react/svelte/vue `useAll` suites stay green
- [ ] `pnpm --filter jazz-tools test:browser` — covers react-core/`useAll` in a real browser
- [ ] New black-box test: a binding built only on `jazz-tools/shared` tracks a live query and reconciles in place
- [ ] `import('jazz-tools/shared')` resolves; the `./shared` exports map points at real JS + `.d.ts`

### DIFF
--- a/packages/jazz-tools/package.json
+++ b/packages/jazz-tools/package.json
@@ -40,6 +40,10 @@
       "react-native": "./dist/react-native/react-core.js",
       "default": "./dist/react-core/index.js"
     },
+    "./shared": {
+      "types": "./dist/shared/index.d.ts",
+      "default": "./dist/shared/index.js"
+    },
     "./permissions": {
       "types": "./dist/permissions/index.d.ts",
       "default": "./dist/permissions/index.js"
--- a/.changeset/shared-binding-utilities.md
+++ b/.changeset/shared-binding-utilities.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Add a `jazz-tools/shared` entry point exposing the framework-agnostic utilities the React, Svelte and Vue bindings use to turn a live query into a reactive, in-place-updated result set: `applyDelta`, `reconcileArray`, `RowChangeKind`, and the supporting types (`SubscriptionDelta`, `RowDelta`, `QueryBuilder`, `QueryOptions`, and the orchestrator's `SubscriptionsOrchestrator` / `CacheEntryHandle` / `UseAllState` shapes). The in-repo bindings now consume this shared surface so an external author can build their own binding (e.g. a signals-based `useAllSignal`) on the same utilities.
+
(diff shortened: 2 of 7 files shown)


packages/jazz-tools/package.json
```diff
@@ -40,6 +40,10 @@
       "react-native": "./dist/react-native/react-core.js",
       "default": "./dist/react-core/index.js"
     },
+    "./shared": {
```
nrainhart, line 43: "shared" sounds kind of generic to me. How do you feel renaming this to "bindings"?

nrainhart (approved):
🚀

packages/jazz-tools/package.json
joeinnes: Earlier discussion on this/similar topic:

https://github.com/garden-co/jazz/pull/934#discussion_r3308124794

Would take the suggestion in #934 and move `createJazzClient` here as well in a later PR.

The problem I have with `bindings` here is that these are not the bindings themselves, they're utils for the bindings.

Not fully happy with any of the options we've discussed so far:

- `shared` (my favourite so far)
- `bindings`
- `web`
- `browser` (my worst idea to date)

@gdorsi any thoughts?

gdorsi: Agree that `shared` is the best candidate.

An alternative would be to put these in the main bundle.

If all the UI frameworks (web and native) depend on it, and we care about bundles size only on the browser, why should this be a separate entry?

gdorsi (approved):

packages/jazz-tools/package.json
joeinnes: React and RN don't really depend on this (types only) because they don't have the same reactivity model (they diff the DOM at the render layer). Main reason for creating this as a separate entry is so that folks can write their own userland bindings for signals-based frameworks on top of Jazz, because otherwise they won't be able to import them from where they are now.

We could expose from the main bundle too, but the vast majority of folks aren't going to want to use these utils so moving them here reduces the API surface for the main bundle.

I don't think bundle size matters here, it's just DX, they'll be imported either way.

PR #1032, opened 2026-06-18
Grow-only set merge strategy for array columns

joeinnes:
## Summary
- Add `merge("g-set")` for non-nullable array columns. Concurrent writes converge to the grow-only union of every replica's elements, deduplicated and sorted into a canonical, byte-identical order.
- An element written by one replica is never dropped by a concurrent write that never saw it; there is no element removal (grow-only).
- The strategy is threaded through the DSL down to the wasm boundary, with construction-time rejection of non-array columns and nullable-array columns in either chaining order.
- Document it in the Advanced Internals reference under a new "Grow-only sets" section, alongside a "Counters" subheading for symmetry.

## Test plan
- [ ] `cargo test -p jazz-tools --features test --test gset_merge` — convergence to the sorted union, grow-only monotonicity, and distinct-float-representation determinism
- [ ] `pnpm --filter jazz-tools test dsl` — DSL stores and exports the strategy to the wasm schema, and rejects non-array / nullable-array columns


crates/jazz-tools/tests/gset_merge.rs
```diff
@@ -0,0 +1,431 @@
...
+        |rows| (rows.len() == 1 && rows[0].0 == doc_id).then_some(()),
+    )
+    .await;
+
+    let alice = Arc::new(alice);
+    let bob = Arc::new(bob);
+    let alice2 = Arc::clone(&alice);
+    let bob2 = Arc::clone(&bob);
+
+    let alice_handle = tokio::spawn(async move {
```
nrainhart: I think we should stay away from having non-determinism in our tests as much as possible, instead controlling the order in which sync messages are sent between nodes.

In this case for example, it'd be nice to assert that the final array converges to the same ordering regardless of the order in which writes are propagated across nodes.

I just merged a `TestingServer::block_messages_to` util that goes in this direction. If we can use it here as well, that'd be great. We may need to extend this API though (and maybe move it? I'm not sure if we want blocking on servers, clients, or in the "connection" itself)

nrainhart (commented):
The implementation looks good to me! Left only one comment around testing

crates/jazz-tools/tests/gset_merge.rs
joeinnes: Makes sense. Adapted the tests to use your helper, what do you think?

crates/jazz-tools/tests/gset_merge.rs
```diff
@@ -0,0 +1,465 @@
...
+    )
+    .await;
+
+    alice.shutdown().await.expect("shutdown alice");
+    bob.shutdown().await.expect("shutdown bob");
+    server.shutdown().await;
+}
+
+#[tokio::test]
+async fn concurrent_write_never_drops_a_peers_element() {
```
nrainhart: Isn't this case redundant with the previous one?

crates/jazz-tools/tests/gset_merge.rs
```diff
@@ -0,0 +1,465 @@
...
+
+/// Drives a deterministic concurrent merge instead of racing two writes.
+///
+/// `second` is blocked from receiving server traffic for the duration, so it
+/// writes without ever observing `first`'s update — a genuine concurrent write
+/// sharing the same ancestor — while the server is forced to process `first`'s
+/// write strictly before `second`'s. Flipping which client is `first` lets a
+/// test assert convergence is identical regardless of propagation order, with
+/// no dependence on the async scheduler.
+async fn merge_concurrently(
```
nrainhart, line 75: ✨

nrainhart (approved):
Thanks for addressing my comment!

crates/jazz-tools/tests/gset_merge.rs
joeinnes: Not any more :)

PR #1033, opened 2026-06-18
test: query retrieves rows required for permission evaluation

nrainhart:
## Description

Validates https://github.com/garden-co/jazz/issues/742 is already fixed.

Also migrates Rust policy integration tests to use the permissions DSL.


gdorsi (approved):
LGTM

PR #1037, opened 2026-06-19
Migrate the opfs-btree benchmarks to use a real browser and realistic data

gdorsi:
## Summary

Rewrite the opfs-btree benchmarks to use a real browser as execution env, realistic data on two different shapes (normal-size and large-size data) and compare it with sqlite.

Done this to have a solid base to use to validate other optimizations I'm working on.

The suite is structured around a shared `bench-core` crate: the workload is declared in `bench-core/src/benchmarks.rs`, each phase's semantics are written once in `phases.rs`, and each engine (opfs-btree, SQLite) is a small `BenchEngine` trait impl. Phase operation streams are generated deterministically from a fixed seed, so both engines replay identical operations and the harness can assert matching checksums.

Profiles: `objects` = ~10.7k Met museum-object records (~900 B each); `wikipedia` = 500 real article-wikitext records (large-value path).

**SQLite is tuned to a fair baseline.** Profiling showed it spends ~91% of its time in OPFS synchronous I/O; matching its page size to opfs-btree's 16 KB (vs SQLite's 4 KB default) roughly halves its total time, so the comparison below is against that tuned baseline rather than a strawman. (`locking_mode=EXCLUSIVE` was A/B-tested and made it slower, so it's not used.)

## Benchmark Results

Command:

```bash
pnpm --dir crates/opfs-btree run bench:compare
```

Median of 3 runs.

| profile | phase | btree_ms | sqlite_ms | btree_ops_s | sqlite_ops_s |
|---|---:|---:|---:|---:|---:|
| objects | load | 61.60 | 109.20 | 173815 | 98049 |
| objects | get_seq | 11.70 | 14.30 | 3660513 | 1497483 |
| objects | get_random | 12.40 | 17.80 | 1612903 | 1123596 |
| objects | get_skewed | 10.40 | 14.10 | 1923077 | 1418440 |
| objects | range_random | 16.40 | 14.70 | 121951 | 68027 |
| objects | update_random | 25.40 | 324.30 | 393701 | 30836 |
| objects | mixed_70_20_10 | 29.00 | 307.40 | 344828 | 32531 |
| objects | cold_get_random | 66.90 | 58.50 | 74738 | 85470 |
| wikipedia | load | 7.30 | 63.00 | 68493 | 7937 |
| wikipedia | get_seq | 10.20 | 10.00 | 849057 | 300000 |
| wikipedia | get_random | 10.00 | 11.80 | 1100000 | 312500 |
| wikipedia | get_skewed | 10.10 | 10.80 | 1800000 | 370370 |
| wikipedia | range_random | 10.20 | 11.60 | 39604 | 51724 |
| wikipedia | update_random | 23.40 | 187.00 | 42735 | 5348 |
| wikipedia | mixed_70_20_10 | 14.50 | 86.80 | 68966 | 11521 |
| wikipedia | cold_get_random | 21.20 | 32.90 | 23585 | 15198 |

Reads are roughly a tie (SQLite even edges ahead on objects `range_random`/`cold_get_random`); opfs-btree leads on writes — bulk load ~2–9×, random updates and mixed ~6–13×. Single in-browser run, directional rather than statistically rigorous.


nrainhart (approved):
Nice work! The harness looks easy to extend, it'll be nice to test other storage engines in the future

PR #1038, opened 2026-06-22
perf(opfs-btree): bitmap freelist allocator

gdorsi:
## Summary

Replace the opfs-btree in-memory free-page list (a sorted `Vec` + `HashSet` pair) with a `FreeBitmap` (`Vec<u64>` + count):

- O(1) `insert`/`remove`
- word-scan `highest` / contiguous-run search
- ascending iteration with **no per-flush sort**

Every large/overflow allocation previously re-collected and sorted all free ids to find a contiguous run, and each flush re-sorted the whole set to serialize it. The on-disk freelist format (linked list of id chunks) is unchanged — this is purely the in-memory working set.

Stacked on #1037 (base: `guido/storage-benchmark`), rebased onto its **page-size-tuned SQLite baseline**.

## Benchmark difference

Median of 3 in-browser runs. The freelist change targets the **overflow / large-value write path** — `wikipedia`, whose values exceed the 4 KB overflow threshold, so updates churn multi-page extents and hit the contiguous-run allocator. That's where it moves the needle, and where opfs-btree's lead over SQLite widens:

| wikipedia phase | btree baseline | btree **+freelist** | freelist Δ | SQLite (tuned) | btree+freelist vs SQLite |
|---|---:|---:|---:|---:|---:|
| update_random | 21.0 | 13.0 | **−38%** | 170.7 | **13.1× faster** |
| mixed_70_20_10 | 14.2 | 12.3 | −13% | 80.2 | 6.5× faster |
| load | 8.4 | 7.1 | −15% | 57.9 | 8.2× faster |

`objects` (~900 B values, never overflow) doesn't exercise the extent allocator, so the freelist is **neutral** there (±2%, noise) — opfs-btree stays ~12–13× faster than SQLite on its update/mixed phases regardless. Read phases are a tie between the engines on both profiles.

Single-machine, median of 3 (±~20% per run); SQLite is the page-size-tuned baseline from #1037 (matching opfs-btree's 16 KB page).

All 73 opfs-btree tests pass with the bitmap allocator.

### DIFF
--- a/crates/opfs-btree/src/free_bitmap.rs
+++ b/crates/opfs-btree/src/free_bitmap.rs
@@ -134,13 +134,51 @@ impl FreeBitmap {
     /// long is free. `count` must be >= 1.
     pub(crate) fn take_run(&mut self, count: usize) -> Option<PageId> {
         let run_start = self.find_run(count)?;
-        // `find_run` just confirmed every bit in the run is set, so each
-        // `remove` succeeds and the count stays accurate.
-        for offset in 0..count as u64 {
-            self.remove(run_start + offset);
-        }
+        self.remove_run(run_start, count);
         self.trim_trailing_empty_words();
         Some(run_start)
+    }
+
+    /// Clears the `count` consecutive free bits starting at `start` using
+    /// word-level masking: at most one masked write per 64-page span the run
+    /// touches, instead of one `remove` (with its own `locate` + bounds check)
+    /// per page. Callers must have confirmed the whole range is free, as
+    /// `take_run` does via `find_run`; `free_count` is decremented by the
+    /// number of bits actually cleared.
+    fn remove_run(&mut self, start: PageId, count: usize) {
+        debug_assert!(count >= 1);
+        let last = start + count as u64 - 1;
+        let start_word = (start / BITS_PER_WORD) as usize;
+        let last_word = (last / BITS_PER_WORD) as usize;
+
+        // Bits at or above `start` within the first touched word.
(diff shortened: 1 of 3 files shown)


crates/opfs-btree/src/free_bitmap.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    /// Finds the lowest run of `count` consecutive free pages, marks them
+    /// allocated, and returns the start page. Returns `None` if no run that
+    /// long is free. `count` must be >= 1.
+    pub(crate) fn take_run(&mut self, count: usize) -> Option<PageId> {
+        let run_start = self.find_run(count)?;
+        // `find_run` just confirmed every bit in the run is set, so each
+        // `remove` succeeds and the count stays accurate.
+        for offset in 0..count as u64 {
+            self.remove(run_start + offset);
```
nrainhart: I wonder if having a "removeRun" method that makes only one lookup instead of N would make any difference

nrainhart (approved):

crates/opfs-btree/src/free_bitmap.rs
gdorsi: Thanks!

Benchmarks:
  - Isolated op — clearing 100k runs: 2.6× faster for 2-page runs, scaling to ~88× for 64-page runs.
  - End-to-end — overflow-churn write path (4k overflow values + 40k random updates via the public OpfsBTree API): within noise (1399ms baseline vs 1399ms changed).

Committed because there is an improvement in isolation, the complexity added is quite low.

PR #1040, opened 2026-06-22
feat(inspector): Data Explorer UI overhaul

nrainhart:
## Description

Each commit encapsulates one change. Most notable ones:
- remove right-hand sidebar. Inserts and edits are now done directly on the table's grid
- inline filter panel so that it always appears on top of the grid
- we can now batch multiple inserts and deletes as part of the same "transaction"
- sidebar width and open/closed state is persisted in localStorage 

See the commit list for the full changelog.

## Demo

https://github.com/user-attachments/assets/637a07e7-54a9-4eb6-ba6a-b24779db02a9


packages/inspector/src/components/data-explorer/TableDataGrid.tsx
```diff
@@ -1078,46 +1229,152 @@ function QueuedCellEditor({
...
         }}
       />
-    );
-  }
+      {nullAction}
+    </div>
+  );
+}
 
+function CrossIcon({ className }: { className?: string }) {
```
nrainhart, line 1289: Not sure what people's take on icons is in this new AI age. I thought of using an icon library to avoid the slop you sometimes see in AI-generated svgs, but tbh the ones Codex generated were pretty good, so didn't feel the need to start using a library just yet

gdorsi (approved):

PR #1045, opened 2026-06-24
refactor: simplify server configuration utils

nrainhart:
## Description

Based on community feedback, I want to add better documentation around how to test Jazz apps. Before doing that, I refactored our server configuration utils to remove duplication and make them simpler to explain and navigate:
- now there's only one way to create local servers:
  - merged `DevServer` and `TestingServer` into `JazzServer`
  - added a new `TestJwtIssuer` test util for scenarios that need to generate JWTs
  - servers are always created in `jazz-tools` using `startLocalJazzServer`
- simplified `jazz-tools/testing` exports
	- removed `pushSchemaCatalog` and replaced all usages with `deploy` (which handles publishing schemas, permissions & migrations) 
	- this is the only util exported for tests, alongside `startLocalJazzServer`
	- we can export `pushSchema`, `pushPermissions` and `pushMigrations` later if needed
- moved all catalogue related logic out of the CLI layer and into `jazz-tools/dev`


gdorsi (approved):
LGTM!

Could you add a changeset?

gdorsi:
PS: CI is failing

PR #1047, opened 2026-06-25
docs: write a design document for opfs-btree

gdorsi:


crates/opfs-btree/design.md
```diff
@@ -0,0 +1,448 @@
+# opfs-btree Design
+
+`opfs-btree` is a single-writer, crash-consistent, embedded B+tree key-value
+store. It is the cross-platform storage engine used by Jazz core. The same
+code runs on native filesystems (`StdFile`), in the browser on top of an OPFS
```
nrainhart, line 5: It's probably worth clarifying that even though opfs-btree technically runs on native platforms, we only use it in the web

antoniomuso (approved):

nrainhart (approved):
The design doc is super clear to follow (and really instructive, I didn't know that much about OPFS internals). Thanks for adding it!!

PR #1050, opened 2026-06-25
Fix jazz-tools MCP startup on Windows

Waquo:
## Summary

Fix `jazz-tools mcp` on Windows by converting the local MCP server path to a `file://` URL before passing it to dynamic `import()`.

Raw Windows absolute paths like `C:\...` are parsed by Node's ESM loader as an unsupported `c:` URL scheme.

## Test

- `node --check packages\jazz-tools\bin\jazz-tools.js`
- `pnpm --filter jazz-tools exec vitest run src/mcp/integration.test.ts --config vitest.config.ts`


nrainhart (approved):
Thanks for the contribution!!

PR #1052, opened 2026-06-25
 feat: unify batches/txs as mergeable/exclusive transactions

nrainhart:
## Description

Unify batches and transactions as two different kinds of transactions:

- `Db.transaction(callback)` and `Db.beginTransaction()` now create "mergeable" transactions (previously `Db.batch` and `Db.beginBatch`), while `Db.exclusiveTransaction(callback)` and `Db.beginExclusiveTransaction()` (previously `Db.transaction(callback)` and `Db.beginTransaction()`) create serializable exclusive transactions.
- Write handles expose `transactionId` instead of `batchId`.
- The `.wait()` method on exclusive transactions does not expect a durability tier, since it resolves when the authority approves the transaction.

This rename was only applied across the TS layer and the Rust bindings. The Rust core preserves the batch/transaction naming.


aeplay (changes requested):

Reviewed briefly alongside #1124. The concern is not the model change itself but
what it does to existing callers silently.

**`Db.transaction()` changes meaning without changing shape.**
`Db.beginTransaction()` now creates a **mergeable** transaction
(`packages/jazz-tools/src/runtime/db.ts:2418-2427`) and `Db.transaction()`
invokes it (`:2440-2460`). Previously those names meant the transactional /
exclusive mode; exclusive behaviour has moved to the new
`beginExclusiveTransaction` / `exclusiveTransaction` (`:2429-2438`, `:2462-2482`).

So existing callers keep type-checking and **silently lose serializable,
authority-validated all-or-nothing grouping** unless every call site is
migrated. The removed `batch` APIs fail loudly; the repurposed `transaction`
APIs do not. Silent semantic substitution under an unchanged signature is the
worst version of a breaking change.

**Retry semantics diverge at the same boundary.** Mergeable waits retain
caller-selected durability (`runtime/client.ts:456-503`); exclusive waits use
the authority/global confirmation path (`:510-540`, `:1083-1086`). A retry after
an exclusive rejection cannot be treated like a retry after a mergeable
transaction whose individual writes have already propagated — so caller retry
logic written for the old semantics is now wrong in a way tests would not catch.

**The changeset declares a `patch` release**
(`.changeset/transaction-kind-terminology.md:1-9`). That does not match a change
in the guarantees of an existing public API.

To be fair to the PR: it preserves the underlying Rust batch modes, so this is a
TS/binding remap rather than a core atomicity rewrite. Tests cover callback
rollback and explicit-handle isolation, and the browser integration covers an
exclusive same-row conflict whose loser is rejected
(`tests/browser/db.transaction-reads.test.ts:273-334`). None of that addresses
the migration risk — there is no compatibility shim and no test showing that a
pre-existing `Db.transaction` caller cannot silently receive mergeable
semantics.

Options, in rough order of preference: keep `transaction`/`beginTransaction`
meaning exclusive and name the new mode explicitly; or rename loudly so old call
sites fail to compile; or, if the remap is intended as-is, make it a major with
a migration note. Any of those is fine — the current combination of "same name,
weaker guarantee, patch release" is the one to avoid.

It is also CONFLICTING and will need a rebase regardless.

aeplay:
🤖 Closing as superseded by #1648, which established the new-core transaction identifiers and vocabulary without compatibility aliases.

PR #1053, opened 2026-06-26
Inspector as a dev-plugin overlay

gdorsi:
## Summary

Makes the Jazz inspector a **dev-plugin overlay**: the dev plugins (Vite, SvelteKit, Next) serve the inspector locally as static HTML and inject it into the running dev app  wired to the app's **live in-process client** over the existing `jazz-devtools-v1` bridge.

Enabled by default in the dev plugind, can be disabled with config.inspector = false

- **Overlay UI** — a draggable floating Jazz-logo toggle (or `Alt+Shift+J`) opens a full-width, resizable **bottom dock** hosting the embedded inspector, served at `/__jazz/embedded`. Built as a **shadow-DOM custom element** so its styles can't leak into (or be restyled by) the host app; reconnect-safe with `AbortController` teardown.
- **Serving** — Vite/SvelteKit attach dev middleware for `/__jazz/embedded/*`. Next has no dev-middleware hook, so the plugin runs a tiny `127.0.0.1` asset server and proxies to it via a **dev-only `rewrites()`**
- **Packaging** — the inspector's embedded build is **bundled into jazz-tools' own `dist` at publish time**, and `jazz-inspector` is a **private, unpublished** workspace package. So jazz-tools is self-contained: **no `jazz-inspector` dependency, no `pnpm.overrides`, no changeset fixed-group entry, no separate publish, and no turbo build cycle**. `serve.ts` resolves the embedded build from jazz-tools' **own `dist`** in any consumer install (works anywhere), falling back to the sibling inspector package only for this monorepo's dev loop.

### Inspector data-path fixes (pre-existing, surfaced by making the inspector easy to reach)

- **Data explorer rows** — forward native row deltas over the bridge so the grid renders rows and live updates (the redundant `__jazzNativeRowDelta` marker was dropped across Rust + TS; native vs wire deltas are discriminated by shape).
- **Editing values** — insert/update/delete over the bridge, run with the inspected page's session so row-level policies authorize them as the app's own writes.
- **Live Query** — enable `devMode` at client creation so mount-time subscriptions are traced.

https://github.com/user-attachments/assets/608cda31-69c1-43ed-a199-3edea538dc38

Screenshot 2026-06-26 at 13 06 56

Verified the overlay end-to-end in real apps (Vite, SvelteKit, Next): toggle mounts by default in dev, `Alt+Shift+J` toggles it (macOS-correct via `e.code`), the embedded inspector loads and shows live tables/rows. The release wiring was validated with a `dry-run` of the publish workflow (jazz-inspector build → guard → pack-verify all green).


packages/inspector/package.json
```diff
@@ -1,37 +1,45 @@
...
   },
-  "dependencies": {
-    "@tanstack/react-table": "^8.21.3",
-    "jazz-tools": "workspace:*",
-    "react": "^19.2.0",
-    "react-data-grid": "7.0.0-beta.59",
-    "react-dom": "^19.2.0",
-    "react-resizable-panels": "^4.9.0",
-    "react-router": "^7.13.0"
-  },
```
gdorsi, line 24: Moved all the deps as devDependencies.

This way making the inspector a dependency of jazz-tools isn't problematic

packages/jazz-tools/src/dev-tools/dev-tools.ts
```diff
@@ -5,6 +5,7 @@ import {
   InsertValues,
```
gdorsi, line 5: Done some changes to the dev-tools protocol because insert/update ops weren't working anymore.

examples/todo-client-localfirst-react/src/App.tsx
```diff
@@ -27,28 +24,6 @@ type AppProps = {
   fallback?: React.ReactNode;
 };
 
-function DevToolsRegistration() {
```
nrainhart, line 30: ✨

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -1782,6 +1782,16 @@ export async function createDbWithRuntimeModule<RuntimeConfig extends DbConfig>(
...
 
   let resolvedConfig = { ...config };
+
+  // Enable subscription tracing in dev so the inspector's Live Query sees the
+  // subscriptions an app creates on mount, with no app config. Keyed on
+  // "development" specifically (not `!== "production"`) to keep tracing off under
+  // `test`/SSR; the runtime is the only client-creation seam shared by every
+  // binding (svelte/vue/solid receive a pre-created client), so it lives here
+  // rather than in the provider. Overridable explicitly via config.devMode.
+  if (resolvedConfig.devMode === undefined && process.env.NODE_ENV === "development") {
```
nrainhart, line 1792: Is this browser-only, or does it also affect RN?

packages/jazz-tools/src/dev-tools/extension-panel.ts
nrainhart, line 527: It'd be nice to add a TS doc explaining what this JazzClient proxy does

nrainhart (approved):
:shipit:

packages/jazz-tools/src/dev-tools/extension-panel.ts
gdorsi: This is going to be dropped by the stacked PRs.

packages/jazz-tools/src/runtime/db.ts
gdorsi: Good catch, we also don't need this in the stacked PRs so going to revert it.

gdorsi: Done on 05b2ddde7

PR #1059, opened 2026-06-30
Inspector dev-overlay: UI polish

gdorsi:
Stacked on #1053. UI polish for the in-app inspector dev-overlay.

## Highlights

**Slimmer chrome** — the 30px title bar is gone, replaced by a 7px drag-to-resize strip; **Close** moved into the inspector's top bar (it posts a close message up to the overlay chrome). **Settings** is now a tab next to Data Explorer / Live Query, and the tables-panel expand/collapse button is gone (the panel stays open and is still resizable).

Data Explorer

**Settings tab** — a launcher toggle (hide the button, reopen with `⌥⇧J`) and a keyboard-shortcuts reference.

Settings

**Popover-API tooltips** on the action buttons — rendered in the top layer, so they escape the overlay's `overflow: hidden` panels and fire even on disabled buttons.

Tooltip

## What's in here

- **Design tokens** (`tokens.css`) replace hand-copied hex across the data-explorer surface and the overlay chrome; inactive-tab contrast bumped to AA.
- **Compact density** pass, **auto-select** the first table on open, and a grid **loading skeleton** distinct from the empty state.
- **Filter toolbar**: the where-clause builder and the row-mutation actions are separated by a divider; delete is disabled until rows are selected.
- Dropped the stale `devMode: true` hint in the Live Query empty state.

## Notes

- The inspector is a private package, so no changeset (the feature changeset lives on #1053).
- Verified: `tsc` (inspector + jazz-tools), oxfmt, oxlint, and the 69 inspector unit tests.


.github/assets/inspector-overlay/data-explorer.png
nrainhart, line 1: We should remove these assets, right?

packages/inspector/src/styles/tokens.css
```diff
@@ -0,0 +1,36 @@
+/*
+ * Inspector design tokens — the single source of truth for the dark theme.
+ *
+ * Before this file the same palette was hand-copied as raw hex across every CSS
+ * module (and the overlay chrome in jazz-tools `loader.ts`, which mirrors these
+ * values on its shadow `:host`). Reach for a semantic token below instead of a
+ * literal; if a value genuinely has no role here, it's a deliberate one-off
+ * (component state tints, animation washes) and stays inline.
```
nrainhart:
```suggestion
 * Reach for a semantic token below instead of a
 * literal; if a value genuinely has no role here, it's a deliberate one-off
 * (component state tints, animation washes) and stays inline.
```

packages/inspector/src/pages/live-query/index.tsx
```diff
@@ -309,7 +309,7 @@ function ExtensionLiveQuery() {
         <section className={styles.emptyState}>
           <p className={styles.emptyTitle}>No active subscriptions</p>
           <p className={styles.emptyText}>
-            Open a page with `devMode: true` and create a live query to see it here.
+            Create a live query in the inspected page to see it here.
```
nrainhart: Do we use the "live query" term anywhere else in Jazz? We use "Subscriptions" way more often: https://jazz.tools/docs/reading/queries#subscriptions

packages/inspector/src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * Hover/focus tooltip built on the native Popover API. The bubble is a
+ * `popover="manual"` element, so it renders in the browser's top layer and
+ * escapes the inspector's many `overflow: hidden` panels — a plain absolutely
+ * positioned tooltip would be clipped by the toolbar or grid.
+ *
+ * Listeners live on a wrapping span (not the trigger) so the tooltip still fires
+ * for disabled buttons, which don't emit pointer events themselves. Positioning
+ * is done in JS rather than CSS anchor positioning for Safari/Firefox support.
+ */
+export function Tooltip({ label, placement = "top", children }: TooltipProps) {
```
nrainhart, line 25: Do we have a stance on using a design system (e.g. https://ui.shadcn.com/) vs building our own? I'd prefer to use something out of the box and avoid subtle bugs (not something we need to change in this PR though)

packages/inspector/src/components/inspector-layout/index.tsx
```diff
@@ -1,67 +1,39 @@
...
 }
 
 export function InspectorLayout() {
+  const { isOverlay } = useDevtoolsContext();
   const standaloneContext = useStandaloneContext();
-  const location = useLocation();
-  const [isTablesPanelOpen, setIsTablesPanelOpen] = useLocalStorageState(
-    TABLES_PANEL_OPEN_STORAGE_KEY,
-    true,
-  );
```
nrainhart, line 42: Any strong reasons for removing this? I liked being able to collapse the sidebar 🥲

packages/inspector/src/components/data-explorer/TableDataGrid.tsx
```diff
@@ -719,7 +720,13 @@ export function TableDataGrid() {
     [queryPropagation],
   );
 
-  const rows = useAll<DynamicTableRow>(queryBuilder, queryOptions) ?? EMPTY_ROWS;
+  // `undefined` means the live query hasn't resolved yet (loading); `[]` means
+  // it resolved and is genuinely empty. Collapsing both to EMPTY_ROWS made
+  // "loading" and "no rows" look identical — keep them apart so the grid can
+  // show a skeleton while the first result is in flight.
```
nrainhart, line 724:
```suggestion
  // `undefined` means the live query hasn't resolved yet (loading); `[]` means
  // it resolved and is genuinely empty. Keep them apart so the grid can
  // show a skeleton while the first result is in flight.
```

nrainhart (approved):
Love the UI tweaks! Left only some minor comments and a plead to keep the sidebar collapse button 😛

packages/inspector/src/components/tooltip/Tooltip.tsx
gdorsi: I went for using the new browser Popover API here which should do most of the heavy lifting.

In general, we can evaluate options, but we should first evaluate the stack (e.g. do we want to stick with Vanilla CSS or move to tailwind?)

packages/inspector/src/components/inspector-layout/index.tsx
gdorsi: I hate having in the tab navigation a button that's specific to one of the tabs.

Can bring it back until I find a better way to handle the layout

packages/inspector/src/pages/live-query/index.tsx
gdorsi: Correct, changed.

.github/assets/inspector-overlay/data-explorer.png
gdorsi: Oh, right, thanks.


Removed.

PR #1060, opened 2026-06-30
transport: in-band server→client Redirect (edge-fallback)

Elfo404:
Adds an opt-in, in-band redirect so a sync server can tell a client to reconnect to a different public URL — without dropping the connection. This is the protocol half of edge-fallback routing: when a client is proxied to the region where its app actually lives, that region can hand the client a direct URL so the long-lived session stops traversing the proxy.

## Protocol (J1)

- New `ServerEvent::Redirect { url }` variant — binary tag 6 + JSON form, with encode/decode and round-trip tests.
- Client honors it in the shared transport (`transport_manager`): re-points its target by swapping the advertised origin onto the existing `/apps/<id>/ws` path (`rebase_ws_url`), then reconnects. In-memory CoValues live in `RuntimeCore`, so the reconnect resyncs via `catalogue_state_hash` — no state loss.
- Consecutive redirects are capped (reset on real sync traffic) as a loop guard. Honored on both the tokio and wasm connected loops.
- **Backward compatible with no capability handshake:** an older client that receives the unknown tag skips the undecodable frame (`decode_payload(...) else { continue; }`) and stays put — it does not drop. Behavior-neutral until a server emits `Redirect`.

## Server (J2)

- New `--public-url` / `JAZZ_PUBLIC_URL`, plumbed `main` → `commands::server::run` → `ServerBuilder` → `ServerState` (mirrors `--upstream-url`).
- On WS accept, after the `Connected` frame, if the request carries `x-jazz-forwarded` **and** a public URL is configured, the server sends `ServerEvent::Redirect { url }` and keeps serving. Gating is a pure, unit-tested `forwarded_redirect_target()` helper. Unset URL or non-forwarded request → unchanged behavior.

Topology stays out of jazz: the server only advertises the URL it's given via config and only acts on a header set upstream — no region/hostname is hardcoded.

## Tests

Protocol round-trip (binary + JSON), `rebase_ws_url` cases, server gating helper, plus 14/14 `edge_server_sync` integration. Builds native + wasm + napi.

## Deferral

The **forced/early close** (server proactively dropping a forwarded connection) is intentionally out of scope — the server only offers the redirect and keeps serving; the client migrates at its own pace and closes its own old socket.

The infra/deployment wiring that sets `x-jazz-forwarded` and `JAZZ_PUBLIC_URL` lives in a separate (private) gitops repo.

### DIFF
--- a/.changeset/edge-fallback-redirect.md
+++ b/.changeset/edge-fallback-redirect.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": minor
+---
+
+Add an opt-in, in-band server→client `Redirect` message: a sync server can tell a client to reconnect to a different public URL without dropping the connection. It is gated by the new `--public-url` / `JAZZ_PUBLIC_URL` server option and an `x-jazz-forwarded` request header — the server keeps serving, and older clients that don't recognize the message simply ignore it (no capability handshake). On receipt, the client re-points its transport to the advertised origin (keeping its `/apps/<id>/ws` path) and reconnects, preserving in-memory state. Enables edge-fallback routing, where a client proxied to the region its app actually lives in is handed a direct URL so the long-lived session bypasses the proxy.
--- a/crates/jazz-tools/src/commands/server.rs
+++ b/crates/jazz-tools/src/commands/server.rs
@@ -18,6 +18,7 @@ pub async fn run(
     in_memory: bool,
     auth_config: AuthConfig,
     upstream_url: Option<String>,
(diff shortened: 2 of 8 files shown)


.changeset/edge-fallback-redirect.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": minor
```
Takeno, line 2: patch

PR #1061, opened 2026-06-30
Inspector overlay: own worker connection

gdorsi:
Stacked on #1059.

**TL;DR:** the embedded inspector overlay no longer reaches the data layer through the postMessage "devtools protocol" bridge (`jazz-devtools-v1`). Instead it opens its **own worker client that joins the host's existing broker SharedWorker** — so it reads the host's actual local store (including unsynced local-only rows) and works offline — and reads connection config + schema from a same-origin handle the host publishes on `window`. Nothing is deleted here: the browser-extension panel keeps working over the bridge, and both the bridge and the extension build are deleted in the stacked follow-up #1069.

## Why

The bridge had two structural problems:

1. **Cross-realm hazards.** Proxying a live `Db` across the iframe boundary breaks `instanceof` / `Date` / `Error` identity across realms, and dead-iframe listeners could fire synchronously inside the host app.
2. **A whole extra build.** The browser-extension/devtools build existed only to host the other side of the bridge.

## Architecture

```mermaid
flowchart TB
  subgraph Browser["Browser — same origin"]
    direction TB
    subgraph Host["Host app window — app's own bundle"]
      HApp["App (JazzProvider)"]
      HDb["Host Db<br/>persistent driver"]
      Handle["window.__jazzInspectorHost<br/>· connection config<br/>· wasm schema<br/>· active subscriptions"]
      HApp --> HDb
      HApp -. "dev plugin: installInspectorHost(db)" .-> Handle
    end
    subgraph Overlay["Inspector overlay — iframe, embedded bundle"]
      OApp["InspectorApp"]
      ODb["Overlay Db<br/>persistent · host's dbName +<br/>broker URL + identity"]
      OApp --> ODb
    end
    Broker["Broker SharedWorker (leader)<br/>owns the OPFS store + server sync<br/>identity = (worker URL, jazz-broker:appId:dbName)"]
    HDb --> Broker
    ODb == "joins — same URL + name,<br/>matching fingerprint" ==> Broker
    OApp -. "reads config + schema<br/>(window.parent)" .-> Handle
    Handle -. "pushes active subscriptions<br/>(one-way postMessage)" .-> OApp
  end
  Server["Jazz sync server"]
  Broker <-->|sync when online| Server
```

## How it works

**The handle.** In dev, the plugin mounts the overlay (a shadow-DOM custom element with the iframe) and calls `installInspectorHost(db, iframeWindow, origin)`, which publishes `window.__jazzInspectorHost` exposing: the connection config (appId, serverUrl, the **resolved** `dbName` and `brokerWorkerUrl`, and the host's credential), the wasm schema, and the active-subscription list. The overlay — same-origin — reads it off `window.parent`.

**Joining the host's store (local + offline).** The overlay opens its own `Db` with a **persistent** driver using the host's resolved `dbName`, broker SharedWorker URL, and identity. A SharedWorker is identified by `(scriptURL, name)`, and the broker additionally checks a config **fingerprint** on join. Three details make the host and the overlay — two different bundles — converge on the same worker *and* the same fingerprint:

- The fingerprint keys on the **resolved** broker worker URL, not the raw `runtimeSources` shape: `baseUrl`/`wasmUrl`/etc. don't affect which SharedWorker gets constructed, so a host that never configured a URL and an overlay that forwards one explicitly still fingerprint identically.
- The default worker URL stays a **statically-analyzable** `new URL("../worker/jazz-broker-worker.js", import.meta.url)` literal, so bundlers (Turbopack, webpack, Vite) emit the worker script as an asset and rewrite the URL. Routing the default through the generic runtime resolver broke Next: the pattern was no longer transformable and the worker script 404'd.
- Turbopack's rewrite yields a **root-relative** path (`/_next/static/…`), so the resolved URL is absolutized against the page before use. Without this, the host fingerprint (root-relative) and the overlay fingerprint (absolute, via the forwarded URL) differed while pointing at the same worker, and the broker rejected the join with `incompatible persistent browser configuration`.

Joining the broker also sidesteps the cross-realm hazards entirely — no live `Db` ever crosses the iframe boundary.

**Schema before any query.** The schema is registered statically at `defineApp` time (`getRegisteredWasmSchema`), so the overlay renders even on a write-only page that never runs a query (e.g. a data generator). When a runtime client exists, its live (engine-normalized) schema is preferred.

**Live Query.** The host pushes a stack-less active-subscription list to the iframe one-way (`onActiveQuerySubscriptionsChange` → `postMessage`); stacks are dropped (host-realm-only).

**Removed: nothing yet.** The extension devtools panel keeps its bridge-connected tree (`ExtensionApp` + the bridge-driven `ExtensionLiveQuery` path, split out of the old shared `InspectorApp`), so it works unchanged. After this PR the overlay no longer uses the bridge; the bridge itself (relay, parent-window ports, the `jazz-devtools-v1` protocol, the `attachDevTools`/`DevToolsAttachment`, `createExtensionJazzClient`, `createEmbeddedJazzClient` exports) and the extension/devtools build are deleted in the stacked follow-up #1069.

## Validation

- **End-to-end (Playwright):** the overlay reads the handle, **joins** the host's broker (which rejected the mismatched client before), renders Data Explorer from the schema, and shows the host's pushed subscription in Live Query.
- **Live on `starters/next-localfirst` (Next 16, Turbopack):** the app boots and the overlay joins the host's broker and displays the host's local rows. This flow is what surfaced both Turbopack regressions above (worker-script 404, fingerprint mismatch); the starter's own Playwright e2e suite is green.
- **Live on `dev/stress-tests/todo-react`:** the Generate Data page (write-only, no `useAll`) renders the `projects`/`todos` tables with data via the statically-registered schema.
- Inspector unit suite + affected jazz-tools tests green; both packages typecheck; embedded build succeeds.

## Not in this PR

- An **admin "see everything"** mode. Admin is a server-side policy bypass — inherently online and not something to write into the host's local store — so it can't be the local/offline default. The overlay uses the host's identity; an explicit, isolated, online-only admin mode is a planned follow-up.


packages/jazz-tools/src/runtime/db.ts
```diff
@@ -1026,6 +1026,18 @@ export class Db {
     return structuredClone(this.config);
   }
 
+  /**
+   * The runtime schema of this Db's live client, normalized. Used by the
+   * inspector overlay (a same-origin iframe) to render columns and build queries
+   * against this connection without bridging or private-field access. Null if
+   * no client exists yet — run a query/subscription (or wait for connection)
+   * first.
+   */
```
nrainhart, line 1036:
```suggestion
   * @internal
   */
```

packages/jazz-tools/src/runtime/db.ts
```diff
@@ -1026,6 +1026,18 @@ export class Db {
     return structuredClone(this.config);
   }
 
+  /**
+   * The runtime schema of this Db's live client, normalized. Used by the
+   * inspector overlay (a same-origin iframe) to render columns and build queries
+   * against this connection without bridging or private-field access. Null if
```
nrainhart, line 1032:
```suggestion
   * against this connection. Null if
```

nrainhart (approved):
Looks good! Qq: do we still need the `DevToolsJazzClient` now that we have an actual Db client in the inspector?

PR #1065, opened 2026-07-01
refactor: catalogue utils to simplify testing apps with multiple schemas

nrainhart:
## Description

- Refactors catalogue utilities so tests can publish schemas and permissions directly, instead of providing paths to the files that contain them.
- Adds an `app.schemaHash` getter to easily get a schema's hash (useful for creating migrations programatically in code)
- Adds test coverage for schema migrations in Typescript.

Stack: PR 4 of 5. Depends on #1064.


packages/jazz-tools/src/schema-hash.ts
```diff
@@ -0,0 +1,32 @@
...
+    const { loadWasmModule } = await import("./runtime/client.js");
+    wasmModulePromise = loadWasmModule();
+  }
+
+  return wasmModulePromise;
+}
+
+export async function computeSchemaHash(schema: WasmSchema): Promise<string> {
+  const wasmModule = await loadSchemaHashWasmModule();
+  const runtime = new wasmModule.WasmRuntime(
```
nrainhart: When moving this code I realized we're creating a whole WASM runtime just to get a schema hash, which is pretty wasteful (and also makes this fn async).

I took a stab at replicating this logic in JS ([link to branch](https://github.com/garden-co/jazz/commit/37acf881619d5bc359d50d4310a044cfcc366b69)), but not a big fan of adding hashing algorithms on the JS side as well, and needing to keep this logic in sync between JS and Rust.

Not sure what's the best approach here, any ideas? I guess we could add this on the `Db` instance, which already has an initialized WasmRuntime (like we do with `mintLocalFirstToken`), but in this case it'd be nice to create migrations without needing a `Db`.

gdorsi (approved):

packages/jazz-tools/src/schema-hash.ts
nrainhart: Decided to add a static method to compute the hash on the Rust side. That way `computeSchemaHash` is still async, but at least we're not creating a whole runtime just to hash a schema.

Also made the from/to hashes optional when defining a schema, so that we don't need to do any hashing for tests

PR #1069, opened 2026-07-03
Inspector: delete the devtools-protocol bridge

gdorsi:
Stacked on #1061.

Now that the inspector overlay opens its own worker connection (#1061), the postMessage "devtools protocol" bridge (`jazz-devtools-v1`) has a single remaining consumer: the browser-extension devtools panel. This PR deletes both. It is the pure deletion, split out of #1061 for reviewability:

- the browser-extension/devtools-panel build: the chrome-extension assets, the devtools html entries, `ExtensionApp`, the bridge-driven `ExtensionLiveQuery` path, and the extension vite build mode — the inspector ships only the `web` (standalone) and `embedded` (overlay) builds
- `dev-tools/`: the bridge implementation — `attachDevTools`, the extension-panel client (`createDbFromInspectedPage` and friends), the protocol, and the parent-window port, including their tests
- the overlay relay (already unused since the loader started installing the host handle directly)
- `createExtensionJazzClient` / `createEmbeddedJazzClient` and the `attachDevTools`/`DevToolsAttachment` re-exports across the react/web/svelte/vue/solid bindings
- the `attachDevTools` usage in the todo-react stress test
- follow-up cleanup: `createSolidJazzClient` loses its `makeCreateSolidJazzClient({ factory })` wrapper, which existed only to share code with the extension variant
- post-migration context cleanup: `isOverlay` is derived from `runtime` instead of being a separate context member, the constant `queryPropagation` and dead telemetry param go away, and stale bridge mocks/comments are cleaned up

**Breaking:** `attachDevTools`, `DevToolsAttachment`, `createExtensionJazzClient`, and `createEmbeddedJazzClient` are no longer exported.


nrainhart (approved):
Nice 🧹

PR #1073, opened 2026-07-06
perf: add persistent batchId->rows index to speed up batch rows lookup

nrainhart:
## Description

Adds a persistent `batch_id -> local batch rows` index so `local_batch_rows` no longer needs to fall back to scanning every row locator and every row history.

The new index is stored as a system raw table and is populated whenever history rows are written through the normal storage mutation paths. `local_batch_rows` now uses only bounded lookup paths:

1. sealed batch submission
2. in-memory local batch record
3. persisted local batch record
4. persisted local batch row index

This removes the previous `scan_row_locators` fallback, which could become `O(rows × history_versions)` per batch fate lookup.


antoniomuso (approved):
LGTM, we could think to delete the lookup entries. Otherwise the index table will continue to grow forever.

PR #1077, opened 2026-07-08
Fix jazz-rn scheduler thread growth

antoniomuso:
## Summary
- Replace jazz-rn's per-schedule thread spawning with one lazy scheduler worker fed by an mpsc channel.
- Preserve tick and mutation-error debounce behavior while keeping scheduler thread count constant.
- Harden shutdown: `RnRuntime::close()` clears the callbacks and drops the worker's channel sender instead of joining the thread.
- Avoid firing a duplicate immediate tick for auto-committed direct writes in jazz-tools.
- Add changesets for `jazz-rn` and `jazz-tools`.

## Root Cause
The previous RN scheduler spawned a fresh thread for each accepted tick or mutation-error delivery. During heavy bare inserts, the debounce flag was cleared before the JS callback returned, so follow-up ticks kept spawning additional threads that all blocked on the busy JS thread — roughly one new thread per millisecond of write burst. Jobs are now queued to a single lazily-spawned worker, and the debounce interlock bounds the queue to at most one pending job per type, so thread count stays constant under any load.

## Shutdown hardening
uniffi-bindgen-react-native dispatches callback-interface calls through `invokeBlocking`: the worker blocks until the JS thread services the call, and `close()` runs on that same JS thread. Joining the worker from `close()` — or locking the callback mutex it held mid-call — could therefore deadlock the app. Instead:
- `shutdown()` drops the channel sender; the worker drains its queue (no-ops once callbacks are cleared) and exits on its own. It is never joined.
- The tick callback is stored as `Arc<dyn BatchedTickCallback>` and cloned out of the mutex before invocation, so `set_callback(None)` never waits on an in-flight callback.
- A failed worker spawn no longer panics while holding the worker mutex (which would have poisoned it and silently disabled the scheduler); the job is dropped and the next schedule retries.
- Mutation-error callbacks are wrapped in `catch_unwind` so a JS exception cannot kill the shared worker.

## Validation
- `cargo test -p jazz-rn` (includes new burst-coalescing and post-shutdown regression tests)
- `cargo test -p jazz-tools --features test`
- `cargo clippy --workspace -- -D warnings`
- `cargo fmt`, `git diff --check`
- The `close()` deadlock scenario requires a real JS thread and cannot be covered by unit tests; recommended on-device smoke test: heavy writes followed immediately by `close()`.

### DIFF
--- a/crates/jazz-rn/rust/src/lib.rs
+++ b/crates/jazz-rn/rust/src/lib.rs
@@ -264,12 +264,54 @@ pub trait MutationErrorCallback: Send + Sync {
 // RnScheduler
 // ============================================================================
 
+#[derive(Clone, Copy)]
+enum SchedulerJob {
+    Tick,
+    DeliverMutationErrors,
+}
+
+struct SchedulerWorker {
+    sender: Option<std::sync::mpsc::Sender<SchedulerJob>>,
+    handle: Option<std::thread::JoinHandle<()>>,
+}
+
+impl SchedulerWorker {
+    fn send(&self, job: SchedulerJob) -> Result<(), std::sync::mpsc::SendError<SchedulerJob>> {
+        let Some(sender) = self.sender.as_ref() else {
+            return Err(std::sync::mpsc::SendError(job));
+        };
+        sender.send(job)
+    }
+
+    fn shutdown(&mut self) {
+        drop(self.sender.take());
+
+        let Some(handle) = self.handle.take() else {
+            return;
+        };
+
+        if handle.thread().id() == std::thread::current().id() {
+            return;
+        }
+
+        let _ = handle.join();
+    }
+}
+
+impl Drop for SchedulerWorker {
+    fn drop(&mut self) {
+        self.shutdown();
+    }
+}
+
 #[derive(Clone, Default)]
 struct RnScheduler {
     scheduled: Arc<AtomicBool>,
     mutation_error_delivery_scheduled: Arc<AtomicBool>,
     core_ref: Arc<Mutex<Option<Weak<Mutex<RnCoreType>>>>>,
     callback: Arc<Mutex<Option<Box<dyn BatchedTickCallback>>>>,
+    worker: Arc<Mutex<Option<SchedulerWorker>>>,
+    shutdown: Arc<AtomicBool>,
 }
 
 impl RnScheduler {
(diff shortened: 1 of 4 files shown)


crates/jazz-rn/rust/src/lib.rs
```diff
@@ -850,12 +977,21 @@ impl RnRuntime {
     /// Flush and close the underlying storage, releasing filesystem locks.
     pub fn close(&self) -> Result<(), JazzRnError> {
         with_panic_boundary("close", || {
-            let mut core = self.core.lock().map_err(|_| JazzRnError::Internal {
-                message: "lock poisoned".into(),
-            })?;
-            let flush_result = core.flush_storage();
-            let flush_wal_result = core.flush_wal();
-            let close_result = core.storage().close();
+            let (scheduler, flush_result, flush_wal_result, close_result) = {
```
nrainhart: I'm not getting why we need this to be a nested block. Is it to drop the mutex on core before we shut down the scheduler?

crates/jazz-rn/rust/src/lib.rs
```diff
@@ -288,6 +330,118 @@ impl RnScheduler {
...
+            }
+        };
+
+        if self.shutdown.load(Ordering::SeqCst) {
+            self.clear_job_scheduled(job);
+            return;
+        }
+
+        if slot.is_none() {
+            *slot = Some(Self::spawn_worker(
```
nrainhart: What's the point of the RnScheduler/SchedulerWorker split? Wouldn't it be simpler to extend the RnScheduler?

Also: should we spawn the thread on scheduler creation, instead on first job send?

crates/jazz-rn/rust/src/lib.rs
```diff
@@ -288,6 +330,118 @@ impl RnScheduler {
...
+        mutation_error_delivery_scheduled: Arc<AtomicBool>,
+        core_ref: Arc<Mutex<Option<Weak<Mutex<RnCoreType>>>>>,
+        callback: Arc<Mutex<Option<Box<dyn BatchedTickCallback>>>>,
+    ) -> SchedulerWorker {
+        let (sender, receiver) = std::sync::mpsc::channel::<SchedulerJob>();
+        let handle = std::thread::Builder::new()
+            .name("jazz-rn-scheduler".into())
+            .spawn(move || {
+                while let Ok(job) = receiver.recv() {
+                    std::thread::sleep(Duration::from_millis(1));
```
nrainhart, line 363: Is this wait still necessary now that we have a single thread? Could we switch to using something like [recv_many](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv_many) and collapse multiple jobs into a single request_batched_tick/deliver_mutation_errors call?

nrainhart (commented):
The fix looks reasonable. I do have questions about some details, though

crates/jazz-rn/rust/src/lib.rs
antoniomuso: I removed it!

crates/jazz-rn/rust/src/lib.rs
antoniomuso: We need it for this reason:
When there's sync work in progress, jazz ticks continuously. Each tick causes the next one, like this:

1. The JS thread calls `runtime.batched_tick()`. The Rust core processes one batch of work.
2. There's still more work left, so before returning, the core calls `schedule_batched_tick()` — "call me again soon, please."
3. That puts a `Tick` job on the worker's queue. The worker picks it up and calls the JS callback `request_batched_tick()`.
4. The JS side of that callback does one tiny thing: it schedules `batched_tick()` to run on the JS thread again.
5. Go back to step 1. Repeat until the core runs out of work.

So each pass through the loop is: **one tick on the JS thread → a hop through the worker → the next tick on the JS thread**.

crates/jazz-rn/rust/src/lib.rs
antoniomuso: I simplified the code now.

crates/jazz-rn/rust/src/lib.rs
```diff
@@ -281,13 +294,134 @@ impl RnScheduler {
...
+            }
+        };
+
+        if self.shutdown.load(Ordering::SeqCst) {
+            self.clear_job_scheduled(job);
+            return;
+        }
+
+        if slot.is_none() {
+            *slot = Self::spawn_worker(
```
nrainhart, line 325: Should we spawn the thread on scheduler creation, instead of on first job send?

nrainhart (approved):
Thanks for the fix!

crates/jazz-rn/rust/src/lib.rs
antoniomuso: Having the spawn here allows us to self-heal the thread when the worker dies, since it can be respawned on every tick.

PR #1080, opened 2026-07-09
feat: migrate React `useAll` to return `{ data, isLoading, error }`

nrainhart:
## Description

Align React and React Native `useAll` with the other framework bindings by returning `{ data, isLoading, error }` instead of a bare `T[] | undefined`. `useAllSuspense` still returns `T[]`.

The return type allows proper type narrowing after validating the load is not pending/errored.

### DIFF
--- a/packages/inspector/src/components/data-explorer/TableDataGrid.tsx
+++ b/packages/inspector/src/components/data-explorer/TableDataGrid.tsx
@@ -1473,7 +1473,7 @@ function RelationCell({
     () => new GenericQueryBuilder(relationTable, schema).where({ id: relationId }).limit(1),
     [relationId, relationTable, schema],
   );
-  const relationRows = useAll<DynamicTableRow>(queryBuilder, queryOptions) ?? EMPTY_ROWS;
+  const { data: relationRows = EMPTY_ROWS } = useAll<DynamicTableRow>(queryBuilder, queryOptions);
   const relationRow = relationRows[0];
   const displayColumn = useMemo(
     () => getRelationDisplayColumn(schema, relationTable),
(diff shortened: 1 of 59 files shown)


packages/jazz-tools/src/react-core/use-all.ts
```diff
@@ -89,27 +116,51 @@ function useAllBase<T extends { id: string }>(
...
+
+  if (state?.status === "fulfilled") {
+    return { data: state.data, isLoading: false, error: null };
+  }
+
+  if (state?.status === "rejected") {
+    return { data: undefined, isLoading: false, error: toError(state.error) };
+  }
+
+  return { data: undefined, isLoading: true, error: null };
```
nrainhart: Note to self: this probably need caching to avoid unnecessary re-renders

packages/jazz-tools/src/react-core/use-all.ts
```diff
@@ -6,17 +6,44 @@ type UseAllOptions = {
...
+  | {
+      data: T[];
+      isLoading: false;
+      error: null;
+    }
+  | {
+      data: undefined;
+      isLoading: false;
+      error: Error;
+    };
```
gdorsi: Since the loading/error states are expressed via isLoading, error, shall we make data non-nullable?

packages/inspector/src/components/data-explorer/TableDataGrid.tsx
gdorsi, line 769: Test failing here

gdorsi (approved):
LGTM.

packages/jazz-tools/src/react-core/use-all.ts
nrainhart: What do you mean, exactly? With the current definition, `data` will be narrowed to the non-nullable `T[]` if you assert `isLoading` is false and there's no `error`. Alternatively, you can choose to ignore loading/error states doing:
```ts
const { data = [] } = useAll(...);
```

Forcing people to always check `isLoaded` and `error` before granting access to `data` feels more cumbersome to me.

gdorsi: Good point.

Defaults in this case are ergonomic enough

PR #1083, opened 2026-07-13
fix: restore migrated row history for deletes

### LINKED ISSUE #1079: Deletes Denied with "Missing Row Content" Error
**Version:** jazz-tools `2.0.0-alpha.52`. Reproduced with the browser client (`jazz-wasm 2.0.0-alpha.52`, OPFS dual-runtime) against both a local server from `startLocalJazzServer()` (same package version) and the cloud sync server at `v2.sync.jazz.tools`.

## Summary

After a schema migration, rows created under the previous schema hash remain readable (lens-composed reads work as documented), but a client on the new schema **cannot delete them** when the table has a row-dependent permission policy (e.g. `isCreator` / `managedByCreator()`). The client-side policy engine allows the delete and applies it optimistically; the server's User-role permission check then rejects it with:

```
permission_denied: "Delete denied on table <t> - missing row content"
```

The rollback syncs down and the row visibly pops back. This recurs for every row that predates every future migration, so any app using creator-owned policies loses the ability to delete pre-migration data permanently.

## Reproduction

1. Deploy schema A; as user `alice` (local-first session), insert a row into a table whose policy is `managedByCreator()` (or any policy whose delete rule needs row content/provenance, e.g. `allowDelete.where(isCreator)`).
2. Deploy schema B plus the A→B migration edge (any column addition) and the permissions.
3. Connect a browser client on schema B as the same `alice`. The row is visible (lens read works).
4. `db.delete(table, id)`.

Observed: the row disappears (optimistic local apply), then `onMutationError` fires with `Delete denied on table <t> - missing row content` and the row reappears.

## Additional findings

1. **Updates on the same rows are accepted**. After one accepted update, a subsequent delete succeeds. This is the basis of the workaround I'm using for now, which is to update a column before deleting (though annoyingly, this requires actually writing something meaningful).
2. **Sequencing inside one batch or one transaction does not help.** `update(touch)` + `delete` in a single direct batch, or in a single transaction, is still denied the same way. Validation seems to read committed pre-batch storage for every op rather than intra-batch state.
3. **Back-to-back separate writes do not help either.** The delete is only accepted if it arrives after the touch is *confirmed* (I wait on the update's `WriteHandle.wait({ tier: "edge" })` before issuing the delete). Fire-and-forget touch immediately followed by delete over the same connection is still denied, so the delete's permission check appears to run against a snapshot that does not yet include a just-queued update.
4. **A rejected delete "poisons" the row.** After the server rejects a bare delete and restores the row, a subsequent plain update on that row is denied with `Update denied by USING policy on table books - no old content`. Later touch attempts from that client do not settle at all (no batch fate I can observe — no `onMutationError`, no apply), leaving the row effectively immortal from the client.

## Expected behavior

Either of these would resolve it:

- The server-side permission evaluator resolves row content/provenance through lens paths / branch ancestry, the same way reads do; or
- A delete on a lens-projected row triggers the same copy-on-write materialization that updates get, so the permission check has content to evaluate.

And ideally: a rejected delete should not leave the row in a state where subsequent updates are also denied (finding 4).

joeinnes:
## Summary
- Resolve migrated row content and provenance through the row's origin history table.
- Rebuild branch tips and rejected-delete restoration across schema branches.
- Add integration coverage for migrated deletes, table renames, policy enforcement, and recovery.
- Add a `jazz-tools` patch changeset.

Fixes #1079

## Testing
- `cargo test -p jazz-tools --features test --test catalogue_sync_integration` — passed (30 tests)
- `cargo test -p jazz-tools --features test --lib rejection_recovery` — passed (19 tests)
- `cargo clippy --workspace -- -D warnings` — passed
- `cargo fmt --check` — passed
- `git diff --check origin/main` — passed
- Temporary benchmark passed; it measured absolute successful-path time, not a before/after delta. The worst measured case was 934 µs, and the benchmark was removed afterward.


crates/jazz-tools/src/query_manager/writes.rs
```diff
@@ -670,6 +670,26 @@ impl QueryManager {
             return entry.branch_frontier;
         }
 
+        let history_table = crate::storage::history_table_for_row(storage, row_id, table);
+        if let Ok(history_rows) = storage.scan_history_row_batches(&history_table, row_id) {
+            let branch_rows = history_rows
+                .iter()
+                .filter(|row| row.branch.as_str() == branch && row.state.is_visible())
+                .collect::<Vec<_>>();
```
nrainhart: I don't get why we need to change this here instead of in `load_visible_region_entry`... what's the case we need to handle differently?

If we do keep this, can we at least avoid scanning ALL branches and then filtering in memory? iirc branches are part of the storage key prefix, so we should be able to filter them on read

nrainhart (approved):
Thanks for adding the integration tests!! They will be really valuable for the core rewrite

crates/jazz-tools/src/query_manager/writes.rs
joeinnes: We need the history from the original table if the row's been migrated. If the row is deleted it won't appear in the materialised visible view, so `load_visible` won't find the previous write. We can definitely do the `row_id:branch` filter though, makes sense.

PR #1084, opened 2026-07-13
docs: correct drift across guides and examples

joeinnes:
## Summary

- document WebSocket cookie authentication and current server configuration
- replace stale Fjall, DevTools panel, and Live Query terminology with SQLite, Inspector overlay, and subscriptions
- correct and type-check framework snippets and example descriptions
- remove the unused remote layout font and label included code samples
- refresh the packaged MCP documentation index and its regression coverage

## Testing

- `pnpm --filter docs build` — passed with Turbopack; 135 pages generated
- `pnpm --filter docs test` — passed; 36 tests
- `pnpm --filter todo-client-localfirst-svelte-docs check` — passed
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/mcp/build-index.test.ts` — passed; 28 tests, 1 skipped
- `git diff --check` — passed


packages/jazz-tools/src/mcp/build-index.test.ts
```diff
@@ -425,6 +425,14 @@ describe("packaged docs index", () => {
     expect(txt).toContain("### Managing JWT changes on a live client");
     expect(txt).toContain("### Reacting to expiry and unauthenticated responses");
 
+    expect(txt).toContain("JAZZ_AUTH_COOKIE_NAME");
+    expect(txt).toContain("### Subscriptions");
+    expect(txt).toContain("NAPI SQLite storage");
+    expect(txt).not.toContain("### Live Query");
+    expect(txt).not.toContain("DevTools panel");
+    expect(txt).not.toContain("attachDevTools");
+    expect(txt).not.toContain("NAPI Fjall");
```
nrainhart: These assertions do not seem very useful, what are we trying to check?

packages/jazz-tools/bin/docs-index.txt
nrainhart, line 1: How did you update this file? If we're running a script, would it make sense to generate this file on CI? (when publishing the package, e.g.)

nrainhart (approved):
Thanks for updating the docs!

packages/jazz-tools/src/mcp/build-index.test.ts
joeinnes: Ensuring that the txt index is current and has been rebuilt properly from the current docs. There's probably a better way to do this...

packages/jazz-tools/bin/docs-index.txt
joeinnes: This was I think implemented before as part of the docs build pipeline, but can't find any evidence of that, so maybe I just *intended* to add it?

Anyway: https://github.com/garden-co/jazz/pull/1084/changes#diff-adfa337ce44dc2902621da20152a048dac41878cf3716dfc4cc56d03aa212a56R6

PR #1086, opened 2026-07-14
fix: make authoritative rejections terminal during replay

joeinnes:
## Summary

- Treat authoritative rejected fates as terminal during reconnect and history replay, failing closed if the fate scan fails.
- Withhold children whose parents have terminal rejected fates and preserve rejection when a delayed row arrives.
- Avoid echoing accepted transaction rows back to the authoritative server that supplied them while still forwarding them elsewhere.
- Record exact late members of a retained rejected seal as rejected, without making them visible.

## Trade-off

A terminally rejected write will not be retried automatically if permissions later change; the application must initiate the write again.

## Testing

- Reconnect, rejected-parent, delayed-row and source-suppression regressions

### DIFF
--- a/crates/jazz-tools/src/runtime_core/tests/accepted_batch_downgrade.rs
+++ b/crates/jazz-tools/src/runtime_core/tests/accepted_batch_downgrade.rs
@@ -61,11 +61,22 @@ fn pump<S: Storage, Sch: Scheduler>(
     server_id: ServerId,
     client: &mut RuntimeCore<S, Sch>,
     client_id: ClientId,
-) {
+) -> Vec<(ServerId, &'static str)> {
+    let mut upstream_row_writes = Vec::new();
     for _ in 0..12 {
         let mut any = false;
         client.batched_tick();
         for entry in client.sync_sender().take() {
+            if let Destination::Server(destination_server_id) = &entry.destination
+                && matches!(
+                    &entry.payload,
+                    SyncPayload::RowBatchCreated { .. }
+                        | SyncPayload::RowBatchNeeded { .. }
+                        | SyncPayload::SealBatch { .. }
+                )
+            {
+                upstream_row_writes.push((*destination_server_id, entry.payload.variant_name()));
+            }
             if entry.destination == Destination::Server(server_id) {
                 any = true;
                 server.park_sync_message(InboxEntry {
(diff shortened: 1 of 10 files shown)


crates/jazz-tools/src/runtime_core/tests/accepted_batch_downgrade.rs
```diff
@@ -312,6 +333,129 @@ fn owner_update_of_backend_created_row_is_not_destroyed() {
...
+
+    let batch_id = crate::row_histories::BatchId::new();
+    let write_context = WriteContext {
+        session: None,
+        attribution: None,
+        updated_at: None,
+        batch_mode: Some(crate::batch_fate::BatchMode::Transactional),
+        batch_id: Some(batch_id),
+        target_branch_name: None,
+    };
```
nrainhart: Can't we use `begin_batch` or some other existing high-level util instead of creating the transaction's write context by hand?

crates/jazz-tools/src/sync_manager/sync_logic.rs
```diff
@@ -43,22 +67,42 @@ impl SyncManager {
         let Ok(row_locators) = storage.scan_row_locators() else {
             return;
         };
+        let authoritative_fates = match storage.scan_authoritative_batch_fates() {
```
nrainhart, line 70: How big is this store? I.e. do we ever clear batch fates for completed batches/transactions, or does it grow unbounded?

nrainhart (approved):
Thanks for the fix!!

crates/jazz-tools/src/runtime_core/tests/accepted_batch_downgrade.rs
joeinnes: Makes sense, will do

crates/jazz-tools/src/sync_manager/sync_logic.rs
joeinnes: Yeah, it's unbounded. We're already doing a (more expensive) unbounded history scan on reconnect. We'd need to keep this as-is because completed batches can still be replayed, so unless we can compact the fates somehow along with the history, it'll probably need to stay unbounded for now.

PR #1087, opened 2026-07-15
Fix `in` filters in included relation queries

joeinnes:
## Summary
- Lower `in` filters inside included relation queries, including empty lists.
- Keep the new `Condition` variants append-only and reserve `NotIn`.

⚠️ This is a sync protocol version change because of the new Condition enum variants.

## Testing
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts tests/ts-dsl/query-api.test.ts` — passed
- Focused Rust query, array-subquery, relation-IR, and protocol tests — passed

### DIFF
--- a/.changeset/quiet-filters-sync.md
+++ b/.changeset/quiet-filters-sync.md
@@ -1,0 +1,9 @@
+---
+"jazz-tools": patch
+"jazz-wasm": patch
+"jazz-napi": patch
+"jazz-rn": patch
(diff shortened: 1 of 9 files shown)


aeplay:
🤖 Superseded by draft #1957. Current Jazz already carries in through its canonical predicate/query engine; #1957 preserves the missing modern include/live-update coverage without restoring this retired protocol path.

PR #1088, opened 2026-07-15
feat(jazz-tools): add `notIn` query filters

### LINKED ISSUE #1082: Add `notIn` operator to typed table query filters
## Summary

Please add a `notIn` operator to the typed table query DSL, with parity across the column types and query contexts that currently support `in`.

For example:

```ts
table.where({
  status: { notIn: ['archived', 'deleted'] },
});
```

## Current behavior

In `jazz-tools@2.0.0-alpha.52`, `in` is available in typed filters, but `notIn` is not. The equivalent query currently requires chaining one `ne` predicate per excluded value:

```ts
table
  .where({ status: { ne: 'archived' } })
  .where({ status: { ne: 'deleted' } });
```

Chained `.where()` calls are combined with `AND`, so this produces the expected result.

## Why add `notIn`

The workaround becomes verbose and awkward when the exclusion list is dynamic, and it does not communicate the intent as clearly as the existing `in` operator. Native support would also make positive and negative membership queries symmetrical.

## Expected behavior

```ts
table.where({ status: { notIn: ['archived', 'deleted'] } });
```

should return rows whose `status` matches neither value, with semantics equivalent to the chained `ne` predicates above.

Ideally, `notIn` would be supported anywhere the typed query DSL currently supports `in`.

joeinnes:
## Summary

- add typed `notIn` membership filters across supported table column types
- lower `notIn` to conjunctive `ne` predicates for root and included relation queries
- expose `notIn` in Inspector filtering and document operator support

Closes #1082

Should probably land in the same release as #1087 for protocol change simplicity.

## Testing

- `pnpm lint`
- `pnpm build:core`
- affected Jazz Tools tests
- Better Auth adapter tests
- Inspector tests
- Jazz Tools and Inspector TypeScript checks
- Oxfmt and Cargo formatting checks

### DIFF
--- a/.changeset/smart-apples-exclude.md
+++ b/.changeset/smart-apples-exclude.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Add typed `notIn` filters to table queries for excluding rows whose column value matches any value in a list.
--- a/docs/content/partials/where-operators-table.mdx
+++ b/docs/content/partials/where-operators-table.mdx
@@ -1,15 +1,15 @@
 Operator support by column type (TypeScript):
 
-| Column kind                   | Operators / shape                          |
-| ----------------------------- | ------------------------------------------ |
-| `id`                          | `eq`, `ne`, `in`                           |
-| `s.string()`                  | `eq`, `ne`, `contains`, `in`               |
-| `s.boolean()`                 | `eq`, `ne`, `in`                           |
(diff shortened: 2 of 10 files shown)


aeplay:
🤖 Related roadmap issue: #1082.

aeplay:
🤖 Superseded by draft #1957, which exposes notIn as canonical Not(In(...)) for root and included queries instead of adding a separate client-side ne-expansion path.

PR #1090, opened 2026-07-16
fix: make persisted batch recovery complete and retryable

joeinnes:
## Summary

- Reapply persisted accepted fates during restart recovery before retiring sealed submissions.
- Use a batch-membership source only when every declared member resolves exactly.
- Retain rejected-batch recovery manifests until every declared member has been handled.
- Retry server rows when fate persistence succeeds but row materialisation fails, with loud failure logging.

## Testing

- Restart, partial-membership and materialisation-retry regressions

### DIFF
--- a/.changeset/recover-incomplete-batch-settlement.md
+++ b/.changeset/recover-incomplete-batch-settlement.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Make persisted batch-fate recovery complete and retryable after restarts, missing batch members, or row materialisation failures.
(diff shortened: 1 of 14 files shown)


nrainhart (approved):
Code changes look good! Do you think we can write any integration tests using only the user-facing APIs? Those are the only ones that will survive the core rewrite and avoid regressions (can be done as a FUP)

aeplay:
🤖 Closing because the old sync-manager recovery implementation was replaced by the new-core fate and replay work in #1292 and #1295. The crash/replay scenario remains useful as a test-audit receipt.

PR #1091, opened 2026-07-16
fix: persist direct batch settlement before publishing rows

joeinnes:
## Summary

- Load every exact direct-batch member before settlement, so a failed load cannot publish a partial batch.
- Persist the sealed submission, local recovery record and authoritative fate before making direct rows visible.
- Retain sealed recovery state when publication fails so retry or restart can finish the batch safely.

## Testing

- Direct-batch regression suite

### DIFF
--- a/crates/jazz-tools/src/runtime_core/writes.rs
+++ b/crates/jazz-tools/src/runtime_core/writes.rs
@@ -1122,6 +1122,7 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
             record.mark_sealed(submission.clone());
             let mut settled_at_commit = false;
             let mut confirmed_tier_at_commit = None;
+            let mut settlement_at_commit = None;
             if record.mode == BatchMode::Direct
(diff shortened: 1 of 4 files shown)


crates/jazz-tools/src/runtime_core/writes.rs
```diff
@@ -1115,40 +1115,68 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
...
-            if let Some(confirmed_tier) = self.local_write_confirmed_tier() {
+        let retry_record = record.clone();
+        let mut seal_persisted = false;
+        let result = (|| {
+            let submission = self.sealed_batch_submission(&record)?;
+
+            record.mark_sealed(submission.clone());
+            let mut settled_at_commit = false;
+            let mut confirmed_tier_at_commit = None;
+            let mut settlement_at_commit = None;
```
nrainhart: Nit: I'd just keep `settlement_at_commit` and deconstruct it when we need the confirmed tier

nrainhart (approved):

PR #1093, opened 2026-07-17
perf: avoid performance pitfalls with rows with deep histories

### LINKED ISSUE #1081: local_batch_rows still falls through to scan_history_row_batches on alpha.53 (post-#1073) — multi-second reconnect freeze from author-side row history
## Summary

On `jazz-tools@2.0.0-alpha.53` (with #1073's persistent `batch_id → rows` index merged), a `jazz-rn` client that has **authored** a deep history on a single row still freezes the JS thread for **multiple seconds** on the next app restart / reconnect. #1073 removed the `scan_row_locators` cross-row fan-out, but `local_batch_rows` still falls through to `scan_history_row_batches`, and that dominates the freeze.

Standalone repro (React Native + docker): https://github.com/Xelson/jazz-freeze-repro

## Symptom

- **Production:** an iOS app writes a presence heartbeat `db.update(users, myId, { onlineTimeUpdatedAtMs: new Date() }).wait({ tier: 'edge' })` every ~15 s. After one night (~2500 writes to one row) the app freezes ~5-6 s on **cold start**.
- **Repro:** author ~3000 single-row updates, then restart the app → **37 s JS-thread freeze** (also observed 2 s / 7 s / 20 s across restarts).

## Native stack (`sample`, the 37 s freeze)

26 170 samples (≈37 s) on the JS/main thread:

```
batched_tick / immediate_tick        26170 / 25866
  └─ local_batch_rows                24554
      └─ scan_history_row_batches    ~24000
          └─ decode_flat_history_row_with_codecs   ~15000
```

## Two manifestations of the same residual scan

1. **Write-time (author-side):** each `db.update(row).wait()` on a deep row runs `scan_history_row_batches(row)` — `O(history depth)` per write → `O(W²)` to author W writes to one row. The writes themselves stall (visibly degrading rate).
2. **Reconnect / cold-start (replay):** on reconnect the runtime replays the accumulated batches; `local_batch_rows → scan_history_row_batches` per fate → the multi-second freeze in the stack above.

## #1073 is not sufficient

The persistent index removed the `scan_row_locators` fan-out, but the sample shows `local_batch_rows` **still** reaching `scan_history_row_batches` heavily (24 554 → ~24 000 samples) on alpha.53. Hypothesis: batches confirmed only at `EdgeServer` (never `GlobalServer`) never retire, so their bookkeeping is retransmitted + re-examined on every reconnect; or this path is not covered by the index.

## Author vs receiver

Only the client that **authored** the deep history freezes. A fresh client that **re-syncs the identical data** from the server does **not** freeze — its pending/replay set is empty. So this is about the depth of *authored* history + accumulated per-batch bookkeeping, not total data volume.

## Repro steps (alpha.53)

1. `cd backend && docker compose up -d` (creates the target row)
2. `cd client && pnpm install && pnpm prebuild && pnpm ios`
3. Tap **BURST 3000** (client authors 3000 updates to one row)
4. **Restart the app** → watch the on-screen `max freeze on timer` climb into seconds on startup.

## Environment

- `jazz-tools` / `jazz-rn` / `jazz-napi` `2.0.0-alpha.53`
- React Native 0.85.3, Hermes v1, iOS Simulator (iOS 26.5)
- Single Core sync server (`jazz-tools … server`), `tier: 'edge'` writes

nrainhart:
## Description

Fixes https://github.com/garden-co/jazz/issues/1081.

- Use the batch-> rows index introduced in https://github.com/garden-co/jazz/pull/1073 to avoid full history scans:
	- avoid scanning a row's full history on every write
	- avoid full-history scan to get batch members when receiving remote tx sealed/accepted payloads
- ⚠️ Temporarily disable full-storage reconciliation when connecting to a server, as the current mechanism is too resource intensive
	- This means rows that couldn't be synced to the server will not be sent on reconnection, instead needing to wait until a query loads them again. Also, the full client storage won't be automatically reconciled when connecting to a new server

### DIFF
--- a/.changeset/fast-batches-rest.md
+++ b/.changeset/fast-batches-rest.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Improve write and sync performance by replacing full row-history scans with batch-indexed or exact row lookups across batch tracking, transaction validation, permission rejection, and common parent resolution.
+
+Disable automatic full-storage reconciliation when connecting to a server, avoiding a replay of all stored rows' history on every connection. This means rows that couldn't be synced to the server will not be sent on reconnection, instead needing to wait until a query loads them again. Also, the full client storage won't be automatically reconciled when connecting to a new server.
--- a/crates/jazz-tools/src/query_manager/writes.rs
+++ b/crates/jazz-tools/src/query_manager/writes.rs
@@ -599,11 +599,9 @@ impl QueryManager {
     ) -> Option<(String, StoredRowBatch)> {
         let table = self.load_row_table_name(storage, row_id)?;
         let row = storage
-            .scan_history_row_batches(&table, row_id)
-            .ok()?
-            .into_iter()
-            .filter(|row| row.batch_id == batch_id && row.branch.as_str() == branch_name)
-            .max_by_key(|row| (row.updated_at, row.batch_id()))?;
+            .load_history_row_batch(&table, branch_name, row_id, batch_id)
+            .ok()
+            .flatten()?;
         Some((table, row))
     }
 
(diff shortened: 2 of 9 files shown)


gdorsi (approved):
Looks good.

gdorsi:
Shall we add a changeset?

PR #1095, opened 2026-07-20
Fix counter merge replay over-counting

joeinnes:
## Summary

- Preserve raw row ancestry when scoped sync delivers a server-computed visible projection, updating only the materialised view rather than treating the projection as a new causal root.
- Translate Counter writes from the visible value the client edited into a delta against its actual raw frontier, including repeated writes in an open transaction.
- Rebase staged Counter rows in both ordinary and index-backed transaction queries so local reads and filters agree.
- Record the visible projection observed by the first transaction write in existing row metadata, then validate it against the authority's reconstructed pre-transaction projection. Rows from older clients retain parent-frontier validation.

## Correctness and performance trade-offs

> [!WARNING]
> Transaction validation now scans and resolves the full raw history of every touched row carrying projection-basis metadata when the batch is sealed. A focused authority-side Criterion benchmark found the basis path 22%-62% slower than the legacy parent-frontier path across the tested matrix. At history depth 10,000 for one row it took 9.48 ms versus 7.53 ms; at history depth 1,000 for 100 touched rows it took 158.3 ms versus 100.7 ms. Both paths also scaled materially with history depth and transaction size, so these results demonstrate a practical performance risk but do not isolate a basis-only asymptotic regression. This PR remains in draft while we decide whether the cost is acceptable or needs a bounded/cached alternative.

- The client deliberately does not attach the server's frontier as causal ancestry. That frontier may include rows outside the query scope or rows the client is not authorised to see.
- The observed projection is carried as a BLAKE3 basis hash in the existing metadata map, avoiding sync-protocol and storage-schema changes.
- Counter transaction writes may perform extra visible-projection and raw-parent loads so an absolute visible edit can be stored as the correct raw delta.
- Each transactional row carries a 64-character basis hash.
- Older clients remain compatible through the existing parent-frontier fallback. A new client talking to an older server can still encounter the original false transaction rejection, but should not corrupt the Counter value.

## Benchmark results

The benchmark enters the real authority-side transaction-sealing path through `SyncManager::process_from_client`. Deterministic history and submission fixtures are prepared outside the measured closure; Criterion times only final seal processing. It compares projection-basis metadata validation with the legacy parent-frontier fallback using `MemoryStorage`.

Command:

```bash
cargo bench -p jazz-tools --features bench-internals --bench transaction_projection_validation -- linear_memory
```

Matrix:

- One touched row at history depths 1, 10, 100, 1,000, and 10,000.
- 1, 10, and 100 touched rows at history depth 1,000.
- 30 samples per case, with deterministic correctness preflights for acceptance, exact visible output, stale-basis rejection, and changed-frontier/same-visible-projection semantics.

Headline results:

| History depth | Touched rows | Projection basis | Legacy frontier | Basis / legacy |
|---:|---:|---:|---:|---:|
| 1 | 1 | 25.9 us | 21.2 us | 1.22x |
| 100 | 1 | 117.9 us | 73.0 us | 1.62x |
| 1,000 | 1 | 869.1 us | 568.0 us | 1.53x |
| 10,000 | 1 | 9.48 ms | 7.53 ms | 1.26x |
| 1,000 | 10 | 8.59 ms | 6.21 ms | 1.38x |
| 1,000 | 100 | 158.3 ms | 100.7 ms | 1.57x |

The basis path was slower in every matched case. Scaling from 10 to 100 touched rows at depth 1,000 was superlinear for both paths, with a clear per-row penalty at 100 rows. The benchmark therefore supports the warning's operational concern, while follow-up profiling is needed before attributing the whole scaling curve specifically to projection-basis reconstruction.

## Testing

- `cargo test -p jazz-tools --features test --test counter_merge --test gset_merge --test history_conflict -- --nocapture` — passed, 27 tests
- `cargo test -p jazz-tools --features test --test transactions -- --nocapture` — passed, 12 tests
- `cargo test -p jazz-tools --features test --lib sync_manager::tests::transaction_sealing -- --nocapture` — passed, 25 tests

### DIFF
--- a/crates/jazz-tools/src/row_histories/mutations.rs
+++ b/crates/jazz-tools/src/row_histories/mutations.rs
@@ -190,7 +190,7 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
     let ApplyRowBatchWithContext {
         object_id,
         branch_name,
-        row,
+        mut row,
         index_mutations,
         row_locator,
         table,
@@ -218,6 +218,68 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
         .as_ref()
         .map(|entry| entry.current_row.clone());
 
+    let existing_history_row = if !is_known_new_object {
+        io.load_history_row_batch(&table, branch_name.as_str(), object_id, batch_id)
+            .map_err(RowHistoryError::StorageError)?
+    } else {
+        None
+    };
+
+    // Scoped deliveries intentionally omit parents. Reattach a new snapshot
+    // to the local frontier so a causal update does not become a new root.
+    // If the raw batch already exists, update only the visible projection:
+    // the merged snapshot must not replace the writer's raw history.
+    if row.state.is_visible() && row.parents.is_empty() && !is_known_new_object {
+        if let Some(existing_row) = existing_history_row.as_ref()
+            && !existing_row.parents.is_empty()
+        {
+            if existing_row == &row {
+                return Ok(ApplyRowBatchResult {
+                    batch_id,
+                    row_locator,
+                    visibility_change: None,
+                });
+            }
+
+            let current_entry = VisibleRowEntry::new(row.clone());
+            let current_visible = Some(row.clone());
+            let visible_changed = previous_visible != current_visible;
+            let encoded_visible =
+                crate::storage::encode_visible_row_bytes_with_context(&context, &current_entry)
+                    .map_err(RowHistoryError::StorageError)?;
+            <H as Storage>::apply_prepared_row_mutation(
+                io,
+                &table,
+                &[],
+                std::slice::from_ref(&current_entry),
+                &[],
+                std::slice::from_ref(&encoded_visible),
+                index_mutations,
+            )
+            .map_err(RowHistoryError::StorageError)?;
+
+            let applied = AppliedRowBatch {
+                row_locator: row_locator.clone(),
+                previous_visible,
+                current_visible,
+                is_new_object: false,
+                visible_changed,
+            };
+            return Ok(ApplyRowBatchResult {
+                batch_id,
+                row_locator,
+                visibility_change: visibility_change_from_applied(object_id, applied),
+            });
+        }
+
+        if let Some(previous_entry) = previous_entry.as_ref()
+            && !previous_entry.branch_frontier.is_empty()
+            && !previous_entry.branch_frontier.contains(&batch_id)
+        {
+            row.parents = previous_entry.branch_frontier.clone().into();
+        }
+    }
+
     if !is_known_new_object {
         for parent in &row.parents {
             if io
@@ -243,16 +305,14 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
     } else {
         let mut patched_history = load_branch_history(io, &table, object_id, &branch)?;
 
-        if let Some(existing_row) = io
-            .load_history_row_batch(&table, branch_name.as_str(), object_id, batch_id)
-            .map_err(RowHistoryError::StorageError)?
-            && existing_row == row
-        {
-            return Ok(ApplyRowBatchResult {
-                batch_id,
-                row_locator,
-                visibility_change: None,
-            });
+        if let Some(existing_row) = existing_history_row {
+            if existing_row == row {
+                return Ok(ApplyRowBatchResult {
+                    batch_id,
+                    row_locator,
+                    visibility_change: None,
+                });
+            }
         }
         if let Some(existing) = patched_history
             .iter_mut()
--- a/crates/jazz-tools/src/row_histories/mutations.rs
+++ b/crates/jazz-tools/src/row_histories/mutations.rs
@@ -47,6 +47,9 @@ pub(crate) struct ApplyRowBatchWithContext<'a> {
     /// True only when a top-level client applies a parent-stripped row from
     /// its upstream server.
     pub(crate) is_scoped_delivery: bool,
+    /// Materialised data to expose while the raw row remains authoritative
+    /// for history and sync.
+    pub(crate) visible_data_override: Option<Vec<u8>>,
 }
 
 pub(super) fn row_locator_from_storage<H: Storage>(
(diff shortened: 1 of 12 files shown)


gdorsi (dismissed):

crates/jazz-tools/src/row_histories/mutations.rs
```diff
@@ -218,6 +218,68 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
         .as_ref()
         .map(|entry| entry.current_row.clone());
 
+    let existing_history_row = if !is_known_new_object {
+        io.load_history_row_batch(&table, branch_name.as_str(), object_id, batch_id)
```
gdorsi: Tiny note: hoisting this means the point-read now runs on every non-new-object apply, including the fast path that previously skipped it.

Probably fine, but maybe we can avoid this.

crates/jazz-tools/src/row_histories/mutations.rs
```diff
@@ -218,6 +218,68 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
...
+            .map_err(RowHistoryError::StorageError)?
+    } else {
+        None
+    };
+
+    // Scoped deliveries intentionally omit parents. Reattach a new snapshot
+    // to the local frontier so a causal update does not become a new root.
+    // If the raw batch already exists, update only the visible projection:
+    // the merged snapshot must not replace the writer's raw history.
+    if row.state.is_visible() && row.parents.is_empty() && !is_known_new_object {
```
gdorsi: This guard fires for every merge strategy, and only counters are covered so far. 

A scoped-replay case in `gset_merge` would be a nice addition, for a gset, the rebuild-from-raw scenario in my main comment would show up as missing elements rather than a wrong sum, so it's a slightly different behaviour to pin down.

crates/jazz-tools/src/row_histories/mutations.rs
```diff
@@ -218,6 +218,68 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
...
+        if let Some(existing_row) = existing_history_row.as_ref()
+            && !existing_row.parents.is_empty()
+        {
+            if existing_row == &row {
```
gdorsi: I think this early return can't currently trigger: the derived `PartialEq` on `StoredRowBatch` includes `parents`, the incoming row is parentless, and the branch above requires the stored one to have parents. If the intent was redelivery dedup, comparing everything except `parents` would get you there — as written, redeliveries re-encode and rewrite the visible entry each time.

crates/jazz-tools/src/row_histories/mutations.rs
```diff
@@ -218,6 +218,68 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
...
+                return Ok(ApplyRowBatchResult {
+                    batch_id,
+                    row_locator,
+                    visibility_change: None,
+                });
+            }
+
+            let current_entry = VisibleRowEntry::new(row.clone());
```
gdorsi: Question: is delivery ordering guaranteed upstream? This path accepts whatever arrives as the new visible value, so if an older merged snapshot shows up late (retry / out-of-order delivery), it would overwrite a newer visible value — and stick until the next time the engine recomputes the visible entry from raw history (a later local write, or a read that misses the visible entry). If sync already prevents that, a one-line comment saying so would help the next reader; if not, an `updated_at` check here would cover it.

crates/jazz-tools/src/row_histories/mutations.rs
```diff
@@ -218,6 +218,68 @@ pub(crate) fn apply_row_batch_with_context<H: Storage>(
...
+            return Ok(ApplyRowBatchResult {
+                batch_id,
+                row_locator,
+                visibility_change: visibility_change_from_applied(object_id, applied),
+            });
+        }
+
+        if let Some(previous_entry) = previous_entry.as_ref()
+            && !previous_entry.branch_frontier.is_empty()
+            && !previous_entry.branch_frontier.contains(&batch_id)
```
gdorsi: This checks frontier membership rather than ancestry. The gap only opens if a batch is an ancestor of the frontier but absent from local history (e.g. after pruning/compaction), so it's narrow — an `updated_at` bound or ancestry check would be cheap insurance if that combination is reachable.

gdorsi (commented):
Thanks for this!
 
 The analysis is spot on, and the tests are nice!

One scenario I'd love to dig into together before this lands (smaller thoughts are inline).

### What happens if a client writes again while still offline?

Borrowing the cast from `counter_merge_does_not_recount_a_writer_after_restart_and_reconnect`: Bob wrote +3 offline, Alice wrote +5, and the server ships Bob the merged 8 as a scoped snapshot.

The path for already-known batches (the first `if let` in the new guard, taken when the incoming snapshot's batch already exists locally with parents) updates Bob's visible value to 8 while leaving his raw history as `{+3}` — Alice's +5 doesn't land in his raw history. That works as long as visible and raw agree, but elsewhere the engine treats visible as a cache rebuilt from raw: the normal apply path recomputes it via `visible_entry_from_history_rows` from raw rows only (`mutations.rs:244–270` on main), and `load_previous_visible_entry` falls back to a rebuild-from-raw when the visible entry is missing or unreadable.

So if we extend the test's timeline by one step — after receiving the merged 8, Bob goes offline again and writes +1 — his local apply rebuilds visible from raw `{+3, +1}` and lands on 4 instead of 9. The current tests always do a server round-trip after the merge, so they don't exercise this; a test that does one purely local write after a scoped merge would pin it down either way.

One related detail: that same path builds the entry with `VisibleRowEntry::new(row)`, which resets `branch_frontier` to `[batch_id]` and clears the tier markers and `winner_batch_pool` (compare `VisibleRowEntry::rebuild`, which derives those from history). Since the fallback below it — the one that reattaches a locally-unknown snapshot via `row.parents = previous_entry.branch_frontier` — later trusts that stored frontier to pick parents, it's probably worth preserving the existing entry's frontier/tier state here regardless.

Two directions that would both work for me:
1. have the scoped merge also reconcile raw ancestry (or mark the row as needing reconciliation before local writes are accepted), or
2. preserve the existing entry's frontier/tier state instead of `VisibleRowEntry::new`, plus the local-write-after-merge test so we know exactly where the line is.

PR #1097, opened 2026-07-20
feat(inspector): allow customizing table columns

nrainhart:
## Description

The inspector now supports hiding columns and reordering them. It also shows the `$createdAt` and `$updatedAt` magic columns by default. `$createdBy` and `$updatedBy` can be included as well.

https://github.com/user-attachments/assets/1fa38514-82a1-44f6-86d1-4c75d17ac3c9

### DIFF
--- a/packages/inspector/src/components/data-explorer/ColumnCustomizationModal.tsx
+++ b/packages/inspector/src/components/data-explorer/ColumnCustomizationModal.tsx
@@ -4,8 +4,7 @@ import styles from "./ColumnCustomizationModal.module.css";
 export interface CustomizableColumn {
   id: string;
   header: string;
-  // `true` unless specified
-  visibleByDefault?: boolean;
+  hiddenByDefault?: true;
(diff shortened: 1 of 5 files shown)


packages/inspector/src/components/data-explorer/ColumnCustomizationModal.tsx
```diff
@@ -0,0 +1,340 @@
+import { useEffect, useRef, useState, type DragEvent, type MouseEvent } from "react";
+import styles from "./ColumnCustomizationModal.module.css";
+
+export interface CustomizableColumn {
+  id: string;
+  header: string;
+  // `true` unless specified
+  visibleByDefault?: boolean;
```
gdorsi: **nit:** this property has effect only when false. Shall we revert the meaning (e.g. `hidden?: true`)

gdorsi (approved):
LGTM

PR #1099, opened 2026-07-20
fix: preserve sorted subscription order through projections

nrainhart:
## Description

When a value in a column used as a query's sort criteria changed:
1. The sort node moved the row to its correct new position.
2. The magic columns/projection nodes preserved the new row content, but not the reordered row sequence.
3. The output node therefore received “this row was updated” without its new ordering.
4. It replaced the row’s content in place at its old index, making it appear that updated rows were unsorted or appended at the bottom.

The fix makes the magic columns/projection nodes preserve the exact ordered tuple list from upstream. They now calculate their delta from the ordered before/after sequences, so the output receives both the updated content and the correct new position.

### DIFF
--- a/.changeset/preserve-sorting-through-projections.md
+++ b/.changeset/preserve-sorting-through-projections.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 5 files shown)


gdorsi (approved):

gdorsi:
Changeset plz

PR #1106, opened 2026-07-21
(new core) Remove duplicate query semantics from the TS layer (SPEC 13.13)

aeplay:
Stacked on #1094. Cleanup/simplification only: one query engine, not two.

SPEC 13.13 (landed on the base branch) states the principle: query semantics
live in the core; binding layers build, serialize, cache, transport, and
hydrate — never re-evaluate. This PR removes the known violations, inventoried
in `dev/TS_SEMANTIC_DUPLICATION_AUDIT.md` (committed here as the work order).

Phased plan (this PR grows through the phases; review per phase is welcome):

1. **Native in-tx reads** — back wasm transaction handles with real core open
   transactions, expose reads via the core's existing `tx_query` /
   `OverlayRef::OpenTransaction` path (identity-aware), then delete the TS
   overlay evaluator (`applyTransactionReadOverlay`, `rowMatchesQuery`,
   `sortRowsForQuery`).
2. **TS-computed semantic deltas → core-authoritative deltas** (the
   `nativeDeltaFromRows` family and `refreshOpenedPlainSubscriptions`).
   The commit-path attribution (`dev/COMMIT_SUPERLINEARITY.md` on the base
   branch work) shows the wasm-only superlinear commit slope lives in this
   sandwich, so this phase is perf-relevant, not just hygiene.
3. **Relation/include materialization → core-emitted shapes** (largest;
   overlaps the relation-parity roadmap; may split into its own stacked PR).

Every phase gates on: jazz-tools suite, browser suite at the 8-red parity
baseline, wire-codec gate, and the differential harnesses on the base branch.


packages/jazz-tools/src/runtime/db.transaction.test.ts
```diff
@@ -112,6 +121,91 @@ describe("Db transactions", () => {
...
+      ).resolves.toMatchObject([{ title: "a" }, { title: "b" }, { title: "c" }]);
+
+      return tx.all(app.todos.where({}), { tier: "local" });
+    });
+
+    await result.wait({ tier: "local" });
+    await expect(db.all(app.todos.where({}), { tier: "local" })).resolves.toEqual(result.value);
+  });
+
+  it("applies non-eq predicates and limit offset inside a mergeable transaction", async () => {
```
nrainhart, line 165: Why do we need to test different types of predicates in conjunction with mergeable transactions? How are these different from regular reads? Is it due to transaction overlays? If so, it'd be better to run all tests in `query-api.test.ts` with and without txs (and maybe also with exclusive txs?)

packages/jazz-tools/src/runtime/db.transaction.test.ts
```diff
@@ -112,6 +121,91 @@ describe("Db transactions", () => {
...
+      return existing.id;
+    });
+
+    await result.wait({ tier: "local" });
+    await expect(db.all(app.todos.where({}), { tier: "local" })).resolves.toEqual([
+      { id: existing.id, title: "committed", done: true },
+    ]);
+  });
+
+  it("uses core ordering and default ordering for in-transaction reads", async () => {
```
nrainhart: What does _core ordering_ mean?

crates/jazz-wasm/src/lib.rs
```diff
@@ -466,6 +467,104 @@ impl WasmDbInner {
         }
     }
 
+    fn begin_exclusive(&self) -> Result<OpenTxId, jazz::db::Error> {
+        match self {
+            Self::Memory(db) => db.begin_exclusive(),
+            #[cfg(target_arch = "wasm32")]
+            Self::Browser(db) => db.begin_exclusive(),
+            Self::Closed => panic!("WasmDb is closed"),
```
nrainhart: Nit: by looking at all these methods it feels like it'd be better to have OOP-style polymorphism on each possible state instead of doing the same exact pattern matching on every method

nrainhart (approved):
Logic-wise makes sense, have some comments on tests and a possible refactoring

PR #1108, opened 2026-07-21
(new core) Edge servers must not act as global fate authorities for client uploads

aeplay:
Stacked on #1094. Fixes the bug @nrainhart reported: the server shell never
passed `NodeRole::Edge` into the database, so client commits on edge servers
took the generic authority ingest path — allocating a `global_seq` and
answering `Accepted/Global` that the core never confirmed. A global-tier
subscriber on the same edge saw rows before core acknowledgment.

**The fix**: edge-role session links are accepted through an explicit
`accept_edge_authority_subscriber_with_claims` entry; the subscriber ingest
context carries `edge_authority`, and commit units route to the existing edge
ingest path (`peer.rs` edge machinery — exactly the path you pointed at) only
when the host is an edge AND the link is a client link. Core-role servers are
byte-identical.

**Review-worthy history**: a first attempt keyed on `PeerRole::EdgeClient`
alone — which every client link on every server carried, because `for_author`
aliased to `edge_client` — and was caught by `cli_dry_run`'s core persistence
test. That led to two fold-ins: the alias is retired (`client_link` /
`PeerRole::ClientLink` for generic links, `edge_client` only where meant), and
SPEC `A_impl_discipline` now has a wiring-canary convention (every shell role ×
dispatch path needs one black-box test through the production shell — the
harness tests hand-wired correct topologies and structurally could not catch
this).

**Tests**: `crates/jazz-server/tests/edge_fate_authority.rs` asserts both
sides through the production shell: edge-role upload reports edge fate and
withholds `Global` until core ack; core-role upload gets `Accepted/Global`
immediately. Full findings in `dev/EDGE_FATE_AUTHORITY.md`.

Open question for you: your trailing thought about non-global-tier
subscriptions — the tier-gating story for edge-local subscribers is asserted
as currently specced, but if you expected different semantics there, that's a
SPEC ch. 9 conversation worth having on this PR.

Gates: jazz, jazz-server, groove, four_tier, all three delivery canaries,
100-seed maintained/one-shot oracle, jazz-tools (--features test), bench
checks, fmt.


crates/jazz/examples/commit_superlinearity_native.rs
nrainhart, line 1: Looks unrelated, did you mean to include it in this PR?

crates/jazz-server/tests/edge_fate_authority.rs
```diff
@@ -0,0 +1,252 @@
...
+    let refs = refs.iter().map(String::as_str).collect::<Vec<_>>();
+    block_on(Db::open_history_complete(DbConfig::new(
+        schema.clone(),
+        MemoryStorage::new(&refs),
+        identity(node_byte, AuthorId::SYSTEM),
+    )))
+    .unwrap()
+}
+
+struct DuplexTransport {
```
nrainhart: Nit: we have this in several test suites, would be nice to extract as a util

nrainhart (approved):

aeplay:
already merged as part of #1156

PR #1110, opened 2026-07-21
(new core) Emit SessionClaims only when claims change

aeplay:
Stacked on #1094. Fixes the wire chatter @gdorsi spotted: SessionClaims pairs
preceding every RegisterShape/Subscribe/CommitUnit.

**Finding**: the spec had no cadence for SessionClaims (now specced in ch. 7:
once per connection, again only on change). The spam was mechanical — the TS
adapter re-applies session claims per operation, and `set_session_claims`
enqueued an upstream wire message every time, identical or not. The server
keeps per-connection claim state and never needed per-message claims, so every
repeat cost bytes plus a server-side claims application for nothing.

**Fix**: `set_session_claims` reports whether the claims actually changed
(also skipping cache invalidation on identical claims), and
`Db::set_identity_claims` only enqueues `SessionClaims` on change. Wire-count
regression test asserts exactly one SessionClaims for repeated identical
applications before subscribe.

Gates: jazz, jazz-server, jazz-tools (--features test), delivery canary,
300-seed maintained/one-shot oracle, fmt. Full investigation notes in
`dev/SESSION_CLAIMS_CHATTER.md`.


dev/SESSION_CLAIMS_CHATTER.md
```diff
@@ -0,0 +1,133 @@
+# SessionClaims chatter investigation
```
gdorsi, line 1: Do we need to keep this file?

crates/jazz/src/db.rs
```diff
@@ -1873,10 +1873,14 @@ where
...
     pub fn set_identity_claims(&self, identity: AuthorId, claims: BTreeMap<String, Value>) {
-        self.node
+        let changed = self
+            .node
             .node
             .borrow_mut()
             .set_session_claims(identity, claims.clone());
+        if !changed {
+            return;
+        }
```
gdorsi, line 1935: nit: from my tests looks like we call this function quite frequently, do we need to clone the claims to check if they need to be updated and synced?

crates/jazz/SPEC/7_authorization.md
```diff
@@ -66,6 +66,11 @@ admission/session layer and must not be client-supplied query bindings. Predicat
 forms outside the supported direct-evaluation subset, such as range and null
 checks, deny until explicitly supported.
 
+Wire cadence: session claims travel upstream once per connection and again
+only when the claim set actually changes; identical re-applications by client
+layers must not produce wire messages. The server holds per-connection claim
+state, so per-message claims are neither required nor expected.
```
gdorsi, line 66: I don't think this is important enough to be part of the auth spec.

How do you feel about making this an invariant instead?

gdorsi (approved):
LGTM

aeplay:
superseded by #1276

PR #1112, opened 2026-07-21
(new core) Default result ordering in the core plan; maintained limit subscriptions

aeplay:
Stacked on #1094. One core mechanism resolves four open threads that turned
out to be the same problem.

## How these relate

1. **Todo example shows random insert order** (@gdorsi). SPEC §6.4.1 decided
   default ordering (Anselm, 2026-07-18: no explicit `order_by` → ascending
   row id, i.e. creation order via uuidv7) — but it was implemented as
   *emergent storage scan order*, never expressed in the lowered plan. Order
   survives pass-through delivery but not merge points (delta application),
   because deltas carry no default-order positions. Audit trail:
   `~/DEFAULT_ORDERING_STATUS.md` — including two order-sensitive TS tests
   skipped since the spec landed, waiting on exactly this.
2. **Stress app: 1s checkbox toggle at 100k rows, degrading with stored size,
   phantom Unsubscribes** (@gdorsi). Verified chain in
   `dev/LIMIT_SUBSCRIPTION_FALLBACK.md`: maintained *unordered* `limit(100)`
   is rejected as unsupported (correctly — unordered limit has
   nondeterministic membership), so the adapter falls back to snapshot
   refresh: **two** `db.all` re-queries per observed write (local-visible +
   settled waiters), each wire-shaped Subscribe→ViewUpdate→Unsubscribe, each
   paying O(stored) evaluation before limiting: 299ms @ 10k → 1.53s @ 50k →
   3.16s @ 100k (probe committed here).
3. The two skipped ordering tests (`query-api.test.ts:112`, `:710`).
4. The per-commit wire churn contributing to the stress app's "blocks all
   subscriptions" experience.

## The resolution

Default ordering already *is* the spec. Once it is injected into the lowered
plan, "unordered `limit(100)`" stops existing — it becomes an ordered-by-row-id
window, and the existing ordered-window machinery (`TopBy` lowering, maintained
window delta positions per SPEC 16) handles it. So:

- Inject default ascending-row-id ordering for relation-valued results without
  explicit `order_by` (root results in this PR; array-subquery/include
  subtrees assessed and either included or staged with a 🔶).
- Lift the maintained rejection for limit/offset over the injected default
  order; deltas carry correct positions, so client layers become pass-through.
- The stress app's `useAll(app.todos.limit(100))` becomes a real maintained
  subscription: delta-driven toggles, no fallback re-queries, no wire churn.
- Re-enable the two skipped TS assertions; update SPEC 14/16 "unsupported"
  notes; update the loud-rejection regression test to assert the new
  supported behavior (intentional, documented change).

Gates: full battery incl. INV-INC-1 canaries (delta positions are the
sensitive surface), 300-seed maintained/one-shot oracle, four_tier,
jazz-tools/jazz-server suites, wire-codec, plus a probe re-run showing
toggle cost at 100k stored rows with the maintained path.


aeplay:
First implementation slice landed (64b0b61d3). **Delivered**: default ascending-row-id ordering injected into root *windowed* plans — maintained `limit/offset` subscriptions now accepted through the standard TopBy window machinery (the stress app's `limit(100)` no longer rejects into the O(stored) snapshot-refresh fallback); a delta-index bug exposed by the work (removals always reported index 0) fixed; the two long-skipped TS ordering assertions re-enabled with expectations corrected to real §6.4.1 semantics (same-millisecond uuidv7 ids sort by random bits — spec gained a precision note). **Staged, tracked in dev/DEFAULT_ORDERING_IMPL.md**: default ordering for unbounded roots (the todo-example case — naive plan-wide injection broke recursive/RLS maintained shapes, needs a scoped pass) and subtree/include ordering, plus subscriber delta-churn polish before window positions settle (ignored tracking test pins it). Full battery green: jazz 724, groove, jazz-tools, jazz-server, four_tier, all three canaries, 300-seed oracle, wire-codec, TS 47/47.

aeplay:
Slice 2 landed: unbounded-root default ordering (the todo-example fix) with the narrowed lowering guard and the churn tracking test now green. Interim honestly documented: unbounded ordered delivery is adapter-side O(result)/event until core deltas carry positions — which is dedup phase 2's mandate (stacked on #1106); that phase deletes this adapter sort.

aeplay:
Superseded by #1278

PR #1116, opened 2026-07-22
(new core) Restore Db.disconnect()/Db.reconnect() (and the offline-path bugs it surfaced)

aeplay:
Stacked on #1094. Restores your #1064 API on the new runtime — the only genuinely missing item from the mainline gap analysis. Wiring verified at node level; your original browser regression is ported and currently skip-marked because it exposes two real engine gaps in the offline path (invalid offset on local reads after disconnected writes; reconnect convergence not settling) — details in dev/DB_DISCONNECT_RESTORE.md. Those two are the actual work this PR now tracks.

### DIFF
--- a/dev/DB_DISCONNECT_RESTORE.md
+++ b/dev/DB_DISCONNECT_RESTORE.md
@@ -1,0 +1,33 @@
+# Db disconnect/reconnect restore
+
+## Restored surface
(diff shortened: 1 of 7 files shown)


aeplay:
Superseded by #1283

PR #1117, opened 2026-07-22
fix: weight was not used in topBy operator

antoniomuso:
## Summary

- make `TopBy` consume weighted multiplicity when calculating window occupancy and output deltas
- represent finite and unbounded retained windows explicitly with `TopByLimit`, instead of using `usize::MAX` as a sentinel
- use architecture-independent `u64` bounds, define `Finite(0)` as an empty window, and update Jazz lowering plus the Groove/Jazz specifications

## Why

`TopBy` receives weighted records, but the window calculation previously treated every positive-weight record as a single ordinal. Duplicate derivations therefore did not consume the expected number of positions in `offset`/`limit` windows.

The follow-up implementation also represented an omitted Jazz limit with `usize::MAX`. That made an "unbounded" window architecture-dependent: it stopped at `u32::MAX` ordinals on wasm32, and even a 64-bit build truncated a stream containing more than `u64::MAX` weighted ordinals.

## Impact

Maintained ordered queries now preserve bag semantics across duplicate rows and partial window boundaries. An omitted limit retains every ordinal after the offset on native and wasm targets, while `Finite(u64::MAX)` remains a genuinely finite and distinguishable bound.

## Validation

- `cargo test -p groove -j 2`
- `cargo test -p jazz -j 2`
- `cargo test -p jazz-tools --features test -j 2`
- `cargo test -p jazz-server -j 2`
- `cargo check -p jazz-wasm --target wasm32-unknown-unknown -j 2`
- `JAZZ_SEED_COUNT=300 cargo test -p jazz -j 2 m3_maintained_one_shot_differential_oracle`
- `dev/benchmarks/smoke.sh` (all 14 scenarios passed)
- pre-commit Clippy, Rustfmt, Oxfmt, and sensitive-data checks


crates/groove/src/ivm/runtime/mod.rs
```diff
@@ -7810,47 +7810,61 @@ fn top_by_window_from_records(
...
+    // order must not depend on arrangement iteration order.
+    ranked.sort_by(|left, right| left.0.cmp(&right.0).then_with(|| left.1.cmp(&right.1)));
+
+    // Bag semantics (INV-QUERY-24/25): a record with multiplicity m occupies m
+    // ordinals, and offset/limit consume copies, not distinct records. The
+    // arithmetic stays in u64 because limit uses usize::MAX as the unbounded
+    // sentinel.
+    let mut window = Vec::new();
+    let mut to_skip = top_by.offset as u64;
+    let mut remaining = top_by.limit as u64;
```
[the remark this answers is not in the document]
antoniomuso: Done!

crates/groove/src/ivm/runtime/mod.rs
```diff
@@ -7810,47 +7808,67 @@ fn top_by_window_from_records(
...
+            window.push((
+                record,
+                i64::try_from(taken).expect("taken copies cannot exceed positive record weight"),
+            ));
+            if let Some(remaining) = &mut remaining {
+                *remaining -= taken;
+            }
+        }
+    }
+    Ok(window)
```
gdorsi, line 7852: Have you benchmarked this change to see if it is causing a perf regression?

gdorsi (approved):
Nice fix!

How do you feel about adding a more "functional" test on the fixes you have done?

Like building a small dataset, and cover the various broken situations.

crates/groove/src/ivm/runtime/mod.rs
antoniomuso: No. But it shouldn't cost more than before, we've just added some arithmetic operations.

antoniomuso: I have done some benchmarking, I did't find any significant regression.

antoniomuso:
> Nice fix!
> 
> How do you feel about adding a more "functional" test on the fixes you have done?
> 
> Like building a small dataset, and cover the various broken situations.

Done, I have added a functional test for it.

PR #1123, opened 2026-07-23
docs: added groove code comments

antoniomuso:
Add rustdoc to Groove code.


aeplay:
🤖 Closing as superseded by the new Groove/core-engine architecture.

The implementation portion of this 20-file stack targets the pre-engine-swap Groove/runtime layout and no longer has a meaningful patch to replay on `codex/jazz-core-engine-swap`. The current engine replaced those seams rather than evolving them in place.

The Rustdoc policy added in the tip commit may still be useful, but it is independent of the obsolete implementation and should be proposed as a small focused `AGENTS.md` change if we still want it.

Preserved historical tip: [`9c8814e56`](https://github.com/garden-co/jazz/commit/9c8814e56).

PR #1124, opened 2026-07-23
fix: properly enforce complex OR permissions in client writes

### LINKED ISSUE #1104: anyOf of exists checks behaves like allOf when writing
If an insert/update policy is an `anyOf(...)` and its branches use `exists` (or `inherits`), the
write path treats it like `allOf(...)`. So a write that clearly matches one branch gets rejected
because some other branch's `exists` is false.

Reads are fine. It only goes wrong on writes, which also means the local check and the server
check can disagree on the same policy.

Repro against 2.0.0-alpha.53, with CI: https://github.com/tehfreak/repro-jazz-anyof-exists-write

Seed only `gate_a` for the user, then insert:

    single exists, gate_a seeded             allowed   (ok)
    anyOf([gate_a, gate_b]), gate_a seeded   denied    (should be allowed)
    allOf([gate_a, gate_b]), gate_a seeded   denied    (ok)
    allOf([gate_a, gate_b]), both seeded     allowed   (ok)

Single exists is fine, allOf is fine both ways, only anyOf is wrong.

I think it comes from how the write path splits a policy. `evaluate_simple_recursive` in
`query_manager/policy.rs` walks the tree and returns the EXISTS/INHERITS clauses as one flat list,
and `writes.rs` then requires every clause in that list to pass. The `Or` arm dumps the clauses
from all the branches into the same list, so once they are ANDed together the anyOf is gone. A
branch that is only exists (no plain column comparison to fail on) never gets dropped, so its
exists always ends up in that AND.

The read path doesn't do this, it goes through `PolicyContextEvaluator` in
`graph_nodes/policy_eval.rs` and evaluates And/Or/Exists directly. Running writes through the same
evaluator would fix it and get the two paths back in sync.

nrainhart:
Fixes https://github.com/garden-co/jazz/issues/1104.

## Background

Policy expressions fall into two categories:
- Simple expressions, such as column comparisons and null checks. These can be evaluated directly against the row being written.
- Complex expressions, such as `EXISTS` and `INHERITS`. These require reading other rows and sometimes constructing query graphs.

The old local-write evaluator tried to optimize policy checks by splitting evaluation into two phases:
- It evaluated simple expressions immediately.
- It extracted complex expressions into a flat `Vec<ComplexClause>` for later batched evaluation.

The problem was that this flat vector did not preserve the OR/AND distinction, turning everything into an AND.

## Fix

Remove the local write optimization to prioritize correctness. Switch to using the `PolicyContextEvaluator` already used by server authorization and local reads.

### DIFF
--- a/crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
+++ b/crates/jazz-tools/src/query_manager/graph_nodes/policy_eval.rs
@@ -399,12 +399,18 @@ impl<'a> PolicyContextEvaluator<'a> {
         let parent_table_name = *parent_table;
         let parent_row = match row_loader(parent_id, Some(parent_table_name)) {
             Some(content) => content,
-            None => return false,
+            None => {
+                visited.remove(&parent_id);
+                return false;
+            }
         };
 
         let parent_schema = match self.schema.get(&parent_table_name) {
             Some(schema) => schema,
-            None => return false,
+            None => {
+                visited.remove(&parent_id);
+                return false;
+            }
         };
 
         let parent_policy = match operation {
@@ -414,29 +420,30 @@ impl<'a> PolicyContextEvaluator<'a> {
             Operation::Delete => parent_schema.policies.effective_delete_using(),
         };
 
-        let parent_policy = match parent_policy {
-            Some(p) => p,
-            None => return false,
+        let result = match parent_policy {
+            Some(parent_policy) => {
+                let parent_row = Row::new(
+                    parent_id,
+                    parent_row.data,
+                    parent_row.batch_id,
+                    parent_row.row_provenance,
+                );
+                self.evaluate_expr_with_context(
+                    parent_policy,
+                    operation,
+                    &parent_row,
+                    &parent_schema.columns,
+                    parent_table_name.as_str(),
+                    io,
+                    row_loader,
+                    depth + 1,
+                    visited,
+                    visited_referencing,
+                )
+            }
+            None => !self.row_policy_mode.denies_missing_explicit_policy(),
         };
-
-        let parent_row = Row::new(
-            parent_id,
-            parent_row.data,
-            parent_row.batch_id,
-            parent_row.row_provenance,
-        );
-        let result = self.evaluate_expr_with_context(
-            parent_policy,
-            operation,
-            &parent_row,
-            &parent_schema.columns,
-            parent_table_name.as_str(),
-            io,
-            row_loader,
-            depth + 1,
-            visited,
-            visited_referencing,
-        );
+        visited.remove(&parent_id);
         if let Some(cache_key) = cache_key
             && let Some(cache) = self.settlement_eval_cache.as_mut()
         {
--- a/crates/jazz-tools/src/query_manager/manager.rs
+++ b/crates/jazz-tools/src/query_manager/manager.rs
@@ -361,11 +361,6 @@ pub(super) struct WriteTableCacheEntry {
     pub(super) indexed_columns: Option<Arc<Vec<ColumnName>>>,
     pub(super) row_layout: Arc<crate::row_format::CompiledRowLayout>,
     pub(super) row_locator: RowLocator,
-    pub(super) insert_policy: Option<Arc<PolicyExpr>>,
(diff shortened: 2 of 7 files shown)


aeplay (changes requested):

Reviewed this as a security fix. **I believe it does close the authorization
hole** — and it also introduces a false-denial regression that should be fixed
before merge.

**What was actually wrong.** The old client-local evaluator split an expression
into a simple result plus a flat list of complex clauses, and `OR` accumulated
the complex clauses from all viable arms (`query_manager/policy.rs:1995-2029` on
the parent) while the write path required every constructed graph to succeed. So
`EXISTS(a) OR EXISTS(b)` was evaluated as `EXISTS(a) AND EXISTS(b)` — too
strict, not permissive.

**The security-relevant direction is negation.** The old `NOT` path discarded
the negation and returned its complex clauses unchanged
(`policy.rs:2032-2048` on the parent). With `NOT EXISTS(a) OR <false arm>` where
`EXISTS(a)` is true, the old client **admitted** the write although the policy
is false. Same flaw for bare `NOT EXISTS(a)`. That is the class that was getting
through.

**Why I think it is closed.** The new code evaluates the actual `PolicyExpr`
tree, with complex leaves evaluated in context and `And`/`Or`/`Not` retaining
boolean structure (`graph_nodes/policy_eval.rs:222-319`). I checked nested OR
inside AND and vice versa (`:270-297`), OR/AND with negation (`:298-309`), an OR
mixing a claim arm with a row-field arm (`:310-318`), and EXISTS / relation
EXISTS / parent inheritance / reverse inheritance (`:223-269`, `:323-550`), with
parent-policy absence still fail-closed in enforcing mode (`:423-446`).

All client mutation entry points route to the shared evaluator — update USING
and WITH CHECK (`writes.rs:775-824`), insert (`:1127-1148`), delete
(`:1780-1801`) — and sync ingest already used the full evaluator
(`server_queries.rs:1756-1782`, `:1887-1963`). No remaining call site of the
deleted split evaluator. The -568 lines are that second write-only evaluator,
not an independent rule.

---

**The regression: `INHERITS REFERENCING` now falsely denies when the FK is
unindexed.**

`PolicyContextEvaluator` finds reverse-inheritance candidates only through the
source FK index — scalar UUIDs via `index_lookup(source, fk, ...)`, UUID arrays
by scanning that index (`policy_eval.rs:153-164`). But a table can deliberately
omit that FK via `index_only` (`types/schema.rs:438-449`, `:540-549`).

The deleted evaluator handled exactly this: it scanned `_id` and
decoded/filter-matched the FK when the scalar FK was unindexed, and did the
equivalent `_id` scan for an unindexed UUID array (`writes.rs:1560-1644` on the
parent).

So a valid `INHERITS REFERENCING` disjunct can now find no candidates and deny
an authorized insert/update/delete. Not a bypass — an availability and
compatibility regression, and it is the answer to "did it become too strict
anywhere". Either restore the fallback in the shared evaluator, or scan when the
FK is unindexed.

---

**Tests: nothing asserts the deny direction.**

The added test is an allow assertion for `OR(EXISTS(gate_a), EXISTS(gate_b))`
(`tests/policies_integration/exists_policies.rs:130-203`). The existing
single-`EXISTS` rejection test (`:75-128`) is not a disjunct test.

That matters more than usual here: **allow-only tests pass trivially against an
over-permissive implementation, which is the bug being fixed.** Worth requiring
deny assertions for:

- `OR(NOT EXISTS(a), false)` with `a` present — the previously weak direction
- nested `AND(OR(...), ...)` with one failing required conjunct
- `OR` mixing a complex arm with a claim/row-field arm, both arms failing
- `OR` using `INHERITS` and `INHERITS REFERENCING` with a denied parent/source row
- an **allowed** `INHERITS REFERENCING` case whose scalar and array FKs are
  omitted from `index_only` — this pins the regression above
- insert, update USING, update WITH CHECK and delete, or a table-driven
  equivalent that demonstrably reaches each entry point

Note the tests need the policy-enforcing local mode; a focused run without
`test-utils` collected zero tests, which is its own trap.

aeplay:
🤖 Closing because the old policy evaluator no longer exists. The complex-policy regression is being carried by the current policy integration work in #1678.

PR #1125, opened 2026-07-24
fix(groove): align aggregate semantics with SQL

joeinnes:
## Summary

- make `SUM`, `AVG`, `MIN`, and `MAX` use stable nullable result types
- implement SQL-style empty and all-null behaviour while preserving grouped and `COUNT` semantics
- support signed `I64` aggregate inputs and carry the new result types through Jazz and its public subscription API

## Testing

- `cargo test -p groove -j 2`
- `cargo test -p jazz -j 2`
- `cargo test -p jazz-tools --features test -j 2`
- `cargo test -p jazz-server -j 2`
- `JAZZ_SEED_COUNT=300 cargo test -p jazz m3_maintained_one_shot_differential_oracle`
- `cargo test -p jazz --test incremental_delivery_canary maintained_relation_include_single_row_changes_are_scale_independent -- --exact`
- `dev/benchmarks/smoke.sh`


aeplay:

The semantic changes here look right to me; the finding in this pair is on
#1126, which I have commented on separately.

This one needs its conflicts resolved against `codex/jazz-core-engine-swap`
before it can go anywhere — it is currently CONFLICTING and it gates #1126.

Two things worth double-checking while you are in there, both because they are
behaviour changes to existing results rather than bug fixes:

1. **Enumerate the semantic changes explicitly in the PR description** — NULL
   handling in SUM/AVG/COUNT, `COUNT(*)` vs `COUNT(col)`, empty-group results,
   and integer division/overflow in AVG. Anyone relying on the old behaviour
   deserves to find that list rather than discover it.
2. **Empty groups.** SQL says a group with no rows vanishes; a group lingering
   with a zero or identity value is a wrong answer. Worth a test that adds rows
   to a group and then removes all of them.

One testing note that applies to both PRs: a test with **one row per group**
exercises projection, not aggregation — it passes identically whether the
aggregate is correct or replaced by "return the only value". Multi-row groups,
with retraction, are what make aggregate coverage mean something.

No changes requested beyond the rebase.

aeplay:
🤖 **Hold — I pushed a rebase to this branch prematurely and it introduces three regressions.** Flagging before you spend time on it.

## What the rebase achieved

- `cargo test -p jazz --features test-utils --test aggregate_subscriptions`: **15 passed, 0 failed** (was 8 passed / 7 failed after the naive rebase; base is 13)
- `m3_maintained_one_shot_differential_oracle_null_semantics` — base's only lib failure — now **passes**, which is exactly what #1125 is for

The conflict it resolved is real and worth recording: #1299 merged and made maintained aggregate subscriptions actually deliver, while #1125 makes non-count outputs nullable. A `NULL` aggregate then produced a result member with no payload, and the delivery path treated payload-less as `invalid stored value` — the two states were indistinguishable. Present-null is now a real `ResultMemberPayloadEntry` carrying `Nullable(None)`; a genuinely missing payload still errors.

## What it broke

`cargo test -p jazz --lib`, both sides extracted by name and diffed:

| tree | failures |
| --- | ---: |
| `codex/jazz-core-engine-swap` | **1** (the oracle arm above) |
| this branch | **3** |

Introduced:

1. `node::query_eval::tests::aggregate_sum_avg_min_max_support_signed_i64_inputs`
2. `node::query_eval::tests::aggregate_sum_min_max_over_filtered_query`
3. `peer::tests::aggregate_policy_oracle_matches_visible_rows_per_identity`

Two are **signed-integer** aggregate tests, which is not obviously in the blast radius of a nullable-representation change — so there may be a second cause. The third concerns **which rows an identity can see**, and I want that understood before anything merges; a nullable-payload change altering visible-row computation would be authorization-adjacent rather than cosmetic.

A lane is diagnosing all three now, with instructions not to weaken or delete any of them, and to stop and report if one encodes an expectation that is genuinely wrong under the new semantics — that call is yours, not mine.

## My mistake

I pushed on the strength of the focused aggregate suites passing, before the full lib run finished. That is the wrong order and I should have waited. The branch is close, but it is not currently a net improvement on `--lib`, and you should not merge it as it stands.

I will follow up when the three are resolved or when there is a decision for you.

aeplay:
🤖 Pushed `576263073` to this branch — a spec section rather than a code change, added at Anselm's request so the aggregate surface has a stated representation contract instead of one implied by the implementation.

## Why here

Three separate bugs this week were all the same missing contract, not aggregation logic:

1. maintained aggregate subscriptions delivered **nothing**, because the facade filtered members by string comparison against `format!("{table}_aggregate")`;
2. `invalid stored value: missing its payload`, because present-but-`NULL` and absent were the same state;
3. this PR's three failing assertions — `U64(6)` vs `Nullable(Some(U64(6)))` — because nothing said which layer is authoritative where.

So `crates/jazz/SPEC/6_queries.md` §6.4.3 now pins the Jazz-side representation, with `INV-QUERY-30`…`35`. Your groove ch.3 text keeps ownership of the operator-level null rules; §6.4.3 only covers what Jazz delivers to a caller. The registry rows land as `target`/`untested` — a lane is implementing against them now, on top of this branch.

The load-bearing one for your three failures is **`INV-QUERY-32`**: the payload layer carries SQL `NULL` as `Nullable(None)`, the public cell layer carries a caller-visible absent value, and the two collapse **exactly once, at the public boundary and nowhere earlier**. That makes the three failures decidable rather than a judgement call — an assertion at the public boundary is stale and should be updated, an assertion at an internal layer means the flattening leaked early and the implementation is wrong. The lane is classifying each of the three individually rather than updating all three the same way.

## One thing you should look at

I recorded an open question because I think your new groove text and ch. 16 §16.6 cannot both hold as written.

Your addition says an empty global aggregate reports `NULL` for `sum`/`avg`/`min`/`max` and `0` for `count`, following SQL. §16.6 places "empty-global-row SQL compatibility" **outside** the maintained subscription surface until its replay semantics and payload shape are specified.

Taken together, a one-shot read of an empty global aggregate returns a row and a maintained subscription over the same query does not — which violates the requirement in §6.4.2 that both reduce to the same result at the same frontier.

Either the maintained surface admits the empty global row (which needs an answer to what its synthetic identity is, since there is no group key to derive one from), or the one-shot path stops emitting it and we diverge from SQL in a documented way. I did not pick — it is recorded as an open question in ch. 6 and it is genuinely your call, since you wrote the SQL-alignment half. It is also the only contradiction in this surface that is visible today rather than latent, so it is worth settling before this merges rather than after.

Also recorded, and explicitly out of scope for the lane: whether `sum` should widen its result type. Today it stays at input width and errors on overflow, so `sum` over a `U8` column fails at 256. Postgres widens instead. Widening is probably where we end up, but it changes the declared descriptor type of every existing `sum`, so it should be a deliberate cut.

PR #1126, opened 2026-07-24
fix(groove): preserve aggregate state incrementally

joeinnes:
## Summary

- keep maintained aggregate arrangements unchanged when an update fails
- stage and commit only touched aggregate groups, avoiding copies of unrelated state
- cover overflow recovery and allocation scale independence through Groove's public database API

## Benchmark

Compared the pre-fix implementation at `9966299fb` with this branch at
`a7fe3fa86`. Each figure is the median across three runs; each run measures
20–100 subscription updates. Timings use Rust's unoptimised test profile, so
the absolute latency is not representative of production builds. Allocation
comparisons are exact.

| Scenario | Pre-fix median | Fixed median | Time cost | Allocation cost |
| 20k groups; update one small group | 75.9 µs | 78.4 µs | +3.3% | +12.2% |
| One 1k-row group | 0.835 ms | 0.889 ms | +6.5% | +19.1% |
| One 10k-row group | 7.87 ms | 8.60 ms | +9.3% | +14.6% |
| One 50k-row group | 41.0 ms | 44.9 ms | +9.5% | +14.3% |
| Ungrouped, 50k rows | 40.7 ms | 44.6 ms | +9.6% | +14.3% |

The fix adds a constant-factor cost from cloning the touched group for rollback
safety. It does not make updates scale with unrelated groups or change the
existing O(touched group size) complexity.

## Testing

- `cargo test -p groove -j 2` — passed
- `cargo test -p jazz -j 2` — passed
- `cargo test -p jazz-tools --features test -j 2` — passed
- `cargo test -p jazz-server -j 2` — passed
- `cargo check -p jazz-sim --benches` — passed
- `dev/gates/ts-wire-codec.sh` — passed
- `JAZZ_SEED_COUNT=300 cargo test -p jazz m3_maintained_one_shot_differential_oracle` — passed
- `cargo test -p jazz --test incremental_delivery_canary maintained_relation_include_single_row_changes_are_scale_independent -- --exact` — passed
- `dev/benchmarks/smoke.sh` — passed

### DIFF
--- a/crates/groove/SPEC/3_queries_operators.md
+++ b/crates/groove/SPEC/3_queries_operators.md
@@ -355,7 +355,8 @@ can be updated by weighted deltas:
 
 - `count(*)`: signed total input multiplicity for the group.
 - `count(expr)`: signed total multiplicity where `expr` is non-null.
-- `sum(expr)`: weighted sum over numeric values.
+- `sum(expr)`: weighted sum over numeric values, including signed `I64`.
+- `avg(expr)`: mean over non-null values, returned as `F64`.
 - `min(expr)` / `max(expr)`: extremum over positive-multiplicity values, backed
   by an ordered value index with deterministic full-record tie accounting.
 - `any_value(expr)`: the value from the deterministic least ordered witness,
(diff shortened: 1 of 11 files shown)


aeplay:

This is a decision rather than a defect, so flagging rather than blocking.

**The state-atomicity fix looks right**, and the MIN/MAX deletion tests are
meaningful coverage.

**What it does not do is bound work by change size.** Only touched buckets are
staged (`crates/groove/src/ivm/runtime/join.rs:405-413`) and updated
(`runtime/mod.rs:5943-5957`), but `aggregate_row_from_records` is then called
for both the after and before states (`runtime/mod.rs:5961-5980`), and that
passes the *entire group* to each aggregate (`:7546-7569`). SUM, AVG and the
extrema all loop over records (`:7632-7670`, `:7705-7722`, `:7733-7759`). So a
single-row change to a large group costs O(group size), not O(1).

The PR says this is accepted, and there is a real reason it might be:

| aggregate | current update | invertible? |
|---|---|---|
| COUNT | recomputed from retained positives | yes, but no inverse accumulator used |
| SUM | recomputed from decoded values | yes for integers, recomputed anyway |
| AVG | recomputed as f64 sum+count | a (sum,count) state could be inverted |
| MIN / MAX | full scan for the extremum | **no** — retracting the current extreme needs the next one |

MIN/MAX genuinely cannot be inverted from the current extreme, so correct
deletion handling *comes from* retaining the whole group multiset. That makes
the rescan defensible for those two. COUNT/SUM/AVG could be accumulators, and
MIN/MAX could use a retained extrema index.

So the choice is: document an explicit INV-INC-1 exception for aggregate
maintenance with a touched-group scaling test that pins the accepted cost, or
implement accumulators plus an extrema index. Either is fine; silently carrying
an O(touched-group) path under a "bounded incremental maintenance" claim is the
option I would avoid, because that is how the invariant erodes.

One smaller thing: I did not find an add-then-retract regression for floating
SUM/AVG. Repeated add/retract on f64 drifts from a recomputed value. Worth
either a test pinning the tolerance or a note that the drift is accepted and
unbounded.

#1125 looks fine on its own once its conflicts are resolved.

aeplay:
🤖 Closing as superseded by #1299. The newer merged aggregate stack preserves incremental aggregate runtime state and delivers maintained aggregate subscriptions through the current terminal path, including empty-group/retraction coverage. This PR targets the pre-terminal runtime/query-evaluation structures and would duplicate or conflict with the replacement implementation.

PR #1128, opened 2026-07-25
fix: clear dropped per-index signals when reconcileArray truncates

wrangelvid:
Truncating with `length =` leaves reactive array proxies (we use deepsignal) holding stale per-index signals, so reads past the new length return old rows. Truncate with splice instead, and export `./reconcile-array` as a subpath so apps can reuse the reconciler directly.


joeinnes (approved):
LGTM! Added a changeset and a behavioural regression test, thanks!

PR #1129, opened 2026-07-25
fix: scope browser client storage by identity for cookie sessions

wrangelvid:
Cookie-authenticated sessions all shared one browser storage namespace, so switching users in the same browser read another identity's local replica (and the sync server then ignored their query subscriptions on the user_id mismatch, which shows up as queries loading forever until a reconnect). Key the storage scope by identity, mirroring resolveDefaultPersistentDbName. Comes with a runtime test.


joeinnes (approved):
Useful fix, thank you!

PR #1130, opened 2026-07-25
fix: remove stale binaries before staging in stage-binary.mjs

wrangelvid:
On macOS, overwriting a staged binary in place keeps the old vnode and the kernel SIGKILLs the relaunched executable because its signature no longer matches the cached file. Delete the destination before copying.


joeinnes (approved):
Thank you!

joeinnes (approved):
Great, thanks!

PR #1132, opened 2026-07-25
fix: resolve array reverse references through element index lookups

wrangelvid:
Reverse references pointing into array columns resolved by scanning even though array elements are indexed individually. Look them up through the element index instead; the content check still verifies every hit.


wrangelvid:
The test-ts failure is the chat-react invite browser E2E timing out; it passes 3/3 locally on this branch with the wasm rebuilt from it, and main had the same kind of E2E timeout on an unrelated TS-only merge (run 30272635847). Looks flaky, could someone rerun it?

aeplay:
🤖 Closing because this optimization targets deleted policy-evaluator and write paths. A current regression or benchmark should be opened only if the new core exhibits the same issue.

PR #1133, opened 2026-07-25
perf: reconcile only tracked rows on re-delivered batch fates

wrangelvid:
Follow-up to #1127. Servers re-deliver settled fates for batches a node never tracked, and each one still paid a store-wide history scan, so repeated bulk imports slowed down as the store grew: by our 8th 100-row import run inserts took ~140 ms each, with this they stay at ~16 ms. Reconcile only the rows the tracked member sources know and keep the full scan as a fallback for replicas that predate the batch row index.

Also applies AcceptedTransaction fates even when the fate itself is unchanged, so a crash between fate persistence and row application heals on replay.


aeplay:
🤖 Closing because the new core uses the bounded fate/replay machinery from #1292 and #1295 instead of the retired tracked-row reconciliation design.

PR #1134, opened 2026-07-25
fix: stop leaking server subscriptions for one-shot queries

wrangelvid:
One-shot reads (a subscribe/unsubscribe pair coalescing in one inbox drain) left their server subscription installed forever, so every read grew the server's subscription set. On our production backend point reads degraded from ~5 ms to ~228 ms after a few thousand reads. Serve the pair once, scope and rows included so cold replicas still get the data, without installing the subscription, and still enqueue the unsubscription so a reconnect replay cannot resurrect one.


aeplay:
🤖 Closing as superseded by the new-core one-shot subscription cleanup work in #1265.

PR #1135, opened 2026-07-25
fix: push session-authorized policy context rows to user clients

wrangelvid:
Stacked on #1134, review the last commit. Clients evaluating read policies with exists/join subqueries never received the rows those policies depend on, so delivered rows stayed invisible on cold replicas. The server now pushes the transitive policy-context closure, filtered by each context table's own read policy so this never widens what a session can read. Covered by an e2e sync test.


aeplay:
🤖 Closing because the old server-pushed policy-context mechanism was replaced by scoped proof/claim routing in #1261, #1319, and #1400. Current integration coverage continues in #1678.

PR #1136, opened 2026-07-25
perf: bound per-connection scope bookkeeping and policy context scans

wrangelvid:
Stacked on #1134 and #1135, review the last commit. Policy context collection scanned whole tables per settle, and per-client scope entries were never dropped on unsubscribe, so long-lived connections kept growing. Iterate the _id/_id_deleted indexes instead (tombstones included so deletions propagate) and drop scope with the client's last subscription, keeping same-drain resubscriptions intact.


aeplay:
🤖 Closing because the per-connection scope structures optimized here no longer exist. Current lifecycle ownership is covered by the new-core subscription cleanup work.

PR #1137, opened 2026-07-27
Source create-jazz starters from published releases

joeinnes:
## Summary

- Clone starter templates from the matching `v<version>` release tag.
- Resolve remote workspace packages from that same release tag.
- Create the matching tag in the alpha publish workflow.

## Testing

- `pnpm --filter create-jazz test` — 71 tests passed.

### DIFF
--- a/.changeset/released-starters.md
+++ b/.changeset/released-starters.md
@@ -1,0 +1,5 @@
+---
+"create-jazz": patch
+---
+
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Superseded by #1952, which preserves the immutable release-snapshot intent while enforcing source/package version equality and safe tag-before-publish ordering.

PR #1138, opened 2026-07-27
Update starters for Better Auth 1.6.24

joeinnes:
## Summary

- Pin Better Auth declarations to exact `1.6.24` across the affected starters and workspace metadata.
- Replace incompatible `jwtClient()` usage with typed `authClient.$fetch("/token")`.
- Keep starter compatibility checks and auth flows aligned with the tested Better Auth release.

## Testing

- `pnpm --filter create-jazz test` — 71 tests passed.
- `pnpm --filter create-jazz-e2e check` — passed.


packages/create-jazz/src/starter-compatibility.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    section: "devDependencies",
+  },
+  {
+    path: "packages/jazz-tools/package.json",
+    section: "peerDependencies",
+  },
+] as const;
+
+describe("Better Auth workspace compatibility", () => {
+  it("requires every Better Auth workspace declaration to use the exact tested release", () => {
```
nrainhart: I'd personally remove this test, don't think it adds much

nrainhart (approved):
Thanks Joe!! If only BetterAuth was not such a moving target...

PR #1141, opened 2026-07-27
(new core) Widen query membership filters to non-text scalars and array elements

aeplay:
Supersedes #808, reimplemented from scratch against the new core (that branch targets the retired pre-swap query engine, so nothing was ported).

## What was broken

`in` and `contains` only worked usefully for text. `contains` on an array column coerced the literal against the **whole array type** rather than the element type, so `contains(numbers, 42)` or `contains(reviewerIds, <uuid>)` could never coerce. The public Rust builder had no `in` pass-through at all.

## What changed

- Array `contains` needles are coerced against the **element** type (unwrapping `Nullable` first).
- `QueryBuilder::filter_in` lowers to `Predicate::In`.
- `in` now covers integer / float / boolean / uuid-reference columns; `contains` covers arrays of numbers, booleans, references, enums, timestamps and text. Text `contains` keeps substring semantics; invalid combinations still error before execution.

Semantics live in the core per SPEC 13.13 — the TS layer carries types and pass-through only.

## Spec

`SPEC/6_queries.md` §6.1.1 now documents the supported column-type × operator matrix and names the narrow compatibility coercions.

**Deliberately not included:** broader literal-vs-column coercion (e.g. string→numeric). That needs its own spec decision rather than accreting case-by-case rules, and is called out as an open question in the chapter and in `dev/QUERY_MEMBERSHIP_FILTERS.md`.

## Coverage

Black-box tests in `crates/jazz-tools/tests/query_membership_filters.rs` through `JazzClient` with builder schemas and `row_input!`: scalar `in`, non-text array `contains`, text substring `contains`, and the rejection cases.

## Gates

jazz · groove · jazz-tools `--features test` · incremental delivery canaries · `JAZZ_SEED_COUNT=300` oracle · ts-wire-codec · jazz-server · jazz-sim bench check · smoke · clippy · fmt · sensitive-data guard — all green, re-verified by the reviewer.

### DIFF
--- a/dev/QUERY_MEMBERSHIP_FILTERS.md
+++ b/dev/QUERY_MEMBERSHIP_FILTERS.md
@@ -1,0 +1,56 @@
+# Query Membership Filters
+
+## Findings
+
+Reference PR #808 (`Support broader query membership filters`) was used only as
+a requirements reference. Its intent is:
+
+- widen typed `where` inputs so `in` works on scalar and reference columns;
+- translate scalar `in` lists element-by-element;
+- cover array-column `contains` for non-text element types;
+- keep TypeScript as typed pass-through, with semantics owned by the Rust core.
+
+Current branch support before implementation:
+
+| Column type                       | `in` validation                                                            | `in` lowering/eval                                    | `contains` validation                                                                | `contains` lowering/eval                                                                                                   |
+| --------------------------------- | -------------------------------------------------------------------------- | ----------------------------------------------------- | ------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------- |
+| Text/String                       | accepts text values; uuid literals are coercible to strings                | lowered as OR of equality predicates                  | accepts text needle as substring search                                              | `TextContains` is not lowered in maintained core; TS/runtime path still covers text contains through existing surface      |
+| Integer / numeric                 | accepts matching numeric values                                            | lowered as OR of equality predicates                  | rejected on scalar columns                                                           | n/a                                                                                                                        |
+| Float                             | accepts matching float values                                              | lowered as OR of equality predicates                  | rejected on scalar columns                                                           | n/a                                                                                                                        |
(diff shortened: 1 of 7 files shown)


dev/QUERY_MEMBERSHIP_FILTERS.md
```diff
@@ -0,0 +1,56 @@
...
+| Bytea                             | accepts byte arrays as whole-column values                                 | lowered as OR of equality predicates                  | rejected by TS adapter; core validation rejects non-array/non-string scalar contains | n/a                                                                                                                        |
+| Json                              | accepts JSON as whole-column values through TS adapter                     | core comparability depends on value representation    | rejected by TS adapter                                                               | n/a                                                                                                                        |
+| Array<Text>                       | accepts whole-array values for `in`                                        | lowered as OR of equality predicates                  | accepts text element needle                                                          | lowered as Groove `Contains` over array membership                                                                         |
```
joeinnes, line 26: Current implementation doesn't quite match this for `in` validation:

https://github.com/garden-co/jazz/blob/4d1f5a148c7f8e524f03c1d7af5bf58899594a11/crates/jazz/src/query.rs#L2861

This would allow a scalar to pass validation if it matches the member type of the array, which would then fail to match anything (because Groove would try to match the scalar with the array column). I think ideally we'd validate and reject this.

joeinnes (approved):

aeplay:
🤖 Agreed, and I'd go further: silently returning empty is the worst of the options, because it reads as "no data" rather than "malformed query".

Picked up in `track/in-filter-scalar-validation`, but I've deliberately briefed it wider than this call site. This is the third time literal-vs-column-type coercion has come up and been settled ad hoc, so the lane has to answer the general question — reject at validation, or coerce to a single-element containment check — argue it in terms of the query invariants, and write the outcome into the owning SPEC chapter so the next case has a rule instead of a precedent. Your lean toward rejecting is in the brief. If it concludes this needs a maintainer call rather than an implementer one, it stops and says so.

I'll link the PR here. Sorry for merging with this unaddressed.

PR #1146, opened 2026-07-27
(new core) Fix non-null writes into nullable columns

aeplay:
## The bug

Writing a **non-null** value into a **nullable** column through the public client produced a row that could not be read back correctly.

`core_cells` in `crates/jazz-tools/src/client.rs` converted public `Value`s to core values without consulting the schema. A nullable column expects `CoreValue::Nullable(Some(..))`, but a non-null public value arrived as the bare `CoreValue`. `Value::Null` was handled correctly (it became `Nullable(None)`), which is exactly why this stayed hidden: only the non-null-into-nullable direction was broken, and the null direction is what tests reached for.

Every public write entry point shared that conversion: `insert`, `insert_with_id`, `upsert`, `update`, and the staged transaction variants of each.

## The fix

Route all of them through a schema-aware conversion that wraps non-null values for nullable columns. The sibling function `core_ordered_values` already had schema access, so the asymmetry was accidental rather than structural.

## Coverage

New `crates/jazz-tools/tests/nullable_writes.rs` does black-box readback across nullable `Text`, `Integer`, `BigInt`, `Boolean`, FK/UUID and array columns, once per affected entry point — direct and staged.

## Gates

`cargo test -p jazz-tools --features test -j 2` green (full crate suite).

## Found in passing, deliberately *not* fixed here

`crates/jazz-tools/tests/catalogue_sync_integration.rs` is an **orphaned test target**: the crate sets `autotests = false` and the file has no `[[test]]` entry, so it has never been compiled or run. Registering it lists 25 tests, of which 21 fail (`websocket session user_id must be a UUID`, and `spawn_local called from outside of a task::LocalSet`).

That is out of scope here and registering it would turn this PR red, so it is left alone. It needs its own decision — the failures may mean the target rotted after an API evolution and should be repaired, or that it was superseded and should be deleted. Filed separately rather than resolved unilaterally.

### DIFF
--- a/crates/jazz-tools/Cargo.toml
+++ b/crates/jazz-tools/Cargo.toml
@@ -212,6 +212,10 @@ required-features = ["test"]
 name = "policy_branch_closure"
 required-features = ["test"]
 
+[[test]]
+name = "nullable_writes"
+required-features = ["test-utils"]
+
 [[example]]
 name = "realistic_bench"
 required-features = ["client"]
--- a/crates/jazz-tools/src/client.rs
+++ b/crates/jazz-tools/src/client.rs
@@ -15,7 +15,8 @@ use crate::public_api::types::{OrderedAdded, OrderedRemoved, OrderedUpdated};
 use crate::public_schema::Schema;
 use crate::public_schema::TableName;
 use crate::public_schema::{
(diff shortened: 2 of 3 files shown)


joeinnes (approved):

aeplay:
superseded by #1160

PR #1148, opened 2026-07-27
(new core) Fix SUM/AVG over public Integer columns, and keep Integer 4 bytes wide

aeplay:
## The bug

`SUM` and `AVG` over a public `Integer` column were wrong. Measured on two rows in one group, scores 10 and 7:

| query | before | correct |
|---|---|---|
| `AVG(score)` | `Double(2147483656.5)` | `Double(8.5)` |
| `SUM(score)` | `Query error: Storage: unsupported operator` | `Integer(17)` |

`2147483656.5` is exactly `((10 ^ 0x80000000) + (7 ^ 0x80000000)) / 2`.

The public client encoded `Integer` (i32) into core `U32` with an order-preserving sign-bit flip (`^ 0x8000_0000`). That is correct for ordering — MIN, MAX and ORDER BY were fine — and it round-trips for a single value, because decode inverts encode. It is not *arithmetic*-preserving, and SUM/AVG do arithmetic in the encoded domain. SUM errored rather than lying only because two encoded values exceed `u32::MAX` and `u32::try_from` fails in the IVM runtime.

## Why no test caught it

Every existing aggregate test put at most **one row per group** (a=10, b=7, delete, repopulate). With one value per group the encoding round-trips and the answer is right by accident. The AVG cases additionally used a BigInt schema — the one integer type that is not sign-flipped.

## The fix

The sign-flip was not wrong in itself; it was applied in the wrong place. As a **key** encoding, an order-preserving transform is correct and necessary. As a **value** encoding it breaks arithmetic. So values are now stored raw and signed, and the order-preserving transform lives only in the key path.

Public `Integer` maps to a core signed 32-bit type, and `BigInt` continues to map to `I64`. **The public API is unchanged** — `Value::Integer(i32)` and `Value::BigInt(i64)` round-trip exactly as before, matching PostgreSQL `int4`/`int8` semantics.

An intermediate step mapped `Integer` to core `I64`, which fixed the arithmetic but doubled every `Integer` column from 4 bytes to 8 (`ValueType::fixed_size` gives `U32 => 4`, `I64 => 8`) and collapsed the `Integer`/`BigInt` distinction into documentation only. Adding a real core `I32` keeps the type distinction meaningful end to end and holds `Integer` at 4 bytes.

## Newly-reachable-variant fallout, also fixed here

Introducing a new signed variant into the reachable set exposed production paths that matched only the unsigned variants and fell through:

- **`maintained_subscription_view.rs`** — `apply_aggregate_value_delta` ended `_ => delta`, so it **overwrote the running total with the latest delta instead of accumulating**. A maintained aggregate view over an `Integer` column would have reported the last change, not the sum. Silent, no error. `zero_aggregate_value` also never zeroed the new variant at initialization, and `aggregate_payload_is_empty` never recognized an all-zero payload, so rows that should retract persisted.
- **`ingest.rs`** — signed counter columns hit `InvalidStoredValue("counter value must be integer")`.
- **`oracle.rs`** — the counter oracle panicked.
- **`window_codec.rs`** — `is_integer_type` excluded the signed variant, so the windowed value-delta encoding was silently skipped for Integer columns. Correct output, but the optimization stopped applying on a hot path.
- **`query_eval.rs`** — result-ordering comparison arms were missing.

A sweep confirmed no remaining production site where the signed variants are reachable and fall through.

## Coverage

Black-box tests, each failing without its corresponding fix:

- SUM and AVG over `Integer` with **multiple rows per group**, negative and mixed-sign values
- maintained aggregate view accumulating **multiple incremental deltas**, plus retraction of an emptied group — the case that would have caught the accumulator bug
- signed counter column merges
- MIN/MAX and ORDER BY over `Integer` across zero, pinning the behaviour the old encoding got right so the fix cannot regress it
- an assertion on the **stored width** of an `Integer` column, so a future change cannot silently widen it again

`QueryBuilder` also gains `avg`/`min`/`max`, needed to express these tests through the public API rather than hand-assembling aggregate specs. That supersedes #1147, which can be closed.

## Gates

`jazz`, `groove`, `jazz-tools --features test`, `jazz-server`, `cargo check -p jazz-sim --benches`, all three incremental-delivery canaries, the m3 differential oracle, `ts-wire-codec.sh`, `cargo check --workspace --examples`, and `dev/benchmarks/smoke.sh`.


crates/groove/src/window_codec.rs
```diff
@@ -1047,6 +1068,7 @@ mod tests {
                 ("u16", ValueType::U16),
                 ("u32", ValueType::U32),
```
joeinnes, line 1069: Do we want to add the `i32` counterpart here?

crates/groove/src/window_codec.rs
```diff
@@ -1286,6 +1336,7 @@ mod tests {
                 Value::U16(rng.next() as u16),
                 Value::U32(rng.next() as u32),
```
joeinnes, line 1366: Same, might want to add `I32` for completeness

crates/jazz-tools/src/public_api/query.rs
joeinnes, line 625: `column` should only be optional if function is `Count`, we should check this during validation too.

crates/jazz/src/node/policy.rs
joeinnes, line 978: This can now return false for numerically equal values if the claim and row are different widths (i.e.: we have an `I32`/`Integer` being compared with an `I64`/`BigInt`). Should we normalise before comparison?

joeinnes (approved):
Looks good, once merged, I'll rebase #1125 and #1126 onto this.

crates/jazz/src/node/policy.rs
aeplay: 🤖 Good catch, and it was worse than the one site — fixed in `96c259d7b`.

You were right that this is a regression the `I32` change introduced: before, public `Integer` mapped to core `U32`, so this particular width mismatch did not arise the same way. Since it is policy comparison, the failure mode is a policy that *should* match not matching — access denied where it should be granted.

Policy equality now compares all integer widths and signedness exactly via `i128`. Two edge cases handled deliberately rather than by blanket widening: `U64` stays exact above `i64::MAX`, and float/int equality remains type-exact rather than being coerced.

**A second site had the same defect**: policy **join correlations** used raw equality on the same values, so a correlated join across an `Integer` column and a `BigInt` claim would have silently failed to correlate. That one was not in your comment — it turned up from auditing the neighbourhood. Also fixed, and covered.

Audited and found clean: policy `IN` already routes through the shared equality helper; ordinary filters are schema-coerced, joins schema-typed, ordering typed — so no further width-mismatch fix applied.

Also preserved signed JWT integer claims as `I64` at the server boundary, with public client/server regressions covering `I64`↔`I32` and `U64`↔`I64` policy matches.

crates/jazz-tools/src/public_api/query.rs
aeplay: 🤖 Done in `96c259d7b`. A non-`Count` aggregate without a column now fails validation with a clear error instead of `.expect()`-panicking at use; `Count` without a column stays valid. Kept the type shape as-is and added validation, since changing `column` to be conditionally-typed would ripple further than the problem warrants — say if you would rather have the type encode it.

crates/groove/src/window_codec.rs
aeplay: 🤖 Added in `96c259d7b`, both here and at :1337, with reachable `I32` round-trip and delta coverage rather than arms added on faith.

PR #1150, opened 2026-07-28
(new core) Fix writes wrongly denied after a table rename, keeping policies lowered and pinned

aeplay:
Stacked on `track/catalogue-sync-target` (the catalogue harness repair).

## The bug

After a v1 `users` → v2 `people` rename, a v2 insert was denied. It never reached predicate evaluation: write-side resolution failed as `TableNotFound("people")`, and that **resolution** error was laundered into `authorization_denied`. Updates had a second instance — the parent row's *unprojected* table name was matched before the rename lens applied, so a v1 `users` parent was skipped before it could become v2 `people`.

## Two wrong turns, and why they were wrong

This took three attempts, and the discarded two are worth recording.

**Attempt 1** removed the compiled query-program path for insert policies and routed everything through the interpreter. It worked — but it reverted `e4fdad384` ("Lower insert write policies through query engine"). Fixing a bug by abandoning the lowered path means the engine swap quietly loses ground every time a defect happens to live in lowering.

**Attempt 2** restored lowering but resolved policies against the **active write schema**. That broke `rls_policy_under_lenses_evaluates_translated_data_against_pinned_policy` — the registered guard for:

> **`INV-LENS-19`**: Policy evaluation under lenses MUST translate data into the pinned permission evaluation schema and MUST NOT translate policy bundles.

Attempt 2 moved the *policy*. The invariant says move the *data*.

## The fix

The write-path projection already existed and already applied table renames via the lens. The missing piece was narrower than either attempt assumed: **its selected pinned schema identity was not carried into lowered evaluation.** The insert evaluator now validates and lowers the unchanged policy against that explicit pinned schema, while the candidate is projected into it.

So the policy stays pinned and unmodified, the row data is projected to meet it, and `INV-LENS-19` holds — which is what it was describing all along.

Also fixed: resolution errors now propagate with `?` instead of being converted to `authorization_denied`. A schema-resolution failure reported as a permission denial cost two debugging cycles here.

## Interpreter signposts

Per review direction — we want to remove the interpreter, not lean on it — the legacy write-policy family now carries signposts recording that lowered coverage *"must shrink this family, never widen it"*, and that it is kept only for not-yet-lowered cases, *"never as a workaround for a lowering defect."* That is precisely how it grew back this time.

## Verification

- `rls_policy_under_lenses_evaluates_translated_data_against_pinned_policy` — passes (independently re-run).
- `rename_write_authorization.rs` — 4/4 pass.
- `multi_hop_table_renames_and_column_rename` now gets **past the write** and fails later at the read-side assertion. That remaining failure is the lens-projection gap, owned by another lane, and is expected.

Gates: `jazz`, `groove`, `jazz-server`, `cargo check -p jazz-sim --benches`, all three incremental-delivery canaries, the m3 differential oracle at seed 300, and the sensitive-data guard — all exit 0.

Not clean: `jazz-tools` still shows the known catalogue integration failures owned by other lanes, and `ts-wire-codec.sh` could not run because workspace JS dependencies are not installed on the build host.


joeinnes (approved):

crates/jazz/src/node/ingest.rs
```diff
@@ -593,7 +593,7 @@ where
             .remove(&tx_id)
             .unwrap_or(stored.tx.made_by);
         for version in &records {
-            if !self.version_satisfies_write_policy(version, permission_subject) {
+            if !self.version_satisfies_write_policy(version, permission_subject)? {
```
joeinnes, line 596: If the transaction errors here due to policy resolution, it'll remain pending. When it's retried, without the override above, it'll fall back to `made_by`, which is a problem if/when a backend is working on behalf of a principal.

aeplay:
🤖 Confirmed, and the mechanism is a bit sharper than described — good catch.

```rust
let permission_subject = self.open_tx.local_permission_subjects
    .remove(&tx_id)                    // consumed here
    .unwrap_or(stored.tx.made_by);     // fallback on retry
```

It's `.remove()`, not `.get()`, so the override is **consumed before the fallible section completes**. Any `Err` after that point (`ingest_rejected_transaction(...)?`, or a later `?`) leaves the transaction `Fate::Pending` with the override already gone — and the `if !matches!(stored.fate, Fate::Pending) { return Ok(()) }` guard above means a retry re-enters exactly this path and takes the `made_by` fallback.

So for a `TrustedBackend` writing on behalf of a principal, the retry evaluates write policy against the *claimed author* rather than the *authenticated link identity* — which is what `INV-RLS-17` and `INV-RLS-18` require. Privilege confusion, latent until trusted-backend attribution is used in anger.

Picked up in `track/retry-preserves-permission-subject`; the fix is to consume the override on success rather than before the fallible section, with a regression that errors on first attempt and asserts the retry still evaluates against the link identity. I'll link the PR here when it opens.

Apologies for merging this with your comment unaddressed — that's on me. I'd built the "ready to merge" list from review state alone and missed that approvals here routinely carry comments. Now checking review bodies and inline comments before calling anything ready.

PR #1154, opened 2026-07-28
(new core) Fix restore-then-read inside a mergeable transaction

aeplay:
## The bug

Inside a mergeable transaction, restoring a deleted row and then reading it back reported the row as **still deleted**. After commit it was correctly restored. Read-your-writes was violated for `restore` specifically.

Found while investigating whether the duplicated mergeable-write representation could be removed — it could not, and this is a concrete instance of the two representations disagreeing.

## Why

A mergeable transaction currently keeps two representations of its pending writes: a core open transaction (which in-transaction **reads** evaluate against) and a legacy `Vec<WasmTxWrite>` buffer (which **commit** replays).

`restoreEncoded` staged only a *content* write into the open transaction, so the read overlay still reported the row deleted. The legacy buffer later replayed a *true restore event* at commit. Reads saw one representation, commit applied the other.

## The fix

WASM restore staging now records both the content write **and** the `Restored` deletion state, so the open transaction's read overlay matches what replay will do. The two representations agree for restore.

Deliberately not fixed by special-casing reads, and not by deleting one representation — a separate investigation established the legacy buffer cannot yet be removed (`OpenExclusive` carries no per-write time, mergeable parent vectors, or transaction identity, and commits as `TxKind::Exclusive`). Making them agree is the available correct move; removal needs core-side mergeable staging first.

## Coverage

Black-box regression through the public TypeScript API: delete a row, open a mergeable transaction, restore it, read it back **inside** the transaction, and assert it is visible — plus a post-commit assertion so a fix cannot trade one for the other. Confirmed failing first (returned `null` inside the transaction), then passing.

## Neighbourhood audit

`delete`, `update`, `upsert`, and insert-after-delete were checked for the same staged-vs-replayed asymmetry and already use matching semantics on both paths. Insert-after-delete correctly remains deleted until an explicit restore. So this was an instance, not a class.

## Gates

Independently re-run rather than taken from the lane, which lost some exit statuses:

`cargo test -p jazz` · `-p groove` · `-p jazz-tools --features test` · `-p jazz-server` · `cargo check -p jazz-sim --benches` · `JAZZ_SEED_COUNT=300 m3_maintained_one_shot_differential_oracle` · sensitive-data guard — **all exit 0**. Transaction Vitest 22/22.

### DIFF
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -2112,7 +2112,7 @@ where
             .map_err(Into::into)
     }
 
-    /// Stage a restore inside an owned exclusive transaction handle.
+    /// Stage a restore inside an owned exclusive transaction handle, applying defaults for omitted columns.
     pub fn exclusive_restore(
         &self,
         tx_id: OpenTxId,
@@ -2120,11 +2120,12 @@ where
         row: RowUuid,
         cells: RowCells,
     ) -> Result<(), Error> {
-        if cells.is_empty() {
-            return Err(Error::new(ErrorCode::Schema, "restore requires row data"));
-        }
         let cells = self.apply_insert_defaults(table, cells)?;
         let mut node = self.node.node.borrow_mut();
+        // Restore needs one content version and one deletion-register version:
+        // `tx_write` rejects a version carrying both. The layers have separate
+        // winners and parent chains; see `restore`'s `local_*_winner_tx_id` pair.
+        // Keep this staged form aligned with the committed restore path.
         node.tx_write(tx_id, table, row, cells, None)?;
         node.tx_write(
             tx_id,
@@ -2165,7 +2166,7 @@ where
             .map_err(Into::into)
     }
 
-    /// Restore a row locally. Data is required by the public API contract.
+    /// Restore a row locally, applying defaults for omitted columns.
     ///
     /// ```rust
     /// # use jazz::db::doctest_support::{block_on, open_todos_db, todo_cells};
@@ -2186,9 +2187,6 @@ where
         row: RowUuid,
         cells: RowCells,
     ) -> Result<WriteHandle<S>, Error> {
-        if cells.is_empty() {
-            return Err(Error::new(ErrorCode::Schema, "restore requires row data"));
-        }
         let cells = self.apply_insert_defaults(table, cells)?;
         self.ensure_row_deleted(table, row, self.identity.author)?;
         let (content_parents, deletion_parents) = {
@@ -2232,9 +2230,6 @@ where
         row: RowUuid,
         cells: RowCells,
     ) -> Result<WriteHandle<S>, Error> {
-        if cells.is_empty() {
-            return Err(Error::new(ErrorCode::Schema, "restore requires row data"));
-        }
         let cells = self.apply_insert_defaults(table, cells)?;
         self.ensure_row_deleted(table, row, identity)?;
         let (content_parents, deletion_parents) = {
@@ -2301,9 +2296,6 @@ where
         cells: RowCells,
         now_ms: u64,
     ) -> Result<WriteHandle<S>, Error> {
-        if cells.is_empty() {
-            return Err(Error::new(ErrorCode::Schema, "restore requires row data"));
-        }
         let cells = self.apply_insert_defaults(table, cells)?;
         self.ensure_row_deleted(table, row, self.identity.author)?;
         let (content_parents, deletion_parents) = self.row_layer_parents(table, row)?;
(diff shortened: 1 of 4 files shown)


crates/jazz/src/db.rs
```diff
@@ -2112,6 +2112,30 @@ where
...
+    pub fn exclusive_restore(
+        &self,
+        tx_id: OpenTxId,
+        table: &str,
+        row: RowUuid,
+        cells: RowCells,
+    ) -> Result<(), Error> {
+        if cells.is_empty() {
+            return Err(Error::new(ErrorCode::Schema, "restore requires row data"));
+        }
```
nrainhart: Shouldn't we make this change after applying the table defaults?

crates/jazz/src/db.rs
```diff
@@ -2112,6 +2112,30 @@ where
...
+        let cells = self.apply_insert_defaults(table, cells)?;
+        let mut node = self.node.node.borrow_mut();
+        node.tx_write(tx_id, table, row, cells, None)?;
+        node.tx_write(
+            tx_id,
+            table,
+            row,
+            BTreeMap::<String, Value>::new(),
+            Some(DeletionEvent::Restored),
+        )?;
```
nrainhart, line 2136: So... a restore triggers two separate writes? Why not just a single `tx_write` operation?

aeplay: 🤖 Good question — the two writes are load-bearing, and the API shape is what makes them look redundant.

Content and deletion are **independent layers with independent parent chains and independent HLC winners**. They resolve to different storage tables (`node/currency.rs:79-82`) and have separate winner queries (`local_content_winner_tx_id` / `local_deletion_winner_tx_id`). The non-transaction restore path does the same two writes and shows why plainly — it passes *different parent vectors*, `content_parents` vs `deletion_parents` (`db.rs:2193-2216`). A single record cannot carry two parent vectors.

Why they have to stay independent:

- **A delete concurrent with a content edit** must hide the row without dominating or erasing content history — so a later restore brings back the *edited* content, not the pre-delete content.
- **A restore concurrent with a content edit** must not become a sibling of that edit on one chain. Sharing a parent set would manufacture an HLC race between two operations that do not actually conflict.
- After a table rename, a shared chain would let a v2 content write parent on (and dominate) a v1 delete — silently un-deleting a row as a side effect of schema migration.

`tx_write` accepting both `cells` and `deletion` in one call is what makes one call look sufficient. The staged path here deliberately mirrors the committed path, which is the point of the PR: reads evaluate the staged representation and commit replays the other, so they have to agree.

Adding a code comment saying this — you should not have had to ask.

crates/jazz/src/db.rs
aeplay: 🤖 You are right, and thanks — this is a real ordering bug.

The empty-cells check runs *before* `apply_insert_defaults`, so a table whose columns are all defaulted rejects a restore that defaults would have satisfied.

Fixing it, and extending the fix rather than just moving the one line: restore and insert should follow the **same** rule, across both the staged/transaction and non-transaction variants. A lane is auditing all of those entry points now and making them consistent, with a test that a fully-defaulted row can be restored (and inserted) with empty cells.

One caveat it has been told to check first: if the early rejection turns out to be deliberate — "restore must be explicit about its data" — that is a semantic decision rather than a bug, and it should come back and say so instead of changing it. I do not think that is the case, but I would rather it verify than assume.

Will follow up on this thread with the commit.

crates/jazz/src/db.rs
aeplay: 🤖 Done in `5df0195ef`. The comment now reads:

```rust
// Restore needs one content version and one deletion-register version:
// `tx_write` rejects a version carrying both. The layers have separate
// winners and parent chains; see `restore`'s `local_*_winner_tx_id` pair.
// Keep this staged form aligned with the committed restore path.
```

One correction to what I said earlier — the constraint is stronger than "a single record cannot carry two parent vectors". `tx_write` **actively rejects** a version carrying both content and a deletion event, so one call was never an option, it would have failed. Verified rather than assumed.

crates/jazz/src/db.rs
aeplay: 🤖 Fixed in `5df0195ef`, and it was worse than the one line you spotted.

Defaults are now applied before validation across **five** paths — exclusive, direct, identity, timestamped, and mergeable transaction — so restore matches insert everywhere rather than just at the site you commented on. Only fixing the one you found would have left four inconsistent siblings.

Coverage: direct and transaction regressions for restoring with empty data on an all-default table.

It also factored the repeated content/deletion parent lookup into a `row_layer_parents` helper, so `db.rs` is net **-22 lines** despite the fix.

Gates: full Rust canonical set, the transaction and restore-api Vitest suites (28 passed), `pnpm build:core`, and the sensitive-data guard — all green.

PR #1155, opened 2026-07-28
(new core) Catalogue projection stack: 4 → 19 of 24 acceptance tests

aeplay:
Takes `catalogue_sync_integration` from **4 passed / 20 failed** to **19 passed / 5 failed** (independently re-verified). The protected acceptance file itself is unchanged throughout — every gain is a production fix.

## Background

This test file was **orphaned**: `autotests = false` plus no `[[test]]` entry meant it had never been compiled or run, for months, while people kept editing it. Registering it exposed 21 failures. Those were not stale tests — they were real defects the file had been failing to catch.

## The stack, in dependency order

Each layer was a genuine product bug, and each was invisible until the one before it was fixed.

1. **Harness repair** — LocalSet wrappers, UUID session ids, target registration. 4 passed.
2. **Permissions-head readiness** — dynamic shells treated the runtime schema as permission-ready before any permissions head existed, so reads settled and writes were ingested early, and later policy changes did not invalidate live subscriber state. **→ 8 passed.**
3. **Runtime catalogue bridge** — schema versions and lenses were published into the admin catalogue but never bridged into the WebSocket-serving runtime shell. A client on an older schema had its shape parked, and the server then **silently dropped** the matching `Subscribe`. One shared bridge now serves both admin routes and the in-memory helper, and parked subscriptions receive `SubscribeRejected` instead of vanishing.
4. **Lens-projected maintained lowering** — same-table projected current sources, with winner selection performed schema-agnostically *before* projection.
5. **Write auth across renames** (#1150) — resolution failures were laundered into `authorization_denied`.
6. **Catalogue-projected runtime refresh** — the remaining projection stack. **→ 19 passed.**

## Additional real bugs found and fixed along the way

- Nested null typing was a general **Groove projection** defect, not a catalogue one.
- Schema-only forward-path checks could translate rows through an **unrelated table lens**.
- Disconnected table lineage could poison otherwise valid schema reads.
- Local `Db` subscriptions lacked runtime-generation invalidation.
- Currency source caching is structurally unsafe across table renames.

## The five that remain are *not* one more layer

The 11 repaired failures formed one coherent catalogue-projection stack. The remaining five are **three unrelated pieces** and should be scoped separately:

| piece | tests |
|---|---|
| public relational-query adapter (join + array-subquery lowering, public result materialization) | 3 |
| projected-current IVM live update | 1 |
| write-side currency lookup across rename lineage | 1 |

That distinction is the point of the exercise: continuing to treat them as a block would have been wrong.

## Gates

`cargo test -p jazz` · `-p groove` · `-p jazz-server` · `cargo check -p jazz-sim --benches` · 300-seed maintained-vs-one-shot oracle · scale-independence canary · `dev/benchmarks/smoke.sh` (all scenarios) · sensitive-data guard — **all exit 0**.

`-p jazz-tools --features test` fails only on the five catalogue cases above.

Includes the branch behind #1150; that PR can land first and this rebases cleanly.


crates/groove/src/ivm/graph.rs
```diff
@@ -558,6 +574,10 @@ impl ProjectField {
 pub enum ProjectExpr {
     Field(FieldRef),
     Literal(LiteralValue),
+    TypedLiteral {
+        value: LiteralValue,
+        value_type: ValueType,
+    },
```
nrainhart, line 580: Can you explain the motivation for adding this new expression type?

Or even better, each a small comment on top of each enum constructor

crates/jazz-server/src/lib.rs
```diff
@@ -307,6 +307,14 @@ impl ShellDb {
         }
     }
 
+    fn publish_lens(&self, lens: MigrationLens) -> ShellResult<Vec<SyncMessage>> {
+        match self {
+            Self::Memory(db) => db.publish_lens(lens).map_err(Into::into),
+            #[cfg(feature = "rocksdb")]
+            Self::Rocks(db) => db.publish_lens(lens).map_err(Into::into),
+        }
```
nrainhart, line 315: Nit: we could have a small macro to avoid this code duplication

crates/jazz-tools/tests/rename_write_authorization.rs
```diff
@@ -0,0 +1,333 @@
...
+    let mut authority = open_node(node(0x90), v1.schema.clone());
+    let mut writer_v1 = open_node(node(0x10), v1.schema.clone());
+    let user_row = row(0x77);
+
+    let (insert_tx, insert_unit) = writer_v1
+        .commit_mergeable_unit(
+            MergeableCommit::new("users", user_row, 1_000)
+                .made_by(alice)
+                .cells(cells(user_row, "alice@example.com", alice)),
+        )
```
nrainhart, line 181: Nit: I'd prefer we use methods closer to the end-user API (instead of the internal `commit_mergeable_unit`) and avoid propagating changes manually across nodes in integration tests. That'll make them easier to read and more resilient to internal refactors.

This is already captured by https://github.com/garden-co/jazz/blob/main/crates/jazz-tools/TESTING_GUIDELINES.md, but i guess agents are not paying attention to it

nrainhart (approved):

aeplay:
🤖 Superseded and split, per @aeplay.

- **#1194** carries the landing half — the 20 commits through `73ae9d791`: catalogue projection stack, catalogue-to-runtime bridge, write-auth and lowered-policy fixes across table renames, permissions-head gating.
- **`track/lens-partition-parked`** holds the four tip commits — authored-schema partition provisioning, removal of eager publish-time materialization, lineage-aware currency lookup, cross-lineage deletion resolution. Three were self-labelled `WIP:`, and all four are storage-layout work that the forthcoming single-table storage is expected to supersede. Nothing is lost.

Splitting at `73ae9d791` also removed an architectural fork from the critical path: `INV-LENS-11` is identical to the base at that commit, and the stay-put-vs-copy-forward divergence arrives only in `9c24bb9e4`, which is parked. So the partition-semantics question stays open for single-table storage to answer rather than being settled prematurely.

Thanks for the review — it carried over to #1194, which is the same content minus the parked commits.

PR #1159, opened 2026-07-28
(new core) Remove the legacy write-policy interpreters

aeplay:
**Net ~1,150 lines removed.** Closes out the top item on the parallel-paths audit's removal backlog.

## Why there were two

Two implementations decided write admission: a legacy interpreter (`policy_allows_insert_candidate` / `policy_base_allows*`) and lowered evaluation through the query engine (`e4fdad384`). Nothing structurally forced them to agree, and that fork produced one regression and one reverted fix in a single evening of work on #1150.

## The precondition

#1153 built a differential harness and measured **three live disagreements**:

| case | interpreter | lowering |
|---|---|---|
| inherited `Insert` | **denies** | allows |
| inherited `Update` | **denies** | allows |
| inherited `Delete` | **denies** | allows |

The interpreter was correct. `InheritsOperation` specifies the matching parent **write**-policy clause; lowering always applied the parent **read** policy — which is generally more permissive, so lowering would have admitted writes that should be denied.

Production was never exposed: inherited policies were explicitly delegated to the interpreter before lowering was reached. **That delegation was load-bearing, not caution** — it was routing around this exact gap. Deleting it without fixing lowering first would have shipped a permissive-write bug.

## What changed

- ordinary and branch write authorization now route through query lowering;
- inherited `Insert`/`Update`/`Delete` select the matching parent write-policy clause;
- **both interpreters deleted** — not deprecated. A kept-but-unused evaluator is the duplication being ended.

The differential harness from #1153 is replaced by direct lowering tests. Once there is one implementation, a *differential* comparison has nothing to compare — so that harness retires having done its job. (Note the contrast with the m3 maintained-vs-one-shot oracle, whose premise survives: it compares incremental accumulation against fresh computation, not implementation A against B.)

Retained coverage spans operations, inheritance, branches, identities, deep chains, and schema/table renames. Authorization spec and invariants updated.

## Gates

`cargo test -p jazz` · `-p groove` · `-p jazz-tools --features test` · `-p jazz-server` · `cargo check -p jazz-sim --benches` · all three incremental-delivery canaries · the exact relation/include canary · 300-seed maintained-vs-one-shot oracle · `dev/benchmarks/smoke.sh` (all scenarios) · `dev/gates/ts-wire-codec.sh` · sensitive-data guard · `cargo fmt --check` — **all exit 0**.

The wire-codec gate initially failed only because `tsc` was absent on the build host; re-run green after `pnpm install --frozen-lockfile`.

Supersedes #1153, which can be closed — its harness lives on here as the lowering tests.


aeplay:
🤖 Rebased on the current base and resolved a subtle interaction worth recording.

## What blocked it

#1148 merged, fixing width-mismatch comparison **in the interpreter** — `policy_values_equal` normalizes integer widths via `i128` so an `Integer(7)` column matches a `BigInt(7)` claim. This PR deletes that interpreter. Taking either side of the merge wholesale would have silently dropped @joeinnes' fix, and it would have looked fine: his regression tests exercise the interpreter path that no longer exists, so they would have reported green while covering nothing.

## What went wrong on the first attempt

Carrying normalization forward is right, but the first pass applied it to **join-arrangement key encoding** as well as policy comparison. That broke five Groove tests — `arg_max_by`, `arg_min_by`, and `aggregate_incremental_matches_recompute_under_seeded_changes` — because those derive group and winner keys through the same key-encoding machinery. Changing key bytes changes grouping and winner ordering.

Verified those pass on the base, so they were caused rather than inherited.

## The distinction now enforced

- **policy value comparison** — normalizes integer widths and signedness. Covers equality, `IN`, array containment, and policy join correlations.
- **arrangement key encoding** — stays exact typed encoding. Keys are compared as bytes for grouping and ordering; normalizing them changes IVM semantics, not policy semantics.

There is now a test proving policy joins normalize while ordinary arrangement keys do not, which is the property that was implicitly violated.

Cross-width lowered policy comparison still passes for equality, `IN`, `contains`, large `U64` above `i64::MAX`, and float type-exactness.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-tools --features test`, `-p jazz-server`, `cargo check -p jazz-sim --benches`, all three incremental-delivery canaries, the 300-seed m3 oracle, `ts-wire-codec.sh`, `smoke.sh`, and the sensitive-data guard — all exit 0. Groove and the interpreter-removal check independently re-verified.

crates/jazz/src/node/query_engine/lowering.rs
```diff
@@ -2780,7 +2780,11 @@ fn lower_linear_plan_steps(
                 available_route_fields.extend(introduced_route_fields);
                 if !matches!(residual, PredicateExpr::True) {
                     let predicate = lower_predicate(&residual, source, root_source, request)?;
-                    graph = graph.filter(predicate);
+                    graph = if uses_policy_value_comparison(request) {
+                        graph.policy_filter(predicate)
```
joeinnes, line 2784: Not sure if this is intentional, but `policy_filter` can evaluate range/null checks, while the spec says we should deny those until support is explicit:
https://github.com/garden-co/jazz/blob/849832626efd49ec6ec28846addd471a7888bed0/crates/jazz/SPEC/7_authorization.md?plain=1#L75.

joeinnes (approved):

PR #1161, opened 2026-07-28
(new core) Separate desired intent from implementation status across 27 SPEC chapters

aeplay:
Markdown only. Applies the pattern established in #1145 across the rest of the spec.

## The pattern

Three destinations for every claim:

1. **Normative invariant text = timeless desired intent.** No rollout state ("not yet", "until X lands"), and no naming of specific operators or cases that go stale. An invariant that names four operators is wrong the moment one changes; one that states the principle cannot go stale.
2. **Implementation-status notes** in the chapter's Details, tied to a **named test** where one exists — that is what makes spec/code agreement checkable rather than aspirational.
3. **🔶 Open questions** in the owning chapter, for genuine unresolved design decisions.

Seven parallel lanes, one per chapter group, 26 files. Merged with **zero conflicts** — the chapters are disjoint.

## Registry updates applied centrally

Lanes were forbidden from editing either `INVARIANTS.md`, because those registries mirror chapter text and seven concurrent editors would conflict on every line. Each reported its replacement lines instead; they are applied here in one pass. **66 Jazz and 5 Groove entries.**

**Bidirectional audit:** no chapter invariant changed by these merges lacked a reported registry update. No lane reworded without reporting.

## One conflict, resolved on ownership — and the code agrees

Two lanes gave incompatible wording for `INV-SHAPE-16`:

- *"until an explicit shape-drop action drops the shape"* (Jazz cross-reference)
- *"while the shape remains registered"* (Groove, the owner)

Resolved by the rule that **the owning crate defines canonical text and cross-references align to it** — so Groove's wording wins.

That turns out to be right on the merits too, not just by procedure: the runtime removes an auto-family prepared shape after its last routed subscription is removed, with **no explicit shape-drop action** (`ivm/runtime/mod.rs:1117`, `:1504`). Registration-state wording matches the code; the action-based wording did not.

Also aligns Appendix D with #1157's box-appropriate `-j` guidance.

## Pre-existing drift, deliberately left alone

A full registry audit found **184 Jazz and 88 Groove rows that do not match their chapter text**, and four Groove ids with **no chapter bullet at all** — including `INV-INC-1`, the incremental-delivery mechanism law.

None of that is caused by this PR and it is not fixed here; mixing 272 legacy corrections into a structural pass would make both unreviewable. Flagged separately as its own piece of work.


crates/jazz/SPEC/E_glossary.md
```diff
@@ -122,7 +125,6 @@ Restored}`) · **global-current overwrite table** — node-local derived current
 - 🔶 **Flat index.** Keep this dependency-ordered grouping, or add a flat
   alphabetical index for lookup as well?
 - 🔶 **Facade spelling.** The high-level facade spells mergeable transactions as
-  _batch_ and exclusive transactions as _transaction_; the implementation has not
-  yet landed those facade spellings.
-- 🔶 **Edge fate authority.** An edge is the mergeable fate authority; the
-  implementation path currently uses the core for mergeable fate authority.
+  _batch_ and exclusive transactions as _transaction_. `Db::transaction` is
+  present; decide whether mergeable writes should expose the corresponding
+  `batch` spelling.
```
nrainhart, line 130: Shouldn't we have gotten rid of the `batch` term completely by now?

nrainhart (approved):
Really like the direction this is going 🚀

PR #1162, opened 2026-07-28
(new core) Unify mergeable transaction staging into one representation

aeplay:
## The duplication removed

A mergeable transaction kept **two** representations of its pending writes: a core open transaction (which in-transaction *reads* evaluated against) and a `Vec<WasmTxWrite>` buffer (which *commit* replayed). Nothing forced them to agree — and #1154 exists precisely because they didn't, for `restore`.

`Vec<WasmTxWrite>` and its replay helpers are now **deleted**. One representation.

## Equivalence proven before deletion, not assumed

The core open-transaction representation gained mergeable staging semantics — restore's paired content/deletion records, coalescing and order, per-write provenance times, parent vectors, author and permission subject — plus a mergeable-open commit path lowering to `commit_mergeable_many`.

A new internal test compares exact `SyncMessage::CommitUnit` batches from both paths: batch id, ordering, timestamps, parents, author, permission subject, intervening commits, rollback. Internal rather than black-box because those protocol details aren't publicly observable.

## A regression this introduced, and what it says about our coverage

The unified path didn't inherit the defaults-before-validation rule from `5df0195ef`, so restoring an all-defaults row failed with `Schema: restore requires row data` — reintroducing the bug @nrainhart caught in review on #1154.

**Every Rust gate passed while that was live.** The regression lives entirely at the public TypeScript surface, and the only thing guarding it was two TS tests. Restored here, passing 28/28, with restore and insert now following the same rule across direct, identity, timestamped, exclusive, legacy mergeable and unified mergeable paths.

Worth stating plainly: a fully green Rust gate set meant nothing for this class of change.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-tools --features test`, `-p jazz-server`, `cargo check -p jazz-sim --benches`, all three incremental-delivery canaries plus the exact relation canary, the 300-seed maintained-vs-one-shot oracle, `ts-wire-codec.sh`, both transaction Vitest suites (22 runtime, 27 browser), `cargo test --workspace`, `dev/benchmarks/smoke.sh` (14 scenarios), and the sensitive-data guard.

Repo-wide `pnpm test` cannot pass due to unrelated pre-existing breakage in the `auth-simple-chat` and `world-tour` example packages — filed separately.


crates/jazz/src/db.rs
```diff
@@ -2061,6 +2061,144 @@ where
...
+    pub fn begin_mergeable_for_identity(&self, author: AuthorId) -> Result<OpenTxId, Error> {
+        self.node
+            .node
+            .borrow_mut()
+            .open_mergeable(author, Some(author))
+            .map_err(Into::into)
+    }
+
+    /// Stage a full row value in an owned mergeable transaction.
+    pub fn mergeable_insert(
```
nrainhart: ⚠️ There's now both `MergeableTx.insert` and `Db.mergeable_insert` (same thing with the other CRUD methods). Isn't it redundant?

nrainhart (changes requested):

aeplay:
🤖 @nrainhart you're right, and it's worse than redundant — it's this PR's own thesis left half-applied. Thanks for catching it.

## What's actually there

The PR description says the `Vec<WasmTxWrite>` buffer is deleted and a mergeable transaction now has one representation. But `MergeableTx` (`crates/jazz/src/db.rs:6693`) holds:

```rust
pub struct MergeableTx<'a, S> {
    db: &'a Db<S>,
    author: AuthorId,
    permission_subject: Option<AuthorId>,
    writes: Vec<PendingMergeableWrite>,   // the buffer, renamed
}
```

with its own coalescing in `stage_value_write` (`:6924`) and `stage_deletion_write` (`:6936`), replayed into `MergeableCommit`s at `commit()` (`:6866`).

Meanwhile `Db::mergeable_insert/update/delete/restore` (`:2083`+) take an `OpenTxId` and write **through `node.tx_write_mergeable`** into the core open transaction.

So the buffer wasn't removed — it was moved and renamed. There are still two staging implementations with two independent coalescing implementations and nothing forcing them to agree, which is precisely the condition that produced #1154.

## It's observable, not just untidy

Writes staged via `Db::mergeable_*` land in the core open transaction, so in-transaction reads see them. Writes staged via `MergeableTx` sit in a private `Vec`, so a read against the same `Db` won't see them until commit. Same logical operation, different read-your-writes semantics.

## What I'm not proposing

Deleting one of the two entry points. `jazz-wasm` (`crates/jazz-wasm/src/lib.rs:501`+) and `jazz-napi` can't hold a `MergeableTx<'a, S>` across an FFI boundary — they need a handle keyed by `OpenTxId`. Two **entry points** is legitimate; two **implementations** is the bug.

## The fix

`MergeableTx` owns an `OpenTxId` and delegates staging to the same `tx_write_mergeable` path, so `PendingMergeableWrite` and both coalescing helpers go away. One representation, one coalescing implementation, with an ergonomic RAII wrapper over it — which is what this PR set out to do; it just stopped at the wasm buffer.

Underway in `fix/mergeable-tx-single-representation`. Two things I've asked for explicitly before the helpers are deleted:

- **proof the core path reproduces the existing coalescing semantics** rather than assuming it. `stage_value_write` replaces a matching non-deletion write for the same `(table, row)`; `stage_deletion_write` retains or drops based on `Deleted` vs `Restored`. If core doesn't already behave identically, that difference is a bug in one of them and I want to know which before either is changed.
- a test asserting **both entry points produce identical results** for the same operation sequence, including restore and delete-then-reinsert — restore being what broke in #1154.

Will push to this branch when it's green.

aeplay:
🤖 Pushed `a955e9d60`. `MergeableTx` now owns an `OpenTxId` and stages through the same `tx_write_mergeable` path as `Db::mergeable_*`; `PendingMergeableWrite`, `stage_value_write` and `stage_deletion_write` are gone — zero occurrences left in `db.rs`. One representation, one coalescing implementation, two entry points.

**The coalescing equivalence held**, which was the thing I didn't want assumed. Core `stage_mergeable_write` replaces same-row non-deletion writes; `Deleted` removes all same-row pending entries; `Restored` removes deletion entries while retaining the content entry — the exact rules the deleted helpers encoded. The strongest evidence is that the six pre-existing coalescing tests (`..._coalesces_insert_then_update_for_same_row`, `..._coalesces_update_then_delete_for_same_row`, `..._coalesces_restore_then_update_for_same_row`, and the rest) still pass unchanged against the new implementation.

Plus the two new ones:
- `mergeable_tx_and_open_handle_have_identical_restore_and_reinsert_results` — both entry points, same sequence, identical results, covering restore and delete-then-reinsert
- `mergeable_tx_read_observes_its_staged_restore` — the user-visible consequence

That second one is worth calling out: `MergeableTx` gained a `read`, backed by the open-transaction overlay. Previously staged writes were invisible to reads on the same `Db` until commit, while the `Db::mergeable_*` path made them visible — so the two APIs disagreed on read-your-writes. They now agree.

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-sim --benches`, and a direct `cargo check -p jazz-napi -p jazz-wasm` all exit 0 — the last one because both bindings consume this API.

`cargo test -p jazz-tools --features test` fails only `numeric_claims_match_integer_columns_across_core_widths`, red on `codex/jazz-core-engine-swap` itself and fixed by #1159.

`cargo check --workspace --all-targets` fails on the stale exhaustive match at `crates/jazz-sim/tests/customer_child_policy_minimal.rs:205` missing `SubscriptionEvent::Rejected` — untouched by this branch, and pre-existing. That one-line gap has now independently blocked four separate pieces of work today; I'm landing it separately.

Thanks again for the catch — the buffer had genuinely survived the PR that existed to remove it.

crates/jazz/src/db.rs
nrainhart: As I mentioned earlier, I don't think we should have a duplicate API (`Db.mergeable_insert`/`MergeableTx.insert` and so on). But at least this is no longer duplicate logic, so feel free to merge. We can clean up as a FUP

nrainhart (approved):

aeplay:
🤖 Pushed the two follow-ups you flagged, on this PR as requested.

**Why both layers exist** — now documented on `Db::mergeable_insert`, since this was the question. The `Db::mergeable_*` primitives are the single implementation; `MergeableTx` is a handle holding an `OpenTxId` that delegates to exactly them, adding a read overlay and abandon-on-drop. Both are needed because callers differ in *who owns the transaction lifetime*: the wasm and napi bindings keep a transaction open across separate FFI calls, so there is no Rust frame to borrow across — and constructing a `MergeableTx` per call would abandon the transaction when that handle dropped at the end of the call. So collapsing to one surface would break the bindings, not just inconvenience them.

**A real bug came out of chasing that.** `commit` consumed `self` without suppressing `Drop`, so `Drop` called `abandon_transaction_handle` on an already-committed transaction. Benign today only because `abandon_tx` tolerates an unknown id, and silent because the result was discarded by `let _ =` — the correctness of commit rested on an error being swallowed. A `committed` flag now suppresses it, set only on success so a failed commit still abandons rather than leaking a handle.

`ExclusiveTx` has the same `db + tx_id` shape but no `Drop`, so it is unaffected — though that means it has the opposite exposure: an abandoned `ExclusiveTx` leaks its open transaction. Noted rather than fixed here, to keep this PR scoped.

Merged the current base in too — the only conflict was `SMOKE_LEDGER.md`, append-only, both sides kept in timestamp order.

`cargo test -p jazz` exits 0, with all 8 `mergeable_tx_*` tests passing including the two added earlier.

PR #1164, opened 2026-07-28
(new core) Permit writes into non-current authored partitions — 20 → 21 of 24

aeplay:
Stacked on #1155. Implements the design decision on schema-version write skew.

## The question this answers

What happens when a client writes using a schema that is **not** the current write schema? A v2 `people` client updating a row physically stored under v1 `users`, when the pointer hasn't flipped.

Three options were on the table:

1. **Writes converge on current** — translate backward through the lens into the current partition. Rejected: lenses are not guaranteed invertible, and one that drops a column can't run backward without losing data.
2. **Permit writes into non-current authored partitions**, with commit arrival as an additional provisioning trigger alongside pointer flip. **Chosen.**
3. **Reject newer-schema writes until the pointer advances.** Rejected: makes a client that upgraded ahead of the server read-only, which doesn't fit local-first.

Read fan-out across partitions is the accepted cost, and the machinery already existed — the cross-lineage winner selection on this branch.

## What changed

- commits provision and write to their **authored** schema partition; no pointer-driven lens rewrite or rejection;
- parked state survives live partition-layout refreshes — an earlier attempt provisioned eagerly at `PublishSchema`, and `rebuild_database_slot` cleared parked shape registrations before they could drain;
- projected reads fixed for non-current schemas;
- lens SPEC and invariant registry updated to state the decided rule.

## Also fixed here: deleted rows reappearing after a rename

Deletion was being applied **within each partition before v1 and v2 were unioned**, so a v2 deletion removed v2 content but let older v1 content resurface. Content and deletion winners are now resolved independently across lineage, honouring HLC ordering, physical-source provenance, policy-schema pinning (INV-LENS-19) and global metadata traversal.

That one is worth noting on its own: a row you deleted coming back because someone renamed a table.

## One test intentionally updated

`old_schema_commit_units_copy_on_write_into_current_schema_partition` asserted **option 1** — that a v1 commit is rewritten into the current partition with lens-translated cells. That semantic was explicitly rejected, so the test now asserts authored-partition behaviour instead. Net +50/−7; it grew rather than shrank.

## Result

`catalogue_sync_integration`: **21 passed / 3 failed**, up from 20/4. The three remaining are the relational-adapter cluster (join and array-subquery lowering), which is blocked on separate work and unaffected by this.

Gates: `cargo test -p jazz`, `-p groove`, `-p jazz-server` — exit 0, independently re-run. `-p jazz-tools --features test` fails only on those three.

Note: clippy is blocked in `currency.rs` by **pre-existing** dead-code and `filter_map` warnings, unrelated to this change — filed separately.


nrainhart (approved):
Let's merge this implementation for now, I'll work on a simplified storage layer for fluid migrations as a follow-up

PR #1166, opened 2026-07-28
(new core) Give INV-INC-1 a performance receipt, and a gate 120x tighter than the canary

aeplay:
Closes the top-ranked gap from `dev/benchmarks/OVERVIEW.md`.

## The problem

`INV-INC-1` is the mechanism law: work to ingest, apply and publish a change must be bounded by the change and its affected keys, **never** by accumulated view size. Corollary 1 says full re-materialization on a maintained path is a defect *even when the observable output is correct*.

It was enforced only by an exact functional canary — one parent, 1k or 20k children, insert one child, assert allocations stay within a **3× band**. That proves the property at one point. No slope, no curve, and no number to set an acceptance target against, which matters because a lot of recent design work is justified against this invariant.

## The receipt

A scale ladder: change held fixed at exactly one child row, accumulated view size varied across five rungs.

| accumulated children | allocations | allocation bytes |
|---:|---:|---:|
| 1,000 | 3,881 | 486,632 |
| 2,500 | 3,882 | 487,176 |
| 5,000 | 3,882 | 487,184 |
| 10,000 | 3,885 | 495,064 |
| 20,000 | 3,882 | 487,184 |

**Allocation ratio 1.001×. Byte ratio 1.017×.** Across a 20× change in accumulated state. The 10k rung is an outlier rather than a trend, so the curve is flat rather than merely bounded.

Reproduced independently before merge.

## The gate

**1.025×**, derived from the measured spread rather than chosen.

Worth stating plainly: the existing canary's 3× band is roughly **120× looser than reality**. A regression that quadrupled delivery work would have passed it. The canary stays — it tests a different thing, exactly and functionally — but the ladder is what would actually catch drift.

Allocation counters are the primary signal, not wall time: they were deterministic across runs on a shared host, while wall time is noise. Recorded as auxiliary only.

## Integration

Emits a smoke-shaped JSONL receipt and is wired into `dev/benchmarks/smoke.sh`, so it joins committed history with previous/delta rather than becoming another disconnected artifact. Methodology in `dev/benchmarks/INV_INC_1_RECEIPT.md`. The scenario itself costs 12.0s.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-tools --features test`, `-p jazz-server`, `cargo check -p jazz-sim --benches`, `ts-wire-codec.sh`, the 300-seed maintained-vs-one-shot oracle, the full `incremental_delivery_canary` suite, `smoke.sh`, and the sensitive-data guard — all exit 0.

### DIFF
--- a/crates/jazz/benches/relation_include_delivery.rs
+++ b/crates/jazz/benches/relation_include_delivery.rs
@@ -1,0 +1,396 @@
+use std::alloc::{GlobalAlloc, Layout, System};
+use std::cell::Cell;
+use std::collections::BTreeMap;
+use std::time::Instant;
+
+mod support;
+
+use jazz::block_on;
+use jazz::db::{Db, DbConfig, DbIdentity, ReadOpts, SeededRowIdSource, SubscriptionEvent};
+use jazz::groove::records::Value;
+use jazz::groove::schema::{ColumnSchema, ColumnType};
+use jazz::groove::storage::MemoryStorage;
+use jazz::ids::{AuthorId, NodeUuid, RowUuid};
+use jazz::query::{ArraySubquery, Query};
+use jazz::schema::{JazzSchema, Policy, TableSchema};
+use serde_json::json;
+use support::{csv_usizes, emit_json_line, env_usize, phase_fields};
+
+struct CountingAllocator;
+
+// The receipt runs each sample on the main thread. Keeping these counters
+// thread-local prevents unrelated allocator activity from becoming part of a
+// sample if the harness evolves to do setup concurrently.
+thread_local! {
+    static ACTIVE: Cell<bool> = const { Cell::new(false) };
+    static ALLOCS: Cell<u64> = const { Cell::new(0) };
+    static BYTES: Cell<u64> = const { Cell::new(0) };
+}
+
+unsafe impl GlobalAlloc for CountingAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let _ = ACTIVE.try_with(|active| {
+            if active.get() {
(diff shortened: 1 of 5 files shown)


crates/jazz/benches/relation_include_delivery.rs
```diff
@@ -0,0 +1,396 @@
...
+        .and_then(|value| value.parse::<f64>().ok())
+        .unwrap_or(DEFAULT_MAX_RATIO);
+
+    let mut rungs = Vec::with_capacity(scales.len());
+    for scale in scales {
+        let mut measurements = (0..samples)
+            .map(|sample| measure_single_child_insert(scale, sample))
+            .collect::<Vec<_>>();
+        measurements.sort_by_key(|measurement| measurement.allocs);
+        let median = measurements[measurements.len() / 2];
```
joeinnes, line 74: Technically this emits *all* measurements from the run that produced the median allocation, not independent medians for each metric.

joeinnes (approved):

aeplay:
🤖 Correct, and this matters more than a normal benchmark nit: this harness produces the `INV-INC-1` receipt and the 1.025x gate threshold is read off these numbers. A statistic labelled "median" that is really "the co-measurements of the median-allocation run" makes cross-metric comparison unsound, and the gate depends on it.

Picked up in `track/bench-per-metric-medians`. Per-metric independent medians are the obvious fix, but the lane also has to check whether any metrics must stay co-measured to remain comparable (a ratio between two would be corrupted by mixing runs) and label those explicitly if so. It then re-derives the gate threshold under the corrected statistic — and if the numbers move the gate, it reports rather than retuning, since that's a maintainer decision.

I'll link the PR here. Sorry for merging with this open.

PR #1168, opened 2026-07-28
(new core) feat: move JSON validation into jazz crate

nrainhart:
## Description

Restored `json_storage` Rust integration tests and in the process found that JSON validation was being (partially) performed on the TS side instead of on the core `jazz` crate. This PR makes that a core `jazz` concern.


crates/jazz/src/schema.rs
```diff
@@ -416,15 +416,114 @@ pub enum MergeStrategy {
     Counter,
 }
 
-fn is_counter_column_type(column_type: &GrooveColumnType) -> bool {
+/// Logical Jazz application-column type.
+///
+/// Most variants lower directly to the corresponding Groove storage type.
+/// JSON remains a Jazz semantic type and lowers to a Groove string.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Deserialize, serde::Serialize)]
+pub enum ColumnType {
```
nrainhart, line 491: I don't think it makes sense to include the Json column type in Groove (at least not right now). This led me to introduce this new lower-level Jazz ColumnType, since now different Jazz column types can map to the same underlying Groove type

aeplay:
🤖 Review findings. Direction is right and the execution is mostly clean — two validation bypasses worth fixing before merge, one decision already resolved, one non-blocking idea.

## Must fix: two paths into storage skip the new validation

Both are the same shape — validation landed on the main path but not every path.

**1. `NodeState::commit_mergeable` (`crates/jazz/src/node/mod.rs:927`)**

The public entry point performs only structural validation, then persists through the unvalidated parts encoder (`mod.rs:1121`). A Rust caller can commit `Value::String("{")` into a JSON column and create invalid local state — which a peer then rejects on the normal wire path. So the node poisons itself locally and only finds out on sync.

**2. Forward-lens ingestion (`crates/jazz/src/node/ingest.rs:1500` and `:4338`)**

Forward-lens ingestion takes an inbound `VersionRecord`, transforms it, then uses `VersionRow::from_parts_with_schema_version` — bypassing the new validation in the no-lens `from_wire` route. An old-schema upload whose value lands in a constrained JSON column can persist malformed or schema-invalid JSON at the receiving node.

Worth noting the restored `json_storage` tests only exercise direct `Db` writes, so neither bypass would be caught by them.

## Resolved: the compatibility break is intentional and fine

The old TS validator implemented a small JSON-Schema subset; `jsonschema` enforces substantially more. Concretely, TS ignored an `additionalProperties` subschema, so `{"x": 7}` under `{type:"object", additionalProperties:{type:"string"}}` used to pass and now fails.

**Breaking compatibility in favour of better JSON validation is explicitly OK** — Anselm confirmed. Worth a line in the PR description or a test pinning the new stricter behaviour, so the next person reads it as a decision rather than a surprise.

## Non-blocking: compile schemas at admission rather than per write

`crates/jazz/src/node/codec.rs:1920` parses and compiles the JSON Schema on **every non-null cell write**. Compiling once when the schema is admitted and reusing it would take that off the write path, and would surface malformed schemas at configuration time instead of on first write. Not a blocker — worth doing if write-path cost becomes material.

## Praise, and it is earned

- The TypeScript validator was **actually removed**, not left dormant alongside the new one (`packages/jazz-tools/src/runtime/value-converter.ts:143` now only serializes). That is the failure mode this class of change usually hits.
- Normal wire ingestion validates before storage and errors stay diagnosable — `InvalidJsonCell` maps to the public `Schema` code.
- `crates/jazz/tests/json_storage.rs:85` is a real behaviour suite — preservation, syntax rejection, schema rejection, failed-update atomicity — not a smoke test.

Checks run: `cargo test -p jazz --test json_storage` (4 passed) and `cargo test -p jazz` — both exit 0. `ts-wire-codec.sh` failed on a missing `tsc` in this environment, not on the PR.

nrainhart:
Addressed the comments. I intentionally didn't fix the _Forward-lens ingestion_ issue because right now we don't have lenses that allow transforming column types (or changing a json column's schema)

aeplay:
🤖 @nrainhart I pushed a commit to this branch (`432d2a3ea`) rather than just replying, because chasing down your answer turned up something bigger than what I originally reported. Short version: **you were right that lenses can't cause this, and I was wrong about where the problem was.** The bypass is not lens-specific — ordinary remote ingest never validated cell values at all.

Everything below is self-contained so it reads cold.

## Why the original finding was mis-scoped

My review said forward-lens ingestion skips validation, and that "normal wire ingestion validates before storage." That second claim was wrong, and it's what made your reasoning look like it only had to cover lenses.

There are two ingest routes:

- `ingest_reset_view_bundle_refs_in_bulk` — has the asymmetry I described. Its no-lens branch uses the validating `VersionRow::from_wire_with_schema_version`; its lens branch does not.
- `stage_transaction_and_versions_with_current_indexes` — the main remote-transaction path, reached from `stage_known_transaction` and `ingest_transaction_and_versions_with_current_indexes`. It builds `target_cells` off the wire and calls `from_parts_with_schema_version` **unconditionally**, lens or no lens.

`validate_cell_value` only ever ran inside `history_values_from_wire` (`codec.rs:1315`). The second route never reaches it.

## The reproduction

Two black-box tests at the public ingest boundary, written before any fix. Both failed on this branch at `10e82da8e`, **with no lens configured anywhere**:

```
remote_ingest_rejects_syntactically_invalid_json ... FAILED
  left: Some({"payload": String("{")})              right: None

remote_ingest_rejects_json_schema_violation ... FAILED
  left: Some({"payload": String("{\"name\":123}")})  right: None
```

So any peer could persist malformed or schema-invalid JSON at any node today, with no lenses deployed. That's why I moved on it rather than leaving it for the lens work.

The tests construct the `VersionRecord` directly. That's deliberate, not a shortcut: the local write path you fixed correctly rejects these values, so the only way an invalid cell reaches storage is from a peer that didn't validate. That untrusted wire record *is* the subject under test. Flagging it explicitly per `TESTING_GUIDELINES.md`.

## Why your specific reasoning still wouldn't have covered it

Even restricted to the lens path, "no lens transforms column types" isn't sufficient, because two of today's ops move a value into a *different* constraint without changing any column's type:

- **`CopyColumn { from, to }`** — copies a value between two JSON columns that already have different schemas. Valid under the source, invalid under the target. No column's schema changed; the value just got measured against a different one.
- **`AddColumn { default }`** — `CompiledLensOp::Add` (`node/mod.rs:3864`) inserts a lens-authored default into cells that go straight to storage. A default that doesn't satisfy the target column's schema silently poisons every row migrated through that lens. Not attacker-driven, just a misconfigured lens.

Both are moot next to the unconditional bypass above, but they're the reason the fix validates *after* lens translation rather than before.

## The fix

`invalid_cell_value_reason`, sitting next to the existing `reject_source_delta_reason` and called at the same three ingest entry points. It reuses **your** `validate_cells_map` helper, and reuses the existing rejection machinery: an offending unit comes back as `Fate::Rejected(MalformedCommit(..))`, so the peer is told and the connection isn't torn down. Deletion records are skipped; lens-chain errors are left to the staging path that already reports them.

It validates against the schema the cells land in, after forward-lens translation, which is what covers the `CopyColumn` / `AddColumn` cases.

**Deliberately not inside `from_parts_with_schema_version`.** That constructor is also called from `maintained_subscription_view.rs:1602/1624` and `branches.rs:258`, which are interior paths. Validating there would be interior revalidation in a hot path — the thing we removed from the serve path in July. Ingest is the trust boundary; that's where the check belongs, and only there.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-sim --benches`, the 300-seed m3 oracle, all three incremental-delivery canaries, `ts-wire-codec.sh`, and `smoke.sh` all exit 0. Smoke is clean, so putting validation on the ingest path didn't move the benchmarks.

`cargo test -p jazz-tools --features test` fails one test, `numeric_claims_match_integer_columns_across_core_widths`. **This is not caused by either commit here.** I re-ran it on unmodified `10e82da8e` and it fails identically, and then on `origin/codex/jazz-core-engine-swap` — where it also fails. So the canonical jazz-tools gate is currently red on the integration base itself, and this branch inherits it rather than introducing it. I've raised that separately; there's an investigation open into whether read-side cross-width claim matching is genuinely broken. Nothing for you to do here, but it means a green jazz-tools run on this PR isn't achievable until that's resolved upstream.

I also merged `codex/jazz-core-engine-swap` into this branch (a merge commit, not a rebase, so your history isn't rewritten under you) to clear the conflicted state. The only conflict was in `SPEC/INVARIANTS.md`, on the six `INV-TEST-*` / `INV-TX-*` rows the base rewrote into normative MUST form; this branch's only edit to that file was a whitespace-only formatter pass on `INV-TEST-4`, so taking the base wording loses nothing.

## One follow-up, non-blocking

I noted on the original review that `codec.rs` compiles the JSON Schema on every non-null cell write, and suggested compiling once at schema admission. That was optional then; it's more load-bearing now that validation also runs on ingest. Smoke didn't regress, so this isn't urgent — but if ingest throughput ever shows up in a profile, that's the first thing to change.

Happy to split the commit out into its own PR if you'd rather keep this one focused on the TS-to-Rust move — say the word and I'll move it.

aeplay (changes requested):

Reviewed this alongside #1180. The forward-lens fix we added earlier is intact
and correct — `invalid_cell_value_reason` rebuilds the source cell map, applies
the same forward-lens condition/translation as staging, and validates against
the resulting *target* schema, on all three authority flows
(`ingest.rs:1062-1065`, `:1262-1265`, `:825-828`). A `CopyColumn`/`AddColumn`
lens moving a cell into an incompatible target column is rejected before
staging. Good.

Three things before this should land.

**1. The boundary is incomplete — two live wire shapes reach storage unvalidated.**

| external admission path | validated? |
|---|---|
| `SyncMessage::CommitUnit` → `ingest_commit_unit_with_context` (`ingest.rs:147-153`) | yes |
| direct global/edge authority entry points (`ingest.rs:507-582`) | yes |
| edge-accepted finalization (`ingest.rs:686-861`) | yes |
| **`SyncMessage::ViewUpdate` / `ViewUpdateChunk`** (`ingest.rs:163-216`) | **no** |
| **public relay admission** (`ingest.rs:864-931`) | **no** |

`apply_view_update_inner` bulk-loads reset bundles (`node/views.rs:766-779`) or
calls `ingest_view_bundle_ref` (`:786-792`, `:1083-1085`), reaching
`ingest_known_transaction` (`:1015-1023`) and the unchecked `from_parts`
staging path. Relay admission stores `Fate::Pending` via
`ingest_transaction_and_versions` (`:925-931`).

The sync-message dispatcher invokes the view-update route, so this is a live
path, not a test-helper concern. Invalid JSON — including a value a forward
lens made invalid for its target — can be stored. That is worse than a returned
write error, because nothing surfaces at all.

So it is neither "validate exactly once at the boundary" nor a redundant
interior check; it is a boundary with two holes. Worth deciding deliberately
which shapes constitute the trust boundary and covering all of them.

**2. Schema compilation per JSON cell, on the ingest hot path.**

`invalid_cell_value_reason` loops every non-deletion `VersionRecord` and all
present cells (`ingest.rs:4565-4585`), and `validate_json_value` parses the
JSON *and* parses the schema and calls `jsonschema::validator_for` **for every
cell** (`node/codec.rs:1914-1934`). No caching per schema/column or per batch.

Per-value constraint checking is inherently per-row and fine. Schema
parsing/compilation is not — it is identical for every cell in a column and is
pure repeated work. This matters more than usual here: our dominant workload is
per-row-cost-bound on ingest, so a new per-cell compile lands squarely on the
thing we are trying to reduce. Caching compiled validators per (schema, column)
should be straightforward.

**3. After rebase, the new test file will silently not run.**

`crates/jazz/Cargo.toml` sets `autotests = false` (`:5-7`) and its hand-declared
targets do not include `json_storage` (`:191-292`). This PR adds
`tests/json_storage.rs` with no `[[test]]` entry. On the current base that file
compiles and runs never, and nothing goes red — we have lost eight test files
to exactly this. Please add `[[test]] name = "json_storage"`.

Coverage gaps worth closing at the same time: the tests cover local writes and
same-schema remote `CommitUnit`s only (`tests/json_storage.rs:193-345`). No
forward-lens case, no `ViewUpdate`/initial-hydration case, and nothing asserting
the structured rejection reason. A lens turning a source-valid cell into a
target-invalid one is the case the fix exists for and is untested.

**Credit where due on the failure mode:** for authority `CommitUnit`s, invalid
input becomes `Fate::Rejected(RejectionReason::MalformedCommit(reason))` with
the reason retained as rejection detail (`codec.rs:1249-1253`). That is exactly
the information channel we want.

**Rebase note:** this predates the consolidation, so it edits the old
`jazz-tools`/`jazz-server` manifests as well as `crates/jazz/Cargo.toml`. Port
the semantic decisions into the consolidated manifest rather than merging both —
duplicate `[[test]]`/`[[bench]]` declarations make cargo reject the whole
manifest and take the crate dark, which has already happened once this week.

aeplay:
🤖 Closing as superseded by the new query/runtime architecture.

This stack moved JSON validation through the old query-manager and native schema/descriptor codec paths. Those paths were removed by the engine swap (notably [`0caac9cc9`](https://github.com/garden-co/jazz/commit/0caac9cc9)), and the current branch has newer JSON-field integration coverage in [`73a088789`](https://github.com/garden-co/jazz/commit/73a088789).

There is therefore no safe direct restack: it would reintroduce deleted architecture rather than preserve a unique current fix. Any remaining JSON-validation gap should be demonstrated against the current engine and opened as a new focused PR.

Preserved historical tip: [`903f2520b`](https://github.com/garden-co/jazz/commit/903f2520b).

PR #1169, opened 2026-07-28
chore: run CI in PRs made against `codex/jazz-core-engine-swap` branch

nrainhart:


aeplay:
🤖 Correct and overdue — approving in spirit. This explains why none of the ~20 PRs currently open against `codex/jazz-core-engine-swap` have had any CI signal at all; the only checks running have been Vercel deploy previews.

Two things worth knowing before it lands, neither an objection:

**1. Expect immediate red across the open PRs**, from problems that predate this change:

- repo-wide `pnpm test` fails in the `auth-simple-chat` and `world-tour` example packages — they report no test files and then fail in `packages/jazz-tools/src/dev/catalogue.ts` with an undefined path argument during browser-test setup;
- `cargo clippy --workspace -- -D warnings` is blocked by pre-existing dead-code and `filter_map` warnings in `crates/jazz/src/node/currency.rs`;
- the browser OPFS suite currently fails before it can run W4.

Lanes are open on the first two. Flagging so a wall of red on merge reads as "CI finally switched on" rather than "this PR broke everything."

**2. `push:` is still `[main]` only.** So PRs *into* `codex/jazz-core-engine-swap` will run CI, but merges *onto* it will not. That is probably what you want while the branch is a staging line, but it means the integrated state is never itself verified — worth a deliberate decision rather than leaving implicit, especially given how much is landing on that branch daily.

aeplay (approved):

PR #1171, opened 2026-07-28
Partition maintained windows by route

tobiaslins:
## What

- Partition every lowered maintained window by the hidden route tuple that identifies a prepared-query binding.
- Add a public-API integration test with two teams sharing one parameterized `ORDER BY … LIMIT 2` shape.
- Assert exact initial snapshots, ordering, independent eviction deltas, and zero unrelated notifications.

## Root cause

A single maintained graph serves all active bindings. `TopBy` was grouped only by user-visible grouping columns, while the per-binding route filter runs downstream. This made the top-N global across tenant bindings: one team could consume another team's slots.

## Impact

Tenant/claim-bound `ORDER BY … LIMIT` subscriptions now maintain an independent window per route tuple. This fixes missing rows and cross-binding eviction behavior for the linear prepared-query path.

## Validation

- Red/green regression: removing the route partition reproduces the two-team failure; restoring it passes.
- `cargo test -p jazz`: 787 passed, 1 ignored.
- `cargo test -p groove`: 431 passed, 1 ignored.
- `cargo test -p jazz-server`: 55 passed.
- `cargo test -p jazz-tools --features test -- --skip numeric_claims_match_integer_columns_across_core_widths`: all remaining tests pass.
- `cargo check -p jazz-sim --benches`.
- `dev/gates/ts-wire-codec.sh`: 86 passed, 1 skipped.
- `JAZZ_SEED_COUNT=300 cargo test -p jazz m3_maintained_one_shot_differential_oracle`.
- Incremental-delivery canaries and `dev/benchmarks/smoke.sh`.
- Full pre-commit `clippy` and `rustfmt`.

## Known base failure

`numeric_claims_match_integer_columns_across_core_widths` times out at the I64-claim/I32-column case both with this patch and with only the route-partition change removed. It does not use a window/`TopBy`, so this PR intentionally does not alter that test or behavior.

The optional `jazz-private` sensitive-data guard was unavailable in this checkout; the public fixture is anonymized and contains no customer data.


aeplay:
🤖 Closing as already merged through the current new-core stack.

The complete effective patch from [`12dddfd78`](https://github.com/garden-co/jazz/commit/12dddfd78b2cb516f4602fabfe1a1e26c44cfca4)—route fields in each maintained `TopBy` partition plus the two-binding regression test—exists on `codex/jazz-core-engine-swap` as [`118ff0dbb`](https://github.com/garden-co/jazz/commit/118ff0dbbc176bb17086f8fb3c6df02a7fbf9aa1).

The two commits have the identical stable patch ID `e89929cc7b82ade6653b979b94c5b44c57c74ce6`, and `118ff0dbb` is an ancestor of the current base. There is no remaining unique diff to restack.

PR #1170, opened 2026-07-28
Benchmark and plan SaaS documents, permissions, and subscription fan-out

tobiaslins:
## Current status (updated 2026-08-05)

This draft is now the live record for the policy, fan-out, and subscription-lifecycle investigation. The original receipts below remain useful historical evidence, but several measurements describe defects or execution paths that have since changed. Do not treat the original timing tables as current-engine performance claims.

### Completed follow-up work

- [x] **Per-route maintained window partitioning:** #1179 keeps shared parameterized and claim-bound `TopBy` windows independent across active bindings.
- [x] **One flush per Jazz refresh cycle:** #1190 removes the per-subscription repeated Groove flush identified by the route-scale write benchmark.
- [x] **Fail-closed claim revision handling:** #1197 rebuilds affected streams when identity claims change.
- [x] **Prepared-claim parameter model:** #1216 specifies claim paths as runtime prepared-query parameters. This completes the specification decision; implementation coverage remains below.
- [x] **Selective Global hydration receipt:** #1227 extracts a focused retained benchmark from the persisted-hydration finding.
- [x] **Indexed selective one-shot hydration:** #1228 keeps eligible indexed one-shot reads selective and adds correctness coverage.
- [x] **Existential authorization before finite windows:** #1238 collapses overlapping policy/grant proofs per protected row and route before `TopBy`, restoring the exact 100-row page while preserving ordinary application join bag semantics.
- [x] **Maintained authorization scope re-entry:** #1239 preserves omitted cells from the storage-visible row during partial updates, restoring exact remove/refill and add/evict deltas when documents move out of and back into an authorized team.

### Still pending

- [x] **Current-core policy cost receipt:** #1240 retains a deterministic five-tier benchmark covering membership, organization-admin, direct-ACL, public/published, and trusted-admin authorization, with exact ordered one-shot and maintained-lifecycle oracles.
- [ ] **Prepared claim routing implementation (ready for review):** #1261 routes claims per binding, collects claims across applicable and transitively reached policies, and covers multiple identities and application bindings against shared prepared shapes.
- [x] **Current-core route/subscription curve:** #1264 retains the post-#1190 public-API scale curve and attributes the remaining attach/memory slope to binding-private full-source witness indexes.
- [x] **Policy churn and reconnect:** #1266 fixes stale fast-cursor membership and payload suppression after disconnected authorization changes; #1267 retains a six-case membership and claim reconnect matrix.
[...]

PR #1172, opened 2026-07-29
(new core) Fix maintained TopBy binding partitions

aeplay:
Reported by an adopter testing the new core on #1170:

> many active team subscriptions are currently not correct: bindings share a global `TopBy(100)` before team routing. The two-team canary returned 0 instead of 100 documents for one team.

Maintained bindings shared one global `TopBy` applied *before* team routing, so the limit was taken across all teams and then filtered per team — leaving one team with none of the 100.

The window must be evaluated per binding partition, so each team gets its own top-N rather than a share of one global top-N.

## Coverage

Two maintained-subscription regressions in `crates/jazz/src/node/tests/policies_rls.rs`:

- two teams x 100 documents: each receives 100
- two owner-policy claim bindings x 100 documents: each receives 100

Both are multi-row-per-group by construction, so they exercise the partitioning rather than degenerating into per-row projection.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-sim --benches` all exit 0.

`cargo test -p jazz-tools --features test` fails **only** `numeric_claims_match_integer_columns_across_core_widths`, which is red on `codex/jazz-core-engine-swap` itself and is fixed by #1159 — not caused by this branch. Verified by running it on the unmodified base.

### DIFF
--- a/crates/groove/Cargo.toml
+++ b/crates/groove/Cargo.toml
@@ -36,3 +36,7 @@ harness = false
 [[bench]]
 name = "micro"
 harness = false
+
+[[bench]]
+name = "saas_documents"
+harness = false
--- a/crates/groove/benches/saas_documents.rs
+++ b/crates/groove/benches/saas_documents.rs
@@ -1,0 +1,1092 @@
+//! SaaS-style team document query benchmark.
+//!
+//! The default fixture contains exactly 500k documents and 5k teams, including
+//! one 30k-document hot team and one 100-document small team. Those exact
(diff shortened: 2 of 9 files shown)


aeplay:
🤖 Superseded by #1179, which landed @tobiaslins' implementation from #1171 with his authorship preserved, plus the claim-derived binding regression this PR added. Nothing here is lost — closing to keep the queue clean rather than carrying a conflicting duplicate.

PR #1173, opened 2026-07-29
(new core) Lower canonical relation gathers instead of rejecting them

aeplay:
Fixes 8 of the 22 browser-suite failures — the group tracked as R1, spanning `db.all`, `db.subscribeAll`, `useAll` and `useAllSuspense`.

## Cause

`collect_relation_facade` (`crates/jazz/src/query.rs`) rejected `Union` / `Gather` / `Distinct` outright, so any multi-hop or gather query failed at the facade even though the engine could serve it. The same rejection showed up independently in the CI sweep at `query.rs:518`.

## Approach

The canonical public `gather` shape is normalized into the ordinary recursive reachability query. `ReachableVia` already has both maintained and one-shot lowering, so relation gathers do **not** need a second evaluator or a separate subscription implementation.

This is deliberately the lower-into-groove direction rather than a facade-side bypass, per `INV-LOWER-12` as settled by @aeplay: query shapes whose storage read crosses partitioned or schema-projected data must lower every representable source projection into groove; schema projection is a source-lowering responsibility. Unsupported shapes still fail loudly.

## Coverage

186 lines of new tests in `crates/jazz/src/db/tests.rs`, covering the gather normalization and its interaction with ordering, offset and limit.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-sim --benches` all exit 0.

`cargo test -p jazz-tools --features test` fails **only** `numeric_claims_match_integer_columns_across_core_widths`, which is red on `codex/jazz-core-engine-swap` itself and is fixed by #1159 — not caused by this branch.


packages/jazz-tools/src/runtime/native-runtime/websocket.ts
```diff
@@ -126,12 +126,12 @@ export class WebSocketCarrier {
         message: "websocket transport error",
       });
     });
-    this.socket.addEventListener("close", () => {
+    this.socket.addEventListener("close", (event) => {
```
joeinnes, line 129: The type definition for this doesn't allow passing an event, we should update it (line 34)

packages/jazz-tools/src/runtime/native-runtime/persistent-browser-runtime.ts
```diff
@@ -546,10 +555,9 @@ export class PersistentBrowserOpfsRuntime implements Runtime {
     if ("subscription" in message) {
       const callback = this.subscriptions.get(message.subscription);
       if ("error" in message) {
-        callback?.(
-          new Error(message.error.message ?? "Persistent browser subscription failed"),
-          null,
-        );
+        const error = new Error(message.error.message ?? "Persistent browser subscription failed");
+        if (message.error.stack) error.stack = message.error.stack;
```
joeinnes, line 559: At this point, `message` is a `PersistentBrowserSubscriptionMessage`, so we should add `stack?: string` to that type (in persistent-browser-protocol.ts)

crates/jazz/src/query.rs
```diff
@@ -296,6 +300,326 @@ pub(crate) fn relation_query_to_query(query: &RelationQuery) -> Result<Query, Qu
...
+/// Normalize the canonical public `gather` shape into the ordinary recursive
+/// reachability query.  The TypeScript query adapter emits this shape for a
+/// same-table forward hop: a seed relation supplies the initial rows and the
+/// step table relates each frontier row to its parent through a scalar FK.
+///
+/// Keeping this conversion here is important: `ReachableVia` already has
+/// maintained and one-shot lowering, so relation gathers do not need a second
+/// evaluator or subscription implementation.
+fn relation_gather_to_query(expr: &RelationExpr) -> Result<Option<Query>, QueryError> {
+    let (expr, order_by, offset, limit) = peel_relation_output_steps(expr)?;
```
joeinnes, line 312: Should we also peel `Filter` here?

joeinnes (approved):

PR #1174, opened 2026-07-29
(new core) bench: add attributed R3 persisted-read receipts

joeinnes:
## Summary

- preserve the historical R3 Criterion measurement while adding an attributed JSON phase receipt
- split persisted reopen/read time into RocksDB storage open, Jazz `Db` open, query preparation, and first materialization
- add opt-in `ci`, `s`, and `m` workload profiles plus warm and Linux best-effort evicted-cache modes
- validate the exact project-board result count for every sample
- document reproduction commands and the initial local receipt

## Why

The combined R3 timing showed that persisted reads scale poorly, but did not identify whether the cost came from RocksDB startup, Jazz initialization, query construction, or materialization. These receipts make that boundary visible while retaining the existing Criterion path for historical compatibility.

The initial local ladder indicates that Jazz open is the dominant scale signal: roughly 3.1–3.2 seconds for the 100,000-task profile, versus 13–35 ms for RocksDB open. First materialization also reaches roughly 340–355 ms while returning 200 rows. These numbers are local characterization, not acceptance thresholds.

## Behavior and methodology

Each sample opens a fresh RocksDB/Jazz instance in the same benchmark process, recreating RocksDB's per-instance cache. It is therefore an instance-cold reopen, not a fresh-process measurement. `warm` leaves OS page-cache state uncontrolled after seeding. On Linux, `evicted` applies `posix_fadvise(POSIX_FADV_DONTNEED)` only to the benchmark's RocksDB files, outside the measured window.

This first instrumentation revision also varied the benchmark node identity between samples, which can add fixture mutation and makes the initial values conservative. #1175 keeps the identity stable. The measurements remain useful as the original worst-case baseline, but later stack comparisons should use the corrected harness.

The default remains the existing small CI profile. Larger profiles, cache eviction, and phase-only execution are opt-in through environment variables.

## Validation

- benchmark compiles with the `rocksdb` feature
- full local `ci,s,m` phase ladder completed in warm and evicted modes
- exact result-row assertions pass for each profile
- existing Criterion benchmark path remains active
- `git diff --check` passes


joeinnes:
CI update after `6d0a1c7b7`:

- [lint](https://github.com/garden-co/jazz/actions/runs/30449342037/job/90567369805) no longer reports the PR-owned `dev/benchmarks/R3_PERSISTED_READ_RECEIPT.md`; the remaining nine formatting failures are outside this PR’s five changed files.
- [test-rust](https://github.com/garden-co/jazz/actions/runs/30449342037/job/90567369953) still fails with `SubscriptionEvent::Rejected { .. }` missing from `crates/jazz-sim/tests/customer_child_policy_minimal.rs:205`. That test is unchanged between the base and this head.
- [test-ts](https://github.com/garden-co/jazz/actions/runs/30449342037/job/90567369877) still times out in the unchanged todo-app server-sync test. This benchmark PR does not touch that path.

The PR-owned formatting failure is fixed. The remaining red checks are base/parent-branch blockers associated with [PR #1094](https://github.com/garden-co/jazz/pull/1094), not changes introduced by this PR. Vercel’s checks are passing.

joeinnes:

This PR establishes the real-world baseline:

| restart | cache | median |
|---|---:|---:|
| clean | warm | 262.2 ms |
| unclean | warm | 273.5 ms |
| clean | evicted | 431.9 ms |
| unclean | evicted | 443.3 ms |

As intended, the value here is reproducible instrumentation and an attributable baseline, not an optimization claim.

aeplay (approved):

PR #1175, opened 2026-07-29
(new core) bench: attribute R3 Jazz open recovery

joeinnes:
🤖

## What changed

- adds a testing-only open receipt for the current physical-lineage recovery pipeline
- attributes catalogue reads, full database reads, state recovery, recovery scans, ahead-current rebuild, and final catalogue alias persistence
- adds an opt-in `r3-open-attribution` benchmark feature
- counts records at the actual physical scan loops rather than the retired logical-table recovery path

## Why

We need trustworthy attribution before optimizing startup recovery. The original PR instrumented an older catalogue-table scan; this restack preserves current recovery semantics and moves the measurements to the physical-schema lineage implementation.

`finalize_catalogue` now measures the real post-recovery alias-resolution and persistence boundary instead of reporting an unused zero-valued field.

## Validation

- independently reviewed after conflict resolution: no P0/P1/P2 findings
- exact receipt test counts physical recovery scans
- planted first-open test proves catalogue finalization timing is nonzero when aliases are absent
- recovery tests: 6 passed
- reopen tests: 8 passed
- pending replay tests: 2 passed
- benchmark builds with `r3-open-attribution`
- formatting, Clippy, and clean-tree checks passed

The historical receipt values in the former PR description are not carried forward as current measurements; the recovery architecture changed and should be remeasured from this implementation.

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -26,6 +26,7 @@ sync-autopsy = []
 # Disabled-by-default counters for the cold-settle attribution bench. They do
 # not alter protocol behavior or production measurement paths.
 cold-settle-attribution = ["groove/cold-settle-attribution"]
+r3-open-attribution = ["rocksdb", "testing"]
 testing = []
 transport-compression-lz4 = ["dep:lz4_flex"]
 transport-compression-zstd = ["dep:zstd"]
--- a/crates/jazz/benches/realistic_phase1.rs
+++ b/crates/jazz/benches/realistic_phase1.rs
@@ -3,6 +3,7 @@
 //! This intentionally exercises `jazz::db::Db<MemoryStorage>` directly, without
 //! the legacy `RuntimeCore`, `SchemaManager`, or `SyncManager` stack.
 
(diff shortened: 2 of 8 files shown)


joeinnes:

Real-world medians at this point:

| restart | cache | #1174 | #1175 |
|---|---:|---:|---:|
| clean | warm | 262.2 ms | 266.3 ms |
| unclean | warm | 273.5 ms | 278.8 ms |
| clean | evicted | 431.9 ms | 430.2 ms |
| unclean | evicted | 443.3 ms | 432.6 ms |

The mixed small deltas are run-to-run variance. This PR remains attribution-only: it identifies the startup scans but has no attributable real-world performance effect by itself.

aeplay:

Clean as far as I can tell. Feature-gated measurement only; the ordinary open
path is unchanged when the feature is off.

Checked: that the instrumentation cannot run in a default build, that no probe
executes inside a region it also reports on, and that the R3 assertions still
verify result counts rather than only timings.

No changes requested. Part of the R3 stack review — the findings are on #1176
and #1203.

PR #1176, opened 2026-07-29
(new core) perf: scan only sequenced transactions during recovery

joeinnes:
🤖

## What changed

- recovery scans only the non-null portion of `jazz_transactions/by_global_seq`
- retains an exact lookup for `u64::MAX`, because Groove range ends are exclusive
- excludes local pending and rejected transactions from sequenced-transaction decoding
- keeps pending replay and ahead-current reconstruction on their existing paths

## Why

The nullable global-sequence index contains both sequenced and unsequenced transactions. Startup recovery only needs sequenced entries for the global recovery pass; scanning the nullable `None` bucket makes work scale with unrelated local pending/rejected history.

## Validation

- independently reviewed after conflict resolution: no P0/P1/P2 findings
- planted former full scan failed the exact receipt test (`2` scanned instead of `0`)
- planted tightened lower bound missed an accepted sequence and failed (`0` instead of `1`)
- recovery tests: 6 passed
- reopen tests: 8 passed
- pending replay tests: 2 passed
- exact reopen replay passed
- benchmark compilation, formatting, Clippy, and clean-tree checks passed

This PR is stacked on #1175 and should be reviewed/merged second.

### DIFF
--- a/crates/jazz/src/node/ingest.rs
+++ b/crates/jazz/src/node/ingest.rs
@@ -624,8 +624,7 @@ where
                 return Ok(());
             }
         }
-        let global_seq = self.clock.next_global_seq;
-        self.clock.next_global_seq = self.clock.next_global_seq.next();
+        let global_seq = self.clock.allocate_global_seq()?;
         self.apply_fate_update(
             tx_id,
             Fate::Accepted,
@@ -670,8 +669,7 @@ where
             self.ingest_rejected_transaction(tx, fate.clone())?;
             return Ok(fate);
         }
-        let global_seq = self.clock.next_global_seq;
-        self.clock.next_global_seq = self.clock.next_global_seq.next();
+        let global_seq = self.clock.allocate_global_seq()?;
         self.apply_fate_update(
             tx_id,
             Fate::Accepted,
@@ -835,8 +833,7 @@ where
             return Ok(updates);
         }
 
-        let global_seq = self.clock.next_global_seq;
-        self.clock.next_global_seq = self.clock.next_global_seq.next();
+        let global_seq = self.clock.allocate_global_seq()?;
         let fate = Fate::Accepted;
         let durability = DurabilityTier::Global;
         let merge_rows = self.merge_rows_for_versions(&versions)?;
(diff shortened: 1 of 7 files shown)


joeinnes:

Incremental effect versus #1175:

|---|---:|---:|---:|---:|
| clean | warm | 266.3 ms | 243.7 ms | -22.5 ms (-8.5%) |
| unclean | warm | 278.8 ms | 248.7 ms | -30.1 ms (-10.8%) |
| clean | evicted | 430.2 ms | 403.6 ms | -26.5 ms (-6.2%) |
| unclean | evicted | 432.6 ms | 407.6 ms | -25.0 ms (-5.8%) |

Cumulative versus #1174: **7.0% clean/warm, 9.1% unclean/warm, 6.5% clean/evicted, and 8.0% unclean/evicted**.

The real-world benefit is smaller than the worst-case fixture—as expected for globally sequenced history—but remains clear and repeatable.

aeplay (changes requested):

Reviewed the R3 stack over the weekend. #1174, #1175, #1178 and #1199 all came
out clean — #1199's not-rebuilt indexes are durably replaced and have reopen,
fallback, deletion and oracle coverage. This one has a finding, and it gates
the other four.

**`GlobalSeq::next` changed from addition to saturation** (`crates/jazz/src/time.rs:26-30`),
and that is not recovery-only: the authority paths allocate through it
(`ingest.rs:604-611`, `:651-658`, `:816-827`, `:1079-1090`, `:2651-2658`), and
recovered state goes through `record_applied_global_seq` (`open_tx.rs:289-301`).

Breaking sequence: recover an accepted transaction at `u64::MAX`, then accept
one more. Saturation leaves `next_global_seq` at `MAX` and assigns `MAX` again —
a duplicated global serialization position.

Being fair about severity: `u64::MAX` is not reachable in practice, so this is
not a live data-loss risk. What has regressed is the failure mode — debug builds
previously failed at exhaustion, and now it silently reuses. An explicit
exhaustion error (or an unrepresentable-next state) restores loud failure.

The new MAX test stops after reopen, so it cannot fail on this. A post-MAX
allocation test would.

The rest of the endpoint handling is sound: the range-scan plus separate
`Some(MAX)` lookup for the exclusive end is correct, fate filtering is
preserved, and `recovery_scans_only_sequenced_transactions_and_preserves_global_gaps`
is real coverage.

PR #1177, opened 2026-07-29
(new core) fix: coerce numeric claims for prepared queries

joeinnes:
🤖 Restacked onto the merged client-local/trusted-serving authorization boundary.

## What changed

- Coerce prepared integer bindings across `U8`, `U16`, `U32`, `U64`, `I32`, and `I64` when the value is representable in the query column's type.
- Preserve the original typed value when conversion would overflow, underflow, or cross an unsupported type, so authorization comparisons fail closed.
- Normalize nullable prepared values exactly once across nullable/non-nullable source and target combinations.

## Why

Prepared claims arrive through JSON-shaped integer representations that do not always match the physical integer width of the policy column. A numerically equal claim could therefore fail a trusted prepared-policy comparison solely because its integer variant differed.

This PR changes only prepared binding normalization. It does not alter authority selection: trusted serving still evaluates policy, while ordinary client-local queries continue to elide permission lowering.

## Validation

- Independent adversarial review: P0 none, P1 none, P2 none.
- Positive matrix across all supported signed/unsigned widths.
- Overflow, underflow, negative-to-unsigned, non-numeric, and nullable `Some`/`None` cases fail closed or preserve shape as intended.
- Reversible unsafe-narrowing, incorrect-I64, and nullable double-wrap mutations fail the new tests.
- `prepared_claim_routing`: 9 passed, 1 known ignored.
- `authorization_scope_reentry`: 4/4.
- Focused prepared integer/nullable tests pass.
- Formatting and strict Jazz library clippy pass.

The old lowering portion of the original PR was obsolete after terminal/auth assembly and is intentionally not replayed.

### DIFF
--- a/crates/jazz/src/node/query_eval.rs
+++ b/crates/jazz/src/node/query_eval.rs
@@ -12427,6 +12427,9 @@ fn prepared_claim_value(path: &ClaimPath, policy: &PolicyContext) -> Result<Valu
 }
 
 fn coerce_prepared_binding_value(value: Value, column_type: &groove::schema::ColumnType) -> Value {
+    if let Some(value) = coerce_prepared_integer_value(&value, column_type) {
+        return value;
+    }
     match (value, column_type) {
         (Value::Uuid(value), groove::schema::ColumnType::String) => {
             Value::String(value.to_string())
@@ -12434,17 +12437,6 @@ fn coerce_prepared_binding_value(value: Value, column_type: &groove::schema::Col
         (Value::String(value), groove::schema::ColumnType::Uuid) => uuid::Uuid::parse_str(&value)
             .map(Value::Uuid)
             .unwrap_or(Value::String(value)),
-        (Value::I64(value), groove::schema::ColumnType::I32) => i32::try_from(value)
-            .map(Value::I32)
-            .unwrap_or(Value::I64(value)),
-        (Value::U32(value), groove::schema::ColumnType::I32) => i32::try_from(value)
-            .map(Value::I32)
-            .unwrap_or(Value::U32(value)),
-        (Value::I32(value), groove::schema::ColumnType::I64) => Value::I64(i64::from(value)),
-        (Value::U32(value), groove::schema::ColumnType::I64) => Value::I64(i64::from(value)),
-        (Value::U64(value), groove::schema::ColumnType::I64) => i64::try_from(value)
-            .map(Value::I64)
-            .unwrap_or(Value::U64(value)),
         (Value::Nullable(Some(value)), column_type) => Value::Nullable(Some(Box::new(
             coerce_prepared_binding_value(*value, column_type),
         ))),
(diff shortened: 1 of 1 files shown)


aeplay (changes requested):

This needs a specification before an implementation, rather than changes to the
patch — so please read this as "let's decide the rule" rather than "this code is
wrong".

**The rule as written** (`query_eval.rs:9695-9756`, mirrored at
`query_engine/lowering.rs:4377-4436`): represent the source value as `i128`; if
the *target declared column type* is one of the six integer types and
`TryFrom<i128>` succeeds, replace the value with that target's variant; if it
does not fit, leave the original value unchanged. Floats and non-integer targets
are untouched, so integer/float is not a coercion relation.

**Finding 1 — the same claim gets different semantics on different paths.** The
established policy comparator compares all six integer variants as exact `i128`
regardless of declared width (`crates/groove/src/ivm/runtime/mod.rs:7423-7458`),
and the policy join-key encoder canonicalises every integer to a common `i128`
key (`crates/groove/src/ivm/runtime/join.rs:780-825`). This PR instead requires
the claim to fit a particular column type. So `U8(0) > I64(-1)` is meaningful
and true under the policy comparator, while binding `I64(-1)` to a `U8` prepared
parameter is rejected. Same values, two answers, depending on which path
evaluates them.

**Finding 2 — it is not scoped to claims.** Adding the integer branch to
`coerce_literal_for_value_type` also changes ordinary field/literal comparison
lowering through `coerce_literal_for_source_field` and `lower_compare`
(`lowering.rs:4291-4301`, `:4357-4374`), not just the authorization-route
literal call at `:5771-5802`.

**Finding 3 — the failure branch does not fail closed.** On a failed conversion
the helper returns the original typed variant (`query_eval.rs:9735-9756`), which
`binding_values_for_plan` then sends to a prepared shape whose descriptor was
declared with `param.ty` (`:9637-9658`, descriptor construction at `:7912-7918`).
The value and its descriptor disagree.

**What the spec should settle**, once, for every comparison operator (`==`, `!=`,
`<`, `<=`, `>`, `>=`, `IN`, containment) across policy/RLS, one-shot lowering,
maintained prepared bindings, and policy joins:

- whether all integer widths and signedness compare by exact mathematical value,
  or comparisons are target-type coercions;
- the observable outcome for out-of-range, negative-to-unsigned, and
  differing-width signed/unsigned values — match, no-match, or structured
  rejection;
- whether any integer/float relation exists, and if so its exact boundary rules;
- the role boundary: filter claims are execution parameters and must not change
  ordinary arrangement encodings, while partition claims need an explicit
  canonical order-preserving encoding plus range/window rules (see
  `crates/jazz/SPEC/6_queries.md`, which #1216 extended with exactly this
  distinction);
- conformance tests across all four paths including boundary and rejection cases.

Then one shared comparison/coercion primitive, adopted by every path.

Happy to draft the spec section if that is useful — it overlaps with the
claim-parameter model that just landed.

PR #1178, opened 2026-07-29
(new core) perf: fuse ahead-current recovery passes

joeinnes:
## What changed

- Rebuild ahead-current in-memory indexes during the existing persisted-layout validation scan.
- Remove the second full scan of ahead-current RocksDB tables.
- Sort and deduplicate recovered keys once, then bulk-construct the full-key, touched-row, and latest-key B-tree indexes.
- Decode content and deletion records through their respective layouts.
- Add a reopen regression proving latest-version selection and fallback after the latest version is rejected.
- Remove the obsolete standalone rebuild timing field and document the local R3 receipt.

## Why

After #1176, startup still decoded 952,620 ahead-current records during layout validation, discarded their keys, then scanned and decoded the same records again to rebuild three in-memory indexes one entry at a time.

This change reuses the first pass and avoids repeated ordered-tree mutations while preserving the existing ordering: reconstructed indexes are complete before unclean-close cleanup runs.

## Local R3 result

Three-sample warm-cache, clean-close medians on the `m` workload:

- Jazz open: 1,104 ms → 832 ms (-24.6%)
- Current-state validation and reconstruction: 1,103 ms → 831 ms (-24.7%)
- Standalone ahead-current rebuild: 763 ms → eliminated
- Total reopen plus first read: 1,459 ms → 1,149 ms (-21.2%)

The optimization still performs work proportional to ahead-current entries and temporarily retains decoded keys during bulk construction.

## Validation

- Full Jazz suite: 730 passed, 1 ignored
- Focused recovery subset in ordinary and attributed builds
- 300-seed M3 differential oracle
- TypeScript wire codec gate: 86 passed, 1 skipped
- R3 attributed benchmark compilation
- Jazz simulation bench compilation
- Full local smoke benchmark suite
- Repository pre-commit Clippy, rustfmt, oxfmt, and sensitive-data gates
- `git diff --check`

### DIFF
--- a/crates/jazz/src/node/mod.rs
+++ b/crates/jazz/src/node/mod.rs
@@ -1672,11 +1672,7 @@ where
         &mut self,
         #[cfg(feature = "testing")] mut receipt: Option<&mut NodeOpenReceipt>,
     ) -> Result<(), Error> {
-        self.ahead_current_keys.clear();
-        self.ahead_current_rows.clear();
-        self.ahead_current_latest.clear();
-        #[cfg(feature = "testing")]
-        let mut entries = 0usize;
+        let mut keys = Vec::new();
         for table in self.catalogue.schema.tables.clone() {
             let storage_tables = table.ahead_current_storage_tables();
             let content_rows = self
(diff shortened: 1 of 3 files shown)


joeinnes:

Incremental effect versus #1176:

|---|---:|---:|---:|---:|
| clean | warm | 243.7 ms | 224.3 ms | -19.5 ms (-8.0%) |
| unclean | warm | 248.7 ms | 237.0 ms | -11.6 ms (-4.7%) |
| clean | evicted | 403.6 ms | 387.4 ms | -16.3 ms (-4.0%) |
| unclean | evicted | 407.6 ms | 395.7 ms | -11.9 ms (-2.9%) |

Cumulative versus #1174: **14.5% clean/warm, 13.3% unclean/warm, 10.3% clean/evicted, and 10.7% unclean/evicted**.

This confirms that fusing the ahead-current recovery pass remains a useful real-world improvement even when transaction history is globally settled.

aeplay:

Clean. It fuses the reconstruction passes without narrowing what is scanned —
same scan, fewer traversals — and the latest/fallback test covers the fused
path.

Checked: that no previously-visited state is skipped by the fusion, that the
fallback still triggers on the same conditions, and that recovery still sees
every transaction it saw before.

No changes requested, but note it is gated behind #1176's endpoint fix, which
is in that PR's stack below this one.

aeplay:
🤖 Closing as superseded by the current recovery architecture, including https://github.com/garden-co/jazz/commit/73434017. The redundant startup validation pass this PR fused no longer exists, so its restack is an empty change rather than a meaningful optimization.

PR #1179, opened 2026-07-29
(new core) Partition maintained windows by route

aeplay:
Fixes the adopter-reported bug from #1170: maintained bindings shared one global `TopBy(100)` applied *before* team routing, so the two-team canary returned 0 documents for one team instead of 100.

## Provenance

The implementation is @tobiaslins' commit from #1171, cherry-picked unchanged with authorship preserved. Opening it here as a separate small PR so #1170 can stay purely investigative and this can land on its own; #1171 is untouched and will merge back trivially, since the content is identical.

## What this adds on top

One regression covering the case the original test does not: partitioning driven by a **policy claim** rather than an explicit route parameter. Two owners under `Policy::owner_only`, 100 documents each, subscribing with the *same* shape and binding — without per-binding partitioning the window is taken across both owners and one of them sees fewer than its 100 rows.

That case matters because claim-derived routing is where the related prepared-query bug lives (finding 2 in #1170, being fixed separately in `track/claim-route-parameterization`). Verified this test passes against the cherry-picked fix, so it is a guard rather than a gap.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-sim --benches` — all exit 0.

### DIFF
--- a/crates/jazz/tests/parameterized_subscription_routing.rs
+++ b/crates/jazz/tests/parameterized_subscription_routing.rs
@@ -1,0 +1,303 @@
+use std::collections::{BTreeMap, BTreeSet};
+
+use jazz::block_on;
+use jazz::db::{
+    Db, DbConfig, DbIdentity, LocalUpdates, PreparedQuery, Propagation, ReadOpts,
+    SeededRowIdSource, SubscriptionEvent, SubscriptionStream,
+};
+use jazz::groove::records::Value;
+use jazz::groove::schema::{ColumnSchema, ColumnType};
+use jazz::groove::storage::MemoryStorage;
+use jazz::ids::{AuthorId, NodeUuid, RowUuid};
+use jazz::query::{OrderDirection, Query, col, eq, param};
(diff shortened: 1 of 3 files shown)


crates/jazz/tests/parameterized_subscription_routing.rs
```diff
@@ -0,0 +1,303 @@
...
+    ))
+    .expect("open parameterized routing db")
+}
+
+fn row(seed: u64) -> RowUuid {
+    let mut bytes = [0_u8; 16];
+    bytes[..8].copy_from_slice(&0x019e_0000_0000_7000_u64.to_be_bytes());
+    bytes[8..].copy_from_slice(&seed.to_be_bytes());
+    RowUuid::from_bytes(bytes)
+}
```
nrainhart, line 54: Should we add a RowUuid method to allow creating a uuid from a seed?

nrainhart (approved):

PR #1180, opened 2026-07-29
(new core) chore: improve Rust build time and artifact size

nrainhart:
## Description

Adds a few optimization to reduce Rust build times and the size of generated artifacts:
- dev/test builds now use `line-table-only` debug information
- minimize rust dependency features to speed up compilation

Current branch is substantially better for development builds; release builds are roughly unchanged:

Metric | codex/jazz-core-engine-swap | Current branch | Change
-- | -- | -- | --
Cold debug build | 107.99s | 73.16s | −32.3%
Cold release build | 130.86s | 128.46s | −1.8%
Debug .node | 114.18 MB | 99.18 MB | −13.1%
Debug .node, gzip | 25.18 MB | 22.06 MB | −12.4%
Release .node | 38.91 MB | 38.67 MB | −238,528 bytes / −0.6%
Release .node, gzip | 14.13 MB | 14.06 MB | −69,794 bytes / −0.5%
Debug target footprint | 5.49 GiB | 2.03 GiB | −63.0%
Release target footprint | 0.956 GiB | 0.943 GiB | −1.4%

This PR also modifies the pre-commit hook to only run Clippy on modified crates (instead of in all of them). CI still runs Clippy for all crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,7 +207,7 @@ dependencies = [
  "serde",
  "sha2 0.11.0",
  "smol_str",
- "thiserror 2.0.18",
+ "thiserror",
  "tinyvec",
  "tracing",
  "unicode-segmentation",
@@ -215,14 +215,14 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.7.9"
+version = "0.8.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
+checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
- "async-trait",
  "axum-core",
  "base64 0.22.1",
  "bytes",
+ "form_urlencoded",
  "futures-util",
  "http",
  "http-body",
@@ -235,15 +235,14 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rustversion",
- "serde",
+ "serde_core",
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
  "sha1",
  "sync_wrapper",
  "tokio",
- "tokio-tungstenite 0.24.0",
+ "tokio-tungstenite",
  "tower",
  "tower-layer",
  "tower-service",
@@ -252,19 +251,17 @@ dependencies = [
 
 [[package]]
 name = "axum-core"
-version = "0.4.5"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09f2bd6146b97ae3359fa0cc6d6b376d9539582c7b4220f041a33ec24c226199"
+checksum = "08c78f31d7b1291f7ee735c1c6780ccde7785daae9a9206026862dab7d8792d1"
 dependencies = [
- "async-trait",
  "bytes",
- "futures-util",
+ "futures-core",
  "http",
  "http-body",
  "http-body-util",
  "mime",
  "pin-project-lite",
- "rustversion",
  "sync_wrapper",
  "tower-layer",
  "tower-service",
@@ -595,7 +592,7 @@ version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fa961b519f0b462e3a3b4a34b64d119eeaca1d59af726fe450bbba07a9fc0a1"
 dependencies = [
- "thiserror 2.0.18",
+ "thiserror",
 ]
 
 [[package]]
@@ -1419,7 +1416,7 @@ dependencies = [
  "serde",
  "smallvec",
  "tempfile",
- "thiserror 2.0.18",
+ "thiserror",
  "uuid",
  "web-time",
 ]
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Reviewed in depth. **No PR-introduced functional blocker.** The dependency trimming is careful and preserves every runtime-only capability I'd have expected to break. One request: substantiate or revise the timing row.

## The feature audit — the part that can't be checked by reading

Removing a dependency feature can compile cleanly and fail only at runtime, so I checked each trimmed feature against actual use rather than against the diff:

- **RocksDB LZ4/Zstd are retained** (`crates/jazz-tools/Cargo.toml:88`). This was my biggest worry, because a missing codec fails when *opening an existing database*, not at build time. Persistent catalogue storage does open LZ4/Zstd DBs (`server/catalogue_storage.rs:92`), and Groove writes both. Nothing removed.
- **`jsonwebtoken` keeps `rust_crypto` and `use_pem`** in both crates, which covers the real RSA/EC/EdDSA/HMAC/PEM/DER/JWK uses in `middleware/auth.rs:400` and `auth_admission.rs:336`.
- **`reqwest` keeps `json` + `rustls-tls`** (`:122`); no `bytes_stream` use exists — the HTTP proxy consumes complete bodies via `.bytes()` (`server/routes/http.rs:211`).
- **WebSocket transport keeps both native and WebPKI rustls roots** (`:101`).
- **`tokio`'s `full` is correctly decomposed**, with `rt-multi-thread` and `signal` scoped to `cli` — SIGTERM handling genuinely needs `tokio/signal` (`commands/server.rs:177`).
- Serde derive and UUID `v4`/`v5`/`v7`/`serde` all still present.

The three source changes are an Axum 0.7→0.8 migration — native `async fn` in traits replacing `async_trait`, `:hash`/`:app_id` → `{hash}`/`{app_id}`, and `Vec<u8> -> Bytes` — all equivalent, not workarounds for removed features.

## Please substantiate the timing row

| Metric | Claimed | I measured |
|---|---:|---:|
| Debug target footprint | 5.49 → 2.03 GiB (−63.0%) | 6.25 → 2.31 GiB (**−63.1%**) |
| Cold debug build | 107.99s → 73.16s (−32.3%) | 232.66s → 222.64s (**−4.3%**) |

The footprint claim reproduces almost exactly in relative terms. The timing claim did not, with fresh target directories and `cargo build -p jazz-napi -j 6`.

**I don't think this means your number is wrong** — I used `-j 6`, and if you used default parallelism on a many-core box the results would legitimately differ, since stripping debug info mostly cuts I/O and link time and that benefit scales with job count. Could you add the exact command, job count, cache policy, host and toolchain to the table? Then it's reproducible rather than contested.

I also did not produce the packaged `.node` via the NAPI CLI, so the `.node` size rows remain unverified by me. Stripping debug info from the built shared libraries as a proxy gave 105.1 → 98.8 MiB (−5.9%).

## Non-blocking

- Worth documenting the debugger trade-off next to `Cargo.toml:37`: dependencies now carry no debug metadata and workspace crates keep only line tables. Panic locations stay useful, but local-variable and type inspection in dependency frames are lost. `--profile debugging` restores it. That's a fine trade, it just shouldn't be a surprise to whoever next attaches a debugger.
- A small CI feature-surface check for the selected TLS/JWT/RocksDB features would be cheap insurance. The current config is right; these are precisely the regressions most likely to be reintroduced later by someone trimming again.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-tools --features test --all-targets`, `--no-default-features` for both crates, and `cargo check -p jazz-wasm --target wasm32-unknown-unknown` all exit 0.

Two failures, both **inherited from the base** — I re-ran each on the base SHA and they fail identically:
- `cargo check --workspace --all-targets`: stale exhaustive match at `crates/jazz-sim/tests/customer_child_policy_minimal.rs:205` missing `SubscriptionEvent::Rejected`.
- `cargo test -p jazz-tools --features test`: `numeric_claims_match_integer_columns_across_core_widths` times out. That one is red on `codex/jazz-core-engine-swap` itself and is fixed by #1159.

aeplay:
🤖 *Automated review by Claude, run by @aeplay. Anselm will give the verdict.*

Clean — and I went looking specifically for the failure mode where build time or
artifact size is bought by turning off something load-bearing. It is not.

Checked and confirmed **not** disabled: debug assertions, overflow checks, panic
symbols, LTO behaviour. The root profile change touches debug information only.

Two dependency narrowings I traced rather than assumed:

- `reqwest` drops `stream`, and I found no `bytes_stream`/streaming-response use
  in production sources.
- Axum 0.7 → 0.8 with explicit server features. The removed defaults are `form`
  and `tower-log`; no `Form` extractor use found. The route-parameter change from
  `:param` to `{param}` (`server/routes/mod.rs:101`, `:114`, `:598`) is required
  by 0.8 and is present.

WASM and NAPI do not consume the optional server/tool dependencies this narrows,
and their shared `jazz` feature set is untouched — worth stating explicitly
because the WASM build was broken recently by exactly that kind of change.

**Rebase note:** this predates the consolidation and edits the now-absorbed
`jazz-tools` and `jazz-server` manifests. Port each explicit feature decision
into the current `crates/jazz/Cargo.toml`, keeping its hand-declared targets —
do not resolve by merging both sets of declarations. `autotests = false` means
duplicate target names make cargo reject the entire manifest, which takes the
crate dark rather than failing loudly on one target.

No changes requested beyond the rebase.

PR #1181, opened 2026-07-29
Clarify the browser schema-mismatch error, and make the fingerprint order-independent

aeplay:
Draft — the second commit is a real decision and I would rather it be deliberate than merged by momentum.

Context: a user on **alpha 53** reported `Failed to start Jazz subscription  Error: incompatible persistent browser schema`, spent hours clearing site data, cookies and caches across two browsers, and eventually worked around it by renaming their schema tables from `family_members` to `familyMembers`.

## What the error actually is

Not a storage problem. `blockTabForSchemaMismatch` fires from `handleSchemaReady` when `namespace.schemaFingerprint !== schemaFingerprint` — two **live** tabs reported different schema fingerprints to the same SharedWorker broker. Nothing persisted is compared, which is exactly why clearing storage did nothing.

## Commit 1 — the message (safe, no behaviour change)

Says what happened and what to do. The leading phrase is kept verbatim because it is a public error string that callers and two browser tests (`shared-worker-broker.test.ts:217`, `worker-bridge.test.ts:2459`) match on as a substring, so both still pass unchanged.

## Commit 2 — order-independent fingerprint (decide deliberately)

`serializeRuntimeSchema` sorted only top-level table names; key order *within* each table value was JS insertion order, which `JSON.stringify` preserves. Two clients describing the same schema but building the objects in different orders produced different fingerprints. Keys are now ordered at every depth.

Arrays are deliberately **not** sorted — column order is positionally significant, so reordering would change what the schema means rather than normalize how it is written. Three tests cover order-independence, column order still distinguished, table names still distinguished.

**This is not a confirmed fix for the report.** The investigation refuted both leading hypotheses: no underscore-specific ordering or second serializer exists (tested with `family_members` vs `familyMembers`, mixed case, reversed construction order, `structuredClone`, `Map -> Object` — all stable), and renaming tables does **not** enter a fresh broker namespace, since the SharedWorker name is `jazz-broker:${appId}:${dbName}` and `schemaHash` is explicitly `null`. Root cause remains undetermined. This closes a real class of spurious mismatch regardless.

**The risk, and why it is a separate commit:** it changes the fingerprint every client computes, once. During a rollout, a tab on the old version and one on the new will disagree, and one will be schema-blocked until the other closes.

## Note for the new core

`codex/jazz-core-engine-swap` deletes this whole mechanism — no SharedWorker namespace, no cross-client fingerprint comparison, a dedicated `Worker` per runtime — so this error class cannot occur there. But its native encoder walks `Object.entries(schema)` **without sorting** (`schema-codec.ts:77`), making table insertion order observable in the native schema encoding. Worth closing there too, separately.

### DIFF
--- a/packages/jazz-tools/src/drivers/schema-wire.test.ts
+++ b/packages/jazz-tools/src/drivers/schema-wire.test.ts
@@ -1,101 +1,22 @@
-import { afterEach, describe, expect, it, vi } from "vitest";
-import type { WasmSchema } from "./types.js";
-import { getRuntimeSchemaCacheKey, serializeRuntimeSchema } from "./schema-wire.js";
+import { describe, expect, it } from "vitest";
+import { serializeRuntimeSchema } from "./schema-wire.js";
 
-afterEach(() => {
-  vi.restoreAllMocks();
-});
-
-describe("serializeRuntimeSchema", () => {
-  it("wraps runtime schema payloads and serializes Bytea defaults as JSON arrays", () => {
-    const schema: WasmSchema = {
-      files: {
-        columns: [
-          {
-            name: "payload",
-            column_type: { type: "Bytea" },
-            nullable: false,
-            default: { type: "Bytea", value: new Uint8Array([0, 1, 255]) },
-          },
-        ],
-      },
-    };
-
-    expect(JSON.parse(serializeRuntimeSchema(schema))).toEqual({
-      __jazzRuntimeSchema: 1,
-      schema: {
-        files: {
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Closing this stale implementation while retaining the underlying diagnostic in #2062. The SharedWorker mismatch path and fingerprint contract changed substantially, and this patch no longer applies. A fresh fix must define rollout compatibility before changing every client fingerprint, then cover nested property insertion order while preserving meaningful array order.

PR #1183, opened 2026-07-29
(new core) Preserve query order in subscription resets

aeplay:
Fixes the 10 browser-suite failures tracked as R2, in `packages/jazz-tools/tests/browser/db.subscribeAll.sort.test.ts`.

## The bug

The packed-reset fast path forwarded reset content in `chunk.delta.added` order — UUID/batch order — rather than the query's rank order. Eight of the ten failures were misordering. **Two (at lines 479 and 575) were window failures**, where a wrong limit/offset window yields a wrong *result set*, not merely a wrongly-ordered one.

## Fixed in core, deliberately not in the adapter

The obvious fix is to re-sort in the TypeScript adapter. That would have been wrong: it duplicates ordering semantics that core already owns — nulls, ties, and the encoded-bytes tie-break in `INV-QUERY-23` — and would drift from them over time. It is exactly the class of duplication `dev/TS_SEMANTIC_DUPLICATION_AUDIT.md` catalogues.

So core emits correctly-ordered reset content instead. The change is 13 lines in `crates/jazz/src/node/query_eval.rs` plus 28 lines of coverage in `crates/jazz/src/db/tests.rs` — additions only, no deletions.

## Provenance

Cherry-picked onto a clean base off `track/browser-consolidated`, deliberately leaving behind that branch's `WIP: bound recursive Inherits lowering` commit, which is unrelated and unfinished.

## Gates

`cargo test -p jazz`, `-p groove`, `-p jazz-server`, `cargo check -p jazz-sim --benches` — exit 0.

Because this changes delivery ordering, the mechanism gates were run explicitly rather than assumed:
- all three `incremental_delivery_canary` tests — ran and passed
- `JAZZ_SEED_COUNT=300 m3_maintained_one_shot_differential_oracle` — 1 passed, 107s

`cargo test -p jazz-tools --features test` fails only `numeric_claims_match_integer_columns_across_core_widths`, red on `codex/jazz-core-engine-swap` itself and fixed by #1159.

### DIFF
--- a/crates/jazz/src/node/query_eval.rs
+++ b/crates/jazz/src/node/query_eval.rs
@@ -65,6 +65,7 @@ pub(crate) struct LocalMaintainedViewSubscription {
     maintained: MaintainedSubscriptionView,
     terminal_schemas: MaintainedTerminalSchemas,
     tables: BTreeMap<String, TableSchema>,
+    result_query: JazzQuery,
     result_table: String,
     result_select: Option<Vec<String>>,
     result_set: BTreeSet<ResultMemberEntry>,
@@ -4473,6 +4474,10 @@ where
             row_keys.insert((row.table().to_owned(), row.row_uuid()));
             rows.push(row);
         }
+        // Result-member ordering is for identity and deduplication, not public
+        // query rank. Membership/windowing is already lowered; only restore the
+        // selected roots to their advertised order before sending a reset.
+        self.apply_query_order(shape.query(), &mut rows)?;
         let root_count = rows.len();
         let mut edges = Vec::new();
         for fact in program_facts {
@@ -6412,6 +6417,7 @@ where
             maintained,
             terminal_schemas,
             tables,
+            result_query: shape.query().clone(),
             result_table: shape.query().table.clone(),
             result_select: shape.query().select.clone(),
             result_set: BTreeSet::new(),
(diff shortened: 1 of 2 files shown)


crates/jazz/src/node/query_eval.rs
```diff
@@ -6665,6 +6671,10 @@ where
                 rows.push(row);
             }
         }
+        // `result_set` is keyed by member identity, so its BTreeSet iteration
+        // order cannot be exposed as a query reset order. Do not re-window: the
+        // maintained program already chose this result set.
+        self.apply_query_order(&local.result_query, &mut rows)?;
```
joeinnes, line 6677: If a query both projects and orders by a field which isn't returned as part of its projection, ordering here will fall back to UUID sorting.

joeinnes (approved):
Comments contain non-blocking but nonetheless important work. Please schedule separately, or fix in this PR.

aeplay:
🤖 @joeinnes sorry for the silence — this was picked up straight away but I never replied here, so from your side it looked ignored. Fixed in **#1208**.

**You were right, and it was worse than a fallback.** A subscription selecting only `title` while ordering by an omitted `rank` returned `[charlie, alpha, bravo]` — insertion/UUID order — instead of `[alpha, bravo, charlie]`. Silently.

The trigger is narrower than the line suggests: it needs the ordering value absent from the internal `CurrentRow`, not merely from the logical projection. The comparator reads the cell at `query_eval.rs:7737`, missing values compare equal, and it falls through to UUID at `:7751`. The maintained reset path was projecting *before* sorting, which is what removed the value.

**Resolution: carry, not reject.** Full rows are carried through sorting and the projection applied afterward — fixing both the maintained path and the authoritative reset path (`:4462`).

I did consider rejecting at validation instead, following the precedent #1182 set for scalar `in` against array columns. A repository audit decided it: no existing Rust, TypeScript, example or benchmark usage selects one field while ordering by another omitted one — every existing select+order pair already includes the order key. So carrying the field through breaks nothing, and rejecting would have been a gratuitous restriction.

Canaries and the 300-seed oracle both pass, since ordering is where delivery semantics drift quietly.

PR #1186, opened 2026-07-29
(new core) URGENT: repair the integration base compile break, and guard against a repeat

aeplay:
## The integration branch does not compile

`codex/jazz-core-engine-swap` at `1ce294908`:

```
error[E0599]: no method named `iteration_cap` found for enum `RecursionBound`
  --> crates/jazz/src/node/query_eval.rs:3671
```

`c2db5a8e4` ("Remove recompute metrics and stale query names") removed `RecursionBound::iteration_cap` while one call site survived. **Neither PR conflicted textually**, so both merged cleanly and produced a tree that does not build.

This is currently blocking every branch. Three separate pieces of work were failing `cargo test -p jazz` on it rather than on their own changes, and two independent lanes rediscovered it before it was traced.

## The fix

Restores the accessor as `depth_steps`, preserving the exact prior mapping — `Fixpoint => 128`, `MaxDepth(n) => n.max(1)` — with a doc comment naming the commit that removed it, so the next person understands why it exists.

If the removal intended the *call site* to change rather than the accessor to survive, that is a fine follow-up; this restores compilation without guessing at that intent.

## The guard

Adds a `build-integration` job running `cargo check --workspace --all-targets`, and extends the `push:` trigger to this branch.

**The integration branch has had no CI run since 2026-07-20**, because `push:` covered only `main`. So the merge *result* was never built. PR-level checks structurally cannot catch this class of break: each PR is built against the base as it was, not against the other PR landing beside it.

The job is **deliberately narrow**. The three existing suites are not green — a one-line clippy fix (#1185) failed all three — so gating on them would force routine overrides and make "red" mean nothing. This checks the one property we can currently guarantee, and is fast and flake-free. The others should be promoted to gates individually as each goes green; `lint` is closest.

## Verification

`cargo check -p jazz` exits 0 with the fix, having genuinely rebuilt (not a cached result). `cargo test -p jazz` and `cargo check --workspace --all-targets` were still running at push time given how much this blocks; I will confirm both on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches: [main, codex/jazz-core-engine-swap]
   push:
-    branches: [main]
+    branches: [main, codex/jazz-core-engine-swap]
 
 concurrency:
   # For pushes, this lets concurrent runs happen, so each push gets a result.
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Verified on this branch:

- `cargo check -p jazz` — exit 0, genuinely rebuilt
- `cargo test -p jazz` — exit 0
- `cargo check --workspace --all-targets` — still running, will follow up

Independent corroboration while this was in flight: a separate lane resolving the #1155 split hit the same break, proved it on untouched `HEAD` of the base, then built a disposable base worktree *with* an equivalent compile repair and confirmed the tree builds. So two lanes converged on both the diagnosis and the fix shape.

aeplay:
🤖 Folded #1184's three-line jazz-sim fix into this PR and **closing #1184 in favour of this one.**

Reason: `cargo check --workspace --all-targets` — the very check the new `build-integration` job runs — was still failing on the non-exhaustive `SubscriptionEvent::Rejected` match that #1184 fixes. Landing the gate without it would have made the gate red on arrival, which is precisely the failure mode this PR argues against.

Verified together:

| check | result |
|---|---|
| `cargo check -p jazz` | exit 0 |
| `cargo test -p jazz` | exit 0 |
| `cargo check --workspace --all-targets` | **exit 0** |

All three genuinely rebuilt rather than served from cache.

PR #1187, opened 2026-07-29
(new core) Attribute cold-settle time by phase

aeplay:
Makes an adopter cold-start profile **reproducible**. The numbers below have been steering demo work but the instrumentation that produced them was never in the repo, so nobody else could re-derive or challenge them.

## What it found, and why it matters

The prior conclusion (2026-07-14) was that cold-start grinding was **exhausted** — that further gains needed design work such as progressive or lazy materialization. That was wrong, and this attribution is what overturned it.

| Settle phase | Time | Share |
|---|---:|---:|
| Receive/decode | 0.039 ms | ~0% |
| **Core ingest** | **17.8 s** | **64.0%** |
| Maintained-view settle | 7.0 s | 25.3% |
| Subscription adapter/materialization | 1.4 s | 4.9% |
| Residual: scheduling, queueing, checks | 1.6 s | 5.8% |

The critical path is **receiver-side bulk view ingest** — not transfer, not connection. Two independent corroborations: a CPU capture of the same shape put `apply_view_updates_in_batch` at 51.6% of foreground samples (3,736/7,235), and a streaming-zstd probe measured 35.3 ms encode / 16.3 ms decode over 20.19 MB raw → 0.49 MB, so serialization is nowhere near the cause.

## Honest framing of the numbers

- **12.409 s** is the isolated low-load cold settle, against the <10 s goal. That is the figure to quote.
- The 27.766 s exact-timer pass ran while host one-minute load rose 2.74 → 16.00, so it is **not** a latency baseline. Its phase ordering and counters are still valid, and the table above comes from it.

Both qualifications are stated in `dev/benchmarks/COLD_SETTLE_PROFILE_20260728.md` rather than buried.

## Fixtures

Anonymized in-repo customer-shape fixture only — 39 subscriptions, 27,518/27,518 final rows. The **jazz-private sensitive-data guard exits 0**, and I verified the guard itself fires on a planted positive rather than trusting a silent pass.

## Gates

`cargo test -p jazz`, `-p groove`, `cargo check -p jazz-sim --benches`, and `cargo check --workspace --all-targets` all exit 0.

## Follow-up already in flight

Because the bottleneck is receiver-side ingest, the flush-once refresh fix (which removes a per-subscription flush from exactly that path, and is worth 254x at 1,000 routes) is now being measured against this profile. At 39 subscriptions the saving is bounded by far less than that, so it will be reported as measured, not projected.

### DIFF
--- a/crates/groove/src/ivm/runtime/mod.rs
+++ b/crates/groove/src/ivm/runtime/mod.rs
@@ -5154,12 +5154,22 @@ where
                 let input = self.update_unary_input(graph_node, node)?;
                 let raw_projection =
                     self.raw_projection_fields(node, project, &input.descriptor, output_desc)?;
-                NodeState::update_map_project(
+                let result = NodeState::update_map_project(
                     project,
                     output_desc,
                     &input,
                     raw_projection.as_deref(),
-                )
+                );
+                #[cfg(feature = "cold-settle-attribution")]
+                if let Ok(output) = &result {
+                    crate::cold_settle_attribution::record_map(
+                        self.context.eval_mode == EvalMode::Hydrate,
+                        self.depends_on_dominant_child(node)?,
+                        input.deltas.len(),
+                        output.deltas.len(),
+                    );
+                }
+                result
             }
             OpType::UnwrapNullable(unwrap) => {
                 let input = self.update_unary_input(graph_node, node)?;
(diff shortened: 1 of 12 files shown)


aeplay:
🤖 Follow-up finding: **the `operator_cardinality` section of this attribution is non-functional.**

While using this instrumentation to investigate the dominant child table, its `operator_cardinality` dominant-child buckets came back all zero — for `keyed_join` and `map_project`, at core→relay, relay→client and client alike. That is tag loss, not a real measurement: the work demonstrably happens.

That matters more than a missing metric, because a table of zeros reads as evidence. In this case it would have supported a wrong conclusion — I was testing whether inherited-read parent visibility is resolved per child row, and "zero per-child join calls" is exactly what a confirmed hypothesis would look like. The investigation only reached the right answer by discarding these counters and dumping the lowered graph instead.

So either the tagging should be fixed or that section should be removed. I would rather it be absent than silently zero. Everything else in the receipt — the phase table, the CPU sampling, the zstd probe — is sound and independently corroborated.

For the record, the investigation it was used for returned a **negative** result worth having: inherited read is already lowered as a semi-join (`normalize_inherited_parent_policy` → `RowSetExpr::Join { mode: Semi }`, with the parent side deduped via `ArgMaxBy` before the keyed join), so existential authorization is not an available cold-start lever. The ~525µs/row is join dataflow plus receiver-side delivery, with `apply_view_updates_in_batch` at 51.6% of samples.

crates/groove/src/ivm/runtime/mod.rs
```diff
@@ -5509,6 +5519,44 @@ where
         Ok(deltas)
     }
 
+    #[cfg(feature = "cold-settle-attribution")]
+    fn depends_on_dominant_child(&self, node: NodeId) -> Result<bool, IvmRuntimeError> {
```
joeinnes, line 5523: This runs inside the benchmark timing, and does a recursive graph walk, and allocates a `Vec` on map/joins. I'm not sure whether it's meaningful but it could overreport the `semantic_tick` time.

crates/groove/src/ivm/runtime/mod.rs
```diff
@@ -5509,6 +5519,44 @@ where
...
+        if matches!(
+            &graph_node.descriptor.operator,
+            OpType::TableSource(source) if source.table == "res_l_child_3"
+        ) {
+            return Ok(true);
+        }
+        // Policy lowering can replace the direct table source with an indexed
+        // source. The anonymous child shape is unique in this benchmark, so
+        // retain the tag through that lowering as well.
+        if ["parent_id", "value_text", "value_json"]
```
joeinnes, line 5537: Are we only supposed to be matching `res_l_child_3` here or any child table? All child tables in this benchmark have these fields, so this is only checking what's coming out of the node, not which table it's from.

joeinnes (approved):
Couple of questions on correctness, not sure we're measuring exactly what we are trying to measure with this.

PR #1188, opened 2026-07-29
(new core) Preserve the permission subject when a mergeable commit is retried

aeplay:
Addresses @joeinnes' review comment on merged PR #1150.

## The bug

```rust
let permission_subject = self.open_tx.local_permission_subjects
    .remove(&tx_id)                    // consumed here
    .unwrap_or(stored.tx.made_by);     // fallback on retry
```

It is `.remove()`, not `.get()`, so the override is **consumed before the fallible section completes**. Any `Err` after that point leaves the transaction `Fate::Pending` with the override gone — and the `if !matches!(stored.fate, Fate::Pending) { return Ok(()) }` guard above means a retry re-enters exactly this path and takes the `made_by` fallback.

For a `TrustedBackend` writing on behalf of a principal, `made_by` is the **user** while write policy must be evaluated against the **authenticated link identity**. So a retry evaluates policy against the claimed author instead of the authenticated subject — privilege confusion, and `INV-RLS-17` / `INV-RLS-18` are the invariants at stake.

## Reproduced red first

With an injected first-attempt error, the unpatched retry was **accepted under `made_by`** where it should have been rejected under the backend subject. The new regression asserts an attributed user write stays denied when the trusted backend identity lacks permission — including on retry.

## The fix

`.get().copied()`, with removal only after a persisted rejection or a complete successful finalization. The override now survives until the operation actually completes.

## Gates

Verified on the current base (after #1186 repaired the base compile break — the earlier run failed on that, not on this change): `cargo test -p jazz` exits 0.

### DIFF
--- a/crates/jazz/src/node/ingest.rs
+++ b/crates/jazz/src/node/ingest.rs
@@ -592,12 +592,14 @@ where
         let permission_subject = self
             .open_tx
             .local_permission_subjects
-            .remove(&tx_id)
+            .get(&tx_id)
+            .copied()
             .unwrap_or(stored.tx.made_by);
         for version in &records {
             if !self.version_satisfies_write_policy(version, permission_subject) {
                 let fate = Fate::Rejected(RejectionReason::AuthorizationDenied);
                 self.ingest_rejected_transaction(stored.tx, fate)?;
+                self.open_tx.local_permission_subjects.remove(&tx_id);
                 return Ok(());
             }
         }
@@ -611,6 +613,7 @@ where
         )?;
         self.create_merge_versions_for(&records)?;
         self.checkpoint_large_values_for_tx(tx_id)?;
+        self.open_tx.local_permission_subjects.remove(&tx_id);
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


crates/jazz/src/node/ingest.rs
```diff
@@ -611,6 +613,7 @@ where
         )?;
         self.create_merge_versions_for(&records)?;
         self.checkpoint_large_values_for_tx(tx_id)?;
+        self.open_tx.local_permission_subjects.remove(&tx_id);
```
joeinnes, line 616: Can we make sure this always runs when the persisted fate is terminal? At the moment an error could return above this line (e.g. if checkpointing errors), even if the persisted fate is already terminal, meaning the map entry would never get cleaned up.

joeinnes (approved):
Approved, but please review comment and either schedule follow up work or fix in this PR.

aeplay:
🤖 @joeinnes sorry for the silence — picked up immediately but I never replied here. Fixed in **#1210**.

**It did leak, confirmed rather than theoretical.** A focused checkpoint-error reproduction persisted `Accepted/Global` and the local override was still present afterward. Since terminal transactions never retry and transaction ids are not reused, repeated attributed writes would grow `local_permission_subjects` unboundedly on a long-lived node.

What you spotted is the mirror of the bug this PR fixed: consuming the override too early caused the retry-under-`made_by` problem, and deferring removal created a path where a post-persistence error strands the entry forever.

The fix moves cleanup into the shared fate-persistence path — terminal status is marked right after the batch commits, and removal happens after all subsequent fallible work returns. I deliberately avoided adding another cleanup call site, since that would only move the gap to the next error path. Errors *before* fate persistence still retain the override, so this PR's pending-retry regression still passes.

PR #1189, opened 2026-07-29
(new core) Count newly allocated pages in the OPFS WAL tail threshold

aeplay:
Addresses @joeinnes' review comment on merged PR #1144: `total_pages` there includes pages allocated during the current commit that live only in the WAL, so subtracting it undercounts the tail and delays checkpointing.

**This partly defeats #1144's own purpose.** That PR exists to bound the OPFS WAL by checkpointing on tail growth; undercounting the tail lets the WAL grow past the bound it was meant to enforce.

## Reproduced red first

An allocation-heavy WAL commit had a true tail of **1,026 pages reported as 514**.

## The fix

The last checkpoint's home-file boundary is tracked separately, so the tail is `persisted_pages - checkpointed_pages` rather than the latest logical `total_pages`.

New regression `flush_wal_counts_newly_allocated_pages_in_tail_threshold` verifies a 1,004-page allocation-only tail, then that a second allocation-heavy commit triggers the checkpoint at the intended threshold rather than later.

## Gates

`cargo test -p opfs-btree` exits 0 — 56 passed, with the new regression named as passing. `cargo test -p jazz` exits 0.

Verified on the current base after #1186 repaired the base compile break; the earlier failing run was that, not this change.

### DIFF
--- a/crates/opfs-btree/src/db.rs
+++ b/crates/opfs-btree/src/db.rs
@@ -122,7 +122,10 @@ pub struct OpfsBTree<F: SyncFile> {
     active: Superblock,
     root_page_id: Option<PageId>,
     total_pages: u64,
-    // Pages >= active.total_pages are WAL tail pages, not home locations.
+    // The home-file boundary recorded by the latest checkpoint superblock.
+    // Physical pages at or beyond this boundary are WAL-only. `active` tracks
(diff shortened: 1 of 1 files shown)


joeinnes (approved):
LGTM

PR #1190, opened 2026-07-29
(new core) Flush Groove once per subscription refresh, not once per subscription

aeplay:
Found via adopter benchmark #1170. Two lines of production change.

## The defect

The Groove flush sat at `crates/jazz/src/node/query_eval.rs:6486`, **inside the per-subscription drain** called from the refresh loop. So N maintained subscriptions meant N flushes per refresh — **independent of how many rows actually changed**.

That is an `INV-INC-1` violation: per-refresh work scaled with accumulated subscription count rather than with change size.

## The fix

Flush once at the start of `refresh_subscriptions_in` (`crates/jazz/src/db.rs:3647`), then drain receivers unchanged.

| routes | before | after |
|---|---:|---:|
| 100 | 123.55 ms | 5.10 ms |
| 1,000 | 10.443 s | **41.10 ms** |

254x at 1,000 routes, and the scaling exponent drops from 84.5x to 8.06x per 10x routes — from super-linear to roughly linear, which is what remains once the only per-route cost is draining receivers. Independently consistent with the adopter's own measurement of 37.55 s → 132 ms.

## Correctness was the bar, not the speedup

A flush-count change must not become a delivery-semantics change. Verified directly:
- all three `incremental_delivery_canary` tests pass
- `JAZZ_SEED_COUNT=300 m3_maintained_one_shot_differential_oracle` passes (1 passed, 141.82 s)

The regression test pins the **invariant** — refresh cost must not scale with unrelated subscription count — rather than a timing threshold, so it will not rot into a flaky benchmark.

## What this does NOT fix, measured

I tested it against the anonymized customer-shape cold-start profile, where receiver-side ingest is 64% of settle time. **No measurable gain: 12.507 s vs a 12.409 s baseline.** That scenario has only ~4 tick cycles, so there are almost no refreshes to deduplicate; its cost is bulk ingest volume (27,518 client + 46,740 relay bundle ingests). Both mechanisms live under "receiver-side ingest", which is why the attribution looked like a match and is not. Recording it so nobody re-runs that experiment.


crates/jazz/src/db.rs
```diff
@@ -3652,6 +3644,7 @@ where
     let pending_authoritative_resets = node
         .borrow_mut()
         .take_pending_authoritative_reset_binding_views();
+    node.borrow_mut().flush_query_runtime()?;
```
joeinnes, line 3837: Should we also flush before we iterate over `coverage_groups` (https://github.com/garden-co/jazz/blob/6c054693a575ff7d2f03824dbfc2e1f864bba4c9/crates/jazz/src/db.rs#L5212)?

We could then get rid of this: https://github.com/garden-co/jazz/blob/6c054693a575ff7d2f03824dbfc2e1f864bba4c9/crates/jazz/src/peer.rs#L779

joeinnes (approved):
Potential perf enhancement in the comments.

aeplay:
🤖 @joeinnes fixed in **#1209** — and your "potential perf enhancement" was actually a second instance of the same `INV-INC-1` violation this PR fixed.

Every coverage group was going through `query_update…` then `drain_maintained…`, executing the peer flush **once per group**. So delivery work scaled with accumulated group count rather than change size — the same shape as the per-subscription flush, in a different loop. This PR fixed one instance of a pattern rather than the pattern.

| groups | before | after |
|---|---:|---:|
| 100 | 138.42 ms | 12.96 ms |
| 1,000 | 11.080 s | **122.03 ms** |
| scaling | 80.04x | **9.42x** |

The after-run was under *heavier* host load (11.92 vs 9.37), so it is not a load artifact. The 80x→9.4x scaling collapse is the same signature this PR produced.

The guard pins the invariant — delivery cost must not scale with unrelated coverage-group count — rather than a timing threshold, so it will not rot into a flaky benchmark.

PR #1192, opened 2026-07-29
(new core) Use independent per-metric medians in the INV-INC-1 receipt

aeplay:
Addresses @joeinnes' review comment on merged PR #1166: the harness emitted *all* measurements from the run that produced the median allocation, not independent medians per metric.

That matters more than a normal benchmark nit because this harness produces the `INV-INC-1` receipt and its **1.025x gate threshold is read off these numbers**. A value labelled "median" that is really "the co-measurements of the median-allocation run" makes cross-metric comparison unsound.

## What changed

Allocations, allocation bytes, and wall time now use **independent per-metric medians**. Delivery-shape counts stay co-measured — but only after asserting they are identical across every sample, so they describe all samples rather than one selected run.

## The threshold still holds, and was not retuned

| metric | corrected ratio |
|---|---|
| allocations | 1.0010306622x |
| allocation bytes | 1.0173272617x |

Both reproduce the original baseline exactly, because the allocation counters are deterministic per rung. **`1.025x` still holds and I did not adjust it** — the lane was explicitly instructed to report rather than retune if the corrected statistic moved the gate.

## Gates

`cargo test -p jazz` and `cargo check -p jazz-sim --benches` exit 0; all three `incremental_delivery_canary` tests pass.

Sorry this one lagged the other three — it was finished and sat unpushed on my side.

### DIFF
--- a/crates/jazz/benches/relation_include_delivery.rs
+++ b/crates/jazz/benches/relation_include_delivery.rs
@@ -67,13 +67,12 @@ fn main() {
 
     let mut rungs = Vec::with_capacity(scales.len());
     for scale in scales {
-        let mut measurements = (0..samples)
+        let measurements = (0..samples)
             .map(|sample| measure_single_child_insert(scale, sample))
             .collect::<Vec<_>>();
-        measurements.sort_by_key(|measurement| measurement.allocs);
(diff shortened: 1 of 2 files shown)


joeinnes (approved):
LGTM

PR #1193, opened 2026-07-29
(new core) Consolidate jazz-tools and jazz-server into jazz

aeplay:
Removes two crates that exist only because the new core was grafted onto the existing repo.

**They could not be absorbed one at a time.** `jazz-server` depends on `jazz`, and `jazz-tools` depends on `jazz-server`, so folding `jazz-tools` into `jazz` alone would have made `jazz` depend on a crate that depends on `jazz` — a cycle Cargo rejects. Absorbing both resolves it.

## Three commits, deliberately separated for review

1. **`5afabf3cc` — pure renames.** 91 files changed, **0 insertions, 0 deletions**. Skim it. Keeping it content-free maximises git rename detection, so in-flight branches editing these files merge to the new paths rather than conflicting.
2. **`7adcc287a` — the wiring.** Module wiring, import rewrites, the merged manifest, and the four external consumers (`jazz-napi`, `jazz-rn`, both example servers). This is the commit to actually read.
3. **`eff9753dc` — docs and the gate command.**

## The canonical gate changes

```
before:  cargo test -p jazz-tools --features test
after:   cargo test -p jazz --no-default-features --features test
```

Updated in `AGENTS.md`, `TESTING_GUIDELINES.md`, `SPEC/D_testing_gates.md` and `dev/CONTRIBUTING.md`. Deliberately **not** updated in `dev/CI_NOTES.md`, `PERF_FINDINGS_DROPDOWN.md`, `REFRESH_SUBSCRIPTIONS_SLOPE.md`, `NATIVE_SELECT_INHERITS.md` or `TS_DEDUP_PHASE1.md` — those are dated receipts recording what was actually run at the time, and rewriting them would falsify the record.

`TESTING_GUIDELINES.md` and `ENV.md` moved to `crates/jazz/`, since the wiring commit left them in a directory with no crate in it while `AGENTS.md` and `CLAUDE.md` both point at that path.

## Target preservation — the quiet risk

`jazz-tools` set `autotests = false` / `autobenches = false`, so every test and bench was explicitly declared. A target omitted from the merged manifest would **silently stop running** with nothing going red.

**27 declared targets before, 40 after, zero dropped.** The 13 additions are required, not scope creep: old `crates/jazz/Cargo.toml` had no `autotests` setting, so its tests were auto-discovered. Under the merged manifest they must be declared by hand or vanish — and that list includes `incremental_delivery_canary`, `four_tier`, `wire_fixtures` and `cli_dry_run`. All 13 verified to exist at their declared paths.

Note the new command runs **more** than the old pair did: `jazz-tools`' library unit and doc tests were suppressed by its `[lib] test = false` / `doctest = false` and are now eligible — 1,074 tests.

## Known reds, both pre-existing and both already fixed elsewhere

Two failures surface here only because consolidation switched on those previously-suppressed library tests: `converts_reverse_inherited_select_with_nested_source_policy` and `converts_policy_graph_perf_public_schema_to_native_fixture_byte_stably`, both in `public_schema_convert.rs`. They have been red all along, hidden by `[lib] test = false`. Fixed by the schema-convert work, which is being rebased separately — **that should land first**, then this.


AGENTS.md
```diff
@@ -22,7 +22,11 @@ For ordinary Rust/core work, the full gate set is:
 
 - `cargo test -p jazz`
 - `cargo test -p groove`
-- `cargo test -p jazz-tools --features test` (matches `crates/jazz-tools/TESTING_GUIDELINES.md`)
+- `cargo test -p jazz --no-default-features --features test` (matches `crates/jazz/TESTING_GUIDELINES.md`).
+  This replaces the former `cargo test -p jazz-tools --features test`: `jazz-tools`
+  and `jazz-server` are now part of `jazz`, so their suites run here. Note it also
+  runs `jazz-tools`' library unit and doc tests, which its old `[lib] test = false`
+  had suppressed.
```
nrainhart, line 29: Do we really need to keep this information?

nrainhart (approved):
Thanks for the cleanup!!

PR #1197, opened 2026-07-30
(new core) Rebind maintained streams when claims change (security)

aeplay:
Fixes finding 3 from adopter PR #1170: **claim revocation leaked future rows.**

> After admin `true -> false`, a one-shot is empty but the existing stream emits no removal and later emits a newly inserted protected row.

So a maintained subscription opened while an identity held a claim kept serving after revocation — and delivered rows created **after** the revocation. The one-shot was correctly empty, so the policy itself was right; the live stream was simply never rebound.

New core is pre-release with no production users, so this was urgent-to-fix rather than urgent-to-patch.

## Reproduced red first

The pre-fix reproduction failed exactly as reported: after `admin: true -> false` the stream emitted neither removal nor close, then received the post-revocation row (literally named `"must remain private"` in the fixture). `0 passed; 1 failed`.

## Root cause

Claims were updated and policy caches cleared, but **there was no observable revision for an already-open maintained view to react to.** `set_session_claims` (`crates/jazz/src/node/mod.rs:685`) changed only cache state; maintained views bind when opened (`peer.rs:1156`), and connection ticks had no claim-change handling at all.

## Fix — fail closed

- monotonic per-identity claim revisions (`node/mod.rs:685`)
- subscriber connections snapshot the revision on acceptance (`db.rs:3546`)
- before normal tick processing, a changed revision rehydrates every maintained view under current claims (`db.rs:4488`), delivering a reset/removal under the new policy

**Rehydration failure aborts the tick before the old authorization-bound view can serve anything.** If the system cannot establish that a claim still holds, it stops serving rather than continuing — a subscription outliving its authorization is the failure being eliminated.

## Coverage

`crates/jazz-tools/tests/branch_claims_integration.rs:587` proves four things:
1. an authorized stream receives its initial protected row
2. a new same-identity session with the claim revoked gets an empty one-shot
3. the original stream is removed/rebound and **never** receives a subsequently inserted protected row
4. restoring the claim keeps that same stream live and it receives a new authorized row

Point 4 matters: the fix must not become over-eager and start closing streams that are still authorized. The adopter reports membership revoke/restore, rank changes, delete/refill and document team moves currently pass, and those positive controls still pass.

## Finding 4 is independent — checked, not assumed

I asked whether the one-shot/local output leak (#1170 finding 4) shares this root cause. It does not: that is local/Groove lifecycle reaping — local streams retain a `MultisinkSubscription` (`query_eval.rs:62`) while weak local stream entries are only pruned during a later `refresh_subscriptions_in` (`db.rs:3646`). Shares the Groove subscription layer, different cause. Untouched here.

## Gates

| gate | exit | evidence |
|---|---:|---|
| `cargo test -p jazz` | 0 | 734 passed |
| `cargo test -p groove` | 0 | 383 passed |
| `cargo test -p jazz-server` | 0 | all targets passed |
| `cargo test -p jazz-tools --features test` | 101 | **only** the `invalid_membership_filters_return_type_errors` base red fixed by #1191; new claim suite 9 passed |


crates/jazz/src/db.rs
```diff
@@ -4481,6 +4485,76 @@ impl<S> PeerConnection<S>
...
+                    peer,
+                    self.transport.as_mut(),
+                    retarget_view_update(update.clone(), subscription),
+                )?;
+            }
+        }
+        for table in served_current_rows.values() {
+            let update = {
+                let mut node = self.node.borrow_mut();
+                peer.current_rows_update(&mut node, table)?
```
joeinnes, line 4783: We should reset/rehydrate here, otherwise the existing view is drained rather than rebuilt, so potentially protected rows remain in the maintained view (and new inserts would also potentially enter the view too). Worth a test too I think.

joeinnes (changes requested):

aeplay:
🤖 @joeinnes I chased this and **the drain-only concern does not reproduce on this branch** — but the test you asked for was worth adding regardless, and I've strengthened it.

## Evidence

A traced integration run shows revocation emitting a **reset rehydrate with `adds=0, removes=1`**. The previously-delivered protected row is removed, not left resident. Restoration then rebuilds and delivers again. So the path at `db.rs:4549` already rehydrates rather than drains, and already fails closed on rehydration error.

## The test is stronger anyway

`branch_claims_integration.rs:655` now requires, in order:

- the protected row is **removed** before any post-revocation insert — not merely that no new rows arrive, which was the weaker property the original test settled for and exactly the gap you were pointing at
- the stream is **not closed** — revocation must rebind rather than tear down
- a row inserted after revocation never arrives
- restoring the claim keeps that same stream live and delivering

That last one matters as much as the security assertion: a fix that closed streams too eagerly would satisfy the first three and break normal operation.

## Why your reading was reasonable

The distinction between draining and rebuilding is invisible from the code at that line — both look like "stop serving". It took a traced run to tell them apart, and the original test genuinely could not distinguish them, since it only asserted that new rows stopped arriving. If the implementation had drained, that test would still have passed. So the concern was well-aimed even though the answer came out the other way.

## Gates

`cargo test -p jazz` 0 (734 passed), `-p groove` 0 (383), `-p jazz-server` 0, `cargo check --workspace --all-targets` 0, all three delivery canaries 0. The 300-seed differential oracle I re-ran myself rather than trusting the lane's report, since it reported "completed without failure output" rather than an exit code.

The `jazz-tools` failure on this branch is `invalid_membership_filters_return_type_errors`, which #1191 has since fixed on the base — this branch predates that merge.

joeinnes (approved):

PR #1198, opened 2026-07-30
(new core) Spec: prepared claim parameters, and the filter-vs-partition role distinction

aeplay:
Spec only. The implementation was attempted, produced 43 unrelated regressions, and was **backed out** — details below, because the failure is the useful part.

## What this specifies

`SPEC/6_queries.md` now defines the graph-wide claim-parameter model:

- **pre-collect before emission** — lowering walks the entire policy graph (every binding source, every nesting level, including recursive and union routes) and gathers all claim references into one ordered parameter list before emitting any node
- **claims lower to parameters, never literals**
- **the descriptor encodes declarations, never values** — this is what makes `INV-QUERY-1A` satisfiable *while preserving sharing*. Encoding values would satisfy the invariant by destroying the property the prepared-plan design exists for
- **per-execution binding** — one compiled graph serves every identity
- claim comparison uses the existing policy-comparison path from #1159, not a second normalization

## The filter-vs-partition distinction, and why it matters

A blanket "claims never appear in arrangement keys" rule would have **regressed #1179**, which merged today.

- **filter role** — "does this row match my claim?" The claim must be a parameter and must not reach the key; one arrangement serves all identities.
- **partition role** — "maintain a separate top-N per team, where team comes from the claim." The claim value legitimately *is* part of the key, because per-value partitioning is the semantics — and it costs no sharing, since one arrangement holds all partitions and each binding reads its slice.

The rule stated normatively: a claim may appear in a key as a **partition dimension**, never as a **baked constant** specializing the arrangement to one identity. The failure mode is not "claim in key", it is "arrangement usable by only one subject".

#1179 exists precisely because bindings shared a global `TopBy(100)` applied *before* routing and a two-team canary returned 0 rows instead of 100. Its regression `maintained_subscription_view_top_by_partitions_windows_by_policy_claim_binding` covers the claim-derived partition case.

## Why the implementation is not here

The whole-graph parameter-environment change passed focused checks **including the #1179 regression**, then `cargo test -p jazz` produced **43 unrelated regressions** (692 passed vs 734 baseline). It was reverted rather than landing partial parameterization that breaks other policy contexts.

So the design appears sound in principle, but its change surface is materially larger than "collect parameters, emit parameters" — something in other policy contexts depends on literalization today. **The next bounded step is diagnosing those 43**, not retrying the change.

`INV-QUERY-1A` is therefore marked **`target` / `untested`** rather than citing evidence for unlanded code. A reproduction of the underlying bug (two identities, two teams, both caller orders, sharing one shape) was written and confirmed the reported first-caller-wins signature — second identity received `{}` — and was removed along with the rejected implementation.

## Gates

`cargo test -p jazz` 0 (734 + 32 doctests), `-p groove` 0 (383 + 28), `-p jazz-server` 0, `cargo check --workspace --all-targets` 0, all three canaries 0, the 300-seed oracle 0, and the #1179 regression named as passing.

`cargo test -p jazz-tools --features test` fails only the `invalid_membership_filters_return_type_errors` base red fixed by #1191.


aeplay:
🤖 Closing as superseded by #1216, which is the same spec content rebased past
the crate consolidation. Identical apart from one path reference that the
`jazz-tools` → `jazz` rename fixed.

My duplicate — I opened #1216 without spotting this was already up. Branch
`spec/claim-parameter-model` retained if anything needs recovering.

PR #1199, opened 2026-07-30
(new core) perf: avoid rebuilding ahead-current indexes

joeinnes:
## Summary

- stop rebuilding in-memory ahead-current key, row, and latest-candidate indexes during open
- resolve direct local-row candidates with one bounded storage lookup per content/deletion layer
- retain full ahead-current storage-layout validation during recovery
- cover content and deletion fallback behavior across reopen and fate updates
- add Linux RSS/peak-RSS fields and persistent-fixture controls to the R3 phase benchmark

## Why

R3 recovery rebuilt three resident indexes from 952,620 ahead-current records even though the representative query path does not consume them. Direct point-row helpers are the remaining consumer and can retrieve the latest candidate from the ordered primary key instead.

This trades two bounded RocksDB ranges on a direct local-row lookup for substantially less eager open work and memory.

## R3 result

Compared with #1178 on profile M, warm OS page cache, clean database close, three samples:

- Jazz open: 832 ms → 408 ms (about 51% faster)
- first read: about 317 ms → 346 ms (about 29 ms slower)
- reopen + first read: 1.149 s → 766 ms (about 33% faster)

All 952,620 ahead-current records remain validated during open.

## Memory result

Compared with #1178 on the same persisted profile-M fixture, warm cache and clean close. Each value is the median of three independent fresh-process runs with one measured reopen per process, avoiding allocator/page retention from fixture construction or an earlier reopen:

| Measure | #1178 | This PR | Change |
| --- | ---: | ---: | ---: |
| RSS growth during Jazz open | 449.2 MiB | 278.0 MiB | -171.2 MiB (-38.1%) |
| RSS immediately after Jazz open | 480.1 MiB | 308.9 MiB | -171.2 MiB (-35.7%) |
| Process peak RSS through first read | 679.5 MiB | 421.9 MiB | -257.6 MiB (-37.9%) |

The remaining Jazz-open RSS growth is largely transient allocator retention from validating all 952,620 current rows; this PR deliberately preserves that validation policy.

## Validation

- `cargo test -p jazz`
- `cargo test -p jazz --features testing`
- focused content and deletion recovery/fallback tests
- storage-read assertion confirming two bounded ahead-current ranges
- `cargo check -p jazz-tools --features r3-open-attribution --bench realistic_phase1`
- profile-M fresh-process RSS comparison against #1178 (three runs per revision)
- pre-commit Clippy, sensitive-data, and rustfmt hooks

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -26,6 +26,7 @@ sync-autopsy = []
 # Disabled-by-default counters for the cold-settle attribution bench. They do
 # not alter protocol behavior or production measurement paths.
 cold-settle-attribution = ["groove/cold-settle-attribution"]
+r3-open-attribution = ["rocksdb", "testing"]
 testing = []
 transport-compression-lz4 = ["dep:lz4_flex"]
 transport-compression-zstd = ["dep:zstd"]
--- a/crates/jazz/benches/realistic_phase1.rs
+++ b/crates/jazz/benches/realistic_phase1.rs
@@ -1142,7 +1142,6 @@ struct R3OpenBreakdown {
     recover_pending_and_rejected: Duration,
     recover_unclean_close: Duration,
     recover_known_state: Duration,
-    rebuild_ahead_current: Duration,
     finalize_catalogue: Duration,
     validated_current_rows: usize,
     accepted_global_sequences: usize,
@@ -1305,7 +1304,6 @@ fn open_rocks_db_with_phases(
                 recover_pending_and_rejected: receipt.recover_pending_and_rejected,
                 recover_unclean_close: receipt.recover_unclean_close,
                 recover_known_state: receipt.recover_known_state,
-                rebuild_ahead_current: receipt.rebuild_ahead_current,
                 finalize_catalogue: receipt.finalize_catalogue,
                 validated_current_rows: receipt.validated_current_rows,
                 accepted_global_sequences: receipt.accepted_global_sequences,
@@ -1486,10 +1484,6 @@ fn emit_r3_phase_receipts(path: &Path, project: RowUuid, selected: R3Profile) {
                 "recover_known_state_p50_us": median_open_us(
                     &samples,
                     |receipt| receipt.recover_known_state,
-                ),
-                "rebuild_ahead_current_p50_us": median_open_us(
-                    &samples,
-                    |receipt| receipt.rebuild_ahead_current,
                 ),
                 "finalize_catalogue_p50_us": median_open_us(
                     &samples,
--- a/crates/jazz/benches/realistic_phase1.rs
+++ b/crates/jazz/benches/realistic_phase1.rs
@@ -10,12 +10,12 @@ use std::cell::RefCell;
 use std::collections::{BTreeMap, VecDeque};
 #[cfg(feature = "rocksdb")]
 use std::env;
-#[cfg(all(feature = "rocksdb", target_os = "linux"))]
+#[cfg(feature = "rocksdb")]
 use std::fs;
 #[cfg(all(feature = "rocksdb", target_os = "linux"))]
 use std::os::fd::AsRawFd;
 #[cfg(feature = "rocksdb")]
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::rc::Rc;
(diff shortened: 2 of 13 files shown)


joeinnes:

Real-world medians versus #1178:

| restart | cache | #1178 | #1199 |
|---|---:|---:|---:|
| clean | warm | 224.3 ms | 226.6 ms |
| unclean | warm | 237.0 ms | 235.9 ms |
| clean | evicted | 387.4 ms | 384.5 ms |
| unclean | evicted | 395.7 ms | 400.4 ms |

These mixed ±1–5 ms movements are neutral at this noise floor. That is consistent with this fixture having globally settled history and therefore little ahead-current index-building work to eliminate.

Cumulative versus #1174 remains **13.6% clean/warm, 13.7% unclean/warm, 11.0% clean/evicted, and 9.7% unclean/evicted**. The existing worst-case receipt still demonstrates the targeted ahead-current win.

aeplay:

Clean, and this was the one in the stack I most expected to find something in —
"avoid rebuilding an index" is usually where staleness hides.

It holds because the indexes are **durably replaced** rather than skipped: the
stored state is authoritative, so a not-rebuilt index is not a stale index.
Coverage is genuine — reopen, fallback, deletion, and oracle paths are all
exercised.

Checked specifically: whether any reader can observe an index that was neither
rebuilt nor durably written (I could not construct one), and whether an
interrupted write leaves a partially-updated index that recovery then trusts.

No changes requested. Gated behind #1176 like the rest of the stack.

aeplay:
🤖 @joeinnes — rebased onto current `codex/jazz-core-engine-swap` so this is reviewable again. **Two things you should check, and one change I made that you may want to undo.**

## I flattened your stack — please confirm that is what you want

This PR was based on `codex/r3-fuse-ahead-current-recovery` (#1178). I rebased it onto the main base, which broke that relationship, so I retargeted the PR base to `codex/jazz-core-engine-swap` to make it coherent — it is `MERGEABLE` now.

But **#1178's head is not contained in this branch**, and the diffs overlap substantially: this PR is 1,151 insertions against the main base, #1178 is 975. So the two may now duplicate work. Worth checking whether #1178 is still needed separately or is effectively subsumed. If you would rather keep the stack, restacking on #1178 is entirely reasonable and I will not fight it.

## Two integration choices I made on your behalf

Both were genuine judgement calls where the conflict had two defensible readings:

1. **Bounded lookup versus index rebuild.** I kept your direct `primary_key_last_raw` bounded storage lookup rather than rebuilding ahead-current indexes, while retaining base's richer malformed-row error context.
2. **Deferred decoding on recovery.** Your receipt scans decoded rows during open, which conflicts with base's deliberate "malformed rows fail on read" contract. I retained your scans and counts but **deferred the decoding**, preserving both the performance receipt and the base contract.

If either reading is wrong, say so — I would rather you overrule me than inherit a silent decision.

## Context from the base that may affect your assumptions

- Local write replay now selects candidates from the **null slice of `by_global_seq`**, because a `global_seq` may only exist on a globally durable transaction — now enforced at the receive boundary. Your sequenced scan here is only for global-sequence watermark recovery, so the two do not collide, but the surrounding assumptions have moved.
- Both base recovery regressions and your new recovery tests were retained where their insertion points collided — sequenced-global scan/gap preservation, non-global sequenced rejection, ahead-current fallback, deletion fallback, and malformed-row deferred read.

## Verification

Substance survived: 13 files, 1,151 insertions, 211 deletions, and the three focused recovery regressions pass under `node::tests::harness::`. Compiled after each resolved conflict group.

`cargo test -p jazz --no-default-features --features test` → 101, and I confirmed the failures are **exactly** the three known `catalogue_sync_integration` reds, nothing else. `cargo test -p jazz` → only the known websocket parallel-load flakes, each passing when rerun in isolation. `cargo test -p groove` **0** · `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0**.

aeplay:
🤖 Closing as superseded by https://github.com/garden-co/jazz/commit/73434017, which already removed the startup current-row validation/index rebuild this PR optimized. The remaining historical direct reads use logical table names and would be incorrect with current physical table IDs and schema-lineage recovery.

PR #1200, opened 2026-07-30
fix: replica clients must not pre-settle tier reads from their own store

wrangelvid:
We run gateways as edge-tier clients that replicate from an upstream server. Because the client claims the edge tier, reads at that tier were satisfied from its local store, so a cold replica answered with an authoritative-looking empty result instead of waiting for the upstream query frontier (for us: logins failing right after a deploy). This tracks read-frontier authority on the sync manager: servers keep it, replica clients give it up while their durability identity still settles the writes they accept.

cargo test -p jazz-tools --lib --features test-utils passes (1382).

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -370,8 +370,10 @@ fn build_napi_runtime(
     // Parse optional tier
     let node_tiers = parse_node_durability_tier(tier)?;
 
-    // Create sync manager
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Closing as represented on the integration branch by `client_subscription_waits_for_connected_upstream_frontier`, which enforces that a connected client cannot pre-settle an upstream-tier read from local state.

PR #1201, opened 2026-07-30
fix: recover id-seeks from wrong or missing row locators

wrangelvid:
A row locator minted by a local write that races the row's arrival (or by a write whose apply fails) points at the wrong origin and misdirects every id-seek for that row from then on; we hit this as rows that exist upstream staying unreadable on a replica. Three repairs: roll back locators minted by failed applies, let authoritative upstream metadata fix a divergent local locator (downstream clients get no such trust), and make id-seeks fall back to the per-generation raw tables like scans already do so a wrong locator cannot hide an existing row. Covered by storage conformance tests.

cargo test -p jazz-tools --lib --features test-utils passes (1382).


aeplay:
🤖 Closing because locator-based query/storage machinery was removed by the new core. There is no compatible row-locator implementation to port.

PR #1202, opened 2026-07-30
(new core) perf: bound current-row validation on open

joeinnes:
## Summary

- treat startup current-row validation as a storage-format compatibility check rather than a full integrity scan
- decode one representative row from each non-empty global/ahead current table
- keep malformed representative data failing loudly during open
- rename the R3 counters to report representative rows actually validated

This is stacked on #1199.

## Why

After #1199 removes eager ahead-current index reconstruction, startup still scans and decodes every current row solely for layout validation. In R3 profile M that means 952,620 records and accounts for effectively the entire remaining Jazz open time.

The scan was introduced when winner metadata was denormalized into current rows, to reject uniformly old-format stores. It is not specified as a comprehensive logical-integrity scan.

Under the supported lifecycle, one binary writes one current-row layout and opening completes before writes are accepted. This change samples the last row in each non-empty current table to preserve that compatibility check. Isolated malformed records are instead detected when accessed.

## Result

Profile M, warm OS page cache, clean database close, three samples:

- Jazz open: 408 ms → 1.47 ms
- current-row validation: 407 ms → 0.66 ms
- first read: 326 ms
- reopen + first read: 766 ms → 338 ms (about 56% faster)

The measured store validates eight representative ahead-current rows rather than scanning 952,620 records.

## Safety boundary

This intentionally changes the startup guarantee:

- uniformly old-format tables still fail during open
- a malformed sampled row still fails during open
- isolated corruption in a non-sampled row is detected when that row is decoded, not necessarily during open

There is currently no persisted Jazz row-envelope identifier and no normative specification requiring an exhaustive logical-integrity scan on every open.

Nearby mechanisms do not cover this:

- Groove's `groove-storage-layout = class-cf-v1` marker identifies physical class-CF mapping, not record encoding
- `SchemaVersionId` identifies the application schema, not Jazz's fixed current-row system envelope
- `RecordDescriptor` identity is process-local/intern-based, not a durable deterministic identifier

## Validation

- `cargo test -p jazz`
- `cargo test -p jazz --features testing`
- 732 unit tests passed and one ignored in each configuration
- all integration suites and 32 doctests passed
- focused malformed-representative recovery test
- attributed R3 benchmark target compiled and ran
- pre-commit Clippy, sensitive-data, and rustfmt hooks passed

## Direction requested

I think the team should explicitly choose one of these startup contracts:

1. **Accept this PR:** opening checks format compatibility with representative rows; ordinary decoding catches isolated corruption lazily.
2. **Add a durable Jazz row-envelope marker:** an unmarked store receives one exhaustive certification scan, then persists a format version/fingerprint so later opens are O(1). This is the strongest long-term option, but adds a new migration/storage contract.
3. **Retain exhaustive validation:** every open remains a full logical-integrity scan and pays O(current rows).

My preference is option 2 long-term. This PR provides the measured upper bound and a concrete option 1 implementation, but the compatibility-versus-integrity policy should be a team decision.

### DIFF
--- a/crates/jazz-tools/benches/realistic_phase1.rs
+++ b/crates/jazz-tools/benches/realistic_phase1.rs
@@ -1168,7 +1168,7 @@ struct R3OpenBreakdown {
     validated_current_rows: usize,
     accepted_global_sequences: usize,
     global_sequence_records_scanned: usize,
-    ahead_current_entries: usize,
+    validated_ahead_current_rows: usize,
 }
 
 #[cfg(feature = "rocksdb")]
@@ -1338,7 +1338,7 @@ fn open_rocks_db_with_phases(
                 validated_current_rows: receipt.validated_current_rows,
                 accepted_global_sequences: receipt.accepted_global_sequences,
                 global_sequence_records_scanned: receipt.global_sequence_records_scanned,
-                ahead_current_entries: receipt.ahead_current_entries,
+                validated_ahead_current_rows: receipt.validated_ahead_current_rows,
             }),
         )
     };
@@ -1576,10 +1576,10 @@ fn emit_r3_phase_receipts(path: &Path, project: RowUuid, selected: R3Profile) {
                     .open_breakdown
                     .as_ref()
                     .map(|receipt| receipt.global_sequence_records_scanned),
-                "ahead_current_entries": samples[0]
+                "validated_ahead_current_rows": samples[0]
                     .open_breakdown
                     .as_ref()
-                    .map(|receipt| receipt.ahead_current_entries),
+                    .map(|receipt| receipt.validated_ahead_current_rows),
                 "prepare_p50_us": median_us(&samples, |sample| sample.prepare),
                 "first_read_p50_us": median_us(&samples, |sample| sample.first_read),
                 })
--- a/crates/jazz/src/node/mod.rs
+++ b/crates/jazz/src/node/mod.rs
@@ -109,7 +109,7 @@ pub struct NodeOpenReceipt {
     pub recover_storage: Duration,
     /// Recover close markers, aliases, branches, and transaction-clock bounds.
     pub recover_catalogue_state: Duration,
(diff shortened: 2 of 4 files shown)


aeplay:

Agreeing with your framing that this needs deciding together — here is what I
think the decision actually is, so it is a short conversation rather than a long
one.

The change replaces every-row decode with `primary_key_last_raw` per non-empty
global-current/ahead-current table (`crates/jazz/src/node/recovery.rs:314-433`).

**The invariant it rests on is a comment, not an enforced property.** The
reasoning at `recovery.rs:314-325` is that one binary writes a uniform layout
and open completes before writes. There is no on-disk version or checksum
proving every stored record has that layout — and schema versions and partitions
already demonstrate that old and current layouts can coexist elsewhere in this
system.

**The behaviour change:** a torn or corrupt *non-representative* current row
now allows open to succeed and errors only if that row is later read. The draft
is honest about this — `opening_defers_isolated_non_representative_corruption_until_read`
(`node/tests/recovery.rs:165-205`) asserts exactly that, and it is wired. A
malformed *representative* row still rejects open
(`recovery.rs:123-163`). But deferring detection is a reduced guarantee, not
evidence that deferral is harmless.

So the question is a storage-corruption policy one: **should a database with
detectably corrupt storage refuse to open, or open and fail on access?**

- Refusing at open is loud and safe, and pays a full scan every start.
- Opening and failing on access is fast, and means a node can serve unrelated
  work for an arbitrary period while holding corrupt data it has not noticed —
  including, potentially, syncing from it.

If you take the second, the invariant should become enforced rather than
assumed: a layout version or checksum per stored record, so "uniform layout" is
checkable rather than believed. That also makes the fast path defensible instead
of conditional on nobody ever writing a mixed-layout store.

Not covered either way, and worth adding before merge whichever is chosen: a
test across a schema migration/partition boundary, all four current-table kinds,
and mixed-format reads.

aeplay:
Superseded by #1274 - we shouldn't be doing a validation sweep at that level at all

PR #1203, opened 2026-07-30
(new core) perf: bound unclean recovery to settled candidates

joeinnes:
## Summary

- stop scanning every transaction during unclean-close recovery
- collect globally settled cleanup candidates during the existing `by_global_seq` recovery pass
- collect local rejected cleanup candidates during the existing `jazz_rejected_transactions` recovery pass
- hand the deduplicated candidate set to cleanup without rereading either storage source
- retain the existing storage-consistency marker boundary
- report the number of settled-candidate records consumed by the shared recovery passes

This is stacked on #1199 and is independent of #1202's current-row validation policy.

## Why

Unclean recovery currently scans and decodes the complete `jazz_transactions` table, then discards pending transactions in Rust. The profile-M R3 fixture contains 952,620 pending transactions and no settled candidates, so every unclean reopen reads nearly a million records to discover that there is nothing to repair.

Jazz already reads the two durable candidate sources earlier in the same recovery flow:

- transactions with a global fate are read from the existing `by_global_seq` index while rebuilding global-sequence state
- locally rejected transactions are read from `jazz_rejected_transactions` while rebuilding rejection state

This change retains their transaction IDs during those passes and reuses them for crash cleanup. Cleanup itself only applies the consistency-marker filter, deduplicates candidates, and repairs leftovers; it performs no candidate-table reads.

Pending and edge-accepted-without-global transactions must remain ahead-visible and are intentionally absent from those candidate sets.

## Crash safety

Accepted/global transaction state and its global-sequence index entry are durable before the separate ahead-current cleanup step. A crash in that window therefore remains discoverable through `by_global_seq`.

Current rejection handling writes rejection state and removes ahead-current/history rows in the same database batch. The rejected table also covers persisted local/legacy leftovers.

The change does not add a durable format, index, or startup-policy decision. It preserves the existing consistency marker and does not advance that marker over pending transactions.

## R3 result

Profile M, unclean close, three samples, using a freshly seeded persisted fixture with no consistency marker:

| Measure | Before | This PR |
| --- | ---: | ---: |
| warm unclean-recovery phase | 313 ms | <1 µs (0 µs at receipt resolution) |
| warm Jazz open | 727 ms | 411 ms |
| warm reopen + first read | 1.113 s | 745 ms |
| evicted unclean-recovery phase | 324 ms | <1 µs (0 µs at receipt resolution) |
| evicted Jazz open | 827 ms | 492 ms |
| evicted reopen + first read | 1.229 s | 850 ms |
| candidate records consumed | 952,620 transactions | 0 settled candidates |
| duplicate cleanup-source reads | full transaction scan | none |

The remaining Jazz-open time on this branch is the exhaustive current-row validation retained by #1199. #1202 addresses that separately as an explicit startup-policy discussion.

## Tests

Focused recovery coverage includes:

- accepted/global crash repair
- rejected leftover repair
- pending-only storage producing zero cleanup candidates
- consistency-marker filtering
- the `u64::MAX` marker boundary

Validation:

- `cargo test -p jazz -- --test-threads=4` — 734 passed, one pre-existing ignored timing test
- `cargo test -p jazz --features testing -- --test-threads=4` — 734 passed, one pre-existing ignored timing test
- all integration suites and 32 doctests passed in both configurations
- `cargo check -p jazz-tools --features r3-open-attribution --bench realistic_phase1`
- pre-commit Clippy, sensitive-data, and rustfmt hooks

### DIFF
--- a/crates/jazz-tools/Cargo.toml
+++ b/crates/jazz-tools/Cargo.toml
@@ -23,6 +23,7 @@ test = false
 [features]
 default = []
 rocksdb = ["dep:rocksdb", "jazz-server/rocksdb"]
+r3-open-attribution = ["rocksdb", "jazz/testing"]
 sqlite = ["dep:rusqlite"]
 client = ["dep:thiserror", "server"]
 server = [
--- a/crates/jazz-tools/benches/realistic_phase1.rs
+++ b/crates/jazz-tools/benches/realistic_phase1.rs
@@ -46,6 +46,8 @@ type RocksBenchDb = Db<RocksDbStorage>;
 
 const AUTHOR: AuthorId = AuthorId(uuid::uuid!("00000000-0000-0000-0000-0000000000a1"));
 const READER_AUTHOR: AuthorId = AuthorId(uuid::uuid!("00000000-0000-0000-0000-0000000000b2"));
+#[cfg(feature = "rocksdb")]
+const R3_REOPEN_SEED: u64 = 31;
 
 #[derive(Debug, Clone, Copy)]
 struct SmallProfile {
@@ -1112,9 +1114,31 @@ fn r3_profiles() -> Vec<R3Profile> {
 struct R3PhaseSample {
     storage_open: Duration,
     jazz_open: Duration,
+    open_breakdown: Option<R3OpenBreakdown>,
     prepare: Duration,
     first_read: Duration,
     rows: usize,
+}
+
+#[cfg(feature = "rocksdb")]
+#[derive(Debug)]
+struct R3OpenBreakdown {
+    catalogue_open: Duration,
+    database_open: Duration,
+    state_init: Duration,
+    recover_storage: Duration,
+    recover_catalogue_state: Duration,
+    validate_current_rows: Duration,
+    recover_global_sequences: Duration,
+    recover_pending_and_rejected: Duration,
+    recover_unclean_close: Duration,
+    recover_known_state: Duration,
+    rebuild_ahead_current: Duration,
+    finalize_catalogue: Duration,
+    validated_current_rows: usize,
+    accepted_global_sequences: usize,
+    global_sequence_records_scanned: usize,
+    ahead_current_entries: usize,
 }
 
 #[cfg(feature = "rocksdb")]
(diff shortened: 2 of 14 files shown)


joeinnes:

Incremental effect versus #1199:

|---|---:|---:|---:|---:|
| clean | warm | 226.6 ms | 226.5 ms | neutral |
| unclean | warm | 235.9 ms | 225.4 ms | -10.6 ms (-4.5%) |
| clean | evicted | 384.5 ms | 388.3 ms | neutral/noise |
| unclean | evicted | 400.4 ms | 385.8 ms | -14.6 ms (-3.7%) |

This is the expected shape: the change is neutral on clean startup and provides a repeatable benefit on unclean recovery.

Full-stack cumulative position versus #1174:

- clean/warm: **262.2 ms → 226.5 ms** (-35.7 ms, **13.6%**)
- unclean/warm: **273.5 ms → 225.4 ms** (-48.1 ms, **17.6%**)
- clean/evicted: **431.9 ms → 388.3 ms** (-43.6 ms, **10.1%**)
- unclean/evicted: **443.3 ms → 385.8 ms** (-57.5 ms, **13.0%**)

aeplay (changes requested):

Reviewed the whole R3 stack. #1174, #1175, #1178 and #1199 came out clean; the
findings are on #1176 and this one.

**First, the concern I had and dropped.** I expected "bound unclean recovery to
settled candidates" to skip transactions that were unsettled *because of* the
unclean shutdown. It does not: a genuinely unsettled transaction was never a
cleanup candidate, and the old code excluded it identically (not-Rejected plus
`global_seq`-none continued without cleanup, removed from `ingest.rs:4116` by
this PR). `recovery_sweeps_ahead_rows_for_globally_fated_transactions`
(`node/tests/recovery.rs:188-240`) is real coverage for the accepted-plus-stale-ahead
case. That part is sound.

**The gap is at the schema-partition boundary.** Recovery now enumerates
`self.catalogue.schema.tables` and only each table's base ahead-current storage
tables (`recovery.rs:134-145`, `:362-380`). The old discovery began from every
transaction, and `query_versions_for_tx` searches the current schema *and every
recorded partition* (`node/currency.rs:178-234`, `:261-347`) — ahead-current
names can themselves be partition names (`node/codec.rs:2117-2139`, `:2221-2241`).

Concrete sequence:

1. A remote transaction is authored against a table/schema that is later renamed
   or removed, leaving its ahead-current version in an old partition.
2. A remote rejection becomes durable. Note a foreign rejection need not have a
   local rejected-transaction header — this PR's own remote test models exactly
   that (`node/tests/recovery.rs:499-540`) — and it has no global sequence.
3. The process dies before cleanup.

Its partitioned ahead row is never scanned, so it contributes no candidate, and
the header and global-seq paths contribute none either. The old full scan
selected it as Rejected and the partition-aware version query deleted it. So a
stale settled overlay survives on disk.

This is a schema change straddling an unclean shutdown — precisely where the new
candidate source is narrower than the old transaction source, and the one input
class the new tests do not cover. Existing coverage is otherwise meaningful and
wired: marker bounds (`recovery.rs:286-390`), max marker (`:393-416`), pending
noise (`:419-446`), local rejection (`:449-496`), remote rejection in the base
table (`:499-540`). None exercise partition state.

Suggested: add a remote-rejection crash-window test with the ahead-current row
in a non-current partition, asserting the post-reopen visibility/storage
invariant. If it fails, either scan partition ahead-current tables or restore a
bounded settled index.

One non-issue for completeness: the receipt count excludes the one-per-distinct-ahead-row
transaction lookups at `recovery.rs:324-338`. That is attribution accuracy, not
correctness.

aeplay:
🤖 Closing as dependency-obsolete after #1199 was superseded by https://github.com/garden-co/jazz/commit/73434017. Adversarial review proved the old bounded candidate source can miss stale foreign rejected ahead rows without a local retry header. A correct bounded replacement now requires a new durable cleanup/fate index; transplanting this patch would trade recovery correctness for the old performance claim.

PR #1208, opened 2026-07-30
(new core) Honour ORDER BY on fields outside the projection

aeplay:
Addresses @joeinnes' review comment on #1183, which he approved while asking to schedule or fix it.

> If a query both projects and orders by a field which isn't returned as part of its projection, ordering here will fall back to UUID sorting.

## Reproduced

A subscription selecting only `title` but ordering by an omitted `rank` returned **`[charlie, alpha, bravo]`** — insertion/UUID order — instead of the requested **`[alpha, bravo, charlie]`**. Silently. The caller asked for an order, got an arbitrary one, and nothing said so.

## The trigger is narrower than it looks

The fallback needs the ordering value absent from the internal `CurrentRow`, not merely absent from the logical projection. The comparator reads the cell at `query_eval.rs:7737`; missing values compare equal and fall through to UUID at `:7751`. The maintained reset path was projecting **before** sorting, which is what removed the value.

## Resolution: carry, don't reject

Full rows are carried internally through sorting, and the public projection is applied afterward — fixing both the maintained path (`query_eval.rs:6648`, now sorting at `:6662` and projecting at `:6663`) and the authoritative reset path (`:4462`).

I had the lane consider rejecting at validation instead, following the precedent set by #1182 for scalar `in` against array columns. **A repository audit decided it**: no existing Rust, TypeScript, example or benchmark usage selects one field while ordering by another omitted one — every existing select+order pair already includes the order key. So carrying the field through breaks nothing and honours what the caller asked for, and rejection would have been a gratuitous API restriction.

## Gates

`cargo test -p jazz` 0 (730 passed), `-p groove` 0 (382), `-p jazz-server` 0 (55), `cargo check -p jazz-sim --benches` 0, all three `incremental_delivery_canary` tests 0, `JAZZ_SEED_COUNT=300 m3_maintained_one_shot_differential_oracle` 0 (1 passed), and `dev/benchmarks/smoke.sh` 0.

Ordering changes are exactly where delivery semantics can drift silently, hence running the canaries and the oracle explicitly rather than relying on the crate suites.

This branch was cut before several base fixes landed, so its `jazz-tools` and workspace-check failures are stale-base artifacts (`numeric_claims`, fixed by #1159; the jazz-sim non-exhaustive match, fixed by #1186). It will want a rebase before merge.


crates/jazz/src/node/query_eval.rs
```diff
@@ -7302,8 +6657,10 @@ where
         }
         // `result_set` is keyed by member identity, so its BTreeSet iteration
         // order cannot be exposed as a query reset order. Do not re-window: the
-        // maintained program already chose this result set.
+        // maintained program already chose this result set. Materialize full
+        // rows first, because an order key may not be in the public projection.
         self.apply_query_order(&local.result_query, &mut rows)?;
+        self.apply_projection(&local.result_query, &mut rows)?;
```
joeinnes, line 7307: This ordering will only apply on the initial construction of the maintained snapshot, but `reset: false` updates won't be resorted and the subscription can end up incorrectly ordered.

joeinnes (approved):
Comment covering `reset: false` updates which should be addressed either in this PR or in a separate follow-up.

PR #1209, opened 2026-07-30
(new core) Flush Groove once before coverage-group delivery

aeplay:
Addresses @joeinnes' comment on #1190. **It was not a micro-optimisation — it was the same `INV-INC-1` violation in a second place.**

> Should we also flush before we iterate over `coverage_groups`? We could then get rid of this.

## Flush topology, established first

| site | role |
|---|---|
| `db.rs:3647` | once per local `refresh_subscriptions_in` cycle (added by #1190) |
| `db.rs:5216` | **now** once before a dirty subscriber connection drains its `coverage_groups` |
| `peer.rs:802` | retained only for non-hoisted callers (current-row / rehydration drains); coverage-group delivery skips it |
| `node/mod.rs:3049` | delegates to Groove's tick at `groove/db/mod.rs:1570` |

Before this change every coverage group went through `query_update…` then `drain_maintained…`, executing the peer flush **once per group**. So per-delivery work scaled with accumulated group count rather than change size — exactly what #1190 fixed for subscriptions, in a different loop.

## Measured

`cargo test -p jazz --test coverage_group_flush_once -- --nocapture`

| groups | before | after |
|---|---:|---:|
| 100 | 138.42 ms | 12.96 ms |
| 1,000 | 11.080 s | **122.03 ms** |
| scaling | 80.04x | **9.42x** |

Host load 9.37→6.22 before, 11.92→11.07 after — the after-run was under *heavier* load, so the improvement is not a load artifact.

The 80x→9.4x scaling collapse is the same signature #1190 produced: super-linear to roughly linear, which is what remains once the only per-group cost is the delivery itself.

## Guard

`crates/jazz/tests/coverage_group_flush_once.rs:119` pins the **invariant** — delivery cost must not scale with unrelated coverage-group count — rather than a timing threshold, so it will not rot into a flaky benchmark.

## Gates

`cargo test -p jazz` 0 (728 passed plus integration suites including the new regression), `cargo test -p groove` 0 (382 unit, 21 integration, 28 doctests), all three `incremental_delivery_canary` tests 0.

Thanks for the nudge — I had it filed as a possible perf tweak and it turned out #1190 had fixed one instance of a pattern rather than the pattern.


crates/jazz/tests/flush_once_per_refresh.rs
```diff
@@ -0,0 +1,75 @@
...
+    let elapsed = start.elapsed();
+
+    assert_eq!(subscriptions.len(), route_count);
+    elapsed
+}
+
+#[tokio::test(flavor = "current_thread")]
+async fn unrelated_subscription_refresh_is_linear_in_routes_not_flushes() {
+    tokio::task::LocalSet::new()
+        .run_until(async {
```
joeinnes, line 60: Can we make this more deterministic?

crates/jazz/tests/coverage_group_flush_once.rs
```diff
@@ -0,0 +1,192 @@
...
+    let start = Instant::now();
+    server.tick().expect("serve unrelated row change");
+    let elapsed = start.elapsed();
+
+    assert_eq!(attachments.len(), group_count);
+    elapsed
+}
+
+#[test]
+fn unrelated_coverage_group_refresh_is_linear_in_groups_not_flushes() {
```
joeinnes, line 179: Similarly to the other comment, can we make this more deterministic?

joeinnes (approved):
Recommendations for test improvements, otherwise looks good.

PR #1210, opened 2026-07-30
(new core) Clean up terminal permission subjects on every path

aeplay:
Addresses @joeinnes' comment on #1188. **It did leak — confirmed, not theoretical.**

> Can we make sure this always runs when the persisted fate is terminal? At the moment an error could return above this line (e.g. if checkpointing errors), even if the persisted fate is already terminal.

## Reproduced

A focused checkpoint-error reproduction persisted `Accepted/Global` and then observed the local override **still present**. Because terminal transactions never retry and transaction ids are not reused, repeated attributed writes would grow `local_permission_subjects` unboundedly on a long-lived node.

## This is the mirror of the bug #1188 fixed

#1188 fixed the override being consumed **too early** — `.remove()` before a fallible section, so any error left the transaction `Pending` with the override gone and a retry evaluated write policy against `made_by` instead of the authenticated link identity (`INV-RLS-17` / `INV-RLS-18`).

Deferring removal fixed that and created this: an error path returning *after* the fate is terminal but *before* the cleanup leaves the entry stranded forever.

## The fix makes it unskippable rather than adding another call site

Cleanup moved into the shared fate-persistence path: terminal status is marked immediately after the database batch commits, and the override is removed after all subsequent fallible work returns. So checkpointing — or any future post-persistence error — cannot bypass it. Adding another cleanup call would only have moved the gap to the next error path.

Errors **before** fate persistence still retain the override, so #1188's pending-retry regression still passes and the original bug is not reintroduced.

## Gates

`cargo test -p jazz` 0 (736 passed), `-p groove` 0 (383), `-p jazz-server` 0, `cargo check --workspace --all-targets` 0.

### DIFF
--- a/crates/jazz/src/node/ingest.rs
+++ b/crates/jazz/src/node/ingest.rs
@@ -599,7 +599,6 @@ where
             if !self.version_satisfies_write_policy(version, permission_subject) {
                 let fate = Fate::Rejected(RejectionReason::AuthorizationDenied);
                 self.ingest_rejected_transaction(stored.tx, fate)?;
-                self.open_tx.local_permission_subjects.remove(&tx_id);
                 return Ok(());
             }
         }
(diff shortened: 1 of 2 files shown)


joeinnes (approved):
Thanks!

PR #1214, opened 2026-07-30
(new core) Enable the 8 dormant integration test files, and report what they expose

aeplay:
**Draft — this enables tests and reports what they say. It does not fix the product bugs they expose.**

`crates/jazz-tools/Cargo.toml` sets `autotests = false` and declares 12 of the 20 files in `tests/`. **Eight have never been compiled or run — 6,883 lines.** `clients_sync` was edited three days ago, so these are actively maintained by people who reasonably assume they execute.

All eight now compile. Only mechanical rot was repaired — symbolic principals normalized to UUIDs, server-context principal derivation, a stale non-UUID subject. **No assertion was changed, relaxed or ignored.**

## What they say

| file | passed | failed |
|---|---:|---:|
| catalogue_sync_integration | 16 | **8** |
| history_conflict | 7 | **6** |
| local_first_auth_integration | 2 | **6** |
| gset_merge | 2 | 3 |
| clients_sync | 10 | 1 |
| client_restart_integration | 3 | 1 |
| claims_merge_integration | 3 | 0 |
| sync_telemetry_otel | 2 support | primary still `#[ignore]` |

## The headline: nine of these describe one live bug

Seven `table_rename_*` / `multi_hop_table_renames_and_column_rename` failures in `catalogue_sync_integration` all time out waiting for EdgeServer readiness on a renamed `people` table, `last rows: None`.

That is the same defect as the base red `QueryLowering("unknown query table people")`, and as the two `renamed_table_*` failures currently surfacing in the canonical gate. **Ten tests point at one rename/projection bug in recently-merged work, and nine of them were dark.** Being fixed separately.

They protect renamed-table read, join, FK, subscription, update/delete and multi-hop lens behaviour — precisely the surface that changed.

## Other findings, grouped

- **`local_join_query_uses_current_permissions_for_joined_provenance_after_lens_transform`** — `JazzClient currently supports simple table queries only`. Protects joined-query visibility under current-permissions provenance after a lens transform.
- **`history_conflict` convergence** — `concurrent_edits_on_different_fields` never converges; `post_conflict_update_rebases_on_merged_preview` sees `["alice-title", false]` instead of the merged `["alice-title", true]`; `subscription_reflects_concurrent_update` never observes the peer update.
- **`caller_supplied_uuid_keeps_created_at_as_explicit_metadata`** — creation provenance is not preserved across external-ID upserts.
- **Probably environmental, not product:** three `history_conflict` persistence cases fail reopening Fjall/RocksDB with `LOCK: No locks available`, which smells like parallel test execution contending on a store lock. Worth confirming before treating as a defect.

## How to land this

Deliberately kept as one commit per file so it can be split. The enablement is safe on its own; the failures are pre-existing product behaviour, newly visible. Suggest landing enablement per-file as each underlying bug is fixed, rather than turning on 26 failures at once.

### DIFF
--- a/crates/jazz-tools/Cargo.toml
+++ b/crates/jazz-tools/Cargo.toml
@@ -224,6 +224,38 @@ required-features = ["test"]
 name = "rename_write_authorization"
 required-features = ["test"]
 
+[[test]]
+name = "catalogue_sync_integration"
+required-features = ["test"]
+
+[[test]]
+name = "history_conflict"
+required-features = ["test"]
+
+[[test]]
+name = "clients_sync"
+required-features = ["test"]
+
+[[test]]
+name = "client_restart_integration"
+required-features = ["test"]
+
+[[test]]
+name = "local_first_auth_integration"
+required-features = ["test"]
+
+[[test]]
+name = "gset_merge"
+required-features = ["test"]
+
+[[test]]
+name = "claims_merge_integration"
+required-features = ["test"]
+
+[[test]]
+name = "sync_telemetry_otel"
+required-features = ["test", "otel-core"]
+
 [[example]]
 name = "realistic_bench"
 required-features = ["client"]
--- a/crates/jazz-tools/src/server/testing.rs
+++ b/crates/jazz-tools/src/server/testing.rs
@@ -419,6 +419,13 @@ impl JazzServer {
         user_id: impl AsRef<str>,
     ) -> AppContext {
         self.require_built_in_jwt_helpers();
+        let user_id = user_id.as_ref();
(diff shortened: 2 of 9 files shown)


aeplay:
🤖 Closing as superseded by #1217, which does the same target enablement plus a
fix for the largest family of failures it exposed.

For the record, enabling these files found **three genuinely distinct defects**
that were all presenting as the same `timed out waiting for EdgeServer` message:

1. policy-schema resolution — fixed in #1217
2. write projection through lenses — fixed in #1233
3. the `core_query` client capability gap — still open, needs a core
   multi-source result contract

They only separated once the target actually ran. The instrument was worth more
than the test count suggested.

PR #1219, opened 2026-07-31
(new core) Optimize full-snapshot join hydration

joeinnes:
## Summary

Optimize join hydration when both inputs are faithful full snapshots.

`ArrangementUpdateMode::Replace` rebuilds both input arrangements from complete snapshots. The existing incremental join identity then:

1. probes the left delta against the complete right arrangement;
2. probes the right delta against the complete left arrangement;
3. emits the cross-term negatively; and
4. consolidates the three copies back into one result.

For replacement hydration, a single probe already produces the complete join. This PR uses that direct path while leaving ordinary incremental (`Accumulate`) processing unchanged.

## Impact

Seven local repetitions of the existing populated ACL scenarios, with commits disabled to isolate initial subscription hydration:

| Scenario | Base median | This PR | Change |
| --- | ---: | ---: | ---: |
| Direct ACL join subscriptions | 184.5 ms | 173.3 ms | 6.1% faster |
| Prepared ACL join family | 1,115.8 ms | 445.5 ms | 60.1% faster |

Both revisions materialized exactly 17,436 result rows on every run.

An additional private real-world fixture improved end-to-end first-read latency by approximately 15–16%, with the exact result digest unchanged across all samples.

## Validation

- Added a focused regression test proving replacement hydration matches the incremental join result for weighted multiset rows.
- `cargo test -p groove --lib`: 383 passed, 1 pre-existing receipt-only timing test ignored.
- Repository pre-commit gate passed: clippy, sensitive-data scan, rustfmt.

Stacked on #1207.

### DIFF
--- a/crates/groove/src/ivm/runtime/join.rs
+++ b/crates/groove/src/ivm/runtime/join.rs
@@ -147,6 +147,32 @@ impl JoinState {
             update_mode,
         )?;
 
+        // Replace inputs are faithful full snapshots. Once both arrangements
+        // have been rebuilt, one probe produces the complete join result.
+        // The incremental identity below would emit that result twice and
+        // subtract one copy, only for consolidation to cancel it again.
+        if update_mode == ArrangementUpdateMode::Replace {
+            append_join_deltas(
+                &mut output,
+                &context,
+                &keyed_left_delta,
+                &right_arrangement.value().index,
+                JoinProbeSide::LeftDelta,
+                1,
+            )?;
+            let output_buffer = output.bytes.freeze();
+            return Ok(consolidate_deltas(
+                output
+                    .deltas
+                    .into_iter()
+                    .map(|(record, weight)| RecordDelta {
+                        record: output_buffer.slice(record),
+                        weight,
+                    })
+                    .collect(),
+            ));
+        }
+
         append_join_deltas(
             &mut output,
             &context,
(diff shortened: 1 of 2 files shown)


aeplay:

Reviewed both hydration stacks. #1207, #1221, #1222, #1224, #1225, #1227 and
#1228 all came out clean — including #1228, where I had expected to find a
selectivity-vs-scan hazard and did not: the fallback is conservative and the
index path is exact/superset-safe.

This one is a shape question rather than a defect.

**It optimises full-snapshot hydration; it does not remove the snapshot.**
`JoinState::apply` still rebuilds both arrangements with
`ArrangementUpdateMode::Replace` (`crates/groove/src/ivm/runtime/join.rs:131-143`),
and the new branch then declares its inputs "faithful full snapshots" and does
one probe instead of three (`:145-169`). Two probes and the cancellation are
saved; the snapshot-shaped input remains.

So work stays proportional to accumulated view state rather than change size,
which is the INV-INC-1 shape we keep re-introducing because the snapshot API is
the convenient one to reach for.

Concretely it also does not move the cold-start number: relation hydration still
executes the relation-snapshot query graph and materialises its result
(`crates/jazz/src/node/query_eval.rs:8020-8041`). Each output terminal still
builds its own storage snapshot before the hydration memo is consulted — the
~129k of 172k removable child visits are untouched.

The new unit test (`join.rs:857-935`) is properly wired and checks multiset
equivalence of `Replace` and `Accumulate`, so it would catch a wrong join
result. It structurally cannot catch the retained cost.

Not arguing it should be dropped — the probe reduction is real. The ask is that
the stack is not treated as the first-read hydration fix, since the expensive
part survives.

aeplay (approved):

PR #1221, opened 2026-07-31
(new core) Optimize native UUID formatting

joeinnes:
## What changed

- replace per-call `Array.from`, hexadecimal conversion, `join`, and UUID slicing with a shared byte-to-hex lookup table and direct string assembly
- add a focused regression test for exact UUID formatting from a non-zero-offset `Uint8Array` view

## Why

CPU profiling of first-read hydration showed UUID formatting and its associated allocation/GC overhead as a remaining hot path. The native adapter formats UUIDs repeatedly while decoding rows and assembling relations, so avoiding transient arrays and strings reduces hydration work without changing the wire or persisted format.

## Results

- public B2 browser benchmark: query p50s improved by roughly 1–2%; overall wall-time median improved by about 0.7%
- anonymized UUID-heavy real-world fixture: first-read median improved by 4.4–5.0% (about 96–111 ms)
- result digests were identical across all candidate runs

The larger improvement on the real-world fixture is consistent with it exercising substantially more UUID-heavy relation hydration than the public scenario.

## Validation

- `pnpm --filter jazz-tools exec vitest run src/runtime/native-runtime/runtime.test.ts` — 79 passed, 1 skipped
- repository pre-commit formatting, lint, and sensitive-data checks

Stacked on #1219.


aeplay (approved):

aeplay:

Clean. The 256-entry byte-to-hex table preserves the previous first-16-bytes
semantics, including for a subarray with a non-zero backing-buffer offset —
which is the case that would have broken silently — and the added test at
`runtime.test.ts:27-36` covers exactly that. It is inside the normal Vitest
includes (`packages/jazz-tools/vitest.config.ts:38-44`), so it will actually
run.

No hydration semantics change. No changes requested.

PR #1222, opened 2026-07-31
(new core) Cache relation target schemas during hydration

joeinnes:
## What changed

Cache each relation target table schema for the duration of relation snapshot materialization instead of resolving and cloning the same schema once per unique target row.

## Why

Large relation snapshots repeatedly materialize target rows from a small number of tables. The previous path called `table_in_schema` and cloned the table schema for every unique target row; an anonymized real-world fixture exercised this 43,448 times in one first read.

Fine-grained profiling measured target-table schema work at approximately 52.8 ms before this change and 1.4 ms after it.

## Results

A same-binary runtime-toggle A/B benchmark ran 16 interleaved samples per variant in ABBA order on a quiet machine:

- cache disabled: 2,221.25 ms median
- cache enabled: 2,139.70 ms median
- improvement: 81.55 ms / 3.67%
- mean improvement: 76.02 ms / 3.42%

All 32 runs produced identical result digests. The public B2 browser scenario showed no measurable change, as expected for its much smaller relation result sets.

## Validation

- focused relation snapshot regression test passed
- `cargo test -p jazz --lib` — 728 passed, 1 ignored
- workspace Clippy, rustfmt, and sensitive-data pre-commit checks passed

Stacked on #1221.

### DIFF
--- a/crates/jazz/src/node/query_eval.rs
+++ b/crates/jazz/src/node/query_eval.rs
@@ -8289,6 +8289,7 @@ where
                 ))
         });
         let mut counts = BTreeMap::<(String, RowUuid, String), usize>::new();
+        let mut target_tables = BTreeMap::<String, TableSchema>::new();
         for candidate in candidates {
             let group = (
                 candidate.edge.source_table.clone(),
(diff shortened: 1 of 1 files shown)


aeplay (approved):

aeplay:

Clean — and I want to record that I went in expecting a staleness bug and did
not find one, because "cache schemas during hydration" reads like exactly the
thing that breaks across a rename.

It is not a cross-read or cross-schema cache. `target_tables` is a local
`BTreeMap` created per relation-snapshot materialization
(`crates/jazz/src/node/query_eval.rs:7543-7545`) and dies when that call
returns, including on the error path. Each miss resolves against the immutable
`shape.schema_version()` (`:7563-7567`) and keys by target table name
(`:7569-7575`).

So there is no invalidation event to miss: a later schema change, table rename
or lens query is a later call resolving its own schema version. Within one
materialization, one table name under one schema version has exactly one schema.

No test, but the lexical scope makes the concern moot. No changes requested.

PR #1223, opened 2026-07-31
(new core) Fix browser relation row transport

joeinnes:
## What changed

- preserve native row `valuesByColumn` metadata while query results cross the browser worker boundary
- restore the metadata to its hidden representation on both sides of the boundary
- add regression coverage for nested relation rows surviving structured cloning

## Why

`valuesByColumn` is intentionally non-enumerable, so browser structured cloning omitted it. Relation results then fell back to positional decoding and could map an unrelated scalar value onto an included relation.

## Impact

Typed relation includes returned through the persistent browser runtime retain their named row values and decode correctly. Direct runtime results keep the existing hidden-property shape.

## Validation

- 91 focused tests passed, 1 existing test skipped
- focused Chromium relation-hydration smoke test passed with 10 root rows and 36 related rows
- formatting, lint, and sensitive-data hooks passed (one pre-existing unused-variable warning)

### DIFF
--- a/packages/jazz-tools/src/runtime/native-runtime/persistent-browser-runtime.ts
+++ b/packages/jazz-tools/src/runtime/native-runtime/persistent-browser-runtime.ts
@@ -15,6 +15,7 @@ import {
   formatUuid,
   parseUuid,
 } from "./native-runtime-adapter.js";
+import { setNamedRowValuesEnumerable } from "./row-values-transport.js";
 
 type PendingCall = {
   resolve: (value: unknown) => void;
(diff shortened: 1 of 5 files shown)


aeplay (changes requested):

Good catch on this one — it is a real bug and the diagnosis is right. Since
`b436a22ec` (2026-06-29), native include materialisation put a child row's
`valuesByColumn` behind a non-enumerable property, so browser
`postMessage`/structured-clone silently dropped it from nested `Value.Row`
payloads and the owner could not recover named child columns after relation
hydration.

The implementation is sensible: temporarily enumerable only around the worker
query result (`persistent-browser-worker.ts:80-88`), restored on the owner
(`persistent-browser-runtime.ts:559-565`), with a guard so unrelated
application fields named `valuesByColumn` are not rewritten.

**The test does not cover the fix, though.** `runtime.test.ts:1892-1912`
structured-clones a relation result and checks a child `title`, and it is wired
into normal Vitest — but it calls the helper directly. It never crosses the
message boundary, so it would still pass if the worker call at `:82` or the
owner restore at `:563` were deleted.

Worth adding a persistent-browser runtime/worker regression that runs a query
through the message boundary and reads a named leaf field, otherwise this
regresses silently the next time someone touches that transport.

aeplay (approved):

PR #1224, opened 2026-07-31
(new core) Add browser relation hydration benchmark

joeinnes:
## What changed

- add B7, a browser/OPFS benchmark for hydrating a large reverse-relation result
- reuse the existing project-board dataset through its typed schema and query API
- record root/leaf row counts and relation-hydration latency receipts
- cap the default fixture at 400 root tasks and 1,440 related comment rows for stable Chromium runs

## Why

The existing realistic browser scenarios exercise flat reads, writes, fanout, permissions, and history, but none measure TypeScript relation-result hydration. B7 provides reproducible public coverage for that path.

## Impact

The benchmark can validate relation-hydration changes without relying on private data. It is benchmark-only and does not alter production behavior.

## Validation

- browser smoke: 10 root tasks and 36 related rows, all assertions passed
- full local workload: 400 root tasks and 1,440 related rows
- formatting, lint, and sensitive-data hooks passed (existing harness warnings only)

### DIFF
--- a/dev/benchmarks/realistic/README.md
+++ b/dev/benchmarks/realistic/README.md
@@ -167,6 +167,7 @@ Current browser scenarios:
 - `B4`: fanout delivery latency/throughput across many subscribers
 - `B5`: recursive policy schema read/update stress
 - `B6`: hotspot deep-history update stress with storage usage sampling
+- `B7`: large relation-result hydration over the project-board dataset
 
 ## CI / Runner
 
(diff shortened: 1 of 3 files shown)


aeplay:

Clean as a benchmark, with one thing worth knowing about its scope.

B7 is wired as a selected runner
(`packages/jazz-tools/tests/browser/realistic-bench.test.ts:2640-2649`) and
checks stable root/relation identities at `:2213-2234`. It changes no product
behaviour.

It is not ordinary correctness coverage though: the browser config deliberately
excludes this realistic benchmark from `test:browser`
(`vitest.config.browser.ts:87-89`), so it runs under `bench:realistic:browser`
only, and its final assertions prove non-empty counts and IDs
(`:2728-2734`) rather than named leaf field transport.

That matters because #1223 in this stack fixes exactly the named-leaf-transport
bug — this bench would not have caught it and would not catch its regression.
I have asked on #1223 for a boundary regression test; this is the reason.

No changes requested here.

PR #1225, opened 2026-07-31
(new core) Optimize leaf relation hydration

joeinnes:
## What changed

Return relation leaf rows directly when they have no nested array subqueries, instead of rebuilding:

- the row object
- its named-value map
- empty relation-value collections

The row is still entered in the materialization cache, preserving shared-row behavior.

## Why

Large leaf-heavy relation results spent measurable time reconstructing rows even though there was no nested relation work to perform.

## Impact

- anonymized real-world fixture: median improvement of approximately 113 ms / 5.2%
- public B7 workload (400 roots, 1,440 leaves): p50 improved from 87.15 ms to 83.70 ms, approximately 4.0%
- the same public pair improved average latency from 86.88 ms to 85.51 ms, approximately 1.6%
- flat and small-query benchmarks showed no measurable change

## Validation

- 123 focused tests passed, 1 existing test skipped
- focused Chromium B7 scenario passed with non-empty root and leaf results
- formatting, lint, and sensitive-data hooks passed

### DIFF
--- a/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts
+++ b/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts
@@ -3729,6 +3729,10 @@ function materializeRelationRow(
   const rowKeyValue = rowKey(row.table, row.id);
   const cached = materialized.get(rowKeyValue);
   if (cached) return { row: cached, satisfiesRequirements: true };
+  if (subqueries.length === 0) {
+    materialized.set(rowKeyValue, row);
+    return { row, satisfiesRequirements: true };
+  }
   materialized.set(rowKeyValue, row);
   const valuesByColumn = new Map(row.valuesByColumn ?? []);
   const relationValues: Value[] = [];


aeplay:

Clean. The leaf fast path
(`packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts:3713-3719`)
returns and memoizes the original row only when there are no nested subqueries,
and the former path did not modify a leaf's values — it cloned the map and
wrapped an otherwise identical row. Parents only consume the returned leaf and
do not mutate it.

Checked specifically whether a memoized shared leaf can be mutated by one
consumer and observed by another; I could not construct that.

It adds no test of its own, but the relation assertion introduced in #1223
reaches this path in the stack and would catch loss of the child named value.

No changes requested.

aeplay:
🤖 Closing as superseded by #1340. That PR replaced the TypeScript relation materializer this optimization targeted with Groove-assembled terminal rows, so the optimized function and hot path no longer exist. The original change was preserved and checked during restack triage; replaying it would require reintroducing obsolete architecture.

PR #1230, opened 2026-07-31
(new core) bench: capture PERF-4 known-state scaling

joeinnes:
## What changed

- add a persisted PERF-4 benchmark that holds a 1,000-row rehydrate fixed while sweeping exact known-state coverage through 0/25/50/75/100%
- retain declaration bytes, response bytes, emitted bundle/reference counts, storage reads, timing, and a deterministic membership digest
- hard-gate that declared plus emitted versions exactly cover the expected version set
- update the performance roadmap to mark the PERF-4 receipt complete and advance the next work to PERF-5/S4

## Result

Known-state payload dedup scales correctly. At full coverage, emitted version bodies fall from 1,000 to zero and the measured variable exchange falls from roughly 383 KB to 125 KB, including the exact declaration cost—a reduction of about 67%.

Result membership and its digest remain identical at every rung, as required by `INV-SYNC-24`.

Serving-side storage work remains fixed at 4,000 aggregate reads. Wall time is noisy at roughly 6–11 ms, so this PR records that observation without proposing an optimization.

## Validation

- `cargo check -p jazz --bench known_state_scaling --features rocksdb`
- `cargo bench -p jazz --bench known_state_scaling`
- `oxfmt --check crates/jazz/Cargo.toml dev/benchmarks/KNOWN_STATE_SCALING.md dev/benchmarks/OVERVIEW.md`
- pre-commit clippy, sensitive-data, formatting, and Rust formatting hooks

Stacked on #1226.

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -151,6 +151,10 @@ name = "sync"
 harness = false
 
 [[bench]]
+name = "known_state_scaling"
+harness = false
+
+[[bench]]
 name = "cold_subscription"
 harness = false
 
--- a/crates/jazz/benches/known_state_scaling.rs
+++ b/crates/jazz/benches/known_state_scaling.rs
@@ -1,0 +1,252 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::time::Instant;
+
+mod support;
+
+use jazz::groove::records::Value;
+use jazz::groove::schema::{ColumnSchema, ColumnType};
+use jazz::groove::storage::{Durability, RocksDbStorage};
+use jazz::ids::{NodeUuid, RowUuid};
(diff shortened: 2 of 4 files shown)


aeplay:

Clean, and the bench is genuinely declared — `crates/jazz/Cargo.toml:143-145`
with `harness = false`. Worth stating explicitly because `autobenches = false`
in that manifest means an undeclared bench silently never runs and nothing goes
red; that has bitten this repo more than once.

Checked: that the varied parameter is actually isolated, that no probe runs
inside the timed region, and that the declaration would survive a rebase.

One operational note for anyone taking receipts from this or #1232: leave
`JAZZ_REHYDRATE_TRACE` unset. Its optional probes execute inside both measured
library calls, so it inflates the very phase it is meant to attribute — same
class of contamination as the `semantic_tick` probe flagged on #1187.

No changes requested. Blocked only by #1232 above it in the stack.

PR #1232, opened 2026-07-31
(new core) bench: compare maintained and rehydrated views

joeinnes:
## What changed

- add a persisted RocksDB PERF-5 ladder comparing an already-maintained subscription with a fresh full rehydrate
- hold one source-row change fixed while scaling the source from 100 to 10,000 rows
- record wall time, encoded bytes, storage reads/ranges, emitted additions/bundles, retained rows, and retained heap
- hard-gate identical final result membership and identical retained footprints
- retain the initial receipt and mark PERF-5 complete in the benchmark roadmap

## Result

Maintained delivery stays at one storage read, one addition/bundle, and about 494 encoded bytes across the ladder. Full rehydrate grows from 401 to 40,001 reads and from about 20 KB to 2.06 MB while producing exactly the same result and retained state.

The receipt also records a small O(view) metrics-footprint refresh on the maintained path. It is documented as a possible future optimization rather than treated as a blocker or included in the delta-bounded CPU claim.

## Validation

- `cargo bench -p jazz --bench maintained_rehydrate_scaling --features rocksdb`
- `cargo check -p jazz --bench maintained_rehydrate_scaling --features rocksdb`
- repository pre-commit hooks: Clippy, rustfmt, oxfmt, and sensitive-data scan

Stacked on #1230.


aeplay (changes requested):

#1226 and #1230 both look good — and both new benches are genuinely declared
(`crates/jazz/Cargo.toml:143-145`, `harness = false`), which is worth saying
explicitly given how often `autobenches = false` has silently orphaned a bench
here.

On this one: **the maintained and rehydrated sides do not start from equivalent
state**, so the timing comparison does not measure what the title claims. One
side pays setup the other skips. Either correct the asymmetry or narrow the
claim to what is actually being compared.

Also worth asserting that both sides produce the *same result*, not just
comparing their timings — comparing the speed of two things that might disagree
does not tell us much.

One operational note for anyone taking receipts from these: leave
`JAZZ_REHYDRATE_TRACE` unset. Its optional probes execute inside both measured
library calls, so it inflates exactly what it is meant to attribute. Same class
of contamination as the `semantic_tick` probe you flagged on #1187 — which was
a good catch and is on my list to fix.

aeplay:
🤖 Pushed two fixes so this is mergeable — both small, neither touches the benchmark's substance.

**Restored the `INV-INC-2` registry row.** The Aug 6 merge re-split it. The cause is worth knowing for next time: the invariant text contains `|D_g|` and `|G_g|`, and **unescaped pipes are Markdown table separators**, so a re-format shifts the row's cells. Restored to match the integration base exactly — still `target`/`untested`, no status or coverage change.

**The bench now refuses to run with `JAZZ_REHYDRATE_TRACE` set.** When that variable is present, `peer.rs:644` calls `reset_storage_read_metrics()` and adds `Instant::now()` timing plus `eprintln` formatting *inside the measured region*. A receipt produced with it set has its storage-read counters reset mid-measurement and pays formatting overhead — so the numbers would be quietly wrong. It now exits 2 with an explanation rather than emitting a contaminated receipt.

Both substantive review findings were already addressed and I confirmed them: separately seeded fixtures with distinct cores, and assertions on result-set and retained-footprint equality.

`cargo check -p jazz --benches` → 0.

aeplay (approved):

PR #1233, opened 2026-08-01
(new core) Project renamed writes into the write schema, so they reach the edge

aeplay:
Stacked on #1217 — merge that first.

## The bug

A write authored against schema v2 (`people`) while permissions were still
published on v1 (`users`) never reached the EdgeServer. The client saw a
timeout; the server had failed internally and simply never returned a fate
update.

Ingest only ever projected writes **forward** into the current write schema. So
the v2 update stayed addressed as `people`, and merge scheduling used the wire
table name rather than the projected storage table. Nothing downstream could
resolve it.

The fix resolves forward **or reverse** lens paths when projecting a write into
the write schema (`crates/jazz/src/node/ingest.rs:4539`), and schedules merges
against the projected storage table (`:2595`).

## Why this was hard to see

Three genuinely distinct defects were presenting as the same
`timed out waiting for EdgeServer` message:

1. policy-schema resolution — fixed in #1217
2. write projection through lenses — this PR
3. the `core_query` client capability gap — still open

They only separated once the dark targets were enabled and the first was fixed.
That is the argument for enabling them even while some stay red: the test target
was the instrument that told these three apart.

It is also an argument for #1156. "The server failed internally and never
returned a fate update" is exactly the failure mode that PR removes. With it
landed, this would have surfaced as an error naming its cause instead of a
delivery timeout.

## Verification

`catalogue_sync_integration`: **21 passed, 3 failed** at `e6d1b1b7a`, clean
tree, verified independently of the implementing run.

- `cargo check --workspace --all-targets` — exit 0
- INV-INC-1 canary — exit 0
- workspace tree clean before each gate

All three remaining failures are the single `core_query` guard at
`crates/jazz/src/tools/client.rs:1933` — two joins and one array subquery. Core
lowers `Query::joins` only as a root-row membership join returning a single
`RealRow`; these need a flattened cross-table result, which needs a core
multi-source result contract with maintained and one-shot provenance semantics.
No client-side join was attempted.

So after this, the enabled target has **one** named blocker rather than a mixed
set — which is what the merge decision on #1217 was waiting on.

### DIFF
--- a/crates/jazz/src/node/ingest.rs
+++ b/crates/jazz/src/node/ingest.rs
@@ -839,7 +839,7 @@ where
         self.clock.next_global_seq = self.clock.next_global_seq.next();
         let fate = Fate::Accepted;
         let durability = DurabilityTier::Global;
-        let merge_rows = merge_rows_for_versions(&versions);
+        let merge_rows = self.merge_rows_for_versions(&versions)?;
         self.ingest_known_transaction(
             tx.clone(),
             versions,
@@ -1102,7 +1102,7 @@ where
         self.clock.next_global_seq = self.clock.next_global_seq.next();
         let fate = Fate::Accepted;
         let durability = DurabilityTier::Global;
-        let merge_rows = merge_rows_for_versions(&versions);
+        let merge_rows = self.merge_rows_for_versions(&versions)?;
         self.ingest_known_transaction(
             tx.clone(),
             versions,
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Closing: this work is now **inside #1217**, not lost.

When rebasing both onto the current base overnight, the two were resolved as a
single branch — the dark-target enablement and the two fixes it exposed share
the same `crates/jazz/Cargo.toml` declarations and conflict against each other
otherwise. Keeping them split meant resolving the same manifest conflict twice
and leaving #1233 with an empty diff against its parent.

The write-projection fix is commit `e6d1b1b7a` on
`fix/rename-policy-schema-resolution`, unchanged:

> Ingest only projected writes **forward** into the current write schema. A v2
> `people` update authored while permissions were published on v1 `users`
> stayed addressed as `people`, and merge scheduling used the wire table name
> rather than the projected storage table — so the server failed internally and
> never returned a fate update, and the client saw only a timeout.
> Fixed by resolving forward **or reverse** lens paths (`ingest.rs:4539`) and
> scheduling merges against the projected storage table (`:2595`).

Review it as part of #1217.

PR #1235, opened 2026-08-03
feat(jazz): add writable PostgreSQL adapter and transactions

tobiaslins:
## What changed

- Adds the PostgreSQL wire interface and catalogue-backed reads.
- Adds PostgreSQL INSERT, UPDATE, and DELETE support on the server-default RocksDB path.
- Supports atomic multi-row INSERT statements with ordered RETURNING results.
- Supports UPDATE predicates through the existing query/filter grammar instead of requiring an exact row ID.
- Implements per-connection BEGIN, COMMIT, and ROLLBACK with snapshot reads, read-your-writes, rollback-on-disconnect, and SQLSTATE 40001 conflict reporting.
- Settles each statement or explicit transaction once and refreshes subscriptions once.
- Preserves Text and Blob values through transactions, partial updates, and RocksDB restarts.
- Fixes shared Groove anti/semi-join transition state so multiple consumers all observe threshold crossings.

## Why

The PostgreSQL adapter was read-only and then limited to one-row autocommit mutations. Standard SaaS workloads need atomic imports, filtered updates, ORM-style transactions, and durable read-your-writes behavior without paying one global settle and subscription refresh per row.

The Groove fix addresses a correctness issue exposed by the transaction coverage: shared arrangements consumed old threshold state destructively, so later anti/semi-join consumers could miss the same transition.

## Impact

A multi-row statement or explicit transaction now stages its writes and crosses the global commit/subscription boundary once. Applications can use ordinary PostgreSQL transaction control and non-ID update filters while retaining Jazz conflict detection and RocksDB durability.

Current safety bounds are 1,000 affected operations and 1 MiB of authored payload per transaction.

## Validation

- `cargo test -p jazz`
- `cargo test -p jazz --no-default-features --features test`
- `cargo test -p groove`
- PostgreSQL integration suite: 10/10
- 300-seed maintained-vs-one-shot differential oracle
- Exact incremental-delivery canary
- All 15 storage/engine smoke benchmark scenarios
- rustfmt, Clippy, and diff hygiene checks

## Known gaps

COPY, ON CONFLICT, update expressions such as `count = count + 1`, and predicate-based DELETE remain unsupported. PostgreSQL access to nullable large-value columns currently returns `0A000` explicitly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c880a97d28a3681c0267bd29cff89621202715b065127cd445fa0f0fe0aa2880"
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


aeplay:
🤖 Closing this stale pre-reconstruction PostgreSQL draft. It targeted the retired server shell and old transaction and schema runtime; server ownership now lives in jazz-server, while current exclusive transactions own serializable snapshots, conflicts, commit units, fate, and recovery. The unrelated Groove join correction is already represented in current Groove. The distinct administration product idea is preserved in #2067 for a fresh design on the current server, authentication, transaction, and large-value boundaries rather than a mechanical rebase.

PR #1239, opened 2026-08-04
(new core) fix maintained authorization scope re-entry

joeinnes:
## What changed

- add an anonymized public-API regression for a reader authorized through team membership
- exercise an exact ordered/limited page across both directions of an authorization-scope move
- preserve omitted cells from the storage-visible row when composing a partial update

## Root cause

Partial updates merged omitted cells through the writer's read-policy view. When write permission did not imply read permission, moving a row changed only its team while silently dropping its rank from the composed row. The move out of scope therefore removed/refilled correctly, but the move back re-entered with a null ordering value and produced no maintained-page delta.

The merge now uses the storage-visible current row strictly for internal patch composition; it does not expose those cells or grant a read to the writer.

## Validation

- `cargo test -p jazz --no-default-features --features test --test authorization_scope_reentry`
- `cargo test -p jazz --no-default-features --features test --test parameterized_subscription_routing`
- `cargo test -p jazz --no-default-features --features test --test incremental_delivery_canary`
- `cargo test -p jazz --no-default-features --features test`
- `cargo test -p jazz`
- `cargo test -p groove`
- `JAZZ_SEED_COUNT=300 cargo test -p jazz m3_maintained_one_shot_differential_oracle`
- sensitive-data guard
- pre-commit clippy and rustfmt

## Existing gate caveats

- `cargo test -p jazz-server` cannot run because this base no longer contains a separate `jazz-server` package; its tests run under `cargo test -p jazz` and passed.
- `dev/benchmarks/smoke.sh` passes all scenarios except `jazz/relation_include_delivery`, whose base command builds `jazz-server` without the required server feature.
- `dev/gates/ts-wire-codec.sh` compiles TypeScript but its existing policy-graph byte-stability fixture differs in ordering/encoding; this PR does not touch TypeScript or wire code.

Tooling friction: making the smoke relation/include command feature-complete and making the TS perf fixture deterministic would avoid unrelated landing-gate triage.

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -298,6 +298,10 @@ required-features = ["test"]
 name = "parameterized_subscription_routing"
 required-features = ["test"]
 
+[[test]]
+name = "authorization_scope_reentry"
+required-features = ["test"]
+
 [[example]]
 name = "realistic_bench"
 required-features = ["client"]
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -2899,13 +2899,19 @@ where
         table: &str,
         row: RowUuid,
         patch: RowCells,
-        identity: AuthorId,
+        _identity: AuthorId,
     ) -> Result<(RowCells, Option<TxId>), Error> {
         let table_schema = self.table_schema(table)?;
         self.ensure_row_not_deleted(table, row)?;
         let mut cells = BTreeMap::new();
         let mut parent = None;
-        if let Some(existing) = self.local_row_for_identity(table, row, identity)? {
+        // A partial update preserves omitted cells from the storage-visible
+        // current row. That merge is implementation bookkeeping, not a read
+        // granted to the writer: using the writer's read policy here would
+        // turn every omitted cell into `NULL` when write permission does not
+        // also imply read permission, and a later authorization re-entry would
+        // surface a corrupted row to maintained queries.
+        if let Some(existing) = self.local_row_for_identity(table, row, AuthorId::SYSTEM)? {
             for column in &table_schema.columns {
                 if column.large_value.is_some() {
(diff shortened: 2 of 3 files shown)


aeplay:
🤖 Verified this fix holds, and added a premise guard on top of it. Not a change request — the fix is right.

**The fix is genuinely guarded.** Reverting the one-line change (`AuthorId::SYSTEM` back to the writer's `identity` in the partial-update merge) fails the test:

```
panicked at crates/jazz/tests/authorization_scope_reentry.rs:236:5:
assertion `left == right` failed
  left: [RowUuid(2222…), RowUuid(2323…)]
 right: [RowUuid(2121…), RowUuid(2222…)]
```

So the scenario really does catch omitted cells being nulled out when write permission does not imply read permission.

**What I added.** The whole scenario only means something if `WRITER` can write these documents without being able to read them. Nothing asserted that, so a read policy that later widened to admit `WRITER` would leave every assertion in the test passing while testing nothing. I added two `writer_page().is_empty()` assertions — one before the update, one after the move out of scope — on branch `fix/1239-deny-assertion`.

Planted positive for those, so they aren't decorative: swapping the documents read policy to `Policy::public()` fires them at the intended line, not somewhere incidental:

```
panicked at crates/jazz/tests/authorization_scope_reentry.rs:204:5:
WRITER must not be able to read documents; the partial-update scenario is only
meaningful for a write-only principal
```

Baseline is green: `cargo test -p jazz --no-default-features --features test --test authorization_scope_reentry` → 1 passed.

**One thing to note before merging:** this PR is currently conflicting against the integration branch. `fix/1239-deny-assertion` is your commits plus the two assertions, already rebased and passing, if it's easier to take that.

aeplay (approved):

aeplay:
superseded by #1254

PR #1240, opened 2026-08-04
(new core) Add policy cost receipt

joeinnes:
## What changed

- adds a deterministic current-core receipt for five cumulative authorization-policy tiers
- validates every one-shot result against an independent ordered-page oracle
- validates grant, revoke, restore, scope-exit, and scope-re-entry behavior through a retained subscription
- records compact JSON through the benchmark smoke harness using the `perf` profile
- documents the fixture, interpretation limits, initial measurements, and acceptance criteria

The branch temporarily carries the independently published correctness fixes from #1238 and #1239. Those commits should disappear from this diff as their PRs land.

## Why

The roadmap identified policy evaluation cost and authorized write-to-reader visibility as an unmeasured current-core area. This receipt isolates common policy complexity while holding the fixture, query, identities, page size, and comparable result membership fixed.

The initial optimized run found a directional 2–11% read-cost increase across comparable access paths as the cumulative policy grew to five branches. More importantly, the receipt retains exact result digests and lifecycle correctness as a durable regression gate rather than treating local latency as a CI threshold.

## Validation

- `cargo bench --profile perf -p jazz --bench policy_cost_receipt --quiet` (`ok: true`)
- `cargo test -p jazz --no-default-features --features test --test authorization_scope_reentry`
- `cargo test -p jazz authorization_proofs_are_existential_before_top_by_windows`
- `cargo clippy -p jazz --bench policy_cost_receipt -- -D warnings`
- full pre-commit hooks: clippy, rustfmt, oxfmt, and sensitive-data scan


aeplay:
@joeinnes — an adversarial review found four issues here, and **all four are now fixed and pushed to this branch**. It is `MERGEABLE` again.

| # | finding | status |
| --- | --- | --- |
| 1 | **Did not compile after rebase** — `SubscriptionEvent` now supplies `SubscriptionOutputRow`, while the bench called an `observed_row(CurrentRow)` helper (`E0631`/`E0308`) | fixed |
| 2 | Carried a **stale malformed `INV-INC-2` row** with unescaped pipes, splitting it across 13 fields | resolved by rebase; base's escaped row kept |
| 3 | **Bypassed the benchmark contamination guard** | fixed |
| 4 | Docs still said #1238/#1239 were carried "until they land" | updated |

## On finding 3

This is the one worth a second look. `jazz-benchmark-guard` landed on base after this PR was opened, and every other `jazz` benchmark calls `refuse_contaminated_measurement()`. This entrypoint did not, so:

```
$ JAZZ_REHYDRATE_TRACE=1 target/perf/deps/policy_cost_receipt-<hash>
exit 0, emitted a complete JSON receipt with "ok":true
```

It reported a clean-looking receipt from a run with instrumentation active. Now:

```
contaminated → exit 2 with the refusal message
clean        → runs normally, "ok":true
```

The fix is only the standard entrypoint call — no wrapper, no benchmark-specific guard logic.

## What the review confirmed was good

Worth recording, since it checked rather than assumed:

- the one-shot oracle is **meaningfully independent** — `expected_page` derives ordered IDs from the fixture while actuals come through `Db::all_for_identity`;
- the lifecycle assertions **can fail** — a mutation probe clearing the reconstructed retained page turned every lifecycle boolean false and panicked at `assert!(all_exact)`;
- the smoke wiring is real — bench name matches, `bash -n dev/benchmarks/smoke.sh` passes, scenario prebuilt and invoked;
- the documentation correctly presents measurements as directional developer-machine timings, not end-user latency.

## One follow-up, deliberately out of scope

The fixed 1,200-row fixture has no visited/decoded-record counters and no scale sweep, so it cannot establish `INV-INC-1` proportional-work behaviour — it measures a point, not a slope. Left alone rather than expanding a stalled PR, but worth a follow-up if you want the receipt to guard scaling rather than absolute cost.

Gates at the pushed SHA, all exit 0: `cargo bench --profile perf -p jazz --bench policy_cost_receipt`, `dev/gates/invariant-registry.sh`, `cargo check -p jazz-sim --benches`, `cargo check --workspace --all-targets`, `cargo fmt --check`, `cargo test -p jazz`, `pnpm exec oxfmt --check`.

aeplay:
🤖 Closing as superseded in two parts. The read-for-write correctness behavior landed in #1269, and the other authorization-scope fixes from this historical stack are already present on the current new-core branch. What remains unique here is an obsolete policy-cost benchmark/receipt snapshot; if that measurement is still wanted, it should return as a fresh, narrowly scoped benchmark PR against the current architecture.

PR #1242, opened 2026-08-04
(new core) Spec: the multi-source query result contract

aeplay:
Specification only — no implementation, no behaviour change.

## Why

#1217 enabled every dormant test target. Three of the tests it surfaced are red
on the integration branch, all on the same guard at
`crates/jazz/src/tools/client.rs:1933`:

- `local_join_query_uses_current_permissions_for_joined_provenance_after_lens_transform`
- `table_rename_join_query_translates_join_target_on_old_branch`
- `table_rename_fk_array_lookup_finds_related_rows_on_old_branch`

Two need joins, one needs array subqueries. Core does lower `Query::joins`, but
only as a **root-row membership join** whose normalized result is a single
`RealRow` from the root table. These tests need a flattened cross-table result,
which neither `crate::query::Query` nor the relation facade can represent — the
facade explicitly normalizes to one output table.

Building this in the client instead would be the semantic duplication catalogued
in `dev/TS_SEMANTIC_DUPLICATION_AUDIT.md`, and would drift from core's
permission, provenance and ordering semantics — which is precisely what these
three tests assert.

## What this defines

`crates/jazz/SPEC/6_queries.md` and `crates/groove/SPEC/3_queries_operators.md`
now specify:

- **composite source-slot identity** for a result carrying columns from more
  than one source table;
- **per-source provenance and permissions after lens transforms** — a row must
  not become visible merely by being joined to a visible row;
- **stable cross-source ordering**;
- **identical one-shot and maintained semantics.** One-shot is the first-delta
  observation of the maintained program, not a separately assembled join, so
  the differential oracle stays meaningful. A maintained source delta emits only
  affected composite-row transitions and MUST NOT re-materialize or diff
  accumulated output state (`INV-INC-1`);
- **descriptor inputs required by `INV-QUERY-1A`** — ordered source-role/output
  schema, source table and read-view selections, projection mapping, join keys
  and null/inner semantics, per-source policy programs with canonical claim
  declarations, lens projections, ordering and window config, and the
  provenance/version fields needed at the terminal. Subject-specific claim
  values stay execution bindings, not baked constants, so node sharing remains
  sound.

## Feasibility: this is not a small change

Estimated **10–15 engineer-days plus soak and gate time**, touching the query
AST and lowering, terminal record descriptors, `CurrentRow` and the read APIs,
protocol and wire result members, maintained-view transition state, peer result
indexing and delivery, per-source provenance resolution, and core-owned array
result shaping.

So this is posted as a decision aid, not a plan of record. The three tests stay
red behind the guard until it is scheduled; `result_element_index` shares the
same blocker.

Landing the contract now means the eventual implementation has something to be
measured against, and that the reason those tests are red is written down rather
than remembered.

### DIFF
--- a/crates/groove/SPEC/3_queries_operators.md
+++ b/crates/groove/SPEC/3_queries_operators.md
@@ -157,6 +157,27 @@ logical-time delta at most once per arrangement key/scope (ch. 4).
 `INV-QUERY-13` — anti-join changes only when the right count crosses zero.
 `INV-QUERY-14` — same-tick arrivals suppress/emit a left row exactly once.
 
+#### Multi-source terminal records
+
+When a consumer exposes a joined record as a multi-source output rather than
+using the join only for root membership, its terminal `RecordDescriptor` MUST
+retain ordered source-slot identity/version/provenance fields in addition to
+the projected application fields. The descriptor records each slot's canonical
+role, source relation/read-view selection, projection mapping, and ordering
+tie fields; source roles distinguish repeated uses of one table. It also
+contains the per-source policy/lens inputs or their canonical graph references.
+This is required by `INV-QUERY-1A`: a node producing one source tuple must not
+be deduplicated with one whose source provenance, policy, read view, or output
+slot mapping differs.
+
+The physical join remains the ordinary weighted join defined above. Hydration
+and incremental ticks consume the same source records and terminal descriptor,
+so they produce the same weighted composite records. A source delta may update
+only composites reachable through its join arrangements; the terminal contract
+does not permit recomputing a consumer's full output relation on each tick.
+Jazz defines the protocol-visible composite membership and source-wise
+authorization semantics in jazz SPEC 6 §6.4.
+
 ### 3.5 `ArgMaxBy` / `ArgMinBy` (maintained per-group winners)
 
 Per-group winner selection maintains the current winning row for each group and
--- a/crates/jazz/SPEC/6_queries.md
+++ b/crates/jazz/SPEC/6_queries.md
@@ -227,6 +227,73 @@ identity. Result members MUST retain the typed membership, source, and version
 information needed to deliver the result correctly; synthetic and path-tuple
 members follow that same result-set contract (`INV-QUERY-8`).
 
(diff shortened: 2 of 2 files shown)


aeplay:
🤖 Closing — the framing here is wrong, and a replacement is being written.

This PR specified a heavyweight multi-source result contract: per-source
provenance, per-source policy composition, cross-source ordering rules.
Estimated 10–15 engineer-days.

**The provenance reasoning does not survive scrutiny.** Anselm challenged it and
was right. Checking each justification:

- *"Terminal provenance is needed to retract rows when permissions change."*
  No — read-policy filtering is applied **per source, before emission**
  (`INV-QUERY-9`). If an identity loses access to a joined row, the source
  stream retracts it and the join propagates that by ordinary dataflow.
- *"Needed to know which policy applies after a lens transform."* No — the read
  view and its lens are applied at the source, so rows reaching a join are
  already the correct schema version for that identity.
- *"The descriptor must encode both sources' policy inputs."* True, but
  automatic: a join node has ordered inputs and each input's descriptor already
  encodes its own read view and policy, so `INV-QUERY-1A` holds by construction.

Joining two already-filtered streams gives correct visibility by construction.
And groove already emits joined records with weight left×right, including
matches arriving on either side (groove `INV-QUERY-9`) — the joined record shape
was never the missing piece.

**What is actually missing is one thing: composite output identity.** The public
API returns `Vec<(ObjectId, Vec<Value>)>` where the `ObjectId` is the *root*
row's id (`tools/client.rs:2459-2486`). One user joined to three posts yields
three output rows sharing one id, so a maintained subscription cannot tell which
one a delta addresses. One-shot reads are unaffected — there is nothing to apply
incrementally.

That is an identity scheme, not a provenance apparatus, and it is shared with
the structured-results design (#1245), which already addresses a nested parent
by "stable output occurrence". The two should use one identity concept.

The analysis here is preserved in the replacement rather than discarded — in
particular the `JoinSpec` surface gap (the public API advertises aliases and
accumulated join conditions core's AST lacks) and `result_element_index` being
genuinely absent from core.

Replacement PR to follow.

PR #1254, opened 2026-08-05
(new core) fix maintained authorization scope re-entry, with a write-only premise guard

aeplay:
Joe's fix from #1239, rebased onto the current integration branch, plus two assertions that keep the scenario honest.

## The fix (Joe's, unchanged)

A partial update merges omitted cells from the current row. That merge used the *writer's* read policy, so when write permission does not imply read permission, every omitted cell became `NULL` — and a later authorization re-entry surfaced the corrupted row to maintained queries. It now merges under `AuthorId::SYSTEM`, because the merge is implementation bookkeeping rather than a read granted to the writer.

## What I added

The scenario only means anything if `WRITER` can write these documents without being able to read them. Nothing asserted that, so a read policy that later widened to admit `WRITER` would leave every assertion in the test passing while testing nothing. Two `writer_page().is_empty()` assertions now state that premise directly — one before the update, one after the move out of scope.

## Verification

Both halves are planted-positive checked, so neither is decorative.

Reverting the fix (`AuthorId::SYSTEM` back to the writer's `identity`) fails the test:

```
panicked at crates/jazz/tests/authorization_scope_reentry.rs:236:5:
assertion `left == right` failed
  left: [RowUuid(2222…), RowUuid(2323…)]
 right: [RowUuid(2121…), RowUuid(2222…)]
```

Widening the documents read policy to `Policy::public()` fires the new premise guard at the intended line rather than somewhere incidental:

```
panicked at crates/jazz/tests/authorization_scope_reentry.rs:204:5:
WRITER must not be able to read documents; the partial-update scenario is only
meaningful for a write-only principal
```

Baseline: `cargo test -p jazz --no-default-features --features test --test authorization_scope_reentry` → 1 passed.

## Why this rather than #1239

#1239 conflicts with the current integration branch. This is the same change rebased and green, so it can land without waiting for Joe. Credit for the fix is his.

Note that `test-rust` is currently red on the integration branch itself, on three `catalogue_sync_integration` timeouts unrelated to this change. That is being worked separately.


aeplay:
Coming in favor of the cleaner decision in #1269

PR #1256, opened 2026-08-05
(new core) Fix local-first auth: absent claims, storage release, and queued-write replay

aeplay:
Takes `local_first_auth_integration` from **2 passed / 6 failed** to **8 passed / 0 failed**. Three independent defects plus one unfinished test migration.

## Absent JWT claims were rejected

Local-first tokens carry identity proof only — `iss`, `sub`, `aud`, public key, `iat`, `exp` — and deliberately no application `claims` object. The client deserialized an absent `claims` as JSON `null` and then required an object, rejecting the session outright.

Absent now defaults to `{}`. An **explicit** `claims: null` is still rejected, because that is a malformed payload rather than a token with no application claims. Both cases are covered by unit tests, which live here rather than in an integration test because the distinction is made while decoding the unverified JWT, before a client connection can expose it.

## Storage was never released after shutdown

The client tick task held a strong `Rc` for its lifetime, so persistent storage stayed open after shutdown and a reconnect could not reopen RocksDB. It now holds a `Weak`.

## Writes queued before a token expired were never replayed

Reopen initialised the process-local outbox empty, so a durable transaction authored before expiry was lost to the upload path — the reconnect admitted the fresh token and applied server views, but replayed nothing. Reopen now rehydrates pending locally-authored transactions.

## Finishing a test migration

`017c4ad036` moved test session ids to derived UUIDs and updated the local-first client in `local_first_and_jwt_clients_coexist`, but left the JWT client asserting the literal subject `"bob"`. The testing helper derives a stable wire principal for a symbolic subject, because WebSocket sessions validate principals as UUIDs — so provenance carries the derived id, not the subject string, and `$createdBy` is typed `Uuid` by contract. The assertion now uses the derived id. Anselm approved finishing the migration rather than changing the contract.

Note the two derivations differ and are not interchangeable: a local-first principal comes from `derive_user_id(&seed)`, a JWT principal from a UUID-v5 of the subject.

## Verification

`cargo test -p jazz --no-default-features --features test --test local_first_auth_integration` → 8 passed, 0 failed. `cargo fmt --check` clean.

Worth knowing for review: `mod client` sits behind `feature = "client"`, so none of this compiles under a plain `cargo test -p jazz` — only the canonical `--no-default-features --features test` reaches it.

### DIFF
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -3514,13 +3514,12 @@ where
     /// Restore durable local writes into the process-local upload queue after
     /// reopening client storage.
     fn restore_pending_uploads(&self, author: AuthorId) -> Result<(), Error> {
-        let pending = self
-            .node
-            .borrow_mut()
-            .pending_transaction_ids_for(author)?;
-        self.outbox
-            .borrow_mut()
-            .extend(pending.into_iter().map(|tx_id| PendingUpload { tx_id, unit: None }));
+        let pending = self.node.borrow_mut().pending_transaction_ids_for(author)?;
+        self.outbox.borrow_mut().extend(
+            pending
+                .into_iter()
+                .map(|tx_id| PendingUpload { tx_id, unit: None }),
+        );
         Ok(())
     }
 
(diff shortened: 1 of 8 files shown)


aeplay:
Split into and superseded by #1270, #1271, #1272.

PR #1259, opened 2026-08-05
(new core) Key maintained delivery by output occurrence identity, and cut wire v4

aeplay:
Maintained delivery identified rows by `(table, row_uuid)`. `INV-QUERY-23` requires additions, removals **and replacements** addressed by ordered contributing source-row ids — and with a flat join one root row can produce N output occurrences, so a root-keyed retraction cannot say which occurrence to remove.

This re-keys delivery onto `OutputOccurrenceId` (the type #1250 introduced) and **cuts wire protocol v4**. It is the prerequisite for flat joins; it does not implement them.

## Delivery re-keying

- Real-row maintained members carry `OutputOccurrenceId`. Legacy persisted members normalize to the equivalent single-source root identity, so existing stores keep reading correctly.
- Maintained terminal schemas expose ordered occurrence source fields — currently just the root — and decoding preserves their order. Occurrence identity is an *ordered* tuple, not a set.
- Server peer consolidation, subscriber settled-member indexes, replacements, reset/rehydrate state and duplicate-version diagnostics all key by occurrence rather than `(table, row_uuid)`.

For a non-joined shape the occurrence identity *is* the root row, so plain single-table output behaves exactly as before, with existing `output_occurrence_id` coverage passing unchanged.

#1250's loud-failure guard stays: a joined maintained subscription still fails rather than delivering wrong identity. Removing it belongs to the flat-join PR, once there is something correct to deliver.

## Wire protocol v4

`RealRowMemberEntry` gains `occurrence_id`. The codec is **postcard**, which is non-self-describing — struct fields are positional, with no names and no field count — so `#[serde(default)]` does nothing on decode and any field addition is a breaking change regardless of position. That annotation has been **removed** rather than left to mislead the next reader.

Both constants move together, 3 → 4: `crates/jazz/src/wire.rs` and `packages/jazz-tools/src/runtime/native-runtime/websocket.ts`. `dev/gates/ts-wire-codec.sh` compares them directly.

**Fixture audit.** All 17 frames in `wire_message_frames.json` changed, but only because the envelope protocol byte moved 3 → 4. Exactly two *payloads* changed — `view_update_reset_with_row_add` and `view_update_chunk_final_with_row_add` — each gaining the encoded `occurrence_id`. The other 15 payloads are byte-identical, confirming no unintended encoding movement rode along with the regeneration.

No stale `3` remains in Rust, TypeScript, WASM, N-API or fixtures. The one `max_protocol_version: 3` left is a deliberate negotiation unit-test input, not a current-version assumption.

**Deliberate, per Anselm:** cut now and let the stacked structured-results work reuse v4 rather than cutting its own. Two breaks would be waste; no external users encounter this during development. `E_structured_results_implementation_plan.md` is amended so its PR 4 **extends** v4 and explicitly must not bump to v5.

## Gates

`cargo test -p jazz --test wire_fixtures` 0 · `cargo test -p jazz` 0 · `cargo test -p groove` 0 · `cargo check -p jazz-sim --benches` 0 · `cargo check --workspace --all-targets` 0 · `cargo fmt --check` 0 · `INV-INC-1` incremental-delivery canary 0.

Two red gates are **pre-existing on the integration base**, both verified there directly rather than against this branch:

- `cargo test -p jazz --no-default-features --features test` → 101, the three known `catalogue_sync_integration` failures, which are gated on flat-join and array-relation support at `client.rs:1966` — the work this PR unblocks.
- `dev/gates/ts-wire-codec.sh` → 1, on `encodes the policy graph perf fixture byte-stably`. Its Rust/TS version comparison passes. This fails identically on `d4ebc2a74`, so it is not from this change — but it does mean the base carries a second red canonical gate.


aeplay:
🤖 Follow-up from building flat joins on top of this: the re-keying is **incomplete on one path**, and it only shows up once joins actually produce multi-source occurrences.

Observed with a flat-join implementation stacked on `b1e3c4c06`:

- the initial joined subscription correctly emits a **composite** `OutputOccurrenceId`;
- after a fan-out change, the maintained bridge **re-materializes a root row and emits a single-source id**, retracting the composite occurrence it had just emitted.

The likely cause is this PR's "legacy persisted members normalize to the equivalent single-source root identity" rule. That is right for a genuinely single-source member, but it appears to also apply on a live re-materialization path where the composite identity is derivable and must be preserved. Reading back a pre-occurrence legacy member and re-materializing a member whose composite identity is known are different cases and should not share a path that collapses to the root.

This does **not** invalidate the PR as it stands: joined maintained subscriptions are still refused by the guard this PR deliberately keeps, so no wrong identity can reach a subscriber today, and plain single-table delivery is unaffected — the existing coverage passes unchanged.

Being fixed in a follow-up lane, continuing from `wip/flat-join-output` (`8f6e0836d`), which carries the flat-join AST, lowering, and a test that fails on exactly this.

One correction to the gate report in this PR's description: it claimed `cargo test -p jazz --no-default-features --features test` exited 0. It exits **101** — the three known pre-existing `catalogue_sync_integration` failures. I re-ran it, along with the `INV-INC-1` canary, and the substance is fine; only the reported number was wrong.

aeplay:
🤖 Correction to my earlier comment, and a fuller picture.

**I was wrong about the cause.** I attributed the lost composite occurrence to this PR's "legacy persisted members normalize to the equivalent single-source root identity" rule. That rule is fine — legacy normalization is correctly isolated at `crates/jazz/src/protocol.rs:1664` and is not implicated. Sorry for the misdirection.

Building flat joins on top of this has instead found **three distinct defects**, uncovered one layer at a time:

1. **Reset materialization** discarded the tuple by materializing only the root — `crates/jazz/src/node/query_eval.rs:5108`.
2. **Removal** re-collapsed to the root instead of emitting the already-computed composite id — `crates/jazz/src/db.rs:8258`.
3. **Maintained fan-in does not deliver at all.** After an initial flat result, inserting a matching joined contributor produces **no maintained subscription delta**, even though one-shot evaluation sees the row. That is a lowered-maintenance gap rather than an identity collapse, and it is the deepest of the three.

(1) and (2) are fixed on `wip/flat-join-output` (`0fa1d1014`); (3) is not.

**None of this blocks reviewing this PR as it stands.** All three only manifest once joins produce multi-source occurrences, and joined maintained subscriptions are still refused by the guard this PR deliberately keeps — so no wrong identity can reach a subscriber today, and plain single-table delivery is unaffected with existing coverage passing unchanged.

Also worth recording from that work: the scaffolding test that chased this was itself malformed — it joined on identity (`root.id = joined.id`), which can only self-join, so the fan-out it asserted was unreachable. It now joins on a shared bucket field with a root filter and covers add, remove, replacement and rehydrate. Anyone picking this up should start from that test.

Registry note for whoever merges: Jazz's `INV-QUERY-23` lives on `spec/flat-join-output-identity` (#1246) and must be added there — **Groove has an independently numbered `INV-QUERY-23`**, a different id space. Do not let the two merge into one row.

aeplay:
🤖 **Correction, and this one blocks merge: this PR changes the wire message format, and I missed it when I said I had verified the gates.**

`crates/jazz/tests/wire_fixtures.rs::wire_message_frame_fixtures_are_current`:

- integration base `d4ebc2a74` → **passes** (2 passed)
- this branch `b1e3c4c06` → **fails**

So carrying `OutputOccurrenceId` on maintained members alters the encoded wire frames. My earlier "gates verified" comment was wrong: I ran `--no-default-features --features test` and read its summary, and this target did not surface in what I checked. Apologies — it is exactly the failure this repo has been bitten by before, when a serialized shape moved and arrived looking like a fixture chore.

**Please do not regenerate the fixtures to clear it.** The questions that need answering first:

- What exactly moved in the encoding — a new field, a reordered one, or a shifted enum discriminant? A shifted discriminant is a silent break for every peer, not just a stale fixture.
- Can occurrence identity ride at the **end** of the record, or behind an optional trailing field, so existing bytes still decode?
- If the layout genuinely must change, this is a protocol version decision and a migration question for ch. 8, not something to absorb into a delivery PR.

The single-source normalization means old members still *read* correctly, so the compatibility story may already be most of the way there — but the encoded frame changing at all needs a deliberate answer.

Everything else I reported stands: the delivery re-keying is sound, plain single-table output is unaffected, and joined maintained subscriptions remain refused by the guard this PR keeps. The follow-up work on `wip/flat-join-output` (`1c4e6444d`) fixes three further regressions it introduced downstream — an array-subquery projection and two policy-fixture byte-stability guards — **without touching any fixture bytes**, which is the right instinct and worth carrying back here.

PR #1261, opened 2026-08-05
fix(jazz): route prepared policy claims per binding

joeinnes:
🤖 Restacked onto the merged terminal and client-local/trusted-serving authorization architecture.

## What changed

- When a prepared authorization union mixes a claimless policy arm with a claim-routed arm, bind the claimless arm to the prepared binding source and project the missing route fields.
- Preserve existing route keys for partially routed arms, avoiding cross-binding products.
- Keep the authorization terminal restricted to row identity plus routing fields.

## Why

One maintained authorization graph serves multiple prepared bindings. A claim-routed arm already carries the binding route, while a claimless arm does not. Without alignment, the claimless alternative cannot produce one authorization result per active binding and may disappear from routed results.

## Validation

- Independent semantic review: no P0/P1 findings; approved with no implementation blocker.
- End-to-end prepared claimless policy integration test passes.
- Compiler-to-terminal regression lowers a real `AuthorizationSubplan` union and verifies:
  - the public claimless arm has an empty-key policy join to `prepared-policy-binding`;
  - the prepared claim route is projected;
  - `policy.authorized_rows` exposes exactly `row_uuid` plus that route, with no evidence fields.
- Reversible old-path mutation fails the route-projection/binding-join assertion while remaining otherwise bound.
- Focused `prepared_claim_routing`, formatting, strict clippy, and diff checks pass.

The existing broad CI failures on the old PR head were inherited from the moving new-core base; this restack keeps the unique two-file implementation and strengthens its regression coverage for the current engine.

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -325,6 +325,10 @@ required-features = ["test"]
 name = "sync_telemetry_otel"
 required-features = ["test", "otel-core"]
 
+[[test]]
+name = "prepared_claim_routing"
+required-features = ["test"]
+
 [[example]]
 name = "realistic_bench"
 required-features = ["client"]
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -590,6 +590,10 @@ where
         let (schema, schema_version) = self.current_write_schema_for_query()?;
         let shape = query.validate_with_schema_version(&schema, schema_version)?;
         let binding = shape.bind(params)?;
+        self.node
+            .node
+            .borrow()
+            .validate_prepared_query_claim_routing(&shape, &binding)?;
         let (local_plan, global_plan) = if should_install_prepared_plan(&shape)
             && !self
                 .node
@@ -7128,6 +7132,8 @@ pub enum ErrorCode {
     Schema,
     /// Query validation or binding failed.
     Query,
+    /// A prepared policy claim predicate cannot be routed safely per binding.
+    PreparedClaimPredicateRoutingUnsupported,
     /// Write was rejected.
     WriteRejected,
     /// Storage failed.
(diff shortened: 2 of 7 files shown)


aeplay:
@joeinnes — an adversarial review of this PR found no defects in it, and I want to be clear about that up front: three separate planted positives all discriminated, so the tests here genuinely guard what they claim.

It did note one residual gap — coverage reaches only a **direct one-hop** policy chain. Closing that gap immediately surfaced a **pre-existing cross-principal read leak**, and I want you to see it because it lives in the area this PR touches.

## The finding

A two-hop chain — document → membership → project, where the project policy uses a reachability seed bound to `sub`:

```
principal B's seed claim must not select principal A's document
  left:  [RowUuid(61616161-6161-6161-6161-616161616161)]
  right: []
```

Principal B sees principal A's document.

## It is not caused by this PR

I ran the same reproducer against `codex/jazz-core-engine-swap` and it fails identically there. **So this is pre-existing and #1261 is not blocked by it** — the PR simply does not reach that case. I wanted to say that explicitly rather than leave a security finding hanging over your review.

The reproducer is on branch `repro/1261-two-hop-claim-leak` (test-only, no implementation change), and it needs the `test` feature:

```
cargo test -p jazz --no-default-features --features test --test prepared_claim_routing \
  prepared_binding_routes_claims_through_two_hop_seeded_policy_sources
```

## What the review did confirm about this PR

The `lower_value_ref` literal-baking path still exists at `crates/jazz/src/node/query_engine/lowering.rs:4670`, but this PR closes it for supported prepared claims: direct column-equality claims are retargeted to binding parameters and intercepted as binding joins before residual lowering, and unsupported claim predicates now fail preparation rather than reaching that path.

It also verified transitive policy collection is a `(schema version, table)`-deduplicated BFS with no fixed depth (`query_eval.rs:6293`), that the predicate validator walks root filters, nested joins, policy branches, reachability edge/access/seed filters and nested arrays, and that `ShapeId` incorporates schema version so schema changes cannot collide in the prepared-plan cache.

The remaining suggestion is coverage rather than correctness: a public-API test for a two-hop chain, and one exercising re-preparation across a schema change. Happy to add those, though the two-hop case currently fails for the reason above — so it may be worth fixing the leak first and landing the coverage with it.

PR #1262, opened 2026-08-05
(new core) Reuse relation snapshots across equivalent terminals

joeinnes:
## What changed

Follow-up to #1219. Equivalent output terminals in one relation-hydration request now share storage snapshot materialisation before terminal-specific hydration begins.

- groups terminals by the complete validated table-source set (table, scan/query shape, and record descriptor)
- materialises one source group at a time, so retained full-snapshot state is bounded
- reuses the existing hydration evaluation memo for shared relation nodes
- preserves each terminal's independent output materialisation
- preserves #1219's one-probe `ArrangementUpdateMode::Replace` join path
- adds opt-in join-time attribution and snapshot/query counters for receipts

The reuse lifetime is one validated runtime hydration call. It cannot cross identities, permission contexts, prepared bindings, or schema runtimes. Different source scans and validated descriptors do not group.

## Why

#1219 removes redundant arrangement probes during replacement, but equivalent output terminals could still independently scan and materialise the same complete relation inputs before the hydration memo had a chance to reuse evaluated nodes. That made first-read work scale with equivalent terminal count.

This moves the reuse boundary ahead of storage snapshot construction without adding a persistent cache or retained incremental arrangements.

## Correctness coverage

Integration/black-box coverage verifies:

- equivalent terminals scan/materialise their shared snapshot once and receive identical correct results
- work remains constant from 1 to 8 equivalent terminals
- different prepared bindings, scan shapes, downstream query shapes, validated schema descriptors, and permission identities do not leak/share incompatible results
- data changes advance the reused state
- weighted duplicates, incremental inserts, negative retractions, and fresh full-rehydration results remain equivalent
- distinct source groups retain at most one complete source-set snapshot concurrently

## Performance receipts

Synthetic inner-join fixture, 4,096 input rows, seven release samples:

| terminals | scans | visited rows | join evaluation | first-read p50 |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 1 | 4,096 | 2,216 µs | 3,285 µs |
| 2 | 1 | 4,096 | 2,235 µs | 3,202 µs |
| 4 | 1 | 4,096 | 2,088 µs | 3,190 µs |
| 8 | 1 | 4,096 | 2,195 µs | 3,316 µs |

The per-terminal digest and 4,096-row result are unchanged. Storage scans, visited records, join evaluation, and total time no longer grow linearly with equivalent terminal count. A single terminal retains its direct path.

Runs against the anonymised real-world relation-heavy fixture returned the same 65 root rows and 43,013 child rows with an unchanged digest and showed an approximately 2.5% first-read improvement. That is a useful positive result, but modest enough to be cautious about generalising the gain: the targeted synthetic fixture shows that the mechanism removes terminal-count scaling, while the end-to-end benefit depends on how many equivalent output terminals a real hydration request contains and how much of its first read is dominated by other work.

Residual cost remains proportional to accumulated source state for each distinct source group. Persisted/incremental arrangement hydration remains a separate possible optimization.

## Validation

- full Groove suite: 394 passed; 2 receipt-only tests ignored
- focused Jazz permission-identity regression
- incremental-delivery canaries
- shared-coverage and warm-reopen differential gates
- maintained-vs-one-shot oracle, 50 seeds
- Clippy, rustfmt, sensitive-data guard
- complete benchmark smoke suite
- independent adversarial review after remediation: no blocker/high/medium findings

The branch was rebased patch-identically onto the live `codex/jazz-core-engine-swap` stack tip before publication.

### DIFF
--- a/crates/groove/src/db/tests.rs
+++ b/crates/groove/src/db/tests.rs
@@ -1685,7 +1685,9 @@ fn assert_direct_record_store_round_trips_array_of_record_values() {
     let results = Value::Array(vec![Value::Record(first), Value::Record(second)]);
     let store = database.direct_record_store("rendered_results").unwrap();
 
-    store.set(&[Value::U64(7)], &[results.clone()]).unwrap();
+    store
+        .set(&[Value::U64(7)], std::slice::from_ref(&results))
+        .unwrap();
 
     assert_eq!(
         store
--- a/crates/groove/src/db/tests.rs
+++ b/crates/groove/src/db/tests.rs
@@ -2174,6 +2174,7 @@ struct ScanCountingStorage {
     inner: MemoryStorage,
     scan_range_count: Rc<Cell<usize>>,
     scan_prefix_count: Rc<Cell<usize>>,
+    visited_records: Rc<Cell<usize>>,
 }
 
 impl ScanCountingStorage {
@@ -2182,6 +2183,7 @@ impl ScanCountingStorage {
             inner: MemoryStorage::new(column_families),
             scan_range_count: Rc::new(Cell::new(0)),
             scan_prefix_count: Rc::new(Cell::new(0)),
+            visited_records: Rc::new(Cell::new(0)),
         }
     }
 
@@ -2191,6 +2193,10 @@ impl ScanCountingStorage {
 
     fn scan_prefix_count(&self) -> usize {
         self.scan_prefix_count.get()
+    }
+
+    fn visited_records(&self) -> usize {
+        self.visited_records.get()
     }
 }
 
@@ -2235,7 +2241,12 @@ impl OrderedKvStorage for ScanCountingStorage {
     ) -> Result<(), StorageError> {
         self.scan_prefix_count
             .set(self.scan_prefix_count.get().saturating_add(1));
-        self.inner.scan_prefix(cf, prefix, visit)
+        let mut counting_visit = |key: &[u8], value: &[u8]| {
+            self.visited_records
+                .set(self.visited_records.get().saturating_add(1));
+            visit(key, value)
+        };
+        self.inner.scan_prefix(cf, prefix, &mut counting_visit)
     }
 
     fn scan_prefix_reverse(
@@ -4368,6 +4379,7 @@ fn equivalent_query_graph_terminals_share_snapshot_materialization() {
     database.commit_batch(batch).unwrap();
 
     let scans_before = counter.scan_prefix_count();
+    let visits_before = counter.visited_records();
     let snapshots = database
         .query_graphs([
             ("first", GraphBuilder::table("albums").project(["id"])),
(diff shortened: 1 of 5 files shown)


aeplay:
Reviewed the grouping and reuse logic. The reuse boundary looks sound; one test does not test what it says it tests.

## The reuse boundary is safe

Worth stating explicitly, since this is the part that would be a security bug if wrong. The group key is `SnapshotSourceSet(HashMap<TableSnapshotSource, RecordDescriptor>)`, and `TableSnapshotSource` is `{ table, scan: Option<StaticScanSpec> }` — so the **scan spec participates in the key**. Two terminals whose visibility differs by a narrowed scan cannot land in the same group, consistent with `G-INV-ARR-6`. Terminals that share a base-table scan but differ in downstream permission filtering correctly share the raw table delta, because the filtering happens downstream in the graph.

The custom `Hash for SnapshotSourceSet` hashes sorted per-entry hashes, so it is order-independent over the map and consistent with the derived `Eq`. And `records.descriptor != output.output → GraphOutputMismatch` is checked per terminal after hydration, so a mis-grouped share cannot silently produce wrong-shaped output.

## `distinct_snapshot_groups_retain_at_most_one_materialized_source_set` cannot fail

`crates/groove/src/db/tests.rs:4258` asserts

```rust
hydration_snapshot_peak_cached_source_sets == 1,
"distinct source sets must be materialized and released one group at a time"
```

but the counter is computed at `crates/groove/src/ivm/runtime/mod.rs:618` as

```rust
self.hydration_snapshot_peak_cached_source_sets = usize::from(!groups.is_empty());
```

That is structurally 0 or 1. It records *whether grouped hydration ran*, not a peak. No implementation — however much it retains — can make it exceed 1, so the assertion holds regardless of the property it names.

**Planted positive, to be sure rather than plausible.** I rewrote the loop to materialize every group up front and hold all of them concurrently:

```rust
let mut retained = Vec::new();
for (sources, outputs) in groups {
    retained.push((materialize_snapshot_table_deltas(&self.schema, storage, sources.0)?, outputs));
}
for (table_deltas, outputs) in retained { /* ... */ }
```

With 64 terminals over 32 distinct source sets all retained at once — precisely what the message forbids — the test still passes:

```
test db::tests::distinct_snapshot_groups_retain_at_most_one_materialized_source_set ... ok
test result: ok. 1 passed; 0 failed
```

I reverted the change; the tree is clean.

So the bounded-retention property, which is the main safety argument for materializing one group at a time, is currently unguarded. If someone later hoists materialization out of the loop for speed, nothing here objects.

**Suggested fix:** make the counter a real peak — increment on materialize, decrement on release, and record the maximum — then the existing assertion becomes meaningful as written. A cheaper alternative is asserting peak *resident bytes* or live snapshot count, but a true peak counter matches the field's name and doc comment (`crates/groove/src/ivm/runtime/mod.rs:254`, "Maximum number of complete source-set snapshots retained concurrently").

## Minor

`crates/groove/src/ivm/runtime/mod.rs:386` guards the descriptor check with `!records.deltas.is_empty()`, while the two new sites at 609 and 624 check unconditionally. Probably deliberate — an empty delta set carries no records to mis-shape — but the asymmetry is worth a word, since all three are the same invariant.

## Not a conflict with the structured work

Flagging for coordination: the "relation snapshots" here are storage snapshots during hydration, distinct from the `RelationSnapshot` **wire** type that the v4 delivery cut replaces. This PR touches no protocol code, so the two do not collide.

joeinnes:
Addressed feedback with a better test.

aeplay:
🤖 Closing as superseded by #1340. The snapshot-sharing stack targets the old multisink relation-hydration/materialization path, which #1340 replaced with maintained terminal assembly and typed terminal edits. The original branch was preserved during restack triage, but there is no truthful equivalent patch to replay onto the current architecture.

PR #1237, opened 2026-08-04
(new core) bench: gate S4 propagation by fixed delta

joeinnes:
## Base

This PR is based directly on `codex/jazz-core-engine-swap`. The fix from #1236 is merged and inherited from that integration base; the two commits unique to this PR are benchmark changes. It is no longer stacked on #1232 and can be reviewed independently.

## What changed

- add an S4 fixed-delta/varying-view benchmark phase using the Payment operation
- propagate from a writer through core and edge to a separate observer client
- hard-gate the exact added and removed row identities, program facts, version bundles and records, wire bytes, storage reads and ranges, and final observer values
- add an attributable receipt and update the benchmark roadmap

## What it demonstrates

For a fixed four-row Payment transaction, propagation work on both hops remains structurally bounded by the delta as the retained whole-table view grows from 15 to 10,005 rows:

- result changes remain 4 additions / 3 removals per hop
- version payload remains 4 bundles / 10 records per hop
- core storage work remains 4 reads / 4 ranges
- edge storage work remains 104 reads / 119 ranges
- encoded bytes grow only from 4,570 B to 4,602 B due to fixed sequence/cursor framing

Wall time grows from approximately 5.9 ms to 28.9 ms because retained-view bookkeeping remains view-sized. Constant CPU time is deliberately not the acceptance claim.

## Validation

- `cargo test -p jazz partial_exclusive`
- `cargo check -p jazz-sim --benches`
- `JAZZ_S4_PROPAGATION_SCALE_ONLY=1 JAZZ_S4_PROPAGATION_CUSTOMERS=2,20 cargo bench -p jazz-sim --bench s4_order_processing`
- `JAZZ_S4_PROPAGATION_SCALE_ONLY=1 JAZZ_S4_PROPAGATION_CUSTOMERS=10,1000,10000 cargo bench -p jazz-sim --bench s4_order_processing`
- repository pre-commit checks: Clippy, sensitive-data scan, oxfmt, and rustfmt

PR #1267, opened 2026-08-05
(new core) Retain policy-churn reconnect matrix

joeinnes:
## What changed

- expand R13 into six independently gated reconnect cases covering unchanged recursive membership, grant, revoke, simultaneous grant/revoke, claim revoke, and claim restore
- measure only the reconnect/catch-up interval while retaining response-byte and exact subscription-delta receipts
- document the methodology, correctness gates, initial local results, and interpretation limits

## Why

#1266 fixes stale fast-cursor state after authorization changes. This stacked receipt makes that contract reproducible and prevents future reconnect changes from silently regressing grants, revocations, or claim transitions while a reader is disconnected.

## Results

All six cases complete in roughly 0.7–2.2 ms on the local single-process, single-threaded, network-free harness. The unchanged control sends 61 B versus an 856 B initial response; each authorization-changing lane emits exactly the expected added/updated and removed row IDs, with no extra queued event.

These figures compare protocol paths and detect accidental full rehydrate; they are not end-user reconnect latency claims.

## Validation

- `cargo bench -q -p jazz --bench realistic_phase1 -- r13_permission_filtered_resume --test`
- `cargo bench -p jazz --bench realistic_phase1 -- r13_permission_filtered_resume --noplot`
- `cargo clippy -p jazz --bench realistic_phase1 -- -D warnings`
- repository pre-commit hooks (clippy, sensitive-data scan, oxfmt, rustfmt)

Stacked on #1266.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -121,10 +121,10 @@ jobs:
       - name: Build native benchmark binaries
         run: |
           set -euo pipefail
-          cargo build --release -p jazz-tools --features client,rocksdb --example realistic_bench
-          cargo build --release -p jazz-tools --features client,sqlite --example realistic_bench
-          cargo bench -p jazz-tools --features rocksdb --bench realistic_phase1 --no-run
-          cargo bench -p jazz-tools --features sqlite --bench realistic_phase1 --no-run
+          cargo build --release -p jazz --features client,rocksdb --example realistic_bench
+          cargo build --release -p jazz --features client,sqlite --example realistic_bench
+          cargo bench -p jazz --features rocksdb --bench realistic_phase1 --no-run
+          cargo bench -p jazz --features sqlite --bench realistic_phase1 --no-run
           cargo bench -p jazz-sim --bench s2_canvas --no-run
 
       - name: Run native benchmark suite
@@ -349,7 +349,7 @@ jobs:
           uname -a > bench-out/browser/uname.txt
 
       - name: Build jazz-tools server binary
-        run: cargo build -p jazz-tools --features cli
+        run: cargo build -p jazz --features cli
 
       - name: Build jazz-napi package
         run: pnpm --dir crates/jazz-napi run build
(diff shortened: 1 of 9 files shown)


aeplay:
The receipt itself is well built — `assert_resume_delta` compares **exact** added and removed row-id sets, not just counts, and each of the six cases asserts its own expected transition. That is the right shape for this contract, and it is a genuine improvement over a timing-only receipt.

One gap, and it undercuts the stated purpose.

## The six cases never run automatically

The description says this "prevents future reconnect changes from silently regressing grants, revocations, or claim transitions". That holds only if something runs them.

`r13_permission_filtered_resume` is **not** in `dev/benchmarks/realistic/ci_benchmarks.mjs`. That file currently lists:

```
r1_crud  r2_reads  r3_rocksdb_cold_load  r4_hot_task_history
r9_subscribed_write  r10_sync_fanout  r11_byte_wire_resume  r12_recursive_permissions
```

`r11` and `r12` — this case's immediate predecessors, and the ones it extends — are both wired. `r13` is wired to nothing. The canonical Rust gates only `cargo check -p jazz-sim --benches`, which compiles benches without running them, so the assertions never fire outside a manual `cargo bench … --test`.

Worth noting the contrast within this same batch: **#1240** adds a new receipt *and* edits `dev/benchmarks/smoke.sh` to run it, and **#1237**'s fixed-delta gating rides `s4_order_processing`, which is already in the CI list. Both are covered. This one is the exception.

**Suggested fix:** add `r13_permission_filtered_resume` to `ci_benchmarks.mjs` alongside `r11` and `r12`. If it is deliberately excluded — six cases may be too slow for the CI budget — then say so in `crates/jazz/benches/README.md` and state where it *is* expected to run, so the next person does not read "gated" as "enforced".

## Otherwise

The interpretation limits are stated honestly — single-process, single-threaded, network-free, "not end-user reconnect latency claims" — which is exactly right for a protocol-path receipt and worth keeping in that wording.

Once the wiring is in, the 61 B unchanged-control versus 856 B initial-response contrast becomes a real accidental-full-rehydrate detector rather than a one-time observation.

PR #1283, opened 2026-08-06
(new core) Restore Db.disconnect()/Db.reconnect() and un-skip the offline suite

aeplay:
Restores `Db.disconnect()` / `Db.reconnect()` and the offline path. **Supersedes #1116.**

## The test work came first, deliberately

#1116 recorded four browser failures as things it surfaced — but **that suite was `describe.skip`**, so they were never fixed, never guarded, and invisible for two weeks. Un-skipping and characterising them was step one, because merging a skipped suite as evidence of offline correctness is how this got lost.

What they turned out to be:

- **`invalid offset` after disconnected local writes** and **missed remote update after reconnect** — real, now fixed, both memory and worker modes converge.
- **Local-only pending reads** — **not** a bug. The old direct-memory expectation was outdated. A local-only read now resolves immediately from the local snapshot, showing pending local writes and omitting remote-only rows until delivery after reconnect.

The un-skipped suite passes **4/4**.

## What was restored

- public `Db.disconnect()` / `Db.reconnect()`;
- intentional-offline state, so lazily-created clients stay offline rather than eagerly connecting;
- an **awaited** persistent-worker disconnect, replacing fire-and-forget;
- native reconnect re-pumps transport, subscriptions, and plain-subscription refresh.

None of this was covered by recent landings: `INV-API-30` is restart/reopen replay, a different lifecycle from an in-process temporary disconnect, and storage release on shutdown is handled separately by shutdown dropping the client.

## One thing to decide before merging

The local-only read contract is written to **`dev/DB_DISCONNECT_RESTORE.md`**. That is the retired habit — contracts belong as normative text in the owning SPEC chapter, here `crates/jazz/SPEC/13_db_api.md`, ideally with an invariant so it is checkable. A `dev/` note is invisible to everyone who later asks "what should a local-only read return while offline?"

I would move it before merging. Flagging rather than doing it, since it changes what the PR asserts.

## Gates

Re-gated on the current base after #1280.

`cargo test -p jazz --no-default-features --features test` → the only failures are the three `catalogue_sync_integration` tests gated on flat joins and array-relation output. The four `known_state` rehydrate failures this branch originally reported are **gone**, fixed by #1280.

`cargo check --workspace --all-targets` 0 · `cargo fmt --check` 0 · `pnpm build:ci` 0 · forced NAPI build 0 · binding load 0.

The full `jazz-tools test:browser` still shows unrelated pre-existing failures — worker `invalid offset` in other scenarios, an operand mismatch, and include-subscription failures. Those were **not** verified against the merged base by the author, so treat them as unconfirmed rather than known-good. `oxfmt` and `ts-wire-codec` report their pre-existing failures.


aeplay:
CI triage: **all three red checks are inherited from the base branch.** None originate here.

`codex/jazz-core-engine-swap` itself is red on the same three (run 31073300632):

- **`test-rust`** — exactly three tests, `21 passed; 3 failed`: `table_rename_fk_array_lookup_finds_related_rows_on_old_branch`, `local_join_query_uses_current_permissions_for_joined_provenance_after_lens_transform`, `table_rename_join_query_translates_join_target_on_old_branch`. The known `catalogue_sync_integration` failures gated on flat joins and array-relation output, blocked on the structured-results work.
- **`test-ts`** — `Error: invalid offset` thrown from `decodeRecordValueWithLayout` (`packages/jazz-tools/src/runtime/native-runtime/native-row-codec.ts:369`), which fails every `todo-client-localfirst-ts` browser test. A real codec defect on base; a fix is in progress on `fix/native-row-codec-offset`.
- **`lint`** — `pnpm format:check` failing on `crates/groove/SPEC/INVARIANTS.md`, whose `INV-INC-2` row had unescaped pipes inside a complexity bound that `oxfmt` read as column separators. Fixed in **#1286**; this will go green once that lands and this branch is rebased.

So the merge decision here rests on the diff, not on the checks.

packages/jazz-tools/src/runtime/client.ts
```diff
@@ -86,7 +86,7 @@ export interface Runtime {
   /** Connect to a Jazz server over WebSocket (Rust transport). */
   connect(url: string, auth_json: string): void;
   /** Disconnect from the Jazz server and drop the transport handle. */
-  disconnect(options?: { rejectWaiters?: boolean }): void;
+  disconnect(options?: { rejectWaiters?: boolean }): void | Promise<void>;
```
nrainhart: Can we just pick one of void | Promise<void>? And if we keep Promise<void>, let's properly document what it means in a TSdoc: the promise is not resolving on disconnect, but rather when the disconnect message is sent to the worker (which tbh I'm not sure in which situation could be useful)

packages/jazz-tools/tests/browser/db.disconnect.test.ts
nrainhart, line 1: There are some tests from the [original test suite](https://github.com/garden-co/jazz/pull/1064/changes#diff-e87d4d9d90f34893c4b9007a1d3fc6a5dde323b5348294eccc6bcf097a965f52) missing, do you think it'd be hard to support them?
- "resolves local waits and keeps edge/global waits pending while disconnected"
- "resolves local reads and defers edge reads while disconnected"

nrainhart (commented):
Thanks for restoring this feature! Two comments before approving

aeplay:
@nrainhart — answering your second comment about the two missing tests from the original suite.

**Both are now implemented in #1301.**

The short answer to "would it be hard to support them?": not hard, but they needed a **semantics change**, which is why they weren't simply ported here. When I tried, both failed *with* the implementation rather than without it — they describe behaviour the runtime did not have:

- *"keeps edge/global waits pending while disconnected"* — `reconnect()` goes through `connect()`, which called `this.disconnect()` unparameterized and so took the `rejectWaiters: true` default. Reconnecting rejected exactly the waiters that `Db.disconnect()` had deliberately preserved with `rejectWaiters: false`. That one turned out to be a **bug**: the intent and the plumbing already existed, they just weren't threaded through the reconnect path or across the worker boundary.
- *"defers edge reads while disconnected"* — `disconnect()` set `connectionReady = null`, and consumers `await this.connectionReady`. Awaiting `null` resolves immediately, so the gate was **absent rather than closed** and an edge read fulfilled from local data. That one is a genuine behaviour change: it now holds until reconnect, while local reads stay ungated.

Rather than land red tests here, this PR keeps your first comment's fix (the `void | Promise<void>` narrowing and its TSDoc) and #1301 carries the behaviour with both of your tests passing.

One thing worth your eye on #1301: with waits now pending rather than rejected, an edge/global wait or read can block **indefinitely** if the client never reconnects. Cancellation (`AbortSignal`) and observable connection state are the intended complement and are recorded as an open question in `SPEC/13_db_api.md` via #1300, but they are not built yet.

nrainhart (approved):
Thanks for addressing the comments!

PR #1284, opened 2026-08-06
(new core) Bring the structured-results plan up to date

aeplay:
The implementation plan is merged and gets read as authoritative, but it had drifted from what we are actually doing. Docs only.

## What was missing

**No stage status.** PRs 1, 2 and 3 are merged (#1260, #1263, #1282) with their invariants in force; the document still read as though none had started.

**Two decisions taken after it was written**, neither recorded:

- *Explicit boundedness* — `INV-QUERY-29` means boundedness must be **declared**, not that a finite limit must be present. `unbounded()` is a first-class mode. That matters because three correctness gates use unbounded child arrays deliberately, and a literal ban would have driven them through a code path production never takes.
- *One unified output terminal* (#1279 spec, #1281 Expand mode) — joins produce flat wide rows, the terminal decides shape, nothing else renders. This postdates the five-stage plan entirely.

**The blocker that stopped two attempts today**, which is the important part.

## The blocker, recorded as an open question

`CollectByOp` carries a single `collection_field` and renders one parent with one `Array<Record>`. `ResultTree` requires named ordered child arrays — siblings — and recursion through parent → child → grandchild. The obvious bridge is composing collectors, and `INV-QUERY-27` forbids exactly that: a collector may not be an input to any graph node, **including another collector**. The rule that makes the terminal clean prevents composing one into a tree.

Separately, the only public maintained carrier is `SubscriptionEvent` exposing flat root rows, related rows and `RelationEdge` deltas (`db.rs:8128`), and the remote carrier encodes `RelationEdgeEntry` (`protocol.rs:1890`). A structured result cannot cross either without flattening back into rows and edges — the facade-side reconstruction the design forbids.

So terminal integration and the wire cut cannot be separated while the carrier is unchanged, which is why both PR 3.5 and PR 4 stopped.

Three candidate resolutions are written into the document: one atomic PR, extending `CollectByOp` to describe a tree within a single terminal (current recommendation, since `INV-QUERY-27` survives intact), or narrowing scope to one-shot reads with maintained delivery left explicitly flat.

No invariant status or coverage changed.

### DIFF
--- a/crates/jazz/SPEC/E_structured_results_implementation_plan.md
+++ b/crates/jazz/SPEC/E_structured_results_implementation_plan.md
@@ -31,6 +31,72 @@ Therefore `INV-STORAGE-27` must remain `target`/`untested` after #1251 and is
 closed by PR 1 below, not credited merely because #1251 adds a `Record` enum
 variant.
 
+## Status (2026-08-06)
+
+| stage | state |
+|---|---|
+| PR 1 — durable-key rejection | **merged** (#1260); `INV-STORAGE-27` now/✓ |
+| PR 2 — Groove `CollectBy` terminal | **merged** (#1263); `INV-QUERY-27`, `INV-QUERY-28`, `G-INV-REC-16` now/✓ |
+| PR 3 — canonical `ResultTree`, explicit boundedness | **merged** (#1282); `INV-QUERY-29` now/✓ |
+| PR 3.5 — terminal integration | **blocked**, see _Terminal integration blocker_ below |
+| PR 4 — atomic structured delivery on v4 | **blocked** on the same question |
+| PR 5 — structured differential oracle | not started; `INV-TEST-5` target |
+
+Two decisions taken after this plan was written, and not part of its original
+five stages:
+
+- **Explicit boundedness.** `INV-QUERY-29` means *boundedness must be declared*,
+  not *a finite limit must be present*. An array subquery declares `limit(n)`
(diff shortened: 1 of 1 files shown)


aeplay:
CI triage: **all three red checks are inherited from the base branch.** None originate here.

`codex/jazz-core-engine-swap` itself is red on the same three (run 31073300632):

- **`test-rust`** — exactly three tests, `21 passed; 3 failed`: `table_rename_fk_array_lookup_finds_related_rows_on_old_branch`, `local_join_query_uses_current_permissions_for_joined_provenance_after_lens_transform`, `table_rename_join_query_translates_join_target_on_old_branch`. The known `catalogue_sync_integration` failures gated on flat joins and array-relation output, blocked on the structured-results work.
- **`test-ts`** — `Error: invalid offset` thrown from `decodeRecordValueWithLayout` (`packages/jazz-tools/src/runtime/native-runtime/native-row-codec.ts:369`), which fails every `todo-client-localfirst-ts` browser test. A real codec defect on base; a fix is in progress on `fix/native-row-codec-offset`.
- **`lint`** — `pnpm format:check` failing on `crates/groove/SPEC/INVARIANTS.md`, whose `INV-INC-2` row had unescaped pipes inside a complexity bound that `oxfmt` read as column separators. Fixed in **#1286**; this will go green once that lands and this branch is rebased.

So the merge decision here rests on the diff, not on the checks.

PR #1285, opened 2026-08-06
(new core) One CollectBy terminal renders a whole tree

aeplay:
One `CollectBy` terminal now renders a whole tree — sibling named arrays and recursion — rather than a single collection slot. This unblocks the structured terminal integration and the v4 delivery cut, both of which stopped on exactly this.

## Why it was needed

`CollectByOp` carried one `collection_field` and produced one parent with one `Array<Record>`. `ResultTree` requires named ordered child arrays — siblings — and recursion through parent → child → grandchild.

The obvious bridge is composing collectors, and `INV-QUERY-27` forbids precisely that: a collector may not be an input to any graph node, *including another collector*. So the rule that keeps the terminal clean is the rule that prevented building a tree from one.

Two attempts to split terminal integration from the delivery cut both stopped here.

## The shape

`CollectBySlotBuilder` / `CollectBySlot` form a **recursive descriptor tree**. Slots are uniquely named within their owner and addressed by field path — `children.grandchildren`. Each slot carries its own group and child projections, order and tie fields, direction, offset and limit.

**`INV-QUERY-27` is unchanged.** Slots are descriptor data inside one boxed `CollectByOp` — never graph nodes. There is still exactly one terminal, so the invariant is not weakened, only satisfied at greater depth.

## Decisions stated rather than left implied

- **Expand rejects tree slots** and remains flat and single-level. The unified terminal has two modes; this says what happens when they meet.
- **`MAX_COLLECT_BY_TREE_DEPTH = 16`**, Groove-owned. Jazz's future receive-boundary depth and width limits stay separate — different trust boundaries deserve different constants.
- **`INV-INC-2` restated** so `R_g(limit)` accounts for selected windows and bytes across the whole rendered tree, without weakening the prohibition on work proportional to accumulated state.

## Coverage

Through `Database::subscribe_one_sink` and `GraphBuilder`:

- sibling arrays with independent ordering and windows;
- parent → child → grandchild rendering;
- a rendered-grandchild insertion producing exactly one root `-old`/`+new` replacement;
- an out-of-window grandchild insertion producing whole-tree byte-equal suppression;
- depth rejection beyond the bound.

**Planted positive, run independently of the author:** disabling tree byte-equal suppression makes `collect_by_tree_grandchild_change_replaces_one_whole_parent_and_suppresses_unrendered_change` fail while the other two tree tests still pass. The `INV-INC-2` property is guarded, not merely asserted.

Existing single-slot Collect and Expand tests pass unchanged.

## Gates

`cargo test -p groove` 0 · `cargo test -p jazz` 0 · `cargo test -p jazz --test wire_fixtures` 0 (no wire bytes change) · `cargo check -p jazz-sim --benches` 0 · `cargo check --workspace --all-targets` 0 · `cargo fmt --check` 0 · `INV-INC-1` incremental-delivery canary 0 · 300-seed maintained-vs-one-shot oracle 0.

Clippy reports only the pre-existing `type_complexity` in `warm_reopen_differential.rs:478`, present on the base — so the recursive descriptor did not widen a shared type. `CollectByOp` needed boxing once before when a large variant blew the stack in an unrelated test, so that was worth confirming.

A `--features test` run of mine showed two websocket failures; both pass in isolation and are the known parallel-load port-binding flake, not this change.


aeplay:
CI triage: the one red check is **fully inherited from the base branch**, not from this change.

`test-rust` fails on exactly three tests:

- `table_rename_fk_array_lookup_finds_related_rows_on_old_branch`
- `local_join_query_uses_current_permissions_for_joined_provenance_after_lens_transform`
- `table_rename_join_query_translates_join_target_on_old_branch`

These are the known `catalogue_sync_integration` failures gated on flat joins and array-relation output, and they fail identically on `codex/jazz-core-engine-swap` itself (run 31073300632, same three names, `21 passed; 3 failed`). They are blocked on the structured-results work this PR is a prerequisite for.

`lint` was red repo-wide for an unrelated reason — a mangled registry row — fixed separately in #1286.

Verified locally beyond CI: a planted positive on `INV-INC-2`, flipping the byte-equality suppression in `crates/groove/src/ivm/runtime/mod.rs` to `if false && before == after`, fails `collect_by_tree_grandchild_change_replaces_one_whole_parent_and_suppresses_unrendered_change` while the other two tree tests still pass — so the guard is real and specific.

PR #1287, opened 2026-08-06
(new core) Refuse contaminated benchmark measurements

aeplay:
Benchmarks silently accepted instrumentation flags that change the very timings they measure. A run with `JAZZ_REHYDRATE_TRACE` set produced a number that looked ordinary and was not.

This makes the refusal shared and total: a contaminated benchmark **exits 2** instead of reporting a figure.

## Mechanism

New crate `jazz-benchmark-guard` owns the single flag list and exits 2 naming the flag and why it contaminates. Wired into **all 17 Jazz benchmark targets, all 13 Jazz Sim targets, both Groove benchmarks**, the Groove/Jazz receipt tests, the in-tree OPFS and native-storage Criterion targets, and the Jazz receipt/benchmark examples.

Criterion targets carry an outer guard only — measurement functions are untouched, so the guard cannot itself perturb a timing.

## Which flags contaminate

Refused: `JAZZ_REHYDRATE_TRACE`, `GROOVE_PROFILE_HYDRATION_OPERATORS`, `GROOVE_TRACE_INDEX_BY`, `JAZZ_CLOSURE_TRACE`, `JAZZ_CAPABILITY_TRACE`, `JAZZ_CAPABILITY_TRACE_FILE`, `JAZZ_PROFILE_OUT`, `JAZZ_CUSTOMER_TRACE_TICKS`, `JAZZ_ALLOC_SITE_SAMPLE_RATE`, `JAZZ_ALLOC_SITE_MAX_SAMPLES`.

Deliberately **not** refused, with reasons: `JAZZ_PROFILE_FREQUENCY` (only configures an already-enabled profiler; the enabling flag `JAZZ_PROFILE_OUT` is guarded), `JAZZ_S3_HEADLINE_PROGRESS` (post-phase output), `JAZZ_PROFILE`/`JAZZ_BENCH_PROFILE` and workload-size flags (select labels and workloads), `JAZZ_BENCH_IGNORE_RESULT_DIRTY`/`JAZZ_RETAIN` (receipt retention metadata), and the OTel/test-server flags (server telemetry, off the benchmark path).

## Verification

Re-run independently of the implementing lane:

- `JAZZ_REHYDRATE_TRACE=1 cargo bench -p jazz --bench maintained_rehydrate_scaling` → **exit 2** with the named refusal
- the same bench with the flag unset → **exit 0**, completing normally — the guard does not false-positive
- `cargo test -p jazz-benchmark-guard` → **0** (2 passed), covering both the clean and contaminated environments

Exit codes read unpiped.

## Gates

`cargo check -p jazz-sim --benches` **0** · `cargo check --workspace --all-targets` **0** · `cargo test -p jazz` **0** · `cargo test -p groove` **0** · `cargo fmt --check` **0**

`cargo clippy --package jazz --all-targets -- -D warnings` → 101, from the pre-existing `clippy::type_complexity` in `crates/jazz/tests/warm_reopen_differential.rs:478`, unrelated to this change.

## Follow-ups for unmerged branches

Three benchmark entrypoints live only on their own branches and will need the guard when they land: `policy_cost_receipt.rs` (#1240), #1262's attributed hydration receipt (which should keep its attribution-only run separate and guard the uninstrumented timing run), and the unmerged scenarios behind #1237's S4 and #1267's `realistic_phase1` entrypoints — those two entrypoints exist on base and are already wired here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1407,6 +1407,7 @@ dependencies = [
  "bytes",
  "hdrhistogram",
  "internment",
+ "jazz-benchmark-guard",
  "opfs-btree",
  "paste",
  "postcard",
@@ -1950,6 +1951,7 @@ dependencies = [
  "hex",
  "insta",
  "internment",
+ "jazz-benchmark-guard",
  "jsonschema",
  "jsonwebtoken",
  "libc",
@@ -1988,6 +1990,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "jazz-benchmark-guard"
+version = "0.1.0"
+
+[[package]]
 name = "jazz-napi"
 version = "0.1.0"
 dependencies = [
@@ -2016,6 +2022,7 @@ dependencies = [
  "diamond-types",
  "hdrhistogram",
  "jazz",
+ "jazz-benchmark-guard",
  "libc",
  "postcard",
  "pprof",
@@ -2778,6 +2785,7 @@ name = "opfs-btree"
 version = "0.1.0"
 dependencies = [
  "criterion",
+ "jazz-benchmark-guard",
  "js-sys",
  "rustc-hash",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = [
+  "crates/benchmark-guard",
   "crates/groove",
   "crates/jazz",
   "crates/jazz-sim",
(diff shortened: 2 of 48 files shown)


aeplay:
CI triage: all three red checks are inherited from the base branch.

- **`lint`** — this branch predates #1286, so it still carries the mangled `INV-INC-2` row that makes `pnpm format:check` fail. #1286 fixes it and is **green on lint**, confirming the cause. Merging #1286 first and rebasing this clears it; nothing here needs changing.
- **`test-rust`** — the three known `catalogue_sync_integration` failures on base, gated on flat joins and array-relation output.
- **`test-ts`** — `Error: invalid offset` from `decodeRecordValueWithLayout` on base; a fix is in progress on `fix/native-row-codec-offset`.

This PR touches only benchmark entrypoints and adds a new guard crate, so it cannot affect any of the three.

PR #1288, opened 2026-08-06
(new core) Fix invalid offset in the native row codec

aeplay:
`test-ts` is red on `codex/jazz-core-engine-swap` itself: every `todo-client-localfirst-ts` browser test fails, each preceded by `Error: invalid offset` from `decodeRecordValueWithLayout`. This fixes it.

## Root cause

The TypeScript decoder classified `I64` and `I32` as **variable-width**. They are fixed-width — 8 and 4 bytes. So for a `Nullable<I32>` or `I64` field the decoder read the *value bytes as a length prefix*, and every subsequent offset was garbage.

Instrumented at the point of failure, from Rust-generated bytes:

```
field user_title, declared Nullable<String>, tag 12
claimed width 1763, remaining 15, current offset 30, end 1793, record length 45
```

`user_title` is merely where it surfaced. The preceding `Nullable<I32>` and `I64` fields had already been misread, so the offset was long past the record by the time a genuinely variable-width field was reached.

Both reported stack paths — `nativeResetDeltaFromBatches` and `refreshPlainSubscriptionRows` — route through `decodeRecordValueWithLayout`, so this is **one shared cause**, not two.

## The corrected Groove `ValueType` table

| tag | type | width |
| --- | --- | --- |
| 3–5 | U64, F64, Bool | 8, 8, 1 |
| 6–8 | String, Bytes, Uuid | variable, variable, 16 |
| 9–12 | Enum, Tuple, Array, Nullable | 1, member sum, variable, inner+1/variable |
| 13 | Record | variable |
| 14 | I64 | **8** |
| 15 | I32 | **4** |

The decoder now also handles `Record`'s descriptor payload, and the fix introduces `storageColumnValueType` where the storage type — not the public output type — is what governs decoding.

**Jazz's schema tags are a separate namespace**: Jazz tag 15 is JSON, not Groove `I32`. Worth stating explicitly, since conflating the two is what silently broke an in-flight PR when `Record` was inserted mid-enum and shifted `I64`/`I32` to 14 and 15.

## Regression guard

A **Rust-pinned cross-language fixture**, so the two sides can no longer drift silently:

- `crates/jazz/fixtures/native_row_codec.json`
- `crates/jazz/tests/wire_fixtures.rs:454`
- `packages/jazz-tools/src/runtime/native-runtime/native-row-codec.test.ts:21`

Wired into `dev/gates/ts-wire-codec.sh:27`. Proven by temporarily removing the tag 14/15 width entries — the test then fails with exactly the instrumented `user_title … 30..1793` error.

## Verification

Re-run independently of the implementing lane, exit codes read unpiped:

- **`todo-client-localfirst-ts` browser suite → exit 0, `14 passed` (2 files).** This is the suite that fails on base.
- `dev/gates/ts-wire-codec.sh` → exit 1, `91 passed`, single failure `encodes the policy graph perf fixture byte-stably` — the pre-existing policy-graph byte-stability failure documented in `CLAUDE.md`, unrelated to row-codec value tags.

Lane-reported, on `7a0dc3dc3`: `cargo test -p jazz --test wire_fixtures` **0** · `cargo test -p jazz` **0** · `cargo test -p groove` **0** · `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0** · `--no-default-features --features test` **101**, the three known `catalogue_sync_integration` failures only.

`pnpm exec oxfmt --check` fails only on the untouched `crates/groove/SPEC/INVARIANTS.md`, verified to fail identically on parent `560a8575a`; that is fixed separately in #1286.


aeplay:
Additional evidence from a large real-world import, which is worth recording because it both **validates this fix and shows it is not the whole story**.

I rebuilt both the client bundle and the server binary from this branch and ran a bulk import whose schema is far wider than the todo example's. The failure moved:

- **without this fix:** `decode cell record: invalid nullable flag byte: 182`
- **with this fix:** `decode cells: Serde Deserialization Error`, on a different table, at a different point

So the `I64`/`I32` width correction is real and load-bearing — the first error is exactly the misread-offset signature this PR fixes, and it is gone. But at least one further encode/decode divergence remains, in a shape the todo example never exercises.

Both failures are staged inserts failing on the **server** while decoding cells the **client** encoded. The surviving one is consistent with a nesting case — `Nullable` of a fixed-width type, an array of records, or a tuple of mixed widths — rather than a scalar, since the scalar widths are now pinned by the fixture here.

**This does not block merging.** The fix is correct, the guard is real, and the browser suite goes from all-failing to 14/14. It does mean `todo-client-localfirst-ts` passing is not sufficient evidence that the codec is sound.

Follow-up dispatched on `fix/native-row-codec-remaining`, stacked on this branch: a round-trip property test over the **whole `ValueType` space including nesting to depth 3**, in both directions, extending the cross-language fixture added here. That turns this class into something that cannot recur silently, rather than fixing one shape at a time.

Separately worth flagging, since it wasted several hours before I caught it: the import runs against `packages/jazz-tools/dist/`, not `src/`. The `dist` in the tree had been built an hour *before* the websocket frame-batch fix landed, so an already-fixed bug kept reproducing and was misattributed to a harness timeout. Rebuild `dist` and `target/debug/jazz-tools` before trusting any measurement or repro that goes through them.

aeplay:
## Correction — do not merge yet

I verified this more thoroughly and found a **real correctness regression** that my earlier verification missed. My "clean" report above was based on the browser suite and `ts-wire-codec.sh`; neither exercises the path that breaks, and CI never reached the suite that does.

### The regression

`packages/jazz-tools` `src/runtime/subscription-manager.test.ts` fails on this branch and **passes on base**:

- `decodes native subscription additions`
- `clears tracked state before applying native reset frames`

The assertion:

```
- Expected
+ Received
-     "count": -42,
+     "count": 2147483606,
```

`2147483606` is `0x7FFFFFD6` — the **order-preserving storage encoding** of `-42`, i.e. the value with its sign bit flipped so that byte ordering matches numeric ordering. So signed integers are now decoded to their *storage* form instead of their *logical* value.

This is the `storageColumnValueType` / `columnValueType` distinction not being carried through: treating the column as `I32` for width purposes is correct, but the decode must then invert the order-preserving transform to recover the signed value. Width was fixed; sign recovery was not.

It would silently corrupt **every negative integer** reaching a client, which is worse than the loud `invalid offset` this PR set out to fix.

### Why I missed it

`pnpm --filter jazz-tools test` was never run — not by the implementing lane, and not by me. On base, `test-ts` short-circuits at the todo example, so CI never reaches `jazz-tools#test` either; fixing the todo suite is what first exposed it.

Two false readings cost me time and are worth recording so nobody repeats them: running that suite inside a lane worktree reports **85** failures, and in another worktree **306**, both artifacts of missing `jazz-napi`/`jazz-wasm` builds there (`Failed to resolve entry for package "jazz-wasm"`). Run in the main checkout with `pnpm --filter jazz-napi build:debug` and `pnpm --filter jazz-tools build` first, the honest numbers are:

| tree | failures |
| --- | ---: |
| base `codex/jazz-core-engine-swap` | 46 (1218 passed) |
| this branch | 46 with **these two tests swapped in** |
| this branch + #1291 | 46 (1220 passed), same two still failing |

The count is flat only because two unrelated timing tests (`keeps fixed-size transaction cost flat as the store grows`) happen to flake the other way. The failure *set* is what matters, and it changed.

### Still true

The `I64`/`I32` width correction is right, the `invalid offset` diagnosis holds, and the todo browser suite genuinely goes from all-failing to 14/14. The fix is incomplete, not wrong.

Fix dispatched. Recommend holding this and #1291 until the sign-recovery case is covered — and the round-trip fixture should gain **negative values for every signed type**, which would have caught this immediately.

aeplay:
**Hold lifted.** The signed-integer corruption is fixed in #1293, which stacks on top of this.

Verified in the main checkout with `jazz-napi`, `jazz-wasm` and `jazz-tools` all rebuilt: the three-PR stack introduces **no new TypeScript failures** relative to base (44 vs base's 46, a strict subset), the two `subscription-manager` corruption tests pass, and the `todo-client-localfirst-ts` browser suite is **14/14**.

Merge order: **#1288 → #1291 → #1293.** Merging the first two without #1293 would ship the negative-integer corruption, so please take all three together.

PR #1291, opened 2026-08-06
(new core) Round-trip the whole ValueType space through the native row codec

aeplay:
Stacked on #1288 — merge that first.

#1288 fixed `I64`/`I32` being read as variable-width. A real bulk import proved that was not the whole story, so this closes the remaining divergence and pins the **entire `ValueType` space** so the class cannot recur silently.

## The remaining defect

TypeScript's descriptor codec wrote and read **only the tag byte** for Groove `Enum` (tag 9). Rust's `ValueType::Enum` carries a serialized `EnumSchema` payload after the tag (`crates/groove/src/records/values.rs:127,171`).

So every field *after* an enum in a descriptor was read from the wrong offset. Rust was correct; TypeScript was wrong. Fixed in `native-row-codec.ts:129` and the Groove-specific type codec at `:179`, with staged-cell encoding now explicitly using the Groove descriptor path (`native-runtime-adapter.ts:3382`) so the **separate Jazz schema-tag namespace is preserved** — Jazz tag 15 is JSON, not Groove `I32`.

## Full round-trip coverage, both directions

| Tag | ValueType | TS → Rust | Rust → TS |
|---:|---|---|---|
| 0–2 | U8, U16, U32 | pass | pass |
| 3–5 | U64, F64, Bool | pass | pass |
| 6–8 | String, Bytes, Uuid | pass | pass |
| 9 | Enum | **pass after fix** | **pass after fix** |
| 10–12 | Tuple, Array, Nullable | pass | pass |
| 13 | Record | pass | pass |
| 14–15 | I64, I32 | pass | pass |

Nesting to depth three: `Nullable<Array<Nullable<I32>>>`, `Array<Record<…>>`, mixed-width tuples with a null member, fixed nullable nulls, and both fixed and variable empty arrays. The enum assertions are **born-red without this fix**.

Fixture and tests extend #1288's: `crates/jazz/fixtures/native_row_codec.json`, `crates/jazz/tests/wire_fixtures.rs:467`, `packages/jazz-tools/src/runtime/native-runtime/native-row-codec.test.ts:46`, all gated by `dev/gates/ts-wire-codec.sh`.

## What I verified, and what I could not attribute

Re-run independently, exit codes unpiped:

- `dev/gates/ts-wire-codec.sh` → **1**, `91 passed`, single failure `encodes the policy graph perf fixture byte-stably` — the pre-existing policy-graph failure documented in `CLAUDE.md`.
- `cargo test -p jazz --test wire_fixtures` → **0**, `3 passed`.

Lane-reported: `cargo test -p jazz` **0** · `cargo test -p groove` **0** (402 passed) · `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0** · `--no-default-features --features test` **101**, the three known `catalogue_sync_integration` failures only.

**Honest limit on attribution.** With this fix the bulk import gets past cell staging entirely — 133 transactions, 26,512 operations, zero errors, where it previously died on the first batch. But that run *also* required rebuilding the NAPI native module, which had been stale since before #1288. I cannot separate this fix's contribution from #1288's in the real-world run. The enum defect is nevertheless real and independently proven by the born-red round-trip assertions above.

### DIFF
--- a/crates/jazz/fixtures/native_row_codec.json
+++ b/crates/jazz/fixtures/native_row_codec.json
@@ -1,4 +1,90 @@
 {
-  "descriptor_hex": "060108726f775f7575696408010a757365725f7469746c650c060109757365725f646f6e650c05010d757365725f7072696f726974790c0f010774785f74696d650e0110757365725f6465736372697074696f6e0c06",
-  "record_hex": "11111111111111111111111111111111010001070000002a000000000000002c00000001427579206d696c6b00"
+  "cases": [
+    {
+      "name": "all_value_types_depth_three",
+      "descriptor_hex": [
+        "15010875385f76616c75650001097531365f76616c75650101097533325f76616c75650201097536345f76616c756503",
+        "01096636345f76616c756504010a626f6f6c5f76616c756505010c737472696e675f76616c756506010b62797465735f",
+        "76616c756507010a757569645f76616c756508010a656e756d5f76616c756509046d6f646502036c6f770468696768",
+        "010b6d697865645f7475706c650a04000e0c050f010b66697865645f61727261790b0f010e7661726961626c655f6172",
+        "7261790b06011a6e756c6c61626c655f61727261795f64657074685f74687265650c0b0c0f010d696e6c696e655f726563",
+        "6f72640d02010b6368696c645f636f756e740f010b6368696c645f6c6162656c0c06010c7265636f72645f61727261790b",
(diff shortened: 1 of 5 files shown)


aeplay:
**Hold lifted.** The signed-integer corruption is fixed in #1293, which stacks on top of this.

Verified in the main checkout with `jazz-napi`, `jazz-wasm` and `jazz-tools` all rebuilt: the three-PR stack introduces **no new TypeScript failures** relative to base (44 vs base's 46, a strict subset), the two `subscription-manager` corruption tests pass, and the `todo-client-localfirst-ts` browser suite is **14/14**.

Merge order: **#1288 → #1291 → #1293.** Merging the first two without #1293 would ship the negative-integer corruption, so please take all three together.

PR #1296, opened 2026-08-07
(new core) [WIP] Numeric aggregate differential oracle, and the maintenance defects it found

aeplay:
> **Currently red by design.** The oracle is in; the fixes for what it found are in progress and will land on this same branch. Do not merge until the arms below are green.

The maintained-vs-one-shot differential oracle mentioned aggregates ~39 times but exercised only **`count`**, over a value space of `Uuid`/`String`/`U64`/`Bool` — no numeric aggregation and no `F64` at all. So nothing checked that a maintained numeric aggregate agrees with a one-shot recompute.

That matters because aggregate output **travels over the wire** as a synthetic result row (`crates/jazz/src/protocol.rs:1555`) rather than being recomputed per replica. The exposure is a client receiving a maintained value, then a later rehydrate recomputing it differently and contradicting what it was already shown.

## What the oracle now covers

Grouped, **multi-row** `F64`/`I64`/`U64` across `count`/`sum`/`avg`/`min`/`max`, a normal-scale F64 control, nullable inputs, and insert→update→retract churn at depths of 10 / 1,000 / 100,000.

Two deliberate design choices:

- **An integer control arm.** `I64`/`U64` aggregates must agree **exactly**. This is what separates "floating point is hard" from "our aggregation is wrong".
- **Tolerance relative to Σ|x|**, not to the result. Under catastrophic cancellation the result approaches zero while absolute error does not, so a result-relative tolerance is unbounded and fails on adversarial input.

`min`/`max`/`count` are asserted **exactly** — they are selection or counting, not accumulation.

## What it found — none of it floating-point

**1. Exact integer sums diverge.** `docs_i64_sum_by_bucket`, seed 11, `fuzz-step-0`: maintained `-5`, one-shot `-1`. `I64 sum` is exact arithmetic; this is a plain correctness bug in incremental maintenance.

**2. `avg` is wrong by a constant, independent of churn depth.**

| depth | avg maintained | avg one-shot | avg error | sum error |
|---:|---:|---:|---:|---:|
| 10 | 0.29333333 | 0.02 | 2.733e-1 | **0** |
| 1,000 | 0.19333333 | 0.02 | 1.733e-1 | **0** |
| 100,000 | 0.19333333 | 0.02 | 1.733e-1 | **0** |

Identical error at 1,000 and 100,000 updates means a **wrong formula**, not accumulated drift — and `sum` being exact in the same runs points at how `avg` is derived from `sum`/`count`.

**3. Normal-scale F64 sum is wrong after one update.** `error=5e-1`, `tolerance=1.599e-13`, `Σ|x|=2.25`, `maintenance_updates=1`. A 0.5 absolute error on a total absolute magnitude of 2.25 after a single update is semantic, not round-off.

**4. Nullable numeric aggregates are rejected** at validation with `OperandTypeMismatch`, so NULL-skipping, all-NULL and empty-group semantics are unsupported on the maintained path — a gap against what #1125 specifies.

## Why "approximate agreement" cannot be the contract yet

A tolerance would have hidden all of this. In the churn arm, Σ|x| ≈ 2e16 gives a computed tolerance of 2.8e7 — the broken `avg` passes comfortably. Specifying a tolerance for values that are wrong for non-numeric reasons would convert real defects into documented noise.

Once the defects are fixed, the supportable statement is a bound proportional to `ε × (input rows + maintenance updates) × Σ|x|`, with update count bounded by recomputation or rehydration after N updates.

## Verification

Reproduced independently of the implementing lane at `JAZZ_SEED_COUNT=20`: three arms fail — `..._oracle`, `..._oracle_f64_approximate_control`, `..._oracle_null_semantics` — with the figures quoted above. `cargo test -p groove` **0** (405 passed), `cargo check --workspace --all-targets` **0**, `cargo fmt --check` **0**, `clippy` **101** (only the known `warm_reopen_differential` lint).

## Relationship to the aggregate PRs

#1125 (SQL semantics) and #1126 (transactional maintenance) both touch this path and neither is merged. The in-progress fix work is checking whether either already addresses any of these defects; if so that changes the plan for both. Worth knowing for #1126 in particular: maintenance is not only non-transactional, it computes wrong values.

### DIFF
--- a/crates/jazz/src/node/tests/m3_differential.rs
+++ b/crates/jazz/src/node/tests/m3_differential.rs
@@ -17,7 +17,10 @@
 // | projection with includes                     | `docs_projected_with_doc_access`            |
 // | relation traversal facade, forward hop       | `docs_relation_facade_direct_access`        |
 // | recursive membership                         | both reachable shapes include transitive hops |
-// | aggregate                                    | `docs_count`                                |
+// | aggregate: count, min, max, sum, avg          | `docs_*_by_bucket`                          |
+// | aggregate numeric value types                 | `F64`, `I64`, `U64`                         |
+// | aggregate nullable / empty inputs              | `m3_aggregate_null_semantics`                |
+// | aggregate cancellation and sustained churn     | `m3_aggregate_churn_curve`                   |
 
 #[derive(Clone)]
 struct DifferentialShape {
@@ -32,7 +35,7 @@ struct DifferentialOracle {
     peers: Vec<PeerState>,
     shapes: Vec<DifferentialShape>,
     rows: Vec<BTreeSet<(String, RowUuid)>>,
-    aggregate: AggregateDifferential,
+    aggregates: Vec<AggregateDifferential>,
 }
 
 struct AggregateDifferential {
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Closing as superseded/already integrated. The extended numeric aggregate oracle is present on `codex/jazz-core-engine-swap`, and `git cherry` confirms this PR’s sole commit is patch-equivalent to the current integration branch. Subsequent merged work also updated the oracle for the collision-free physical aggregate namespace. Any remaining red oracle arms belong in the active test burndown rather than this stale draft.

PR #1297, opened 2026-08-07
(new core) Retain maintained collector output (structured PR 3.6)

aeplay:
The prerequisite that stopped the fourth attempt at PR 4. With this, structured results are constructible **and retained** on the maintained path, so the delivery cut becomes purely a carrier change.

## What was missing

`CollectByTree` was built only inside `if let Some(app_rows) = &request.output.app_rows`, but `app_rows` was populated only for one-shot outputs:

```rust
app_rows: matches!(
    output,
    CurrentQueryProgramOutput::AppRows | CurrentQueryProgramOutput::RelationSnapshot
).then(|| ...)
```

`MaintainedView` was absent from that match, so the collector was never constructed on the **only** path that produces `ViewUpdate`. Earlier work wired the collector into the one-shot path; the maintained path was never connected.

## What this does

- `MaintainedView` now requests recursive collector output for array queries (`crates/jazz/src/node/query_eval.rs:2659`).
- `MaintainedSubscriptionView` (`crates/jazz/src/node/maintained_subscription_view.rs:52`) retains collector records by root, tracks changed roots per tick, and exposes point lookup via `structured_app_row(root)`.
- The peer update-builder boundary already reaches this state; `crates/jazz/src/node/views.rs:169` documents the carrier-ready access path.

**No wire code is touched.** Wire vocabulary, `SyncMessage`, chunking, the v3/v4 question and all four bindings are unchanged, so observable behaviour is unchanged and this lands independently.

## The two risks, addressed

**Authorization.** Collector output stays downstream of policy-filtered query sources — authorization is not re-implemented here. The test plants both an allowed and a **denied** child under the same parent; only the allowed one appears, and the denied child's parent retains an empty collection rather than vanishing.

**`INV-INC-1`.** Per-change work is bounded to collector output for affected parent roots plus map updates for those roots — no scan or re-render of accumulated view state. Retaining a whole rendered tree and re-rendering each tick would be exactly the defect the invariant forbids. The scale canary still passes.

## Tests

`maintained_array_collector_retains_authorized_parent_trees_incrementally` (`crates/jazz/src/node/tests/queries.rs:1415`) covers a true nested `users → todos → comments` retained tree; allowed child present and denied child absent; empty optional collection preserved; and updating one child marking and replacing only its parent's tree.

## Verification

Re-run independently of the implementing lane; exit codes read unpiped.

**Planted positive:** disabling the new `MaintainedView` condition in `query_eval.rs` fails the test at `queries.rs:1481` — so the guard is load-bearing, not incidental.

- the new test → **0**, `1 passed`
- `cargo test -p jazz --test incremental_delivery_canary … -- --exact` → **0**, `1 passed`

Lane-reported: `cargo test -p jazz` **0** · `--no-default-features --features test` **101** (only the three known `catalogue_sync_integration` reds) · `cargo test -p groove` **0** · 300-seed maintained-vs-one-shot oracle **0** · `cargo check -p jazz-sim --benches` **0** · `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0** · clippy **101** (only the known `warm_reopen_differential` lint).

## What remains for PR 4

Only the carrier: wire vocabulary, the four bindings, and fixtures. The structured state now exists on the maintained path and is reachable from the view-update builder.


aeplay:
## Regression found by adversarial review, now fixed (`76264e582`)

An adversarial review lane flagged that this PR could break **parameter-routed maintained array subscriptions**. Confirmed, reproduced through the public API, and fixed on this branch.

**The repro:** `Db::prepare_query_bound` with a root `rank = param("rank")` plus an unbounded `children` array subscription.

| tree | result |
| --- | --- |
| `origin/codex/jazz-core-engine-swap` | **1 passed** |
| this PR, unfixed | **fails** — `collect_by is terminal-only and cannot feed another graph node` |
| this PR, fixed | **1 passed**, including initial reset and incremental related-child delivery |

So my "observable behaviour is unchanged" claim was **wrong**: a subscription that worked on base failed here.

**Correction to my own diagnosis.** I predicted the failure would come from the guard at `lowering.rs:4862` ("tree app projections with routed parameters are not lowered yet"). It did not. For correlated array paths, `available_fields` can omit the introduced route (`__jazz_route_rank`), leaving `root_route_fields` **empty** — so that branch was not entered, and the failure arrived later as a terminality violation instead. Policy claim routes do enter `ResolvedSource.routing_fields` and are included in `source_fields`, so the guarded case is reachable too; there were two failure modes, not one.

**The fix** suppresses the unsupported tree app-row terminal whenever the lowering parameter domain has routed fields, leaving the established maintained fact terminals in place. It falls back to pre-#1297 behaviour rather than erroring, and preserves the existing tree/routed guard.

**Regression guard:** `maintained_array_subscription_with_root_parameter_lowers_and_delivers` (`crates/jazz/tests/structured_result_tree.rs:192`). Verified independently by reverting only `lowering.rs` and keeping the test — it fails with the exact error above, and passes with the fix restored.

## Gates re-run at `76264e582`

`cargo test -p jazz` **0** (`1110 passed`) · `--no-default-features --features test` **101** (only the three known `catalogue_sync_integration` reds) · `cargo test -p groove` **0** (`405 passed`) · incremental-delivery canary **0** · 300-seed maintained-vs-one-shot oracle **0** · `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0** · clippy **101** (known `warm_reopen_differential` lint).

## On the review that caught it

The same review also reported, as CONFIRMED, that this PR's test is orphaned because `crates/jazz/src/node/tests/mod.rs` declares only `mod harness;`. That was **incorrect** — `harness.rs:32` contains `include!("queries.rs")`, so the test compiles and runs under `node::tests::harness::`. The reviewer queried `node::tests::queries::…`, got `running 0 tests`, and read that as absence. Noting it because the same trap is worth avoiding in future reviews: a zero-match filter exits 0.

PR #1301, opened 2026-08-07
(new core) Keep edge/global waits pending and defer edge reads while disconnected

aeplay:
Implements the two behaviours from PR #1064's suite that @nrainhart asked about on #1283:

> There are some tests from the original test suite missing, do you think it'd be hard to support them?
> - "resolves local waits and keeps edge/global waits pending while disconnected"
> - "resolves local reads and defers edge reads while disconnected"

**The answer was: not hard, but it needed a semantics change** — both tests failed *with* the implementation, not without it, because they describe behaviour the runtime did not have. Rather than land red tests on #1283, that PR kept the type-signature fix and this branch carries the behaviour.

## The principle

**Tier is a correctness boundary, not a hint.** An `Edge`/`Global` read or wait asserts "this reflects state the edge/global server acknowledged". While disconnected that assertion cannot be satisfied, so it must not be. And **disconnection is temporary** — outstanding waits survive it and resume on reconnect.

## Two changes, different in kind

**1. Reconnect destroyed the waiters disconnect preserved — a bug, not a design gap.**

`disconnect()` already takes `{ rejectWaiters }`, and `Db.disconnect()` deliberately passed `rejectWaiters: false` (`client.ts:1067`) — the intent was there. But `connect()` called `this.disconnect()` **unparameterized**, taking the `true` default, and `reconnect()` goes through `connect()`. So reconnecting rejected exactly the waiters disconnecting had kept.

Fixed by passing `rejectWaiters: false` on the reconnect path and forwarding the option across the worker boundary. Inconsistent plumbing, not a new contract.

**2. The disconnected read gate was absent, not closed — the real change.**

`PersistentBrowserRuntime.disconnect()` set `connectionReady = null`, and consumers `await this.connectionReady`. Awaiting `null` resolves immediately, so an edge read passed straight through and fulfilled from local data. Replaced with a **reconnect gate**: edge/global waits, reads and server-tier subscriptions hold until the next connection. **Local reads and waits remain ungated** and unchanged.

## Verification

- Both #1064 tests ported: browser suite **6 passed**.
- Focused runtime suite: **16 passed**, adding coverage for a wait outstanding across disconnect *and* reconnect, and for local-versus-edge reads.
- **Red check**: removing the read gate fails `runs local reads and defers edge reads until reconnect` with `expected … length of 1 but got 2` — the edge read was sent while disconnected. The waiter test separately asserts the worker-boundary payload is `{ rejectWaiters: false }`, so the plumbing itself is guarded.
- `cargo test -p jazz` → `1116 passed, 1 failed`: `ws_reader_query_covered_empty_when_existing_row_hidden_by_read_policy`, which I re-ran in isolation → **ok, 1 passed**. Known parallel-load flake.
- `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0** · `pnpm exec oxfmt --check` **0**.

## What can now block indefinitely — read this before merging

This is the accepted cost of "waits stay pending", stated plainly. If a client never reconnects, these never settle:

- edge/global `waitForTransaction` and public write-handle waits;
- edge/global one-shot reads;
- edge/global subscription setup and initial server-backed delivery.

**No timeout, cancellation, or connection-state API is added here** — deliberately. Those are recorded as a 🔶 open question in `SPEC/13_db_api.md` (#1300): the core waits indefinitely, cancellation is caller policy via `Future` drop in Rust and `AbortSignal` in TypeScript, and observable connection state is the complement that lets an application decide.

That follow-up matters. A bulk import that hung on an unbounded global-tier wait was undiagnosable precisely because it could not be cancelled or observed.


nrainhart (approved):
🚀

PR #1308, opened 2026-08-07
(new core) fix: preserve app rows for implicit routed paths

joeinnes:
## What changed

- Restrict the routed tree fallback introduced in #1297 to requests that actually contain a correlated-path projection.
- Preserve the ordinary app-row terminal for routed queries whose tree paths come only from implicit reference closure.

## Root cause

The #1297 fallback used the presence of any projected tree path plus any routing parameter as a proxy for a routed maintained array. Ordinary policy reads can also have both: claim-based policy routing supplies a parameter, while schema references supply implicit tree paths. Those reads incorrectly skipped their app-row terminal and returned an empty result for otherwise public rows.

## Impact

Mixed policies such as `public OR member` return their public rows again, while parameter-routed maintained array subscriptions retain the fact-terminal fallback added by #1297.

## Validation

- `cargo test -p jazz read_policy_branch_or_join_allows_public_or_membership_reads -- --nocapture`
- `cargo test -p jazz --test structured_result_tree maintained_array_subscription_with_root_parameter_lowers_and_delivers -- --nocapture`
- `cargo clippy -p jazz --lib -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/jazz/src/node/query_engine/lowering.rs
+++ b/crates/jazz/src/node/query_engine/lowering.rs
@@ -4861,13 +4861,20 @@ fn lowered_terminals(
     // domain rather than only `root_route_fields` to keep routed maintained
     // array queries on their existing fact-terminal path; the tree collector
     // cannot retain any routed binding fields yet.
-    let has_routed_tree_app_projection = matches!(
-        &request.output.app_rows,
-        Some(AppRowOutputRequest {
-            projection: PayloadProjection::Tree(tree),
-            ..
-        }) if !tree.paths.is_empty() && !routing_param_fields.is_empty()
-    );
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Closing as superseded by #1340. The workaround skipped  because the previous collector could not retain routed fields. The terminal collector now carries route fields explicitly (while hiding them from public rows), and current regression coverage exercises initial plus incremental routed-array delivery. Replaying this fallback would bypass the now-supported  path.

aeplay:
🤖 Clarification: the superseded path is app_rows. The terminal collector introduced by #1340 now retains routed fields explicitly while keeping them out of public rows, so the old fallback that skipped app_rows must not be replayed.

PR #1309, opened 2026-08-07
refactor: shared storage across schema versions

nrainhart:
## Description

Jazz currently gives each schema version its own physical storage tables. This keeps every stored row tied to the descriptor under which it was written, but it also makes a logical-table read span every schema-version partition that may contain data. Physical tables and their indexes accumulate as the schema evolves, even when a migration only renames a table or column.

This proposal replaces that layout with one shared physical storage lineage for each logical table. Schema versions remain immutable application-facing views, but they no longer define separate physical tables. Table and column renames change the logical-to-physical mapping; adding a column extends the shared physical representation; dropping a column removes it from newer logical views without immediately removing its stored values. Writes authored against any supported schema are translated into this shared representation, while reads resolve the relevant current or historical row once and project it into the requested schema.


aeplay (approved):
Directionally great - merging.

Main concerns explored/addressed in followup PRs
- atomic schema & lens publishing #1333 
- direct comparison to lowering to first-class discriminated unions (WIP, will stay only an alternative design until discussed)

PR #1310, opened 2026-08-07
(new core) Unify ColumnType and ValueType into one tag space

aeplay:
🤖 `ColumnType` (schema) and `ValueType` (records) were identical for tags 0–12 and diverged only because `Record` sat at 13 in one and not the other, pushing `I64`/`I32` to 14/15. Two overlapping-but-different tag spaces, distinguished only by which writer you happened to call.

That gap caused real bugs: the row codec read stale `ValueType` positions, and the schema codec called the `ValueType` writer for a `ColumnType` value and asked for a record descriptor on an `i64` column — `missing inline record descriptor for tag 13`.

## Columns can hold records — the restriction was never about columns

`ColumnType` lacking `Record` was not principled. `G-INV-STORAGE-27` says a record-valued type "MUST NOT appear, directly or recursively, in a **durable primary key**" — that is about *position*, and it is enforced by `validate_durable_key_schema`. Both `Value::Record` rejections raise `InvalidDirectRecordStoreKey`. Nothing barred a record in an ordinary column; the variant was simply missing.

So this merges the two types and keeps the real constraint where it belongs.

## The new tag space

```
0 U8   1 U16   2 U32   3 U64   4 I32   5 I64   6 F64   7 Bool
8 String   9 Bytes   10 Uuid
11 Enum   12 Tuple   13 Array   14 Nullable   15 Record
```

Integers contiguous and ascending, then remaining scalars, byte-like, then structures.

**This is a deliberate breaking encoding cut**, coordinated across Rust and TypeScript in one change. The standing rule is "append, never insert" — violating it is what shifted `I64`/`I32` and silently broke an in-flight PR — and it is suspended for this change only.

## Record columns round-trip

Verified, with a public database insert/scan test. Durable-key rejection remains position-based.

## Removed duplication

The duplicate TypeScript schema/row readers and writers, `ColumnType::value_type`, and the Jazz lowering conversion helper.

## A known red went green

**`dev/gates/ts-wire-codec.sh` now exits 0 — `94 passed | 1 skipped`.** The policy-graph byte-stability failure, which has been in every gate brief for weeks as an accepted red, was the fixture missing signed-default storage encodings. Regenerating it from source under the unified encoding fixed it. Verified independently.

## Codec fixes confirmed individually

Rather than inferred from a green suite:

- `I32`/`I64` fixed-width — native-row fixture tests
- `Enum` carries its serialized `EnumSchema` after the tag (now 11) — TS descriptor test
- signed values recovered from order-preserving storage form — native-row signed test plus 10 NAPI schema-default cases
- `ArraySubquery` emits `unbounded` and `offset` — native-query fixture

## Fixture audit, frame by frame

- `native_row_codec.json` — exactly one descriptor frame changed; all substitutions are the new tags. Record value bytes did not move.
- `native_query_codec.json` — unchanged.
- `wire_message_frames.json` — 2 of 18 frames changed: the embedded descriptor and the published schema, both retag consequences.
- policy-graph native schema fixture — regenerated from source; its 21 extra bytes are the required signed-default storage encodings, not unrelated movement.

## Gates

`cargo test -p groove` **0** · `cargo test -p jazz` **0** · `cargo test -p jazz --test wire_fixtures` **0** (4 passed) · `dev/gates/ts-wire-codec.sh` **0** · `cargo check --workspace --all-targets` **0** · `cargo fmt --check` **0** · browser suite **0**.

`clippy` exits 1 solely on the documented `warm_reopen_differential` `type_complexity` lint.

**Two caveats, stated rather than glossed:** the lane reported `--no-default-features --features test` as exit 0, but I re-ran it and the `m3_..._null_semantics` oracle arm still fails — that is base's own red, fixed separately in #1125, not something this PR resolves. And `pnpm --filter jazz-tools test` could not produce a trustworthy exit code here because its workers outlived the command wrapper, so no base-diffed TS failure set is claimed.


aeplay:
🤖 Adversarial review of this PR, plus a golden regenerated in `b4aa8d235`.

## Result: exactly one regression, now fixed

Ran `cargo test -p jazz --no-default-features --features test --no-fail-fast` on this branch and on base with the identical command, and diffed the failure **sets** rather than counts. Both exit 101, 46 binaries each:

| | base `6d796d770` | this PR `cca26c16f` |
| --- | --- | --- |
| failing tests | 15 | 16 |

- **Only on this PR:** `query::tests::canonical_bytes_stability_golden`
- **Only on base:** none

So 15 of the 16 were already failing on base. This PR breaks exactly one thing and fixes none.

## The one failure, and why it is not a defect

The canonical shape encoding hashes column-type tags (`crates/jazz/src/query.rs:4204`), so unifying that numbering with `ValueType` — `0..=15`, `Record` at 15 — changes every query shape id:

```
left:  "3525b634-18b9-5967-8773-96b1f87a39ef"
right: "dd92ae54-eeec-57e1-be75-f3957227bed8"
```

Confirmed deliberate (Anselm). `b4aa8d235` regenerates the golden and records the reasoning at the assertion.

Two things worth having written down, since neither was in the PR description:

- **Every query shape id changes.** `ShapeId` travels in `protocol.rs` subscription messages, so peers either side of this cut compute different ids for the same query and will not match each other's subscriptions. Fine as part of one breaking cut; surprising if it were to land quietly.
- **The blast radius inside the tree is one line.** The golden was the only absolute shape id pinned anywhere; every other `shape_id` assertion compares two shapes to each other and is therefore numbering-independent. That is a good property of the suite and worth preserving.

The `native_query_codec.json` fixture being unchanged is what makes this easy to miss: the query *encoding* is stable while query *identity* moved. The golden was the only thing that caught it.

## What was attacked and held up

Everything else in the review survived, and the two highest-severity attacks were run rather than reasoned about:

- **Old stores are safe.** A RocksDB store seeded on base with old `String`/`I32` schema tags and `I32(-7)`, reopened on this revision, returned the original row unchanged. This holds structurally rather than by luck: catalogue schemas persist as JSON variant names, and historical partitions retain their old schema ids instead of being decoded with the new descriptor. Tag 12 meaning `Nullable` before and `Tuple` after cannot reach stored data.
- **The record round-trip test can actually fail.** With record encoding deliberately disabled in a scratch worktree, `groove db::tests::record_valued_columns_round_trip_through_table_storage` failed at the insert; restored, it passed.
- Pushed past the PR's own coverage: nested records, `Array<Record>`, `Nullable<Record>` holding NULL, and durable-key smuggling attempted direct, inside `Nullable`, inside `Array`, and nested two deep — all correctly rejected. `Tuple<Record>` is unconstructible because tuple members must be fixed-width, so that path is closed by the type system rather than by a check.
- Tag audit across both languages found no missed site. Fixtures changed exactly as claimed: one descriptor in `native_row_codec.json` with `record_hex` byte-identical, `native_query_codec.json` untouched, two of eighteen wire frames.
- No callers of `ColumnType::value_type`, the lowering conversion helper, or the duplicate TS readers/writers remain.
- `dev/gates/ts-wire-codec.sh`: **94 passed, 1 skipped** — this PR fixes the policy-graph byte-stability failure that is red on base.

## Note on the base failures

The 15 pre-existing failures are worth a separate look — they include `scope_revocation_removes_edge_results_without_redacting_local_copy` and four merge/convergence tests across `gset_merge` and `text_document_merge`. Both runs agree on them, so they are consistently failing rather than flaky under parallel load. Not this PR's problem, but larger than the known-red list we have been circulating.

PR #1311, opened 2026-08-07
(new core) Specify policy subplan authorization as INV-RLS-21

aeplay:
🤖 States the limit on an authorization suspension that previously had none. **Specification only — no code changes.** The fix is in flight against it.

## The gap

A policy deciding visibility must often read tables the requesting identity cannot read directly: a document policy consults membership rows, a membership policy consults a project. Those reads run **unfiltered** — policy evaluation does not recurse into policy. That is deliberate and correct; if it recursed, a policy could never see the rows it needs to reach a decision.

But the suspension had **no stated limit**, and the limit is exactly where a bug lives.

## `INV-RLS-21`

> A read performed inside a policy subplan MUST NOT apply the read target's own read policy; policy evaluation does not recurse into policy. The subplan MUST, however, apply every constraint the **evaluating** policy expresses — its filters, its join conditions, and its claim bindings — at every step of that subplan's evaluation, including each hop of a recursive reachability traversal and every step beyond the seed. A claim bound at the seed constrains all subsequent hops. For a recursive policy the constraint applies to each iteration, not to a flattened transitive closure computed without it.

The corollary carries the weight:

> Suspending the target's read policy MUST NOT suspend the evaluating policy's own predicate. A subplan that binds its claim only at the seed and then traverses freely violates `INV-RLS-21`, even though every individual read along that traversal was legitimately unfiltered. "Each read was permitted" is not a defence; the question is whether the evaluating policy's own predicate held throughout.

That last sentence exists because the leak is genuinely defensible read-by-read. Every individual read in the failing case was legitimately unfiltered. The property being violated is about the *whole traversal*, which is why per-read reasoning missed it.

## The known violation, recorded

A two-hop chain — document policy → membership table → project, via a reachability seed bound to `claim("sub")` — lets one principal's seed claim select another principal's document:

```
principal B's seed claim must not select principal A's document
  left:  [RowUuid(61616161-…)]   right: []
```

`PolicyContext::AuthorizationSubplan` resolves source authorization to `System` (`lowering.rs::source_authorization_for_source`), which correctly suspends the target's policy but also drops the evaluating policy's claim binding past the first hop.

Reproducible on `codex/jazz-core-engine-swap` — **pre-existing, not introduced by any open PR**. Reproducer on `repro/1261-two-hop-claim-leak`.

## Why this needed writing down first

An earlier attempt to fix the leak stopped without changing anything, judging that propagating the route alone would produce a correctly-labelled but still insufficient authorization proof. That judgement was sound — but it stalled because the *intended* semantics were nowhere stated. `SPEC/16` covers policy dependency tables only for incrementality, and `SPEC/7` said nothing about transitive policy application at all.

So the implementation's answer to "does a dependency's policy apply?" was held only in code, with no record that it was a decision. This makes it one, and bounds it.

## Registry

`INV-RLS-21` added as `open`/`untested` with `NONE-FOUND`, to be flipped to `now`/`✓` with citations when the fix lands. `dev/gates/invariant-registry.sh` reports 432 rows and the same 30 pre-existing missing citations — this adds none.

`pnpm exec oxfmt --check` **0**. The formatter reflowed chapter 7; I compared word sequences to confirm nothing was lost — **3,640 words before and after, identical order**.


aeplay:
superseded by #1326

PR #1320, opened 2026-08-08
fix(jazz): preserve branch merge-back causality

aeplay:
🤖

## Summary

- make branch merge-back return a typed `Accepted`, `Pending`, or `Rejected` outcome instead of closing the branch optimistically
- reserve the exact merge-back transaction and version payload durably, reuse it across retries and reopen, and block further branch writes/discard while it is pending
- bind settlement to the locally reserved commit unit rather than trusting peer-supplied `source_branch` provenance
- reconcile accepted and rejected reservations safely during recovery, including canonical payload rewrites after schema/lens translation
- observe the complete private/public parent frontier before minting the squash transaction and reject oversized or malformed reservations before recovery allocation

## Correctness boundary

An accepted reservation closes the branch only when the durable transaction and complete admitted version payload match the frozen local unit. A rejected reservation checks the exact durable transaction envelope, clears the reservation, and leaves the branch open. Same-TxId substitutions conflict across normal authority, relay, edge-authority, accepted-finalization, known/staged-known, and shared-staging ingress paths.

The branch spec now records these settlement, retry, recovery, translation, and provenance rules.

## Tests

Coverage includes:

- private-frontier clock observation before squash minting
- accepted, pending, rejected, retry, and reopen behavior
- accepted/rejected durable recovery asymmetry
- canonical reservation recovery after schema/lens translation
- exact commit-unit mismatch and foreign-provenance rejection across ingress routes
- reservation arity, encoded-size, and recovery decode bounds
- seeded merge-back row/deletion and strict-version oracles

An independent adversarial review exercised the provenance, translated-recovery, reopen, and bounds paths and found the final commit safe to publish.

## Out of scope

Ordinary peer sync still does not transport private branch-partition ancestors. A squash received by a peer that lacks its private branch-tip parent therefore remains parked; transport of those ancestors is separate follow-up work.

### DIFF
--- a/crates/jazz/SPEC/11_branches_timetravel.md
+++ b/crates/jazz/SPEC/11_branches_timetravel.md
@@ -181,8 +181,30 @@ merge-back and discard have graduated:
   partition table names.
 - ✅ **Merge-back / discard** (`INV-BRANCH-17`, `INV-BRANCH-18`). Merge-back emits
   one atomic mergeable squash of the branch's net effects into the parent,
-  records typed `Transaction.source_branch` provenance, then flips state to
-  `Merged` with overlay history retained. Content and deletion-register overlay
+  records typed `Transaction.source_branch` provenance, and returns a typed
+  `Accepted`, `Pending`, or `Rejected` outcome. A pending squash has one durable,
+  node-local reservation of its exact `(Transaction, VersionRecord[])` commit
+  unit which is reused across retries and reopen; branch writes and discard are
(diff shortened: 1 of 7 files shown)


aeplay:
🤖 Superseded by #1324.

The replacement makes a branch merge an ordinary local-first transaction targeted at an explicit lineage. Source branch transaction IDs are merger-authored provenance rather than causal parents that downstream peers must hydrate, so the witness/reservation mechanism proposed here is no longer needed.

Keeping this PR closed preserves the design and review history; continued review should happen on #1324.

PR #1325, opened 2026-08-08
fix(jazz): deliver maintained multi-source terminal rows

aeplay:
🤖 Delivers maintained flat-join tuple rows through the existing Groove terminal and Jazz subscription pipeline.

## Why

The core already maintained composite query occurrences, but payload and reset paths lost or replaced the terminal tuple. Public subscriptions therefore could not reliably address fan-out additions, replacements, removals, rehydration, or same-batch netting.

## Changes

- carry composite output occurrence fields and tuple payloads through maintained view lowering, settlement, reset, and protocol updates
- prevent aggregate fact payload handling from overwriting ordinary tuple payloads
- decode authoritative reset payloads as their actual row-shaped flat occurrences
- feed flat app-row terminals from the final policy-filtered joined graph
- preserve stable composite occurrence identity for additions, replacements, and removals
- reject one-shot flat joins because their `(ObjectId, values)` API cannot represent composite occurrence identity safely
- reject direct-core flat-join combinations outside the implemented executable envelope, including empty source lists
- keep the new projected terminal path strictly limited to actual flat joins

## Correctness boundary

This PR enables the same-schema maintained-subscription carrier. It does not claim the three existing cross-schema catalogue cases: those currently settle with zero source facts because joined-source/lens production is missing upstream of this terminal path.

The existing browser hop witness-prefix failure is also inherited: it reproduces identically on the exact parent commit after a clean release WASM rebuild.

## Validation

- independent adversarial review: safe to publish within the stated boundary
- maintained occurrence suite: 4 passed
- direct-core capability-envelope matrix: 1 passed
- fan-out additions, replacements, removals, rehydration, and same-batch add/remove netting
- commit hooks: clippy with warnings denied, sensitive-data guard, formatting
- invariant registry has one canonical `INV-QUERY-23` entry

## Stack

This draft is based on #1323 (which is based on #1322). It is intentionally a sibling of #1324 rather than depending on the branch-lineage redesign.


aeplay:
🤖 Superseded by #1335, which keeps this maintained multi-source carrier and adds the missing public opaque ResultKey, named one-shot joined output, transaction read-your-writes, and real multi-hop identity. Closing this draft so review stays on the complete stack.

PR #1339, opened 2026-08-09
Fix private policy evidence settlement

aeplay:
🤖 Fixes permissions that depend on related rows, especially for live subscriptions and writes.

## What was broken

A policy can allow access to one row by inspecting another row.

Example: a file is readable only if an attachment connects it to a message the user may read.

For a live query to keep evaluating that rule after later changes, the client-side database needs the relevant supporting data. Previously Jazz either omitted that data, causing valid updates to be denied or missed, or mixed it into the query's normal result set.

## What this changes

The server now sends two separate things:

- **query results** — rows returned to application code;
- **authorization evidence** — supporting row versions needed to keep evaluating the approved policy.

The evidence has its own membership and lifetime tracking. It is added only while it contributes to an authorized policy evaluation, removed when that contribution disappears, and never becomes a normal query result merely because it was used by a policy.

This fixes:

- live results whose permission depends on a newly added relationship;
- writes whose permission depends on related data the client had not queried first;
- nullable and claim-bound policy branches;
- revocation when supporting data is removed or replaced.

## Security boundary

This does **not** rely on hiding received bytes from application code. Jazz clients are untrusted, so excluding evidence from query results is an API-semantics boundary, not a confidentiality boundary.

The server must first restrict evidence to rows that participate in the requesting session's authorized policy proof. Sending those bytes is therefore an intentional disclosure made by the enclosing policy, even when the supporting row would not be readable as a standalone query result.

Rows unrelated to an authorized proof are not sent. Tests cover unrelated rows, dangling relationships, other users' rows, revocation, replacement versions, and bounded retention.

If a policy must use a supporting row without disclosing that row to the client at all, raw client-side evidence is the wrong mechanism; that proof would need to remain on a trusted server or be represented by a derived proof instead.

## Validation

- Browser private-read scenarios: 5/5
- Rust policy tests: 89/89, plus the websocket case
- Four-tier policy matrix: 13/13
- Seeded public-result oracle passes
- Wire and fixture tests pass
- Independent adversarial review: no remaining P0/P1/P2 findings

Stacked on #1326. #1338 builds on this PR.


aeplay:
This is an incorrect approach (clients should simply not evaluate permissions) and will be superseded by a clearer implementation of that idea.

PR #1343, opened 2026-08-09
Fix maintained join allocation scaling

aeplay:
🤖

## What changed

When both sides of a maintained join share the same arrangement, update only the buckets for join keys touched by the incoming deltas. The previous implementation cloned the entire accumulated arrangement for every child change.

Touched keys are sorted and deduplicated before buckets are moved and replaced. Deduplication is required because moving a bucket consumes it; processing the same left/right key twice would drop the replacement.

## Why

A single relation/include child write performed O(total accumulated relation state) allocation inside `db.insert`, even though the maintained query emitted one terminal operation. At 20,000 accumulated rows, allocated bytes grew by roughly 10x.

## Result

Exact 20k canary after the fix:

- small: 5,100 allocations / 653,606 bytes
- large: 5,101 allocations / 654,256 bytes

The allocation and byte cost are effectively constant with accumulated state.

## Validation

- all incremental-delivery canaries: 3/3
- Groove: 389 passed, 1 ignored; doc tests passed
- structured result tree: 4/4
- authorization scope re-entry: 3/3
- prepared claim routing: 9 passed, 1 ignored
- parameterized subscription routing and reconnect coverage passed
- planted old full-clone, missing-dedup, and wrong-replacement variants exercised the intended failure boundaries
- strict Groove Clippy, formatting, sensitive-data and clean-tree checks passed

### DIFF
--- a/crates/groove/src/ivm/runtime/join.rs
+++ b/crates/groove/src/ivm/runtime/join.rs
@@ -26,6 +26,18 @@ pub(super) type JoinKey = SmallVec<[u8; 64]>;
 type JoinBucket = HashMap<Bytes, i64>;
 type JoinIndex = HashMap<JoinKey, JoinBucket>;
 
+pub(super) fn touched_join_keys(
+    descriptor: &RecordDescriptor,
+    fields: &[String],
+    deltas: &[RecordDelta],
+    comparison: ValueComparison,
+) -> Result<Vec<Vec<u8>>, IvmRuntimeError> {
+    Ok(keyed_join_deltas(descriptor, fields, deltas, comparison)?
+        .into_iter()
+        .map(|delta| delta.key.into_vec())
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 CI attribution complete with fresh matching N-API/WASM artifacts. Rust: the base fails the relation/include allocation canary at 6.36x allocations / 10.39x bytes, while this PR passes all three incremental canaries and the shared-coverage, warm-reopen, and M3 gates. Browser: base and head both produce exactly 12 failed / 187 passed / 1 skipped / 8 unhandled, in the same inherited terminal codec/subscription clusters; the two order-dependent differing tests pass alone on both revisions. Therefore #1343 removes one deterministic base failure and adds no deterministic Rust/browser failure. The inherited browser cluster is being repaired separately.

PR #1348, opened 2026-08-10
refactor(groove): represent schema versions as enum variants

aeplay:
🤖 Stack 1/2: replace the special-purpose schema-version row representation with first-class Groove enum variants.

## What changes

- Groove can encode tagged enum payloads whose cases carry record descriptors.
- Every enum occurrence has its own durable append-only registry identity; nested enum columns evolve independently without Cartesian flattening.
- Enum projection works recursively for an enum stored in a column or used as the entire physical row.
- Jazz lowers physical schema versions to internal whole-row enum cases.
- The old Groove versioned-row codec and `TableSchemaVersion` path are removed rather than retained as a compatibility path.
- Jazz's public row model is unchanged: whole-row enums are an internal lowering detail and are not exposed to Jazz users.

This deliberately reuses the useful registry, descriptor and projection work from the prior multi-schema implementation while replacing its special-purpose representation. A follow-up stack PR will expose the same enum machinery as a user-facing per-column Jazz type.

## Validation

- `cargo check -p groove`
- `cargo check -p jazz`
- full Groove gate: 391 unit tests plus integration and doctest targets
- Groove variant-table tests: 4 passed, 1 ignored
- Groove versioned-row migration tests: 10 passed
- Jazz physical schema variant pointer/reopen tests
- alternating evolution of two enum columns across three row layouts, including reopen/lens behavior
- append/reorder/removal and caller-supplied registry-ID collision guards
- independent Terra-high adversarial review, including planted noncanonical-varint, registry-rebind and epoch regressions
- public sensitive-data guard

This is intentionally a draft while the current merged Jazz failure matrix is being repaired and the broader canonical suite is rerun on top of this representation change.


crates/groove/SPEC/2_storage_model.md
```diff
@@ -36,6 +36,31 @@ Invariant digest:
...
+Engine-owned schema catalogues and operator-supplied native schema files are trusted
+durable formats: their serialized enum registry identities round-trip exactly. Public
+Jazz schema JSON is a separate, deliberately narrower model and cannot carry a Groove
+enum registry identity; public conversion creates fresh internal descriptors instead.
+
+### Enum registries and variant records
+
+An `EnumSchema` is an ordered set of `EnumCase`s. A selected `EnumValue` carries a
+canonical bounded `u32` case tag and that case's record payload. Tags are dense in
+declaration order; appending a case is compatible, while changing an existing case,
```
nrainhart, line 50: @aeplay can appending a case be forward-compatible? What do we show an old client that's trying to read an enum it doesn't know about? Don't think we can solve this by having a "backwards default" value that returns one of the previous enum values

crates/groove/SPEC/2_storage_model.md
```diff
@@ -36,6 +36,31 @@ Invariant digest:
...
+its tag, or payload descriptor is not. Scalar enum columns use the same declaration
```
nrainhart, line 51: What do you think about supporting enum renames? It's pretty easy to do, and Postgres supports it as well: https://www.postgresql.org/docs/current/sql-altertype.html#:~:text=RENAME%20VALUE,-This

crates/jazz-napi/src/lib.rs
```diff
@@ -2330,6 +2330,10 @@ fn core_subscription_event_to_json(event: &SubscriptionEvent) -> napi::Result<se
                 "reset": reset,
                 "delta": delta,
                 "terminalOperations": terminal_operations,
+                // Terminal operation bytes originate from Rust CurrentRow
+                // records, whose app cells retain their nullable carrier.
+                // WASM's older logical terminal payloads omit that envelope.
+                "terminalPayloadEncoding": "current-row",
```
nrainhart: Do we really need this? We don't need to preserve backwards compatibility

crates/groove/SPEC/2_storage_model.md
nrainhart, line 50: what happens if two concurrent migrations both add a tag?

crates/groove/SPEC/2_storage_model.md
aeplay: Yeah we should allow this, but this seems like a jazz-level / lens concern

crates/jazz-napi/src/lib.rs
aeplay: thanks for catching this

aeplay: addressed in https://github.com/garden-co/jazz/pull/1368

crates/groove/SPEC/2_storage_model.md
aeplay: 🤖 Addressed in #1370, primarily by [`8b845f9cb`](https://github.com/garden-co/jazz/commit/8b845f9cb1cdfae704bce0523127cf7260f0f217) and the maintained-subscription follow-up [`c5ad59cb8`](https://github.com/garden-co/jazz/commit/c5ad59cb866f8c2cff4be194e94f1b533a84d44e).

The rule is projection-sensitive: an old schema view can still use a row when its query does not read, filter, join, order, group, aggregate, or authorize through the unknown enum column. If the query semantically requires the unknown case, projection fails explicitly unless an authored downgrade lens supplies a nullable/default mapping. We never silently substitute an older case or hide the whole row merely because an unused column is newer. Equality against a known case may treat an unknown newer case as non-matching.

That behavior is now normative in `14_lowering_to_groove.md` and covered for scalar/payload values, policies, relations, one-shot queries, and maintained snapshot/reset/delta paths. #1355 adds the public Jazz scalar/payload-enum API on top of these semantics; it does not defer this compatibility behavior.

aeplay: 🤖 Addressed in #1370. The semantic identity is no longer the naked authored tag: it is `GlobalCaseId = (introducing SchemaVersionId, introducing authored ordinal)` for each physical enum occurrence.

The core pieces are the normative model in [`dbe27aba2`](https://github.com/garden-co/jazz/commit/dbe27aba2a42bf89ce91eb9da2ff02cef0627f46), scalar interning/remapping in [`827556e5d`](https://github.com/garden-co/jazz/commit/827556e5db68b7d3e5ac1668a5c48cb9f35ac7df), payload/nested identity lowering through [`5fdd8aa91`](https://github.com/garden-co/jazz/commit/5fdd8aa91823abd59e7d9451650728b6dd52bb77) and [`37d1c09d5`](https://github.com/garden-co/jazz/commit/37d1c09d5b4b1ff18f8568f64f422762d03845ae), and first-introduction ordering in [`f22db736a`](https://github.com/garden-co/jazz/commit/f22db736ad5cb2f113c48ce632b6f51f5e0e02ce).

So concurrent children may both author ordinal `2` without aliasing. The catalogue lineage resolves them to distinct global identities; each node then interns those identities into durable compact local tags in authoritative `CatalogueSeq` order and remaps at write/projection boundaries. Tests cover sibling collisions, payload and nested occurrences, activation/reopen, and a deep branch followed by a later sibling without retagging earlier cases.

aeplay: 🤖 Correction after clarifying the compatibility invariant: my previous reply described the implementation currently in this branch, but an additive schema change must not make an old client query/subscription error.

The intended behavior is: unused unknown enum fields do not affect the row; if an old schema view selects or otherwise depends on an unknown case, that row is not representable and is skipped. Policy dependencies hide the row, and optional includes omit the incompatible child while preserving the parent. We still must not substitute an older case or invent a default.

I am reopening this thread until the spec, tests, and implementation use row-level incompatibility rather than a surfaced query error.

aeplay: 🤖 We’ll handle this compatibility correction in a focused follow-up: unknown appended cases must never surface as an old-client query/subscription error. Rows remain readable when the enum is unused; when the old schema view semantically requires an unrepresentable case, the row is skipped, policy access is hidden, and an incompatible optional child is omitted without dropping its parent. No implicit substitution or invented default.

I’m leaving this conversation unresolved until that follow-up has landed with one-shot, maintained-subscription, policy, include, ordering/grouping/aggregate, and pagination coverage.

aeplay: 🤖 The focused follow-up is now implemented and independently adversarial-reviewed in [#1374](https://github.com/garden-co/jazz/pull/1374).

It changes unknown appended cases from surfaced old-client errors to row-level incompatibility: unused enum fields remain readable; semantically required unknown values omit the row; policies hide it fail-closed; optional incompatible children are omitted without dropping the parent; no older case/default is invented.

The implementation covers direct and nested scalar/payload enums, winner-before-omission across Global/Ahead state, maintained snapshot/reset/deltas, live registry refresh, indexed reads, pagination/aggregation, historical/branch/including-deleted reads, and rejects unrelated projection errors. Independent review found no remaining P0/P1 issues.

PR #1349, opened 2026-08-10
feat: make permission preflights advisory

aeplay:
🤖 Replace ambiguous local permission booleans with explicit advisory results.

## What changes

- Public permission preflights return `allowed | denied | unknown` instead of `boolean`.
- Client-local and offline paths return `unknown` without evaluating policy locally or accessing supporting policy data.
- Authenticated clients can request an authoritative decision from the serving authority; it responds only when history and permission state are ready.
- Requests support cancellation, timeout-to-`unknown`, replay isolation and bounded deduplication without returning supporting rows or policy reasons.
- Trusted policy evaluation remains server-side; ordinary client reads and optimistic writes are unchanged.
- Rust, WASM, native and TypeScript APIs, examples and authorization specs use the same result model.

This prevents a local `true` from being mistaken for authoritative permission when the client may not possess all supporting information. It also avoids turning local policy evaluation into an oracle for data the client is not allowed to receive. Online authoritative requests return definitive `allowed` or `denied`; unavailable or incomplete authority returns `unknown`.

## Validation

- focused Rust permission-advice tests: 7 passed
- targeted native/browser/persistent TypeScript tests: 111 passed, 1 skipped
- TypeScript test typecheck
- affected async subscription/channel suite
- `cargo check -p jazz -p jazz-wasm`
- invariant registry
- independent Terra-high security review
- planted tests for partial authority, update semantics, cancellation, replay, timeout, supporting-data privacy and local policy-oracle escape
- public sensitive-data guard

This remains draft while the broader current-base Jazz suite finishes its separately tracked green-up pass.


packages/jazz-tools/src/runtime/db.ts
```diff
@@ -1390,13 +1391,43 @@ export class Db {
     return client.canInsert(table._table, values, context?.session);
   }
 
-  canRead<T, Init>(table: TableProxy<T, Init>, id: string): boolean {
+  async requestCanInsert<T, Init>(
```
nrainhart: @aeplay can you explain why we need both can* and requestCan* methods?

aeplay: collapsed them, thanks!

PR #1354, opened 2026-08-10
Honor backend default durability tier

### LINKED ISSUE #1353: createJazzContext ignores defaultDurabilityTier and replaces it with node identity
### Affected versions

- `jazz-tools@2.0.0-alpha.52`
- `jazz-tools@2.0.0-alpha.53` (latest published version when verified)

The same behavior is also present on the current `main` branch.

### Description

`BackendContextConfig` accepts `defaultDurabilityTier` through `AppContext`, but the backend context does not pass that configured value to `JazzClient`.

In `packages/jazz-tools/src/backend/create-jazz-context.ts`, `createClient()` currently sets:

```ts
const nodeTier = this.config.tier ?? "edge";

const context = {
  // ...
  tier: nodeTier,
  defaultDurabilityTier: nodeTier,
};
```

As a result, `defaultDurabilityTier` supplied by the caller is silently ignored.

### Reproduction

```ts
const context = createJazzContext({
  appId,
  app,
  permissions,
  driver: { type: "memory" },
  serverUrl,
  backendSecret,
  tier: "local",
  defaultDurabilityTier: "edge",
});

// A query without an explicit tier uses the local tier instead of edge.
await context.asBackend().one(query);
```

This configuration is useful when a backend process should be treated as a local client of its edge sync server, while unannotated query results should still wait for edge durability.

### Expected behavior

The configured default should be used independently from the node identity:

```ts
defaultDurabilityTier: this.config.defaultDurabilityTier ?? nodeTier,
```

Then a caller can use `tier: "local"` for node identity and `defaultDurabilityTier: "edge"` for query defaults.

### Actual behavior

The query default is always forced to `nodeTier`, even when the caller explicitly supplies `defaultDurabilityTier`.

husseinraoouf:
## What changed

- honor `defaultDurabilityTier` when creating a backend Jazz client
- retain the backend node tier as the fallback when no query default is configured
- add regression coverage for a local-tier backend whose queries default to edge
- add a patch changeset for `jazz-tools`

## Why

`BackendContextConfig` exposes `defaultDurabilityTier`, but `createJazzContext` replaced it with the
node's durability tier. This prevented backend clients from using a local node identity while making
unannotated queries wait for edge durability.

Closes #1353.

## Validation

- `pnpm exec vitest run --config vitest.config.ts src/backend/create-jazz-context.test.ts`
- `pnpm exec oxfmt --check packages/jazz-tools/src/backend/create-jazz-context.ts packages/jazz-tools/src/backend/create-jazz-context.test.ts .changeset/backend-default-durability-tier.md`
- `pnpm exec oxlint packages/jazz-tools/src/backend/create-jazz-context.ts packages/jazz-tools/src/backend/create-jazz-context.test.ts`

### DIFF
--- a/.changeset/backend-default-durability-tier.md
+++ b/.changeset/backend-default-durability-tier.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Honor `defaultDurabilityTier` independently from the backend node's durability tier.
--- a/packages/jazz-tools/src/backend/create-jazz-context.test.ts
+++ b/packages/jazz-tools/src/backend/create-jazz-context.test.ts
@@ -230,6 +230,35 @@ describe("backend/create-jazz-context", () => {
     );
   });
 
+  it("BC-U01a: keeps query durability independent from the node tier", () => {
+    const context = createJazzContext({
+      appId: "server-app",
+      app: { wasmSchema: SCHEMA_A },
+      permissions: {},
+      driver: { type: "persistent", dataPath: "/tmp/jazz.db" },
+      tier: "local",
+      defaultDurabilityTier: "edge",
+    });
+
+    context.db();
+
+    expect(mocks.runtimeCtor).toHaveBeenCalledWith(
+      expect.any(String),
+      "server-app",
+      "dev",
+      "main",
+      "/tmp/jazz.db",
+      "local",
+    );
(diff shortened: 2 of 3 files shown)


joeinnes (approved):

joeinnes:
@husseinraoouf Thank you for your contribution!

PR #1355, opened 2026-08-11
feat(schema): expose enum and payload-enum columns

aeplay:
🤖 Stacked on #1373, completing the enum stack above #1348.

## What changed

- lowers string enum columns to Groove `EnumTag` values with a durable registry per physical column occurrence
- adds object-map payload enums such as `{ type: "message", body: string }`, including nullable, defaulted, and array fields
- carries payload cases through public schema/value wire types, catalogue codecs, native row codecs, NAPI/WASM adapters, reads, writes, and subscriptions
- adds typed payload-case query matching with case-local field validation and literal coercion
- keeps declaration order schema-significant and rejects unsafe reorder/removal, invalid payload fields, unsupported nested references/modifiers, and tag overflow
- preserves independent registry identities across schema evolution and derived history/global/ahead storage
- adds exact Rust↔TypeScript structural-hash and registry-ID fixtures, including lossless `u64` decoding

## Why

Enum alternatives need stable physical tag identity across schema versions without combining independent columns into a central row-variant registry. Payload enums extend that model to Rust-style discriminated unions while retaining ordinary object-shaped Jazz rows and precise TypeScript inference.

## Validation

- focused TypeScript payload codec/query/runtime suite: 175 passed, 1 skipped
- `pnpm exec tsc --noEmit`
- `cargo check -p jazz -p groove`
- Jazz enum tests: 30 passed
- Groove enum tests: 14 passed
- all-target Jazz/Groove Clippy with warnings denied
- scalar, nullable, array, default, query-match, subscription, and codec round-trip coverage
- planted regressions for tag/string fallback, descriptor layout, selected-case validation, registry rebinding, and unsafe schema evolution

The restack is based on the independently reviewed concurrent-enum identity, projection, subscription, and introduction-order stack.

### DIFF
--- a/packages/jazz-tools/src/dev/schema-utils.ts
+++ b/packages/jazz-tools/src/dev/schema-utils.ts
@@ -203,9 +203,8 @@ function hashColumnType(writer: StructuralHashWriter, columnType: WasmColumnType
       return;
     case "Enum": {
       writer.byte(9);
-      const variants = [...new Set(columnType.variants)].sort();
-      writer.u64(variants.length);
-      for (const variant of variants) {
+      writer.u64(columnType.variants.length);
+      for (const variant of columnType.variants) {
         writer.stringBytes(variant);
         writer.byte(0);
       }
--- a/packages/jazz-tools/src/runtime/native-runtime/schema-codec.ts
+++ b/packages/jazz-tools/src/runtime/native-runtime/schema-codec.ts
@@ -8,6 +8,11 @@ import type {
   WasmSchema,
 } from "../../drivers/types.js";
 import { PostcardWriter, writeValueType, type ValueType } from "./native-codec.js";
+import {
+  encodeNativeRowValues,
+  storageColumnValueType,
+  writeDescriptor,
+} from "./native-row-codec.js";
 
 const OUTER_ROW_SESSION_PREFIX = "__jazz_outer_row";
 
@@ -187,6 +192,23 @@ function writeDefaultValue(writer: PostcardWriter, columnType: ColumnType, value
       writer.u64(6); // groove::records::Value::String
       writer.string(value.value);
       return;
+    case "EnumPayload": {
+      if (value.type !== "Enum") throw new Error("expected payload enum default");
+      const caseIndex = columnType.cases.findIndex((entry) => entry.name === value.value.case);
+      const entry = columnType.cases[caseIndex];
+      if (!entry || entry.fields.length !== value.value.values.length) {
+        throw new Error("invalid payload enum default case or width");
+      }
+      writer.u64(17); // groove::records::Value::Enum(EnumValue)
+      writer.u64(caseIndex);
+      const descriptor = entry.fields.map((field) => ({
+        name: field.name,
+        valueType: storageColumnValueType(field),
+      }));
+      writeDescriptor(writer, descriptor);
+      writer.bytes(encodeNativeRowValues(entry.fields, value.value.values));
+      return;
+    }
     case "Uuid":
       if (value.type !== "Uuid") throw new Error("expected Uuid default");
       writer.u64(8); // groove::records::Value::Uuid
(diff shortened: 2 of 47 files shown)


packages/jazz-tools/src/dev/schema-utils.ts
```diff
@@ -203,9 +202,8 @@ function hashColumnType(writer: StructuralHashWriter, columnType: WasmColumnType
       return;
     case "Enum": {
       writer.byte(9);
-      const variants = [...new Set(columnType.variants)].sort();
-      writer.u64(variants.length);
-      for (const variant of variants) {
+      writer.u64(columnType.variants.length);
+      for (const variant of columnType.variants) {
```
joeinnes, line 207: This is problematic because the DSL sorts the array: https://github.com/garden-co/jazz/blob/c6b123347bd46b77b047c628dbf2cc16e49e1b89/packages/jazz-tools/src/dsl.ts#L27-L40

When enums were string-valued, this was structurally irrelevant, but with this change, the array order matters, so if someone edits a schema like this: `s.enum('complete', 'incomplete')` -> `s.enum('complete', 'incomplete', 'blocked')`, it'll fail the compatibility check, meaning devs would only be able to add new members which sort after existing members.

On a separate point, we probably shouldn't `localeCompare` in `dsl.ts` because that makes the sort order dependent on the runtime.

packages/jazz-tools/src/runtime/native-runtime/schema-codec.ts
```diff
@@ -251,19 +273,52 @@ export function columnTypeToValueType(type: ColumnType): ValueType {
       return { tag: 6 };
     case "Text":
     case "Json":
-    case "Enum":
       return { tag: 8 };
+    case "Enum":
+      if (type.variants.length > 256) {
```
joeinnes: Should we also check this in `s.enum()`? Other similar checks (empty/duplicates) are there.

joeinnes (changes requested):

aeplay:
🤖 Rebased onto the completed new-core stack and addressed the outstanding review concerns.

- Scalar enum declaration order is now preserved by the DSL and by both Rust and TypeScript structural hashing; no locale-dependent sorting remains.
- Empty, duplicate, and more-than-256 scalar enum definitions are rejected locally.
- A shared Rust↔TypeScript fixture proves reordered variants have distinct exact hashes.
- Payload-enum terminal layouts now validate complete case/payload shapes and safely normalize only registry-rebound descriptors with identical layouts.
- Adapted payload query parameters and NAPI/WASM terminal validation to the current typed-claim and early-bound layout contracts.

Independent adversarial review found no remaining P0/P1/P2 issues. Focused Groove/Jazz enum, schema evolution, maintained subscription, DSL/hash, native codec/query/value, and Rust↔TypeScript fixture gates pass. One broader Jazz enum test and one websocket runtime case are pre-existing failures reproduced on the exact parent.

packages/jazz-tools/src/runtime/native-runtime/schema-codec.ts
aeplay: 🤖 Yes—this is now enforced at `s.enum()` / `col.enum()` construction time alongside the existing empty and duplicate checks.

The shared normalizer rejects an empty list, empty strings, duplicates, and more than the native 256-tag space before schema marshalling: https://github.com/garden-co/jazz/blob/d6b6e221d/packages/jazz-tools/src/dsl.ts#L28-L48

The direct DSL tests are here: https://github.com/garden-co/jazz/blob/d6b6e221d/packages/jazz-tools/src/dsl.test.ts#L15-L35

packages/jazz-tools/src/dev/schema-utils.ts
aeplay: 🤖 Addressed in the rebased tip.

The scalar-enum DSL now preserves the supplied declaration order directly (no `sort()` / `localeCompare`), because position is the durable tag meaning: https://github.com/garden-co/jazz/blob/d6b6e221d/packages/jazz-tools/src/dsl.ts#L28-L48

The TypeScript structural hash consumes variants in that same order: https://github.com/garden-co/jazz/blob/d6b6e221d/packages/jazz-tools/src/dev/schema-utils.ts#L204-L210

Coverage now asserts that an append preserves existing tags while a reorder changes schema identity: https://github.com/garden-co/jazz/blob/d6b6e221d/packages/jazz-tools/src/dsl.test.ts#L37-L52

The Rust side was also corrected to make order significant, and both runtimes consume one shared golden fixture with distinct exact hashes for reordered schemas: https://github.com/garden-co/jazz/blob/d6b6e221d/packages/jazz-tools/src/dev/schema-utils.test.ts#L20-L45 and https://github.com/garden-co/jazz/blob/d6b6e221d/crates/jazz/src/tools/public_api/types/tests.rs#L287-L350

Restoring the old Rust sort/dedup behavior fails that shared fixture.

PR #1382, opened 2026-08-12
ci: pilot representative builds on Latitude runner

aeplay:
🤖 Runs a representative Rust workspace check and correctness-artifact build on the new persistent `jazz-ci` runner.

This is intentionally an experimental, non-required workflow. It records compiler-cache state before and after the build so we can compare the first run with a warm rerun before moving required CI jobs away from Blacksmith.

Base: `codex/enum-pr-1`.


aeplay:
🤖 Closing the standalone pilot now that cold and warm runner validation succeeded. The normal CI workflow will adopt the dedicated runner directly.

PR #1395, opened 2026-08-12
docs: define a shared compiler-cache pilot

aeplay:
🤖 Proposes a measurable pilot where compilations produced on trusted developer machines can warm protected integration CI, using one realizable `sccache` topology.

## Motivating loop

A developer usually compiles a change before opening or updating its PR. Today CI recompiles the same Rust inputs. The pilot tests whether both trusted environments can use one content-addressed R2 bucket so the developer build becomes a CI cache hit.

This deliberately accepts a trust tradeoff: only explicitly trusted devboxes and protected integration jobs may read or write that shared bucket. Untrusted fork and PR jobs receive no remote-cache credential and run behind separate OS/container boundaries.

## Isolation model

- Trusted devboxes and protected integration CI share one dedicated R2 bucket because one ordinary `sccache` daemon cannot read one S3 prefix while writing another.
- Both persistent-runner PR jobs, including `test-ts` and `build-integration`, remain credentialless unless the code is explicitly trusted and isolated.
- Release builds use a separate checkout, `CARGO_TARGET_DIR`, sccache L0, daemon/socket, and release-only bucket. This prevents Cargo or sccache from reusing an object materialized by the developer/CI trust domain.
- Cargo `target`, registries, pnpm stores, and writable network filesystems are never shared across trust boundaries.

The proposal accurately distinguishes long-lived bucket-scoped R2 tokens from short-lived S3 credentials that may add action or prefix restrictions. The dedicated bucket remains the security boundary. Cloudflare audit logs cover control-plane changes; bucket operations metrics and machine-readable build receipts provide data-plane evidence.

## Acceptance evidence

The pilot requires an empty-local-cache, cross-host developer-to-CI hit experiment with receipts recording toolchain, platform, cache namespace, daemon version, hit/miss counts, bytes, timing, and source revision. It also requires denied cross-boundary probes for daemon sockets, local cache, checkout, and `CARGO_TARGET_DIR`, plus explicit cost and latency stop criteria.

Motivating measurements:

- dedicated-runner artifact build: approximately 55 seconds on a recent warm run
- devbox sccache: 822 Rust hits / 264 misses (75.69%)
- observed 0.17 CLI connected to a stale 0.15 daemon, demonstrating why daemon lifecycle and provenance receipts are required

Validation:

- `pnpm exec oxfmt --check dev/CI_REMOTE_CACHE_PILOT.md`
- `git diff --check`
- sensitive-data guard and pre-commit hooks
- independent adversarial review of sccache topology, R2 credentials/telemetry, PR isolation, and release isolation

### DIFF
--- a/dev/CI_REMOTE_CACHE_PILOT.md
+++ b/dev/CI_REMOTE_CACHE_PILOT.md
@@ -112,9 +112,18 @@ cache-poisoning boundary GitHub documents for Actions caches.
 ## Telemetry and acceptance criteria
 
 Add an `always()` CI step that emits `sccache --show-stats` to the job summary
-and a compact receipt (backend, Rust hit/miss count, read/write errors, bytes,
-and elapsed time). Do not report cache "wins" without both a cold baseline and
-two warmed repetitions on the same commit.
+and a compact receipt. The sccache part records its executable/server versions,
+safe configuration identity (never credentials), local `SCCACHE_DIR`, local
+cache size, Rust hit/miss counts, cache read/write errors, and elapsed time.
+Those are compiler/local-disk measurements; neither the displayed cache size
+nor sccache's stats establish R2 requests, bytes transferred, egress, or cost.
+
+The corresponding storage receipt comes from the bucket provider's per-prefix
+metrics and billing/audit exports: stored bytes, object reads/writes/lists,
+download bytes, egress bytes/cost, request cost, and denied/error operations.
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Closing this historical planning PR because it is not part of the integration release and has no implementation to port. A compiler-cache pilot can be proposed afresh when the infrastructure decision is active.

PR #1398, opened 2026-08-12
Fix typed NAPI subscription event marshalling

joeinnes:
🤖 Replaces JSON-shaped NAPI subscription events with typed native objects and makes Rust and TypeScript agree on terminal-row decoding by construction.

## Motivating problem

The query engine can produce several legitimate terminal record shapes: complete CurrentRow carriers, explicitly projected logical rows, flat collectors, and nested tree collectors. TypeScript previously tried to infer which shape it received from descriptor names and nullable wrappers. That was both brittle and hostile to fast decoding: internal layout changes could either reject valid rows or silently swap same-typed fields.

## Early-bound decoding contract

Rust now builds an immutable `TerminalRootLayout` when the maintained result shape is prepared, at the point where the authoritative public `AppRowSchema` and the emitted record representation are both known.

The layout records:

- a stable identity hashed from all decoding semantics
- row-key and every public output slot index, including nested collection fields
- exact public column types and ordering
- the authoritative encoded representation for each slot (`Logical` or `CurrentRow`)
- authoritative hidden fields, so binding/claim route metadata remains internal

NAPI and WASM publish the layout definition before any operation references its ID. TypeScript validates and installs it atomically, compiles a monomorphic decoder once, and caches that decoder per subscription manager. Per-event decoding then uses only the layout ID and fixed slots; it performs no field-name or carrier inference.

Each slot representation is derived during lowering:

- flat collect-all scalar slots that preserve storage wrappers are `CurrentRow`
- explicitly unwrapped projections are `Logical`
- nested/tree collection slots are `Logical`, even when sibling scalar slots remain `CurrentRow`
- bound-query route fields stay in the internal descriptor for result partitioning but are excluded from the public layout

Logical layouts recursively strip TypeScript-only sparse/wildcard metadata. CurrentRow layouts use that same normalized logical shape plus exactly one outer carrier wrapper, preserving a declared nullable value as a distinct inner nullable layer.

## Typed NAPI events

- packed subscription deltas are `Uint8Array`
- delta, rejected, and closed events are discriminated TypeScript unions
- descriptors, terminal ordering, edit payloads, reset semantics, and nested producer fields are preserved
- unknown layout IDs and immutable-ID redefinitions are protocol errors

Direct raw NAPI consumers now receive `delta` as `Uint8Array` instead of `number[]`.

## Stack cleanup

The old PR tip contained a broad claim-routing workaround and unrelated policy-test changes. Those are removed; merged #1396 owns the core claim-routing behavior. No policy tests are skipped and no serving-host escape hatch is included.

## Validation

- native codec and subscription-manager contract: 48/48
- fresh NAPI registered-layout integration: 1/1
- fresh-WASM `alpha-public-flow-gate`: 12/12
- Rust flat/explicit/nested collector carrier regressions
- NAPI and WASM layout serialization checks
- planted wrong-carrier, sparse-wrapper, reordered-slot, unknown-ID, and ID-redefinition failures
- `cargo check -p jazz-napi -p jazz-wasm`
- TypeScript runtime build and typecheck
- full artifact build
- formatting, lint, clippy, rustfmt, diff, and sensitive-data checks

Known friction: artifact provenance includes Git HEAD, so even a TypeScript-only amendment requires rebuilding both NAPI and WASM before claiming fresh cross-language evidence; release WASM optimization took approximately four minutes in this review loop.

### DIFF
--- a/.changeset/typed-native-subscriptions.md
+++ b/.changeset/typed-native-subscriptions.md
@@ -1,0 +1,5 @@
+---
+"jazz-napi": patch
+---
+
+Marshal raw native subscription deltas as `Uint8Array` and publish typed subscription event declarations.
--- a/crates/jazz-napi/index.d.ts
+++ b/crates/jazz-napi/index.d.ts
@@ -108,8 +108,8 @@ export declare class Tx {
 export declare class Write {
   get batchId(): string
   get payload(): Uint8Array
-  wait(tier: string): Promise<undefined>
   writeState(): any
+  wait(tier: string): Promise<undefined>
   close(): boolean
 }
 
@@ -124,6 +124,7 @@ export interface SubscriptionDeltaEvent {
   reset: boolean
   delta: Uint8Array
   terminalOperations: Array<SubscriptionTerminalOperation>
+  terminalLayouts: Array<SubscriptionTerminalLayout>
   settled: boolean
   tier: 'None' | 'Local' | 'Edge' | 'Global'
 }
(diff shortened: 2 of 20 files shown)


aeplay:
🤖 CI is currently stopping before the browser suite on a TypeScript narrowing error in `packages/jazz-tools/src/runtime/napi.core.test.ts:672`: `Property 'item' does not exist on type 'RowDelta<WasmRow>'`.

I reproduced it locally on the PR commits. A narrow guard before reading the item is sufficient:

```ts
if (!delivered || !("item" in delivered)) {
  throw new Error("expected a delivered row item");
}
const deliveredValue = delivered.item?.values[0];
```

Separate verification: stacking #1398's NAPI marshalling commits over #1396 does **not** resolve #1396's remaining private-invite subscription failure; that appears unrelated to this PR.

PR #1437, opened 2026-08-13
test: quarantine concurrent server reopen teardown

aeplay:
🤖 Quarantines one newly measured load-sensitive Rust teardown red: the server-reopen test completes its body successfully, then aborts with `pthread lock: Invalid argument` under concurrent CI. It passes in isolation.

This is stacked on #1433 and keeps the failure explicitly owned in TEST_BURNDOWN rather than allowing intermittent CI red.

Verification:
- exact 45 active + 7 dormant inventory bijection
- default run skips; ignored-only run passes
- swapped-marker adversarial plant rejected
- independent review: no findings


aeplay:
🤖 Superseded: this exact quarantine is already present on the integration branch through the merged Rust burndown stack.

PR #1458, opened 2026-08-13
Release NAPI callbacks on runtime shutdown

mdlawson:
## Summary

After `await JazzContext.shutdown()` resolves, a Node.js process can remain alive because a `napi_rs_threadsafe_function` owned by `NapiRuntime` is still referenced. Storage has been closed, but the native JavaScript callback is not released until the runtime wrapper is garbage-collected.

This PR:

- releases JavaScript-owned callbacks when `NapiRuntime.close()` runs
- clears mutation-error, auth-failure, rejected-batch acknowledgement, and active subscription callbacks before storage shutdown
- adds child-process regressions for `JazzContext.shutdown()` and direct N-API callback ownership
- adds a `jazz-napi` patch changeset

## Minimal reproduction

```js
// repro.mjs
import { mkdtemp, rm } from "node:fs/promises";
import { tmpdir } from "node:os";
import { join } from "node:path";
import { createJazzContext } from "jazz-tools/backend";

const dataPath = await mkdtemp(join(tmpdir(), "jazz-exit-repro-"));

const app = s.defineApp({
  items: s.table({
    text: s.string(),
  }),
});

const context = createJazzContext({
  app,
  appId: "exit-repro",
  driver: {
    dataPath: join(dataPath, "db.jazz"),
    type: "persistent",
  },
  permissions: {},
});

context.db();

await context.shutdown();
await rm(dataPath, { force: true, recursive: true });

console.log(
  `shutdown resolved after ${Math.round(performance.now() - started)}ms`,
);

if (process.env.FORCE_GC === "1") {
  globalThis.gc?.();
}

process.on("exit", () => {
  console.log(
    `process exited after ${Math.round(performance.now() - started)}ms`,
  );
});
```

Without forced garbage collection:

```sh
timeout 3s node repro.mjs
echo $?
```

Observed:

```text
shutdown resolved after 5ms
124
```

The process did not exit before `timeout` killed it.

With forced garbage collection:

```sh
NODE_OPTIONS=--expose-gc FORCE_GC=1 node repro.mjs
```

Observed:

```text
shutdown resolved after 5ms
process exited after 20ms
```

The fact that forced GC immediately allows the process to exit points to a native wrapper/callback lifetime issue rather than unfinished JavaScript work.

## Diagnostics and root cause

The issue originally appeared as a roughly six-second pause after a Vitest integration suite completed.

- `perf` showed no CPU activity during the pause.
- `strace` showed the main thread sleeping in `epoll` and native Tokio threads sleeping in `futex`/`epoll_wait`.
- `process._getActiveHandles()` showed no relevant JavaScript handles.
- Async-hook tracing found one remaining `napi_rs_threadsafe_function` created while constructing `JazzClient`.
- Forcing GC destroyed that handle, after which the process exited immediately.

`JazzClient` always registers a mutation-error callback. The N-API binding wraps it in a `ThreadsafeFunction` and stores it on `RuntimeCore`, but `NapiRuntime.close()` previously only flushed and closed storage. The callback therefore remained referenced until the complete runtime wrapper was garbage-collected.

The React Native close path already clears its mutation-error callback explicitly. Auditing the N-API path also found the same lifecycle issue for optional auth-failure callbacks and active subscription callbacks.

## Fix

`NapiRuntime.close()` now clears callbacks owned by the host binding before performing the fallible storage flush/close operations. This deterministically drops their `ThreadsafeFunction` values even if storage shutdown later returns an error.

## Validation

- confirmed both regression tests time out without the cleanup and pass with it restored
- `pnpm --filter jazz-napi build` (release)
- `pnpm --filter jazz-napi build:debug`
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/napi.shutdown.test.ts`
- targeted Oxlint and Oxfmt checks
- `cargo clippy -p jazz-napi --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/.changeset/napi-shutdown-callbacks.md
+++ b/.changeset/napi-shutdown-callbacks.md
@@ -1,0 +1,5 @@
+---
+"jazz-napi": patch
+---
+
+Release native JavaScript callbacks when `NapiRuntime.close()` is called so Node.js processes can exit without waiting for garbage collection.
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -822,10 +822,11 @@ impl NapiRuntime {
     /// Flush and close the underlying storage, releasing filesystem locks.
     #[napi]
     pub fn close(&self) -> napi::Result<()> {
-        let core = self
+        let mut core = self
             .core
             .lock()
             .map_err(|_| napi::Error::from_reason("lock"))?;
+        core.clear_host_callbacks();
         let flush_result = core.storage().flush();
         let flush_wal_result = core.storage().flush_wal();
         let close_result = core.storage().close();
--- a/crates/jazz-tools/src/runtime_core/mod.rs
+++ b/crates/jazz-tools/src/runtime_core/mod.rs
@@ -527,6 +527,15 @@ impl<S: Storage, Sch: Scheduler> RuntimeCore<S, Sch> {
         self.schedule_mutation_error_delivery_if_needed();
     }
 
+    /// Release callbacks owned by the host binding during runtime shutdown.
+    pub fn clear_host_callbacks(&mut self) {
+        self.mutation_error_callback = None;
+        self.rejected_batch_acknowledged_callback = None;
+        self.auth_failure_callback = None;
+        self.subscriptions.clear();
+        self.subscription_reverse.clear();
+    }
+
     pub fn set_rejected_batch_acknowledged_callback(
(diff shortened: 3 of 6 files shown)


aeplay:
🤖 Superseded by #1730, which ports the still-relevant retained-schema-view shutdown bug to the current N-API runtime and adds planted-sensitive memory and persistent process-exit coverage.

PR #1487, opened 2026-08-14
feat: add ordinary-row text documents

aeplay:
🤖 Reintroduces rich plain text as an ordinary-row feature after the Large Value subtraction.

## Design

- mutable document head plus immutable balanced rope nodes and version roots
- bounded patch frontier for keystroke-level inserts
- flat-string behavior remains preferable for small documents
- historical versions load directly from their root without ancestor-history replay
- ordinary Jazz transactions, permissions, branches, and synchronization only

## Correctness

- Unicode-safe insertion and consolidation
- format-level frontier caps: 64 patches / 16 KiB
- fail-closed ownership, cycle, height, balance, and malformed-frontier validation
- permission-denied rollback and competing-head coverage

## Measurements

For a 100 KiB document with 300 edits, the frontier measured roughly 6.5–6.8 ms p50 versus 15.5–16.4 ms for whole-string rewrites. Closed compressed RocksDB storage was 0.50–0.64 MB versus 0.62–0.96 MB. For 4 KiB end-heavy text the flat string remains faster and smaller, so the receipt recommends a hybrid threshold.

Draft for design and API iteration.

### DIFF
--- a/.changeset/ordinary-row-text.md
+++ b/.changeset/ordinary-row-text.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Add an experimental ordinary-row `TextStore` with Unicode-safe inserts,
+bounded inline patch frontiers, immutable rope snapshots, and exact-version
+reads that do not replay ancestor history.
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Closing this exploratory implementation as superseded by the shared typed-content stack. The corresponding text implementation is #1495, built on the common foundation in #1494.

PR #1488, opened 2026-08-14
feat: add ordinary-row persistent streams

aeplay:
🤖 Reintroduces appendable streams as an ordinary-row persistent tree after the Large Value subtraction.

## Design

- mutable stream head plus immutable bounded byte parts and tree nodes
- append, full read, range read, subscription, and independent historical roots
- ordinary Jazz transactions, permissions, and synchronization only
- full fail-closed tree validation before reads and appends

## Correctness

- validates heights, extents, fanout, cycles, traversal bounds, and physical part lengths
- rejects corrupted roots before authority can advance the stream head
- covers concurrency, deep trees, corruption, and historical snapshots

## Measurements

The compressed RocksDB sweep shows a persistent tree is the safe binary default. A small inline tail is promising only for tiny appends: 128–256 bytes was best for repeated 32-byte binary appends, while 256–512 bytes was best for repetitive text. A 64 KiB tail remains prohibitively expensive for incompressible history. The public default is intentionally unchanged pending policy iteration.

Draft for design, API, and performance iteration.

### DIFF
--- a/crates/groove/Cargo.toml
+++ b/crates/groove/Cargo.toml
@@ -39,3 +39,7 @@ harness = false
 [[bench]]
 name = "micro"
 harness = false
+
+[[bench]]
+name = "stream_layout_storage"
+harness = false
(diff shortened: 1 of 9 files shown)


aeplay:
🤖 Closing this exploratory implementation as superseded by the shared typed-content stack. The corresponding stream implementation is #1496, following the foundation in #1494 and text layer in #1495.

PR #1489, opened 2026-08-14
feat: add ordinary-row queryable JSON documents

aeplay:
🤖 Reintroduces queryable JSON documents as an ordinary-row feature after the Large Value subtraction.

## Design

- mutable document head plus immutable part sets and independently readable roots
- declared scalar projection rows for efficient cross-document queries
- localized edits atomically advance the root and only the affected projection
- no special storage, protocol, synchronization, or history-scan path

## Correctness

- fail-closed root validation for duplicate IDs/pointers, missing parents, malformed payloads, and noncanonical arrays
- exact projection cardinality, old-value, and declared-pointer validation before mutation
- retained historical roots reconstruct without ancestor replay

## Measurements

With compressed RocksDB, 20 × 10 KiB documents and ten edits each used about 2.25 MB with immutable roots versus 4.89 MB for whole-row rewrites. At 100 KiB, immutable roots used about 2.66 MB versus 12.38 MB. Declared projection filtering remained fast; current hydration still scans the document part relation and is an explicit optimization target.

The storage receipt is warm-handle-only because a separate post-reopen transaction-conflict issue currently blocks a valid cold-update receipt.

Draft for design and API iteration.


aeplay:
🤖 Closing this exploratory implementation as superseded by the shared typed-content stack. The corresponding structured JSON implementation is #1498, the final layer above #1494–#1497.

PR #1490, opened 2026-08-14
feat: add ordinary-row mutable files

aeplay:
🤖 Reintroduces mutable files as an ordinary-row feature after the Large Value subtraction.

## Design

- mutable file head plus immutable bounded binary parts and tree nodes
- append, overwrite, insert, full/range reads, and independently readable snapshots
- overwrite path-copies affected leaves and ancestors
- arbitrary insert currently rebuilds the tree and is explicitly tracked as a performance gap
- future blob-backed immutable parts can preserve this logical model while moving bulk egress off the database path

## Correctness

- validates ownership, heights, extents, cycles, traversal bounds, and the 256 KiB part cap
- FileRow reads reload the current head; FileSnapshot is the historical value type
- immutable child tables require explicit update/delete denial and ownership isolation
- authority-settled tests prove denied mutations leave no head, node, or part changes

## Measurements

The native RocksDB receipt uses LZ4 roots, Zstd parts, and bottommost Zstd, and reports apparent and allocated bytes with preallocation/compaction caveats. The initial small receipt measured 7 µs p50 / 8 µs p95 at the storage-layout layer; end-to-end Jazz latency remains the more relevant follow-up benchmark.

Draft for design, API, path-copy insertion, and blob-egress iteration.

### DIFF
--- a/crates/groove/Cargo.toml
+++ b/crates/groove/Cargo.toml
@@ -39,3 +39,7 @@ harness = false
 [[bench]]
 name = "micro"
 harness = false
+
+[[bench]]
+name = "file_layout_storage"
+harness = false
--- a/crates/groove/benches/file_layout_storage.rs
+++ b/crates/groove/benches/file_layout_storage.rs
@@ -95,10 +95,17 @@ fn main() {
     let mut opts = Options::default();
     opts.create_if_missing(true);
     opts.create_missing_column_families(true);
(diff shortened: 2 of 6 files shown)


aeplay:
🤖 Closing this exploratory implementation as superseded by the shared typed-content stack. The corresponding file implementation is #1497, following the foundation, text, and stream layers in #1494–#1496.

PR #1492, opened 2026-08-14
refactor(storage): share deletion history and combine current rows

aeplay:
🤖 Draft intent: land the storage simplification for deletion history and current rows before the new-core alpha release.

## What changed

- Stores deletion/register history in one schema-independent relation keyed by branch lineage, stable physical table ID, row ID, and transaction identity.
- Replaces per-layer current-row composition with a derived combined current row carrying the independently selected content winner, deletion winner/event, visibility, and projected content cells.
- Routes root and branch deletion history, catalogue rename/reopen, recovery, and clock reconstruction through the shared representation.
- Preserves logical-table identities on sync while keeping the shared physical layout node-local.

## Why

Content and deletion are independent histories, but the previous physical layout duplicated deletion-register storage across lineages and required separate current-content/current-register tables plus an anti-join. This branch makes the storage model match the simpler semantic model: immutable content history + one shared deletion register, producing one derived current row.

This is intentionally a draft while independent adversarial review and launch-topology verification continue. It is one of the two planned architectural changes before the next v2 alpha; correctness findings will be fixed on this branch before it is marked ready.

## Validation

- Focused deletion, combined-current, global, recovery, catalogue rename/reopen, physical-table collision, branch, time-travel, history, restore-witness, and replay tests pass.
- Full FD-capped Jazz library gate: 1,303 passed, 3 ignored, 0 failed.
- Jazz library clippy with warnings denied passed.
- Rust formatting and `git diff --check` passed.
- The full gate caught and fixed a legacy three-part in-batch deletion-history lookup; the final implementation uses the full lineage/table/row/transaction key.

## Review focus

- branch-lineage isolation and physical-table identity across compatible schema renames;
- current-row equivalence before and after restart/rebuild;
- deletion/restore winner selection under concurrent content writes;
- sync remaining logical-table-addressed despite the shared local representation;
- no visibility regression in subscriptions, pagination, time travel, or policy evaluation.

### DIFF
--- a/crates/jazz/SPEC/11_branches_timetravel.md
+++ b/crates/jazz/SPEC/11_branches_timetravel.md
@@ -28,6 +28,7 @@ Invariant digest:
 - `INV-BRANCH-16`: A branch-scoped subscription MUST include BranchId in its identity.
 - `INV-BRANCH-17`: A branch merge MUST be calculated locally from readable source and target views and emitted as one ordinary atomic mergeable transaction on the target, with no branch-specific fate or authority admission path; successful merge MUST leave the source branch open.
 - `INV-BRANCH-18`: `Discarded` MUST be the only terminal branch state; discard makes a branch read-only while retaining overlay history, and merge MUST NOT close or otherwise mutate source lifecycle state.
+- `INV-BRANCH-38`: Branch deletion events MUST share the universal deletion history but be keyed by `BranchLineage`; a branch event MUST update only that branch's combined current row and MUST never affect root or sibling visibility.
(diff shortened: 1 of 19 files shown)


crates/jazz/SPEC/11_branches_timetravel.md
```diff
@@ -137,11 +138,14 @@ and siblings: branch overlay writes never affect parent/main current reads
...
 
-Branch overlays are stored in one content-history and deletion-register table
-per `(PhysicalTableId, BranchId)`. Rows retain their authored
-`SchemaVersionAlias`, while `jazz_branch_partitions` records only the stable
-physical table and branch identities. Reads and merge-back project mixed-schema
-winners into their requested schema.
+Branch overlays are stored in one content-history partition per
+`(PhysicalTableId, BranchId)`. Their sparse deletion events instead live in the
+universal deletion history under `BranchLineage::Branch(BranchId)` and the same
+physical-table id. A branch write updates that branch lineage's combined current
```
nrainhart: This sounds weird... why keep branch content history separate but not the deletion history?

PR #1495, opened 2026-08-14
feat(content): add persistent text manifests

aeplay:
🤖 Adds the persistent text implementation on top of the shared typed-content foundation.

## What changed

- Adds content-addressed immutable UTF-8 rope nodes and leaves.
- Applies Unicode-scalar edit tails and materializes full or ranged text from `root + editTail`.
- Promotes bounded tails with localized path copying and AVL rebalancing while preserving old roots.
- Registers `text-v1`, requires byte-tail metadata, derives tail-aware index values, and enforces intrinsic 64-entry / 16 KiB limits at schema activation.
- Adds a complete public text API chapter with runnable setup, publication, editing, materialization, and indexing examples.

## Validation

- Focused text/schema tests and real Node runtime coverage.
- Independent adversarial review of balancing, integrity, registry startup, schema bounds, and tail typing.

This is PR 2 of 5 and depends on the foundation PR.

### DIFF
--- a/crates/jazz/SPEC/12_text_content_adapter.md
+++ b/crates/jazz/SPEC/12_text_content_adapter.md
@@ -1,0 +1,139 @@
+# 12. Text content adapter
+
+## 12.1 Scope and foundation boundary
+
+This chapter specifies only the text adapter for the ordinary content-manifest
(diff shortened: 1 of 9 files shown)


aeplay:

PR #1496, opened 2026-08-14
feat(content): add persistent stream manifests

aeplay:
🤖 Adds the append-oriented stream implementation on top of text and the shared foundation.

## What changed

- Adds content-addressed immutable stream parts and fanout-32 tree nodes.
- Keeps one bounded 256-byte logical append tail and promotes via right-spine path copying.
- Supports full/range/historical materialization, tail-aware length indexing, and conservative merge behavior.
- Fails closed on malformed fanout, height, aggregate-length, conversion, and overflow metadata.
- Adds the stream contract and a complete public API chapter with runnable append/read/index setup.

## Validation

- Focused stream tests plus real Node integration.
- Multi-leaf split/reuse/range/history coverage and extreme-length mutation-sensitivity plants.
- Independent adversarial review; Clippy and formatting pass.

This is PR 3 of 5 and depends on the text PR.

### DIFF
--- a/crates/jazz/SPEC/20_stream_content_adapter.md
+++ b/crates/jazz/SPEC/20_stream_content_adapter.md
@@ -1,0 +1,17 @@
+# jazz — Specification · 20. Stream content adapter
+
+`stream-v1` is declared as one required content-manifest record with a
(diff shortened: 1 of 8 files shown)


aeplay:

PR #1497, opened 2026-08-14
feat(content): add persistent file manifests

aeplay:
🤖 Adds mutable file manifests on top of the shared persistent-content stack.

## What changed

- Adds a content-addressed fanout-32 extent tree with bounded, base-relative file edits.
- Supports overwrite, insert, delete, full/range reads, merge/index derivation, and direct historical roots.
- Consolidates same-length edits with selected-path copying and preserves untouched leaf IDs.
- Authenticates selected parent/child length edges while retaining bounded range-read behavior.
- Adds the file contract and a complete public API chapter with runnable publication, mutation, range-read, and indexing setup.

## Validation

- Focused file tests plus Rocks-backed Node admission/materialization/index coverage.
- Counting-store sensitivity proves bounded tail ranges and localized consolidation.
- Independent corruption, equal-offset, history, and path-reuse review; Clippy and formatting pass.

External object-store signing remains a separate authority-boundary follow-up.

This is PR 4 of 5 and depends on the stream PR.

### DIFF
--- a/crates/jazz/SPEC/21_file_content_adapter.md
+++ b/crates/jazz/SPEC/21_file_content_adapter.md
@@ -1,0 +1,17 @@
+# jazz — Specification · 21. File content adapter
+
+`file-v1` is declared as a required manifest record with `Bytes` edit-tail
+entries. Use `ContentManifestSchema::with_tail_entry_type("file-v1",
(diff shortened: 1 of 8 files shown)


aeplay:

PR #1506, opened 2026-08-15
fix(sync): preserve provisional openings safely

aeplay:
🤖 Preserves provisional opening state across restart and prevents stale read-view traffic from publishing or clearing the wrong snapshot.

- persists opening markers before result/cursor advancement
- clears them only after authoritative non-pending state
- orders eviction cleanup conservatively across crash boundaries
- recovers markers on reopen and database-slot rebuild
- cleans detached exact-key tombstones without installing state
- rejects stale read-view true/false/unsubscribe traffic after binding reuse
- avoids redundant durable true-marker rewrites
- validates marker records strictly

Coverage includes injected write failures, reopen at each eviction phase, A/B read-view reuse, late true/false/duplicate false, stale unsubscribe, detached cleanup, marker churn, and sensitivity plants.

Verified with focused tests, the full Jazz library suite modulo two host file-descriptor failures that passed serially, Clippy, formatting, hooks, and independent adversarial review.

### DIFF
--- a/crates/jazz/src/node/mod.rs
+++ b/crates/jazz/src/node/mod.rs
@@ -1259,6 +1259,10 @@ where
         self.register_physical_current_variant_projections()?;
         self.groove_runtime_token = next_groove_runtime_token();
         self.invalidate_runtime_handles_after_database_rebuild();
+        // The rebuilt Groove registry invalidates runtime handles, not durable
+        // query facts. Rehydrate those facts (including conservative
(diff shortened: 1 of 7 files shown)


aeplay:
🤖 Superseded by #1520 after clarifying the contract: `settled_through` is a durable GlobalSeq cache/dedup cursor, while `settled` is valid only after confirmation from the current live authority connection. This older durable-opening-marker approach should not be merged.

PR #1519, opened 2026-08-15
test: restore default root ordering coverage

aeplay:
🤖 WIP test-burndown follow-up stacked on #1516.

Restores `edge_server_mode::public_root_default_order_and_windows_are_stable_across_reset`, removes its exact active Rust burndown row, and updates the expected active count. The test proves stable default RowUuid ordering and reset/window behavior; a planted title-order mutation fails the root-order assertion.

This is intentionally stacked on #1516 and should be rebased after #1516 is refreshed onto the current integration base. The separate concurrent-history subscription test remains deterministically red and quarantined.

Validation: complete edge_server_mode target (10 passed, 2 ignored); Rust burndown self-test and exact bijection; cargo fmt; clippy; sensitive-data guard.


aeplay:
🤖 Closing as superseded by #1549, which restores the same default-root ordering coverage and adds stricter exact reset, settlement, and window assertions.

PR #1540, opened 2026-08-15
(parked) spike(jazz): model pollable persistent IVM commits

aeplay:
🤖 Draft design spike only. This does not convert the production Jazz runtime or IndexedDB backend yet.

## Question explored

Can the persistent worker's IVM become async/pollable while preserving the existing browser topology and the invariants that:

- main-thread clients remain normal synchronous in-memory Jazz instances using the normal sync protocol
- a whole Jazz `CommitUnit` is atomic at persistence
- subscription/view publication, local Fate, edge fan-out, and upstream/core messages remain invisible until persistence succeeds
- cold persistent page loads may suspend and resume

Leader election, the broker, transport roles, and tab topology are intentionally unchanged.

## Executable seams

- pollable page loads with store-issued request identity
- scheduler-owned cleanup for dropped loads, including ambiguous cancellation and FIFO retry
- serialized persistent batch ownership with explicit request tokens
- terminal quarantine on failed or ambiguous write cancellation
- owned `WorkerTickOutput` carrying IVM notifications, `ViewUpdate`, local Fate, and upstream/core output behind the commit gate
- controlled pending/failing stores proving no early publication or partial batch visibility

## Important production boundary

Current `NodeState` ingestion commits data/IVM writes and then a separate finalize/consistency-marker batch. Production must either merge those into one IndexedDB transaction or persist a recoverable `finalizing(tx)` state and suppress all publication across restart until finalization completes. The spike documents both valid directions; it does not pretend the current path is already atomic.

## Verification

- focused pending, same-poll, failure/retry, ambiguous-cancel, dropped-load, and multi-token cleanup tests pass
- reversible plants proved serialization and cancellation sensitivity
- rustfmt, clippy, diff check, and sensitive-data guard pass

## Deliberate non-goals

- no production `Db::tick` / `NodeState` async conversion
- no IndexedDB adapter integration
- no leader-election or multi-tab topology change
- no higher-level API change

### DIFF
--- a/crates/jazz/SPEC/async_persistent_runtime_spike.md
+++ b/crates/jazz/SPEC/async_persistent_runtime_spike.md
@@ -1,0 +1,107 @@
+# Async persistent runtime spike (design note)
+
+This note is an executable-design spike, not an approved storage API. It holds
+the existing browser leader election, SharedWorker broker, Web Locks,
+`LeaderWorkerConnectionRole`, follower `MessagePort`s, and the normal Jazz sync
+protocol fixed.
+
+## Current seams
+
+`groove::storage::OrderedKvStorage` is synchronous. Its `get`, scans, and
+`write_many` are used directly below `groove::Database::commit_batch`. The
+`WriteOperation` batch is the encoded atomic unit. `NodeState` owns history,
+IVM maintained graphs, and storage; `Db::tick` drives peer messages and turns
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Closing this explicitly parked persistent-IVM spike. The production-oriented successor direction is tracked in #1661.

PR #1571, opened 2026-08-16
test: make direct subscription fixtures local-only

aeplay:
🤖 Makes direct local subscription fixtures state their local-only propagation contract after provisional remote empty openings became authority-gated.

Five failures on the combined stack were stale assumptions, not product regressions:
- two in-process relation tests requested full propagation despite having no authority peer;
- the dynamic server test expected the provisional fail-closed empty event that is intentionally no longer public;
- S1 and S2 direct-Db smokes opened full subscriptions without an authority and blocked until CI timeout.

The direct fixtures now use explicit LocalOnly propagation. The dynamic server test waits for the first authority-admitted add after permissions and Edge settlement, then still proves retightening removes it. No sleeps or remote-settlement weakening were added.

The S2 historical implicit-include QueryCapability diagnostic remains intact and visible.

Evidence:
- both relation tests, dynamic integration, S1, and S2 pass;
- restoring Full propagation makes the local fixtures fail at the missing initial event;
- planting allow-all instead of retightening makes the removal assertion time out;
- independent adversarial review found no P0–P2;
- Rust burndown remains exactly 9 active + 9 dormant.

### DIFF
--- a/crates/jazz-sim/benches/s1_saas.rs
+++ b/crates/jazz-sim/benches/s1_saas.rs
@@ -6,8 +6,8 @@ use std::time::Instant;
 
 use hdrhistogram::Histogram;
 use jazz::db::{
-    Db, DbConfig, DbIdentity, ReadOpts, RowCells, SeededRowIdSource, SubscriptionEvent,
-    SubscriptionStream,
+    Db, DbConfig, DbIdentity, Propagation, ReadOpts, RowCells, SeededRowIdSource,
+    SubscriptionEvent, SubscriptionStream,
 };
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Closing as superseded by #1570. The updated contract is that a Local-tier subscription publishes its truthful local opening, including an empty result, even when propagation is Full; propagation may request upstream work but does not raise the requested observation tier. The LocalOnly fixture changes in this PR were workarounds for the behavior corrected in #1570, and the dynamic-server expectation would now contradict that contract. No commits from this PR should be merged.

PR #1578, opened 2026-08-16
fix: preserve live subscriptions across schema registry growth

aeplay:
🤖 Preserves live subscriptions while installing compatible schema-registry growth, without weakening cold stale-runtime reconstruction.

Dropped multisink receivers now have explicit Arc/Weak liveness leases and can be pruned without waiting for a later notification failure. When live receivers remain, new table/variant/projection cases are installed in place so their channels and runtime token survive. When pruning reaches zero, the runtime performs the full cold rebuild required by stale edge projection state.

All prune, synchronization, and rebuild failures remain inside the catalogue activation rollback/fail-closed boundary.

The cumulative reconnect race also exposed a pre-existing contradiction at the offline Edge boundary: a validated durable Edge may serve before its Core reconnects, but an Edge-accepted write previously retained no downstream fate route. Offline writes now retain a bounded, deduplicated, unbound client obligation. The first authenticated authority binds it and redrives the canonical commit; stale authority epochs remain unable to settle it. Public `wait_for_batch` keeps its fixed 25-second API deadline, while the dedicated negative test uses an explicit short timeout.

Evidence:
- both cumulative CI regressions pass (live history channel and enum/layout subscription);
- dropped-receiver cold rebuild test is mutation-sensitive to disabling rebuild;
- offline route regression proves bind, redrive, exact client delivery, retransmit deduplication, and route retirement;
- removing offline route registration fails at Local vs Edge durability; disabling deduplication fails with two routes instead of one;
- stale persisted-edge reconnect passes 50/50 sequential local stress runs at multiplier 1;
- Jazz lib suite passes 1,404/1,404, with 2 ignored;
- authority handoff, stale-epoch, route-capacity, formatting, clippy, sensitive-data, and commit hooks pass;
- independent adversarial review found no P0–P2.

Fresh CI for the amended SHA is in progress.

### DIFF
--- a/crates/groove/src/db/mod.rs
+++ b/crates/groove/src/db/mod.rs
@@ -1842,6 +1842,16 @@ where
             .unwrap_or(false)
     }
 
+    /// Retire subscriptions whose receiving handles have already been
+    /// dropped, even when no later data delta exists to discover the closed
+    /// notification channel.
+    pub fn prune_dropped_subscriptions(&mut self) -> Result<usize, Error> {
+        self.ensure_not_poisoned()?;
+        self.ivm_runtime
+            .prune_dropped_subscriptions_with_storage(&self.storage)
+            .map_err(Error::IvmRuntime)
+    }
+
     /// Run one IVM tick without base-table writes.
     ///
     /// Commiting a batch ticks automatically; `flush` is useful after creating
--- a/crates/groove/src/db/tests.rs
+++ b/crates/groove/src/db/tests.rs
@@ -2999,6 +2999,23 @@ fn dropping_subscription_receiver_unsubscribes_on_next_message() {
 }
 
 #[test]
+fn dropped_subscription_receiver_can_be_pruned_without_a_later_message() {
+    let temp_dir = tempfile::tempdir().unwrap();
+    let storage = RocksDbStorage::open(temp_dir.path(), &["albums"]).unwrap();
+    let mut database = Database::new(albums_schema(), storage).unwrap();
+    let subscription = database
+        .subscribe_one_sink(GraphBuilder::table("albums"))
+        .unwrap();
+    let subscription_id = subscription.id();
+    assert_eq!(database.runtime_stats().active_subscriptions, 1);
+    drop(subscription);
+
+    assert_eq!(database.prune_dropped_subscriptions().unwrap(), 1);
+    assert_eq!(database.runtime_stats().active_subscriptions, 0);
+    assert!(!database.unsubscribe(subscription_id));
+}
+
+#[test]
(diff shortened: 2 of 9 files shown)


aeplay:
🤖 Open design question for follow-up: define the intended invariant for live subscriptions during schema publication. In particular, which compatible descriptor/registry changes must preserve the existing channel and occurrence identity; which incompatible changes require an explicit reset or recompile; how settlement and known-state advance across that transition; and whether the answer differs for history, maintained, routed, and branch subscriptions. This PR deliberately implements the narrow current boundary—compatible live receivers stay attached, while a cold runtime rebuilds and every activation failure rolls back/fails closed—but does not claim that this is the complete long-term schema-publication contract.

aeplay:
🤖 CI follow-up: the full workspace shard twice timed out in `TestingClient::wait_for_batch` after exactly 25 seconds, while the identical stale-reconnect test passes on this rebased tip in 0.55–0.60 seconds with the multiplier explicitly set to 1. I amended this PR with a one-line test-harness correction: `wait_for_batch` now uses the existing tolerant test-timeout multiplier, matching the neighboring query/connect waits. There is no retry or production/runtime semantic change. Focused stale reconnect, fmt, diff-check, clippy, rustfmt, and the sensitive-data guard pass. A fresh full CI run is underway.

aeplay:
🤖 Follow-up correction: the first one-line multiplier change made the intentional unattainable-tier negative test exceed Nextest’s 180-second cap, so it was reverted. The amended fix now gives normal `wait_for_batch` calls the existing load-tolerant timeout while exposing an exact-timeout test seam; the negative test uses 100 ms and now completes in 0.11s. The stale reconnect test remains green at multiplier 1 in 0.60s. Focused workspace-feature tests, fmt, diff-check, clippy, rustfmt, and the sensitive-data guard pass. Fresh trunk-targeted CI will be the final receipt.

aeplay:
🤖 Root-cause follow-up: the expanded timeout exposed that the stale reconnect was genuinely stranded, not merely slow. On an Edge, the upstream connection is visited before the later client subscriber. Ingesting that client upload queues the Edge→Core outbox after upstream already had its turn; the bounded second pass from #1575 revisited subscriber links only. This amendment also revisits an earlier upstream link when subscriber inbound changed shared state. A new deterministic one-tick regression fails 0-vs-1 Core-bound CommitUnit under the old behavior and passes with the fix; the quiet follow-up emits no replay. The exact stale reconnect, #1575 Core peer regression, 100 ms unattainable-tier regression, workspace feature build, clippy, fmt, and sensitive-data guard pass. Fresh full CI is required before merge.

PR #1631, opened 2026-08-17
(new core) Let React JazzProvider own local-first authentication

joeinnes:
## Summary

- Add `auth="local-first"` to React's `JazzProvider`, so it loads the stored secret before creating the client.
- Render the configured fallback while loading and share the resulting identity with descendant hooks.
- Keep existing configured authentication modes unchanged.
- Simplify the React and Next.js local-first starters and update the canonical documentation.

## Testing

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.react.ts src/react/provider.local-first.test.tsx src/react/provider.devtools.test.tsx` — 5 tests passed
- `pnpm --filter react-localfirst build` — passed
- `pnpm --filter todo-client-localfirst-react-docs build` — passed
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/mcp/build-index.test.ts` — 28 tests passed


aeplay (changes requested):
🤖 Requesting changes: `auth="local-first"` currently discards application scoping.

`LocalFirstJazzProvider` calls `useLocalFirstAuth()` without options even though the provider config contains `appId` and the existing hook supports app/user/session-scoped stores. Consequently, two providers for different apps on the same origin use the same global `jazz-auth-secret`. Applications migrating from `useLocalFirstAuth({ appId })` plus an explicit secret also cannot preserve their existing scoped identity through the new provider API.

Please expose/forward auth-store options or deliberately scope the provider-owned store by `appId`, ensure descendants use that same store, and add a test proving distinct app IDs derive distinct stored identities. Existing focused and full React suites pass, which confirms this isolation boundary is currently untested.

aeplay (approved):

PR #1632, opened 2026-08-17
(new core) Remove the legacy query propagation option

joeinnes:
## Summary

- Remove the legacy `propagate` query option and keep `propagation` as the single public API.
- Mark the removal as a breaking alpha change. Use `propagation: "local-only"` to prevent upstream forwarding.

## Testing

- `pnpm --filter jazz-tools exec tsc --project tsconfig.tests.json` — passed
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/client-tests/for-request.test.ts src/runtime/db.dev-mode.test.ts src/package-root-public-api.test.ts` — 28 tests passed
- `pnpm build:core` — passed

The repository-wide format check only reported the unrelated ignored `.tokensave/config.json`; every changed file passes formatting.

### DIFF
--- a/.changeset/query-propagation-option.md
+++ b/.changeset/query-propagation-option.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
-Remove the legacy `propagate` query option. Use `propagation: "local-only"` to prevent upstream forwarding.
+Breaking change: remove the legacy `propagate` query option. Use `propagation: "local-only"` to prevent upstream forwarding.
--- a/.changeset/query-propagation-option.md
+++ b/.changeset/query-propagation-option.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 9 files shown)


aeplay (changes requested):
🤖 Requesting changes: the legacy `propagate` behavior remains active in lower runtime paths.

This PR removes `propagate` from `QueryExecutionOptions` and TypeScript normalization, but `native-runtime-adapter.ts`, `jazz-napi`, and `jazz-wasm` still recognize `propagate: false` and translate it to `LocalOnly`. A reversible N-API parser test passing `{ "propagate": false }` succeeded, proving that the supposedly removed behavior remains reachable through raw/native runtime option payloads.

Please remove the legacy parsing/forwarding from the native adapter, N-API, and WASM paths, define whether unknown legacy input is rejected or ignored at each boundary, and add focused coverage for those paths.

aeplay (approved):

PR #1633, opened 2026-08-17
(new core) Pass upsert IDs as a positional argument

joeinnes:
## Summary

- Change the upsert signature to `upsert(table, id, data, options?)`.
- Keep mutation options separate from the row ID, consistently with the other row mutation APIs.
- Update the documentation and existing call sites.

## Testing

- `pnpm build:core` — passed


aeplay:
🤖 Non-blocking test-hygiene finding: `packages/jazz-tools/src/runtime/db.schema-order.test.ts` still types its `upsert` mock with the removed `(table, values, { id })` shape. The `as unknown as JazzClient` cast hides the mismatch, so this test cannot statically guard the new positional-ID signature. Please update the mock tuple to include positional `id`.

No functional blocker was found: tracked TypeScript and documentation call sites use the new signature, focused API tests pass, and a planted forwarded-ID regression was caught.

aeplay (approved):

PR #1634, opened 2026-08-17
(new core) Tighten insert and authentication config types

joeinnes:
## Summary

- Rename `CreateOptions` to `InsertOptions` so the public type matches `db.insert`.
- Make `secret`, `jwtToken` and `cookieSession` mutually exclusive in `DbConfig`.
- After local-first authentication resolves, return the minted JWT from `db.getConfig()` without exposing the original secret.

## Testing

- `pnpm --filter jazz-tools build` — passed
- `pnpm --filter jazz-wasm build:fast` — passed

### DIFF
--- a/.changeset/tidy-api-config-types.md
+++ b/.changeset/tidy-api-config-types.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Rename `CreateOptions` to `InsertOptions` and reject conflicting `secret`, `jwtToken`, and `cookieSession` values in `DbConfig` at compile time. After local-first authentication is resolved, `Db.getConfig()` now returns the minted JWT without also exposing the original secret.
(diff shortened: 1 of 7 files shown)


aeplay (changes requested):
🤖 Requesting changes for two test-sensitivity gaps around the tightened public/runtime boundaries.

1. `public-api-types.typecheck.ts` no longer proves that the breaking `CreateOptions` export removal remains enforced. Temporarily re-exporting `CreateOptions` passed the current typecheck; restoring negative `@ts-expect-error` imports made that mutation fail as intended. Please retain negative type assertions for the removed export.

2. `db.local-first-auth.test.ts` covers only two of the three runtime auth conflicts. Temporarily deleting the `secret + cookieSession` guard left all five focused tests green. Please add an untyped/runtime test for that combination so JavaScript callers cannot bypass the tightened TypeScript union unnoticed.

No current implementation defect was found. The focused auth suite and public typecheck otherwise pass; the React Native typecheck failure reproduces unchanged on the base.

aeplay (approved):

PR #1636, opened 2026-08-17
test(jazz): port query integration tests

nrainhart:
## Description

Ports query integration tests we have in `main` to the new core.

This uncovered several regressions, which are partially fixed in this PR:
- remove the old `jazz::tools::{Query,QueryBuilder}` and replace them with `jazz::query::Query`
- allow nullable keys in multi-hop joins
- fix ordered subscription emitting no delta when its sort key changes (Rust-client-only bug)
- decode subscription rows to improve test assertions (only in tests)
- allow applying filters to any of a join's tables (not just the root table), automatically routing unambiguous columns to the right table and failing on ambiguous columns

Remaining failing/ignored tests will be addressed in follow-up PRs.


aeplay (changes requested):
🤖 Requesting changes: the query test target has two reproducible failures on this branch.

1. `crates/jazz-testkit/tests/query/common.rs` decodes subscription `Row.data` with the legacy todo descriptor. Terminal subscription rows now contain the projected terminal payload, so `subscriptions::local_subscription_preserves_final_state_under_rapid_updates` panics with `decode maintained todo title: unexpected end of record`. Reading the projected `title` field through `Row::get` made the focused test pass.

2. `crates/jazz-testkit/tests/query/subqueries.rs` passes `groups.member_ids` as the outer correlation field. The canonical query layer qualifies it again as `groups.groups.member_ids`, so `subqueries::array_subquery_requires_all_array_refs_to_resolve` times out after reporting the unknown column. Passing the unqualified `member_ids` field made the focused test pass.

The full query target reported 18 passed, 2 failed, and 21 ignored. `cargo fmt --all -- --check`, diff whitespace validation, and query-target compilation passed. I also independently reproduced the rapid-update panic during coordinator verification.

Please fix both test call sites before merging. The 21 ignored query tests remain a substantial residual coverage gap, particularly around pagination, recursion, and several subquery paths.

aeplay (approved):
🤖 Follow-up fixes verified after restacking onto #1635. Subscription rows use decoded projected fields, the correlation key is correctly unqualified, and the full query target passes with 20 active tests, 21 pre-existing ignores, and 0 failures. Approving.

aeplay (approved):
🤖 Re-reviewed the current restacked head.

The query target now completes with 19 active tests passing, 22 precisely documented quarantines skipped, and no failures. The Rust burndown gate enforces an exact 22-active + 10-dormant compiled identity bijection. Two tests that are now green were re-enabled; both restored nullable/array-key validation cases are mutation-sensitive. The two newly discovered reds reproduce their documented timeout/hang signatures.

No remaining findings.

aeplay (changes requested):
🤖 Changes requested: the new flat-join filter schema hard-codes every qualified `scope.id` as `Uuid` and removes the table's declared `id` column (`crates/jazz/src/query/validation.rs:340-367`). That conflicts with the stack's declared-ID contract: when a table declares `id`, `scope.id` must use that authored column and `_id` is the explicit physical-row UUID spelling for flat joins.

I confirmed this with a reversible validation test using `users.id: String`, a physical join on `users._id`, and `.filter(eq(col("users.id"), lit("external-user")))`. It fails with `OperandTypeMismatch`, while normalization would resolve `users.id` as the declared source field. Please build the synthetic filter schema from each table's effective declared-ID type (and expose physical `_id` consistently), then add this case as a regression test for both root and joined scopes.

The full current PR range was reviewed at `c68a657a4d2cfbc01c9dbfd643e21b63721f3ca6`. Existing joined-source/alias filter tests pass, integer-width normalization passes, and the burndown gate passes with an exact 18 active + 10 dormant bijection. The checkout is restored clean.

aeplay (approved):
🤖 Fix verified at `f0ec0aa0875d2209c84a6e71c57e83353df7092f`: flat-join filters now preserve authored declared-`id` types on root and joined sources, while `_id` remains a flat-join-only physical UUID alias. Validation, binding revalidation, and normalization agree; ordinary queries do not acquire `_id` alias semantics.

Focused declared-ID, `_id`, existing joined-filter, integer-normalization, and burndown gates pass. An independent adversarial review exercised nested predicate routing and a planted alias regression, found no remaining issue, and restored clean.

PR #1645, opened 2026-08-18
(new core) Remove deprecated API options

joeinnes:
## Summary

- remove the deprecated `authSecretStorageKey` option; use `key` instead
- remove the deprecated `permissions` option from `publishStoredSchema`; use `publishStoredPermissions` separately
- update the affected testing example

## Testing

- `pnpm exec vitest run --config vitest.config.ts src/runtime/auth-secret-store.test.ts src/runtime/schema-fetch.test.ts` — 28 tests passed
- `pnpm --filter jazz-tools build` — passed


aeplay (changes requested):
🤖 Requesting changes for three API-removal gaps.

1. `docs/content/docs/recipes/testing.mdx` replaces `authSecretStorageKey` with `key` on an undocumented `mountApp` configuration. `key` belongs to `useLocalFirstAuth` / `BrowserAuthSecretStore`; it is not a documented `createDb` option. Please show a working `useLocalFirstAuth({ key })` migration or explicitly document and typecheck how this example's `mountApp` forwards it.

2. Schema publication lacks a regression test proving the removed `permissions` field stays out of the wire payload. Reintroducing `permissions: "regression"` into the JSON body left all 12 schema-fetch tests green. Add a mocked `publishStoredSchema` test that asserts the exact POST body is `{ schema }`.

3. The changeset removes public TypeScript members but does not identify the change as breaking. Repository precedent for this alpha line retains a patch bump while explicitly labeling such removals “Breaking change.” Please mark it accordingly and retain concrete migration guidance.

The focused 57 tests, jazz-tools checks, React Native typecheck, and diff checks otherwise pass.

aeplay (approved):

PR #1647, opened 2026-08-18
(new core) Add reactive single-row helpers for framework bindings

joeinnes:
## What changed

Jazz already has `db.one()` for one-time reads, but framework users had to build the reactive equivalent with `useAll(query.limit(1))` and then pull the first item from the array.

This adds a direct reactive single-row API:

- React and React Native: `useOne()` and `useOneSuspense()`
- Vue: `useOne()` and `useOneSuspense()`
- Solid: `useOne()`
- Svelte: `new QuerySubscriptionOne()`

These helpers apply a limit of one automatically.

The result distinguishes between:

- `undefined`: no query was provided, or the query is still loading
- `null`: the query finished and no row matched
- a row: the first matching result

The subscription stays active, so the value updates when the first matching row changes, disappears, or is replaced.

## Why

Developers already expect a framework-level counterpart to `db.one()`. Making it explicit removes repeated array handling and gives loading, empty, and populated states clear types.

## Validation

- Added focused tests for React, Vue, Solid, and Svelte
- Verified loading, empty, populated, and later reactive states
- Verified existing query limits are replaced with `limit: 1`
- Ran formatting and lint checks
- Built the Svelte and Solid bindings
- Ran the todo Svelte browser suite (7 tests)


aeplay (changes requested):
🤖 Requesting changes: Solid `useOne` can leak the previous session's row across a subscription reset.

`solid/useOne` delegates to `solid/useAll`, whose subscription callbacks do not include `onReset`. When the orchestrator resets an entry to pending during a session/auth transition, React, Vue, and Svelte clear their data and return to loading. Solid instead retains the old row with `isLoading: false` until the replacement subscription produces data, transiently exposing prior-session state.

A reversible `useOne` reset test confirmed `callbacks.onReset` is absent. Please register an `onReset` handler that atomically sets `{ data: undefined, isLoading: true, error: null }` and cover the behavior through `useOne`. Existing framework-focused tests, package checks, React Native typecheck, and Solid declaration checks otherwise pass.

aeplay (approved):

PR #1648, opened 2026-08-18
(new core) refactor: use transaction IDs in the public Rust API

joeinnes:
## Summary

- rename public Rust API types, fields, and methods from batch IDs to transaction IDs
- retain `BatchId` within the query engine and `batch_id` in wire and catalogue formats
- update native bindings, examples, and tests to use the public transaction vocabulary

## Testing

- public transaction ID API tests, with and without the runtime feature
- all-target, all-feature Jazz compilation
- Groove and Jazz Server test suites
- Jazz simulator bench compilation
- TypeScript/native wire-codec gate
- core package build


aeplay (changes requested):
🤖 Requesting changes: the public transaction-ID rename is incomplete across bindings, errors, and authoritative specs.

1. Rust serializes `LocalTransactionRecord.transaction_id` as `transactionId`, but the published TypeScript `LocalTransactionRecord` still requires `batchId`. Rejection settlement is also explicitly serialized as `batchId`, and the updated worker-bridge assertion locks in an event containing both vocabularies. Typed `onMutationError` consumers therefore receive an object without the declared `transaction.batchId`. Please make the TS contract, settlement serializers, and tests consistently use `transactionId` (or intentionally preserve the old serialized field everywhere).

2. `wait_for_transaction` still returns user-visible `unknown batch`, `batch was rejected`, and batch timeout messages. Please update this public error surface to transaction terminology.

3. `SPEC/3_transactions.md`, `SPEC/13_db_api.md`, and `SPEC/INVARIANTS.md` still prescribe removed `OpenBatchId`/`BatchId` public symbols and old test names. Specs and the numbered invariant registry are authoritative and need to move with the API.

Focused Rust API tests, workspace check, doctests, and diff checks pass. Existing tests do not exercise rejection-event serialization or wait failure text, which is why these inconsistencies remain undetected.

aeplay (approved):

PR #1650, opened 2026-08-18
(new core) resolve declared id columns before row ids

joeinnes:
## What changed

Resolve a schema-declared `id` column before falling back to Jazz's physical row UUID in predicate normalization, access-path selection, and join-key lowering.

## Why

Several query paths treated every column named `id` as the hidden physical row identity. Queries returned no rows when an application declared its own `id` whose value differed from the physical UUID. The primary-key optimizer could also incorrectly narrow the scan using the declared value.

## Impact

Filters and parent-reference join shapes use the application column when it exists. Tables without a declared `id` retain the existing physical-row-ID behavior.

## Validation

- Added direct declared-ID filter regression
- Added parent-reference join regression

Stacked on #1649.

### DIFF
--- a/crates/jazz/src/node/policy.rs
+++ b/crates/jazz/src/node/policy.rs
@@ -278,15 +278,19 @@ where
                 ))?
                 .schema
         };
+        // `id` resolves to a declared user column when a table has one, so an
+        // internal physical-row probe must use the dedicated access-path API.
         let shape = crate::query::Query::from(table_name)
-            .filter(crate::query::eq(
-                crate::query::col("id"),
-                crate::query::lit(Value::Uuid(row_uuid.0)),
-            ))
             .validate_with_schema_version(schema, schema_version)?;
         let binding = shape.bind(BTreeMap::new())?;
-        self.query_rows_for_link(&shape, &binding, DurabilityTier::Local, identity)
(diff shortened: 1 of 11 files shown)


aeplay (changes requested):
🤖 Requesting changes: declared `id` resolution is still bypassed in multiple query paths.

1. `join_via_column(..., source_column: "id")` unconditionally normalizes a root source `id` to physical `RowId` in `join_via_root_key`, bypassing the new declared-ID lowering. A focused inverse-direction probe with a root declared ID and matching child FK returned no rows.

2. `Predicate::In` and `Predicate::IsNull` still call the old `normalize_operand` path with `has_declared_id = false`. Adapting the new declared-ID filter test to `IN (declared_id)` returned no row; nullable/missing declared IDs are likewise evaluated against the non-null physical UUID.

The reachable-filter access planner also still promotes `id` equality to a physical primary-key scan without checking for a declared `id`; it should share the new guard used by the root path.

The PR's two added focused tests pass, but they cover only equality and the opposite join direction. Please make declared-ID resolution consistent across joins, `IN`, `IS NULL`, and access planning, with mutation-sensitive coverage for each relevant path.

aeplay (approved):

PR #1651, opened 2026-08-18
(new core) use point reads for write policy checks

joeinnes:
## What changed

Use the existing local point-read path when checking update and delete policies for one row.

## Why

Authorization previously materialized every current row in the table and searched the result for the requested UUID. The cost therefore grew with unrelated table contents.

## Impact

Single-row update/delete policy checks now perform bounded lookup work.

## Validation

- Added a storage-read-metrics regression with 40 unrelated rows

Stacked on #1650.

### DIFF
--- a/crates/groove/src/db/query.rs
+++ b/crates/groove/src/db/query.rs
@@ -487,6 +487,16 @@ where
             .unwrap_or(false)
     }
 
+    /// Retire a prepared graph after all bindings have been unsubscribed.
+    /// Long-lived callers should retain and reuse their shapes; one-shot
+    /// callers must retire dynamically compiled shapes to release graph nodes.
+    pub fn retire_prepared_shape(&mut self, shape: PreparedShapeId) -> Result<(), Error> {
+        self.ensure_not_poisoned()?;
+        self.ivm_runtime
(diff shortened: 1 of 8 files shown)


aeplay (changes requested):
🤖 Requesting changes: the new declared-ID point-read path executes a claim-bound authorization graph without preparing and binding it.

`query_rows_for_link_at_row` directly invokes `database.query_graph(lowered_app_rows_graph(...))`, unlike the normal `query_rows_for_link` path that prepares and binds claim-bearing programs. A focused owner-only RLS probe on a table with a declared `id` showed owner access succeeding but a denied read returning `Groove(IvmRuntime(BindingSourceRequiresPrepare))` instead of `false`.

This path is production-reachable through permission-advice reads and repair authorization. It fails closed as an error, but breaks ordinary denied authorization evaluation for declared-ID tables with claim-bound RLS. Please use the prepared/bound path (while preserving the intended point-read optimization) and add owner/denied coverage for this exact configuration.

The scoped update-performance test, recursive write-policy test, and `cargo check -p jazz` otherwise pass.

aeplay (approved):

PR #1652, opened 2026-08-18
(new core) recover from transient storage commit failures

joeinnes:
## What changed

Restore Groove's pre-tick runtime after a failed storage transaction and restore Jazz's sequence clock when authority ingestion fails before durability.

Publication scopes now distinguish a failure before any successful batch from a failure after partial durable progress.

## Why

Groove advances maintained in-memory state before committing storage. A transient storage error previously poisoned the database permanently, even when the atomic transaction wrote nothing and an exact resend would have been safe.

## Impact

An exact resend can succeed after an initial transient storage failure. A multi-batch scope that already committed an earlier batch remains poisoned because it cannot be safely rolled back by restoring memory alone.

## Validation

- Added exact-resend recovery regression
- Updated Groove's failed atomic commit contract test
- Verified nested partial-publication failure remains fail-closed

Stacked on #1651.


aeplay (changes requested):
🤖 Requesting changes: transient-storage recovery currently treats two ambiguous/durable failure states as safely retryable.

1. `ingest_commit_unit` restores `self.clock` after every error from `ingest_commit_unit_once`, including errors that occur after the source transaction was durably committed and assigned a global sequence but a later derived-merge write failed. A reversible three-unit failpoint probe committed source units at sequence 1 and 2, failed the derived merge after unit 2 was durable, then assigned an unrelated unit sequence 2 instead of 3. Reusing a durable global sequence breaks ordering and replay. Restore the clock only across a proven non-durable boundary, or recover/preserve it once source publication succeeded; add this exact regression.

2. `OrderedKvStorage::write_many` exposes only `Result<(), Error>` and does not prove that `Err` means nothing was committed. The new generic rollback/retry path therefore mishandles ambiguous acknowledgements. A probe storage durably applied the batch and then returned `Err`: the first authority ingest discarded its subscription delta; exact resend observed the transaction already durable and emitted no tick, leaving an existing subscriber permanently empty despite durable history. In-process rollback/retry requires a definite-not-committed outcome; unknown or possibly committed results must poison/reopen and recover coherently under `INV-TX-25`.

The existing transient-failure Jazz test and Groove final-failure retry test pass, but neither exercises these post-source-commit or ambiguous-acknowledgement boundaries.

aeplay (approved):

PR #1653, opened 2026-08-18
(new core) cover rejected ancestry and global read sync

joeinnes:
## What changed

Add executable coverage for two synchronization guarantees from the upstream-defect audit.

## Why

The new core rearchitected both problem areas, but lacked focused regressions proving the old failures cannot recur.

## Covered guarantees

- A rejected update does not silence the next fresh commit for that row; the fresh commit reaches authority and is accepted.
- A pending local write appears at local tier but not global tier, and a later authoritative remote winner is used by global reads.

No production behavior changes are included in this layer.

## Validation

- Both new focused regressions pass

Stacked on #1652.

### DIFF
--- a/crates/jazz/src/node/tests/sync/commit_causality.rs
+++ b/crates/jazz/src/node/tests/sync/commit_causality.rs
@@ -91,6 +91,67 @@ fn authority_rejects_later_child_of_rejected_parent_with_cascade() {
             .is_empty()
     );
 }
+
+#[test]
+fn rejected_update_does_not_silence_the_next_fresh_row_commit() {
+    let (_client_dir, mut client) = open_node_with_uuid(node(1));
+    let (_core_dir, mut core) = open_node_with_uuid(node(9));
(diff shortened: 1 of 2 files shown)


aeplay (changes requested):
🤖 Requesting changes: the new global-read test is insensitive to the tier behavior it claims to protect.

The test applies a fate update that turns its “pending local” write into an Accepted/Global write before performing the read, then imports another Global accepted winner. Changing `PeerState::current_rows_update` rehydration from `DurabilityTier::Global` to `Local` still leaves this test passing because both paths ultimately see the same accepted remote winner.

Please retain a higher-HLC unacknowledged local write through the read (and, if server-tier selection is part of the contract, a conflicting local-only server write), assert the observable Local-vs-Global difference and settlement sequence, then verify the authoritative Global winner. The rejected-ancestry test is mutation-sensitive; this finding is limited to the global-read witness.

aeplay (approved):

PR #1657, opened 2026-08-18
(new core) fix Moon Lander write mocks

joeinnes:
## What changed

Updated Moon Lander's `reconcileDeposits` test double to implement the current fluent write API:

- `insert(...).wait({ tier })`
- `update(...).wait({ tier })`

## Why

The production example was migrated away from the removed `insertDurable` and `updateDurable` methods, but its test mock still exposed those old methods. Four tests therefore failed before exercising their assertions.

This is a test-only compatibility update with no runtime behavior change.

## Validation

`pnpm --filter moon-lander-react exec vitest run --config vitest.config.browser.ts tests/browser/writes.test.ts --reporter=verbose`

Result: 18 tests passed.

### DIFF
--- a/examples/moon-lander-react/tests/browser/writes.test.ts
+++ b/examples/moon-lander-react/tests/browser/writes.test.ts
@@ -170,6 +170,38 @@ function mockDb() {
   };
 }
 
+type DeferredWait = {
+  resolve: () => void;
+  reject: (error: unknown) => void;
+};
+
+/**
+ * A db mock whose write handles stay pending until the test settles them.
+ *
+ * This makes the reconciliation promise observable independently from when
+ * the insert/update calls themselves are issued.
+ */
+function mockDbWithDeferredWaits() {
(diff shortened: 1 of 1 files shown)


aeplay (changes requested):
🤖 Requesting changes for a test-sensitivity gap in the repaired write mocks.

The `insert` and `update` mock handles use `async wait` implementations that mutate their capture arrays synchronously and never reject. They therefore cannot prove that `reconcileDeposits` actually awaits durability or propagates wait failures. Replacing `await Promise.all(ops)` with an unawaited `Promise.all(ops)` still left all 18 tests passing.

Please add deferred/rejectable wait handles and assert that `reconcileDeposits` remains pending until every wait resolves and rejects when any wait rejects. The API migration itself is otherwise correct, and removing the `.wait` calls entirely is caught by four existing reconciliation tests.

aeplay (approved):

PR #1660, opened 2026-08-18
Replace core branches with schema-defined branch views

aeplay:
## Before → after

| | Before: core-owned branches | After: schema-defined branch views |
|---|---|---|
| **Storage** | Transactions targeted a core branch lineage backed by dedicated branch metadata/lifecycle tables. | Every content/deletion version carries a canonical schema-derived `BranchKey`; current winners and indices are keyed by branch key. Empty keys are shared data, and one transaction may span keys. |
| **Semantics** | Jazz owned branch identity, parentage, lifecycle, routing, and merge entry points. | Applications model branches as ordinary rows/columns and authorize them with ordinary policies. Jazz only composes a head over an optional live or frozen base, reducing content and deletion independently before queries, policies, joins, and indices. |
| **Probable API** | Branch-specific creation/query/merge methods. | Tables declare stable named dimension bindings through `branchBy`; normal reads accept `branch` plus an optional live or snapshot-qualified `base`; exact and view-relative mutations name the branch view. Merge is an optional high-level helper producing an ordinary atomic transaction. |

`RowUuid` remains global object identity, while each branch key can hold a different incarnation. Tables may bind subsets of uniformly named dimensions; unbound tables are shared. Uniqueness remains scoped to an exact branch key, with distributed uniqueness explicitly left as an open question.

This PR removes the unreleased legacy branch subsystem and implements branch-keyed history, deletion registers, current state, indices, live/frozen branch views, policy traversal, subscriptions/sync, copy-on-write mutations, cross-key transactions, and contribution merges across Rust, TypeScript, WASM, and NAPI. React Native is intentionally out of scope.

Validation includes the full Rust workspace and all-target compile coverage, release `pnpm build:core`, the no-default/testing/zstd core gate, TypeScript/native wire-codec tests, all benchmark smoke scenarios, incremental-delivery canaries, and the 300-seed maintained-vs-one-shot differential oracle. The invariant registry passes while continuing to report pre-existing documented coverage debt.


nrainhart (commented):
A couple of questions:
- Impact on full-table scans: do we need to query over all branches' data and filter out rows that do not belong to the queried branch?
- Same concern as with exclusive transactions: snapshotting reads at the moment the branch was created currently requires traversing the whole history for each row

PS: not a big fan of the term *branch incarnation*, I'd prefer to use *branch version*.

aeplay:
🤖 Changes requested — remote branch-key validation is a correctness/security blocker.

Incoming `VersionRecord.branch_key` values are decoded and persisted without verifying that they are the complete, sorted, typed canonical key for the authored table, or that content cells agree with that key. Local commits enforce this agreement, so remote ingest currently bypasses INV-BVIEW-1/3/4.

Adversarial reproduction: mutate a valid remote branch-A commit so its wire `branch_key` is empty. The commit is accepted and assigned `GlobalSeq(1)`. This permits a peer-supplied version to be stored/routed in a different incarnation from its branch-dimension cells, with absent dimensions potentially reinterpreted as migration defaults later.

Please validate every incoming content and deletion branch key against the authored schema before any batch write: require exact dimension coverage, canonical order and types, and content-cell agreement. Add atomic no-residue coverage for malformed multi-version units plus missing, duplicate, extra, out-of-order, wrong-type, and deletion keys.

Focused existing branch tests passed; the planted malformed-key test demonstrated the missing boundary and was restored.

dev/BRANCH_VIEW_IMPLEMENTATION_PLAN.md
nrainhart, line 1: Is it worth keeping this files checked out in the repo?

crates/jazz-napi/index.d.ts
```diff
@@ -49,14 +49,25 @@ export declare class NapiDb {
   subscribeRelationQuery(queryJson: string, opts?: { tier?: string; local_updates?: string; propagation?: string; include_deleted?: boolean } | undefined | null): Subscription
   subscribeRelationQueryForIdentity(queryJson: string, author: Uint8Array, opts?: { tier?: string; local_updates?: string; propagation?: string; include_deleted?: boolean } | undefined | null): Subscription
   insertWithIdEncoded(table: string, rowId: Uint8Array, cells: Uint8Array, updatedAtMs?: number | undefined | null): Write
+  insertWithIdEncodedInBranch(table: string, rowId: Uint8Array, cells: Uint8Array, branch: JsonValue): Write
+  insertWithIdEncodedInBranchForIdentity(table: string, rowId: Uint8Array, cells: Uint8Array, branch: JsonValue, author: Uint8Array): Write
```
nrainhart, line 53: The public Rust API keeps growing; it'd be good to consolidate it at some point. Especially considering this is what future bindings will need to integrate with

packages/jazz-tools/src/runtime/client.ts
```diff
@@ -380,6 +480,18 @@ function encodeQueryExecutionOptions(options: InternalQueryExecutionOptions): st
...
+    read_view?: {
+      source: {
+        BranchView: {
+          head: WireBranchSelector;
+          base?:
+            | { Current: WireBranchSelector }
+            | { Snapshot: { branch: WireBranchSelector; snapshot: unknown } };
+        };
+      };
+      schema: "Current";
```
nrainhart: What's the reason for having this `schema` param?

docs/content/docs/reference/inspector.mdx
```diff
@@ -22,7 +22,8 @@ Open [v2.inspector.jazz.tools](https://v2.inspector.jazz.tools/) and enter:
 - **App ID**&hairsp;—&hairsp;Jazz app namespace to inspect
 - **Admin secret**&hairsp;—&hairsp;app admin credential
 - **Env**&hairsp;—&hairsp;logical environment such as `dev`, `staging`, or `prod`
-- **Branch**&hairsp;—&hairsp;logical branch, usually `main`
+- **Branch**&hairsp;—&hairsp;legacy connection namespace, usually `main` (this does not select a
```
nrainhart, line 25: Should we just get rid of this?

docs/content/docs/reference/inspector.mdx
```diff
@@ -48,14 +49,14 @@ is useful when you want to confirm what policy bundle the server is enforcing.
...
-mode, branches, and compiled query JSON, then jump straight into the matching table view in Data
+mode, branch-view options, and compiled query JSON, then jump straight into the matching table view in Data
 Explorer.
 
 ### Schema switching
 
 If an app has multiple published schema hashes, the inspector lets you switch between them without
 reconnecting. This is useful when checking migrations, comparing stored shapes, or debugging data
-that still lives on older schema branches.
+that was authored under an older physical schema.
```
nrainhart:
```suggestion
that was authored under an older schema.
```

docs/content/docs/reference/internals.mdx
```diff
@@ -340,7 +340,8 @@ Migrations in Jazz produce **lenses** — bidirectional transformations between
 such as adding, removing, or renaming columns and tables.
 
 At query time, Jazz can use lens paths to read older stored data through the current schema. At
-write time, updates to older rows are written back into the current schema branch via copy-on-write.
+write time, it projects updates through the lens path while retaining the row's physical schema
+identity. Schema evolution and application branch views are separate mechanisms.
```
nrainhart:
```suggestion
identity.
```

docs/content/docs/schemas/migrations.mdx
```diff
@@ -14,15 +14,16 @@ Most migration systems are one-way: rewrite every row to the new shape, then cut
 
 Jazz keeps every schema version addressable by hash and translates rows between them on read and write. Clients on different versions stay interoperable, and nothing on disk is rewritten when you ship a new schema.
```
nrainhart:
```suggestion
Jazz keeps every schema version usable and translates rows between them on read and write. Clients on different versions stay interoperable, and nothing on disk is rewritten when you ship a new schema.
```

docs/content/docs/schemas/migrations.mdx
```diff
@@ -14,15 +14,16 @@ Most migration systems are one-way: rewrite every row to the new shape, then cut
...
 
 <LensDiagram />
 
-Storage is partitioned by schema hash: rows written under a given schema live in
-that schema's own branch (`env-{hash}-userBranch`) and stay there unchanged.
-Non-adjacent reads compose lenses in sequence to bridge multiple schema versions.
+Rows retain the physical schema identity under which they were written. A read in another compatible
+schema uses lenses to project those rows into the requested shape; schema versions are not
+application branches. Non-adjacent reads compose lenses in sequence to bridge multiple schema
+versions.
```
nrainhart, line 25:
```suggestion
Rows retain the physical schema identity under which they were written. A read in another compatible
schema uses lenses to project those rows into the requested shape. Non-adjacent reads compose lenses in sequence to bridge multiple schema
versions.
```

docs/content/partials/create-jazz-client-reference.mdx
```diff
@@ -11,7 +11,7 @@
 | `driver`                | no       | Storage driver: `{ type: "persistent" }` (default) or `{ type: "memory" }`.                                                                                                                 |
 | `dbName`                | no       | Local database name. Defaults to `appId`.                                                                                                                                                   |
 | `env`                   | no       | Environment label (for example `"dev"`, `"prod"`).                                                                                                                                          |
-| `userBranch`            | no       | Branch name. Defaults to `"main"`.                                                                                                                                                          |
```
nrainhart, line 14: Ditto, let's drop this

docs/content/docs/concepts/branches.mdx
```diff
@@ -1,84 +1,216 @@
...
+`base_branch_id` or `status`; the app resolves them before constructing a view. This keeps workflows
+such as review, archival, nested drafts, and environment promotion in userland.
+
+## Declare branch dimensions
+
+A branch dimension has a stable UUID identity, a schema-wide selector name, a key-encodable type,
+and a migration default. Tables bind ordinary columns to any subset of those named dimensions.
+
+```ts title="runtime schema excerpt"
+const branchDimension = {
```
nrainhart: It'd be nice to show an example using the end-user API instead of the JSON representation

docs/content/docs/concepts/branches.mdx
```diff
@@ -1,84 +1,216 @@
...
+  id: "31313131-3131-3131-3131-313131313131",
+  name: "branch",
+  columnType: { type: "Uuid" },
+  migrationDefault: {
+    type: "Uuid",
+    value: "00000000-0000-0000-0000-000000000000",
+  },
```
nrainhart: I have several questions about how you modelled branch dimensions:
- why do we need the `id` to be explicit? Couldn't we support renaming branches via a migration, as we do for tables and columns?
- why do we need all of this separate information? Couldn't we just have a reference to an existing row?

nrainhart (commented):
Did a first pass over the ts side. My biggest concern right now is that we've designed the whole user-facing API without having actually used it in a real app. We should do that or mark this API as experimental.

crates/jazz-sim/benches/s9_durable_execution.rs
```diff
@@ -1048,7 +1048,12 @@ fn open_worker(node_uuid: NodeUuid, edge_uuid: NodeUuid, schema: JazzSchema) ->
         outbound,
         inbound,
         _upstream: upstream,
-    }
+    };
+    worker.edge_peer.set_ship_complete_exclusive_payloads(true);
```
nrainhart, line 1052: Why is this a setting? Isn't this something we always want?

packages/jazz-tools/src/runtime/testing/wasm-runtime-test-utils.ts
```diff
@@ -90,20 +90,20 @@ export async function createWasmRuntime(
   opts?: {
     appId?: string;
     env?: string;
-    userBranch?: string;
+    identityScope?: string;
```
nrainhart: What's this needed for?

crates/jazz/SPEC/11_branch_views_time_travel.md
```diff
@@ -0,0 +1,440 @@
...
+    "todos": {
+      "columns": [{ "name": "workspace_id", "column_type": { "type": "Uuid" }, "nullable": false }],
+      "branchBy": ["workspace_id"]
+    }
+  }
+}
+```
+
+Entries naming missing, nullable, duplicate, or non-key-encodable columns fail
+schema admission. A branch column is immutable after insertion. Across all
```
nrainhart, line 118: Do we validate that branch columns are not modified?

nrainhart (commented):
Overall this looks good! I'll take another look next Monday

crates/jazz-sim/benches/s9_durable_execution.rs
aeplay: 🤖 I audited this rather than making it unconditional. The setting distinguishes trusted writer relays, which need the complete accepted exclusive transaction to safely author later exclusive transactions, from ordinary readers, which intentionally receive only view-scoped payloads. Publication also ignores the preference for identity-scoped client links. I kept the behavior and clarified that distinction on the setter.

packages/jazz-tools/src/runtime/testing/wasm-runtime-test-utils.ts
aeplay: 🤖 It gives multiple logical test peers deterministic but distinct node/author identities when they share an app and environment. I renamed the option from `identityScope` to the more direct `peerId`, kept NAPI/WASM helpers consistent, and removed the unused `tier` and `useBinaryEncoding` compatibility options and their stale caller.

crates/jazz/SPEC/11_branch_views_time_travel.md
aeplay: 🤖 Yes: mutation admission derives the canonical branch-column cells from the selector and rejects an explicitly authored disagreement; parents cannot cross canonical branch keys; replicated versions are revalidated against their stored key and branch-column cells. I added a public TypeScript API regression test that attempts to update `branch: "draft"` to `branch: "published"`, asserts rejection, and verifies neither branch view was corrupted. The spec now spells out these enforcement points.

PR #1663, opened 2026-08-19
fix premature browser edge subscription seeding

joeinnes:
## Summary

- make server-backed browser subscriptions wait for the worker's confirmed initial result instead of reporting local-storage data too early
- preserve local optimistic state for apps that explicitly call `Db.disconnect()`, with server-backed updates resuming after reconnect
- add regression tests covering initial-result timing and rows entering and leaving filtered queries across two independent browser workers

## Root cause

Browser `edge` and `global` subscriptions deliberately defer native subscription creation until the worker's durable path is ready. The separate JavaScript local-storage seed path could still complete during that wait and invoke the public callback with a provisional empty result. Consumers treated that callback as the fulfilled initial result and could issue writes before remote coverage existed, causing later filtered-query membership changes to be missed.

An explicitly disconnected client is the intentional exception: it continues to show unresolved local inserts, updates, and deletes while offline, and starts its server-backed subscription after reconnecting.

## Validation

- `cargo test -p jazz --test browser_relay_durability` — 13 passed
- full `packages/jazz-tools/tests/browser/db.disconnect.test.ts` and `worker-bridge.test.ts` — 67 passed, 1 expected failure
- full `examples/moon-lander-react/tests/browser/multiplayer.test.tsx` — 4 passed
- TypeScript type-check, oxfmt, oxlint, and sensitive-data checks passed locally
- all GitHub CI checks passed

### DIFF
--- a/crates/jazz/tests/browser_relay_durability.rs
+++ b/crates/jazz/tests/browser_relay_durability.rs
@@ -20,6 +20,13 @@ fn schema() -> JazzSchema {
     )])
 }
 
+fn filtered_membership_schema() -> JazzSchema {
+    JazzSchema::new([TableSchema::new(
+        "deposits",
+        [ColumnSchema::new("collected", ColumnType::Bool)],
+    )])
+}
+
 fn included_relation_schema() -> JazzSchema {
     JazzSchema::new([
         TableSchema::new("profiles", [ColumnSchema::new("name", ColumnType::String)]),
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Superseded by #1675.

The replacement preserves the intended authoritative first-result behavior without making explicit disconnect a hidden tier-semantic switch. It also separates the contracts cleanly: `tier` selects the durability/result gate, while `propagation` only controls upstream forwarding. The broader question of an explicit provisional local fallback for edge/global subscriptions is recorded as an open specification question.

PR #1664, opened 2026-08-19
(new core) fix(auth): support WorkOS JWT subjects

joeinnes:
## Summary

- route server auth admission through the shared principal mapper introduced in #1656
- support external JWT subjects such as WorkOS `user_*` IDs while preserving UUID subjects
- use identical UUID-recognition rules in Rust and TypeScript
- test realistic WorkOS permissions and cross-language mapping edge cases

## Testing

- `pnpm --dir examples/auth-workos-chat test` — passed
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/author-id.test.ts` — passed
- `cargo test -p jazz --features runtime,testing non_canonical_uuid_subjects_are_mapped_from_their_exact_bytes` — passed
- `cargo test -p jazz-server ws_session_identity_must_match_peer_identity` — passed


aeplay (changes requested):
🤖 Changes requested: the shared subject mapper still has two principal-identity blockers.

1. Rust and TypeScript remove every hyphen before UUID recognition. That makes a malformed opaque subject such as `123e4567-e89b-12d3-a456426614174000` collide with canonical UUID principal `123e4567-e89b-12d3-a456-426614174000`. Recognize only exact supported UUID spellings; map every other subject via UUIDv5 over its exact bytes. Add matching Rust/TS vectors proving malformed variants remain distinct.

2. Subject normalization differs across admission paths. Browser/default runtime and the regular server trim JWT `sub`; CLI loopback maps the raw value; static first-frame validation checks a trimmed value but maps the untrimmed input. Define and enforce one contract across all paths, including whitespace-only, surrounding whitespace, case, canonical/bare UUID, malformed UUID-like, and opaque WorkOS subjects.

Focused current tests pass, but they do not cover these seams. The #1656 + #1664 source restack is otherwise clean.

PR #1672, opened 2026-08-19
test: port policy integration tests

nrainhart:
## Description

FUP to https://github.com/garden-co/jazz/pull/1636. Migrates the remaining Rust integration tests that were lost in the migration to the new core.

Several tests are currently failing (and marked as `#[ignored]`), but we can fix them as a follow-up.


aeplay:
🤖 Changes requested: the port currently adds 51 compiled `#[ignore]` policy tests without registering them in `TEST_BURNDOWN.md`.

The executable Rust burndown gate fails because the workspace now contains 79 compiled ignores versus 32 documented entries: the 51 new policy skips are undocumented, and four existing ledger rows are stale. The focused target reports 54 passed / 51 ignored, and no CI checks are currently attached to this head.

Please either make passing tests active or add every remaining skip to the burndown ledger with its exact FQN and current reason/receipt, resolve the stale rows, and update the gate's exact inventory expectation. The NULL-policy coverage itself is mutation-sensitive and looks sound.

aeplay:
🤖 Changes requested: the policy port and burndown accounting now review cleanly, but the exact current head still fails the target's Clippy gate with three warnings promoted to errors:

- `simple_policies.rs:340`: redundant closure `|rows| Some(rows)`; use `Some`.
- `recursive_policies.rs:723`: `rows.len() > 0`; use `!rows.is_empty()`.
- `session_cases.rs:2010`: `expect(&format!(...))`; construct the panic lazily with `unwrap_or_else`.

Otherwise the focused result is good: 54 runnable policy tests pass, all 51 ignored tests have concrete ledger entries, the exact 69-active + 10-dormant burndown bijection passes, NULL lowering is covered end-to-end, and the branch merges cleanly with current integration.

PR #1676, opened 2026-08-20
feat: persist and sync only the Jazz public schema

nrainhart:
## Motivation

Policies have two representations: the public `PolicyExpr` authored by an app developer and the lowered `Query` used by the engine. Only the public form should be persisted and synchronized between core and edge servers. Lowered policies are compiled in memory when a database opens or installs a trusted catalogue snapshot.

This keeps persisted data independent of the engine's internal policy shape, eliminates duplicate Rust and TypeScript lowering, and allows lowering changes without storage migrations.

The durable source catalogue contains:

- structural public schemas, including tables, columns, references, defaults, merge strategies, and indexes;
- table and branch policies expressed as `PolicyExpr`;
- schema lineages and migration lenses;
- the current write-schema pointer.

The compiled `SchemaVersion { JazzSchema }` remains an internal runtime type. It is no longer serialized into `jazz_catalogue` or carried in catalogue snapshots.

## Changes

1. Move the versioned public schema and `PolicyExpr` catalogue codec into the `jazz` crate for shared use.
2. Replace persisted and snapshot `JazzSchema` payloads with source catalogue payloads; compile them at node-open and trusted-snapshot boundaries.
3. Change Rust and NAPI schema registration to accept the source form.
4. Make TypeScript encode the public schema AST and remove its policy-to-`Query` lowering.
5. Remove the server bridge that persists a second compiled permissions schema.


nrainhart:
Closing in favor of https://github.com/garden-co/jazz/pull/1677 (changes need to be based on top of https://github.com/garden-co/jazz/pull/1672)

PR #1677, opened 2026-08-20
feat: persist and sync only the Jazz public schema

nrainhart:
## Motivation

Policies have two representations: the public `PolicyExpr` authored by an app developer and the lowered `Query` used by the engine. Only the public form should be persisted and synchronized between core and edge servers. Lowered policies are compiled in memory when a database opens or installs a trusted catalogue snapshot.

This keeps persisted data independent of the engine's internal policy shape, eliminates duplicate Rust and TypeScript lowering, and allows lowering changes without storage migrations.

The durable source catalogue contains:

- structural public schemas, including tables, columns, references, defaults, merge strategies, and indexes;
- table and branch policies expressed as `PolicyExpr`;
- schema lineages and migration lenses;
- the current write-schema pointer.

The compiled `SchemaVersion { JazzSchema }` remains an internal runtime type. It is no longer serialized into `jazz_catalogue` or carried in catalogue snapshots.

## Changes

1. Move the versioned public schema and `PolicyExpr` catalogue codec into the `jazz` crate for shared use.
2. Replace persisted and snapshot `JazzSchema` payloads with source catalogue payloads; compile them at node-open and trusted-snapshot boundaries.
3. Change Rust and NAPI schema registration to accept the source form.
4. Make TypeScript encode the public schema AST and remove its policy-to-`Query` lowering.
5. Remove the server bridge that persists a second compiled permissions schema.


aeplay:
🤖 Changes requested: two compatibility blockers remain.

1. Existing durable databases cannot reopen. Catalogue rows currently persist `SchemaVersion` JSON as `{ id, schema }`; the new decoder accepts only `{ id, source_json }`. Both normal and dynamic-edge reopen paths deserialize those rows directly, so a pre-PR database fails with `missing field source_json`. `INV-LENS-8` requires durable catalogue schemas to survive restart. Please add a tested legacy conversion/migration strategy.

2. The existing public Rust schema API now fails on a first `Db::open`. `JazzSchema::new(...)`, `TableSchema`, and `Policy` remain public and are used by docs, examples, benches, and callers, but produce `source: None`. Genesis persistence now rejects those schemas as lacking public source. Either migrate/deprecate that public surface and all consumers in this PR, or provide a supported public-source conversion that preserves it.

The intended new direction looks sound: public schema and `PolicyExpr` are persisted/synced, decoding recompiles and validates IDs, and lowered policy internals are not included. The focused new snapshot/reopen and protocol tests pass.

The branch also needs a stack update after #1672; the rehearsal conflicts are limited to two test files and should preserve both sets of coverage.

PR #1678, opened 2026-08-20
fix: policy integration tests

nrainhart:
## Description

Addresses bugs and limitations uncovered by porting policy integration tests from `main` (https://github.com/garden-co/jazz/pull/1677).

See individual commits for changes introduced in this PR.


aeplay:
🤖 Closing #1678 as obsolete rather than merging its pre-reconstruction 33-commit stack. None of its patches applies to current main. Its already-landed behavior was reconstructed in #1958, #1976, #2015, #2045, #2046, and #2049. The old write-hidden-row behavior is superseded by the chosen read-for-write contract in #1762 and #2034. The remaining integration receipts still compile on current main and remain explicitly tracked in #1759, #1760, #1761, #1763, and #1764. We will reconstruct only the still-live fixes against the current architecture, beginning with rejection-cascade stack safety and policy-dependency subscription reconciliation. The current policy integration target compiles, and the two receipts enabled by reconstructed work pass.

PR #1679, opened 2026-08-20
test: port regression suites from an old-core production deployment

wrangelvid:
We ran the previous core in production for a while and accumulated a set of regression tests around the bugs we hit: subscription lifecycle hygiene, delivery of policy-dependency rows to cold clients, merged-row redelivery, replica settlement across connection states, write durability across reconnects, policy evaluation across schema migrations, mixed-generation row history, array-reference policy grants, and fate replay idempotency. This ports those tests (tests only, no engine changes) onto the new core, rewritten against the jazz-testkit and protocol-test harnesses. One commit per suite.

Suites that pass document behavior the new core already gets right. The ones it doesn't pass yet are registered per the burndown convention: 8 Rust and 3 TS quarantines, each with the concrete observed behavior in its ignore reason. The notable ones:

- dropping a `SubscriptionStream` never tears down the core subscription (`JazzClient::unsubscribe` is a no-op), and detaching the upstream leaves held authority-tier reads and global durability waits pending forever
- cross-generation updates wedge at the edge under session-scoped read policies, and Local-tier reads fail on lens-added column defaults under a two-generation catalogue
- a migration published to a live server makes the edge reject the current-schema shape (`UnsupportedShapeCapability`), so post-migration writes never settle
- `reconcileArray` truncates with a `.length` assignment, leaving proxy-cached per-index state stale, and `deploy` doesn't replay a chain of local migrations when no direct file connects head to release


aeplay:
🤖 Thank you for contributing these production-derived regression suites. We have adopted the complete test series into Garden-owned PR #1682 so we can maintain the follow-up engine-fix stack ourselves while preserving your authorship and exact commit provenance.

All 13 commits retain David von Wrangel as author with `-x` references to their originals. The Garden-only adaptation ports three fixtures to the current public-schema API and records the two disconnect/fallback cases as questions whose desired semantics we will confirm with adopters; their test bodies remain unchanged and quarantined.

Closing this PR as superseded by #1682. We have not modified this contributor branch.

PR #1680, opened 2026-08-20
fix: preserve optimistic rows across stale worker view updates

nrainhart:
### Bug

Opening more than one tab at the same time or mounting the Jazz inspector exposed an ordering race between optimistic writes and worker view updates. When issuing multiple writes, the main thread could receive a worker membership snapshot that included previous writes but had not yet incorporated the latest one. It treated that snapshot as fully authoritative and temporarily retracted the newest optimistic row, leaving the UI one row behind.

### Fix

Reconciliation now:

- Advances the authoritative generation only for actual membership changes.
- Preserves local rows whose settle position is newer than the worker view’s frontier.
- Applies this protection to the browser’s Local-tier worker handoff.

A browser regression test covers a host app sharing its broker with an inspector client and verifies that consecutive inserts never disappear.

### DIFF
--- a/crates/jazz/src/db/node_runtime.rs
+++ b/crates/jazz/src/db/node_runtime.rs
@@ -1028,7 +1028,7 @@ where
 
 fn unsettled_transaction_row_keys_for_query<S>(
     node: &Rc<RefCell<NodeState<S>>>,
-    cache: &mut BTreeMap<(AuthorId, Option<GlobalSeq>), BTreeSet<(String, RowUuid)>>,
+    cache: &mut BTreeMap<(AuthorId, Option<GlobalTime>), BTreeSet<(String, RowUuid)>>,
     shape: &ValidatedQuery,
     binding: &Binding,
     author: AuthorId,
--- a/crates/jazz/src/db/node_runtime.rs
+++ b/crates/jazz/src/db/node_runtime.rs
@@ -1026,28 +1026,39 @@ where
     }
 }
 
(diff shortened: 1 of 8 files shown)


aeplay (changes requested):
🤖 Changes requested: the new browser regression tests are not sensitive to the behavior this PR protects.

As a reversible planted-negative probe, we disabled optimistic-overlay preservation at both new handoff gates (`node_runtime.rs:1219` and `node_runtime.rs:1470`), rebuilt the real jazz-wasm artifact, and reran the exact two new tests. Both still passed. That means the tests do not currently force the stale worker `ViewUpdate` interleaving and would not catch removal or regression of the fix.

Please make the race deterministic—e.g. with a transport/worker interleaving hook or an explicitly scripted stale `ViewUpdate`—and demonstrate a planted positive where the test fails with preservation disabled.

The production implementation otherwise has no blocking correctness or authorization finding in this pass. Follow-up coverage for update/filter exit, delete+reinsert, rejection, reconnect/failover, and a remote same-row race would be valuable but is not blocking once the core race is deterministic.

aeplay:
🤖 Superseded by #1728. Current `main` already contains the production reconciliation behavior; the replacement keeps the useful intent as a small deterministic regression that asserts the optimistic row remains present in the actual post-handoff query result, with planted sensitivity. The older branch is substantially behind the current SharedWorker runtime and should not be revived.

PR #1683, opened 2026-08-21
fix: invalidate reactive array indexes on truncation

aeplay:
🤖 Fixes the adopted `reconcileArray` truncation regression from #1682.

## Behavior

When reconciliation shrinks an array, explicitly delete each removed numeric index before assigning the shorter `length`. This triggers reactive proxy `deleteProperty` traps, so cached per-index reads cannot reappear when the array later grows.

Identity preservation, keyed reordering, deep reconciliation, and explicit sparse-array non-goals are unchanged.

## Coverage

- activates the adopter-supplied truncation regression
- adds truncate-then-regrow stale-cache coverage
- removes exactly the matching TypeScript burndown entry
- planted removal of the delete loop makes both regressions fail
- focused Node: 14/14
- real Svelte `$state`: 17/17
- TypeScript check and TS burndown gate

Independent adversarial review found no remaining issue.


aeplay:
🤖 Closing as superseded: the reactive-array truncation behavior and active regression coverage are already on `main` via the later `target.splice(result.length)` repair. The focused reconcile-array suite passes 15/15 on current `main`; reviving this older branch would only reintroduce stale conflicts.

PR #1684, opened 2026-08-21
fix: scope browser storage controls by principal

aeplay:
🤖 Fixes the adopted cookie-session browser storage namespace regression from #1682.

## Behavior

DevTools/window storage controls now use the same canonical principal-aware namespace resolver as the actual OPFS and browser-broker runtime.

- the same authenticated principal keeps a stable namespace
- different cookie/bearer principals are isolated
- anonymous sessions remain app-scoped
- local-first sessions retain their stable identity scope
- an explicit `dbName` remains an intentional caller-controlled override

A running client keeps the namespace selected at registration across auth refresh; a newly created client resolves its current principal.

## Coverage

- activates the adopter-supplied cookie-session isolation regression
- covers same-user continuity/deduplication and different-user separation
- removes exactly the matching TypeScript burndown entry
- a planted regression back to app-only naming fails the isolation assertions
- focused storage/runtime tests and TypeScript/burndown gates pass

Independent adversarial review found no blocking issue.


aeplay:
🤖 Closing as superseded by `fe0f11985` on `main`, which carries the principal-scoped browser-storage behavior and promoted regression test. The focused `window-client-storage.test.ts` test passes on current `main`; the remaining rebase conflict here is only an obsolete burndown count.

PR #1696, opened 2026-08-21
Replace GlobalSeq with GlobalTime HLCs

aeplay:
## Summary

- document the `GlobalTime` model first: a core-authority 48-bit physical-ms + 16-bit logical HLC, sparse by design
- rename `GlobalSeq` and related wire/storage/query vocabulary to `GlobalTime` without adding a point-in-time public API yet
- allocate timestamps atomically with global acceptance and recover the authority HLC/frontier from durable accepted history
- keep whole-history frontiers core-only while edges/clients retain explicit sparse snapshot dots and query-scoped settlement receipts
- preserve branch and exclusive-transaction behavior when a partial node's snapshot contains sparse accepted timestamps
- leave shard identity outside the timestamp; future sharded snapshots can carry per-shard `GlobalTime` values externally

## Commit structure

1. `spec: replace global sequences with settlement HLCs`
2. `feat: replace global sequences with settlement HLCs`

## Validation

- `cargo test -p jazz --no-default-features --features testing,transport-compression-zstd --lib` (1271 passed, 2 ignored)
- `cargo test -p groove`
- `cargo test -p jazz-cli --features test`
- `cargo test -p jazz-otel`
- `dev/gates/invariant-registry.sh`
- incremental relation/include delivery canary
- `JAZZ_MICRO_ITERS=1 cargo bench -p jazz-sim --bench micro`
- pre-commit clippy, formatting, and sensitive-data gates

## Existing gate issues observed

- the documented default-feature `cargo test -p jazz` / soak command currently compiles `tests/edge_fate_authority.rs`, which imports `jazz::serving` while the `runtime` feature is disabled
- smoke's `jazz/sync` and `jazz/validation` fixtures currently use unsupported `session.sub` policy references; all other smoke scenarios passed after fixing the sparse-snapshot exclusive validation regression found by `jazz-sim/micro`

Point-in-time query APIs are deliberately deferred to a follow-up; this PR preserves the current read/snapshot surface while making its global coordinate time-shaped.


aeplay:
🤖 Changes requested — adversarial review confirmed three blocking partial-snapshot/authority-time defects.

1. Filtered exclusive reads reject their own sparse snapshot dots. Dot-aware validation is applied only to degenerate whole-table predicates; filtered shapes compare solely at `global_base`. On a partial client that base remains zero while received global rows are represented as dots. A planted scenario read a delivered `title = "watched"` row, wrote another row, and was incorrectly rejected with `ExclusiveConflict`. Predicate validation needs the full dotted snapshot for every shape and binding.

2. Partial branches cannot sync. Branch creation serializes received global transactions as snapshot dots, while core session admission rejects any dotted v1 branch base as `unsupported v1 branch snapshot shape`. A direct client/core reproduction failed after receiving one global write and creating a local branch. This also conflicts with the current branch spec and makes branch creation scale with the number of received transactions. Please choose and implement one coherent contract: keep externally admitted v1 bases global-only, or explicitly support and authorize dotted metadata end to end.

3. The production client-link path makes authority physical time client-controlled. `apply_sync_message_with_ingest_context` derives authority `now_ms` from the incoming transaction timestamp. The future-skew check therefore cannot reject, and that same value feeds authority HLC allocation. An authenticated client can push the core settlement clock arbitrarily far into the future. Please thread a core-controlled clock through authority ingress, or explicitly defer physical-time semantics without presenting the current check as an authority boundary.

Nonblocking follow-ups: reconcile the snapshot spec’s contradictory statements about foreign dots, and specify/test logical-counter saturation behavior.

Existing focused HLC, whole-table sparse-dot, and frontier tests pass. Two planted partial-client regressions failed as described and were restored.

aeplay:
🤖 Re-review update for `9bed20c86`: all three previously reported P1 blockers are resolved; no new merge blocker found.

- Filtered exclusive validation now evaluates the full dotted snapshot. The original partial-client regression passes; reverting only snapshot evaluation reproduces `Rejected(ExclusiveConflict)`.
- Session branch admission now parks until each dotted transaction is present, then verifies accepted/global fate. The direct partial-branch sync regression passes, and a planted always-pending mutation fails the expected admission assertion.
- Authenticated client ingress now supplies authority wall time before skew validation and HLC allocation. The production transport regression rejects a far-future client clock; reverting to client transaction time makes the planted test fail as intended.

The snapshot, branch, and authority-clock specs now align. Logical-counter saturation remains deliberately fail-not-wrap and is documented; a focused non-maximum saturation-capacity test would be a useful nonblocking follow-up.

Focused tests and diff checks passed; all adversarial probes were restored.

nrainhart (approved):

PR #1702, opened 2026-08-22
Add large values and partial value APIs

aeplay:
🤖 Implements the draft storage and API foundation for large values in ordinary Jazz string, bytes, and JSON columns.

## Behavior

- Small values remain inline; large values transparently use immutable content-defined chunk trees plus a bounded byte-edit tail.
- Default reads and writes use native `string`, `Uint8Array`, and JSON values.
- Partial projections return plain primitives, not mutable handles.
- TypeScript text coordinates use UTF-16 code units; Rust exposes explicit UTF-8 byte and UTF-16 variants.
- Bytes support slices, splices, and append. JSON supports JSON-pointer projection and native replacement writes.
- CAS is deliberately outside this PR.

## Jazz-row storage model

Each application table with large-value-capable columns receives a hidden, versioned node table. Immutable tree nodes are ordinary Jazz rows with owner references and inherited authorization. Owner changes and generated nodes commit together and use ordinary Jazz history, persistence, schema lineage, subscriptions, and synchronization. There is no parallel object store, permission system, or content protocol.

The latest adversarial passes hardened this model by:

- admitting only canonical, owner-reachable generated nodes;
- rejecting forged mutation/deletion of generated nodes;
- requiring descriptor roots to exist and authenticating their byte/UTF-16 metrics;
- shipping the complete reachable node closure in synchronized views.

## Adversarial-review status

Two independent adversarial review/fix iterations are complete. They found two architectural release blockers that this draft now records explicitly rather than papering over:

1. Groove currently sees the physical descriptor for a promoted cell. Terminal reads materialize afterward, but filters, policies, ordering, joins, aggregates, and indices must instead see the assembled logical value. The intended direction is a query-local logical relation/source that materializes only plan-required large columns while durable current/history relations remain physical and authoritative.
2. Promotion is implemented in the TypeScript client lowering path, but direct Rust `Db`/node mutation paths do not yet share that lowering boundary. A trial admission rejection correctly caught bypasses but also rejected existing valid `Db` writes, so it was reverted; a common core lowering step is required.

The docs now state that partial APIs may still materialize a complete value internally. The PR remains draft and is not release-ready until these blockers are implemented and re-reviewed.

## Verification

- Focused generated-node admission, descriptor-root, synchronization-closure, and large transport tests pass.
- `cargo test -p jazz --no-default-features --features testing,transport-compression-zstd --lib --no-fail-fast`: 1259 passed, 0 failed, 2 tracked ignored.
- Canonical gates were rerun after the review fixes; Rust/Groove gates are green. The TypeScript gate retains the integration base's unavailable `better-auth` and `@vue/server-renderer` dependency failures.
- Docs type-check, formatting, clippy, and the sensitive-data guard pass.


aeplay:
superseded by #1706

PR #1704, opened 2026-08-22
Use read-policy indexes for owner-filtered reads

tobiaslins:
## Why

Large multi-tenant tables need `get my/org documents` to scale with the tenant-owned slice, not with the whole server table. Before this change, the read-policy authorization path could still scan every Global current row and then join authorized row ids, so a user owning 500 rows in a 1M-row table paid the 1M-row cost. That is exactly the hot path for sync nodes serving user/org document lists.

This PR lets simple indexed read policies, such as `owner == session.user_id`, drive the physical read path. Both the authorization proof and the protected base source can now use the declared owner index, keeping the work proportional to authorized candidates.

## What changed

- Derive Global current access paths from normalized read-policy authorization subplans.
- Compile policy authorization graphs with the access-path-aware compiler.
- Reuse the policy-selected access path for the protected visible-current base source when the user query itself has no narrower path.
- Add `owner_filter_diagnostic` to cover the multi-tenant owner-policy benchmark directly.

## Performance

Baseline from `codex/async-ordered-storage-v2` at 100k total / 500 owned rows:

| case | query time | Global current row reads |
| --- | ---: | ---: |
| policy only | ~914 ms | 100,000 |
| owner predicate | ~843 ms | 100,000 |
| owner + order/limit | ~841 ms | 100,000 |

This PR:

| total rows / owned rows | policy only | owner predicate | owner + order/limit | row reads | index reads |
| --- | ---: | ---: | ---: | ---: | ---: |
| 10k / 500 | 15.9 ms | 14.7 ms | 17.6 ms | 500 | 500 |
| 100k / 500 | 14.5 ms | 14.9 ms | 17.2 ms | 500 | 500 |
| 1M / 500 | 17.8 ms | 15.1 ms | 18.0 ms | 500 | 500 |

So the owner-policy path is now flat across 10k -> 1M total rows in this benchmark.

## Verification

- `cargo check -p jazz --benches --features testing`
- `JAZZ_OWNER_FILTER_ROWS=10000,100000 JAZZ_OWNER_FILTER_OWNED_ROWS=500 JAZZ_OWNER_FILTER_RESULT_ROWS=500 cargo bench -p jazz --features testing --bench owner_filter_diagnostic --quiet`
- `JAZZ_OWNER_FILTER_ROWS=1000000 JAZZ_OWNER_FILTER_OWNED_ROWS=500 JAZZ_OWNER_FILTER_RESULT_ROWS=500 JAZZ_OWNER_FILTER_BATCH_ROWS=10000 cargo bench -p jazz --features testing --bench owner_filter_diagnostic --quiet`
- Pre-commit ran `cargo clippy --package jazz -- -D warnings` and `rustfmt`. The sensitive-data guard warned it was unavailable because `jazz-private` is not present in this temp worktree.

## Known boundary

This PR fixes the owner/read-policy path. The generic non-policy `selective_global_hydration` benchmark still fails, so arbitrary prepared-query index selection remains separate follow-up work.

Tooling-friction: the 1M receipt spent ~660s seeding; a reusable persisted fixture would save nearly all benchmark wall-clock.

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -106,6 +106,11 @@ harness = false
 required-features = ["testing"]
 
 [[bench]]
+name = "owner_filter_diagnostic"
+harness = false
+required-features = ["testing"]
+
+[[bench]]
 name = "observer_write_path"
 harness = false
 
--- a/crates/jazz/benches/owner_filter_diagnostic.rs
+++ b/crates/jazz/benches/owner_filter_diagnostic.rs
@@ -1,0 +1,245 @@
+mod schema_fixture;
+mod support;
+
+use std::collections::BTreeMap;
+use std::path::Path;
+use std::time::Instant;
+
+use jazz::db::{
+    Db, DbConfig, DbIdentity, LocalUpdates, MergeableTxOps, Propagation, ReadOpts,
+    SeededRowIdSource, block_on,
+};
+use jazz::groove::db::StorageReadMetrics;
+use jazz::groove::records::Value;
+use jazz::ids::{AuthorId, NodeUuid, RowUuid};
+use jazz::query::{OrderDirection, Query, col, eq, lit};
+use jazz::schema::JazzSchema;
+use jazz::tools::{ColumnType, SchemaBuilder, TablePolicies, TableSchemaBuilder};
+use jazz::tx::DurabilityTier;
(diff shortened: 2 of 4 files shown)


aeplay:
🤖 Superseded by #1708, which preserves Tobias's original commit and performance receipts while replacing the policy-specific planner with one normalized-program access-path mechanism shared by ordinary queries and authorization policies.

PR #1705, opened 2026-08-22
Bound one-shot indexed first-page hydration

tobiaslins:
## Why this matters

This is the server/sync-node multi-tenant shape we have been testing:

- one physical table can contain millions of rows
- one user/org owns a much smaller slice
- the common read is "give me the first page of my rows"

Before this PR, a row-projected index scan could find the `owner = X` prefix, but it still had to hydrate every row in that prefix before the query-level `LIMIT` was applied. For a 500-row page over 10k owned rows, that meant 10k row hydrations.

## What changed

- Adds `StaticScanSpec::PrefixLimit` to Groove.
- Adds `OwnedStorage::scan_prefix_limit`.
- Adds an indexed-row storage request that scans only the bounded prefix entries and hydrates only those rows.
- Lets Jazz attach that source limit for a narrow safe query shape:
  - root indexed literal equality
  - `LIMIT`, no `OFFSET`
  - no explicit `ORDER BY`
  - no joins/includes/reachable/inherits/flat joins
  - no read policy
  - no access-edge implicit policy
- Carries root index access paths into include-deleted one-shot reads where the source limit is safe.
- Explicitly does **not** apply source limits before visible-row tombstone filtering or before read-policy filtering.

## Benchmark

Command:

```sh
JAZZ_BENCHMARK_DIRTY_OK=1 JAZZ_OWNER_FILTER_ROWS=100000 JAZZ_OWNER_FILTER_OWNED_ROWS=10000 JAZZ_OWNER_FILTER_RESULT_ROWS=500 cargo bench -p jazz --features testing --bench owner_filter_diagnostic --quiet
```

Data shape:

| Setting | Value |
| --- | ---: |
| total rows | 100,000 |
| rows owned by target user | 10,000 |
| requested page size | 500 |

Result:

| Case | What it represents | Result rows | Index reads | Row hydrations | Logical reads | Query time |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| `owner_predicate_ordered_limit` | current protected, visible, ordered app-facing control | 500 | 10,000 | 10,000 | 20,000 | 215.905ms |
| `public_owner_predicate_limit_include_deleted_system` | new bounded exact-prefix source path | 500 | 512 | 500 | 1,012 | 8.638ms |

Source-level upside shown by the bounded path:

| Metric | Before/control | Bounded path | Change |
| --- | ---: | ---: | ---: |
| row hydrations | 10,000 | 500 | 20x fewer |
| logical reads | 20,000 | 1,012 | ~19.8x fewer |
| query time | 215.905ms | 8.638ms | ~25x faster |

The 512 index reads for a 500-row page are expected today because the storage cursor batches in chunks of 256. Row hydration is capped exactly at 500.

## Important scope note

This PR is **not** the full app-facing fix for:

```text
owner == current_user ORDER BY updated_at DESC LIMIT 500
```

under read policies and normal visible-row semantics.

That case is intentionally still shown in the benchmark as `owner_predicate_ordered_limit`, and it still reads/hydrates all 10k owned rows. Making that fast needs a follow-up visibility/policy-safe live-row or composite ordered index path, so we can bound the source without under-filling pages after tombstone filtering and without capping rows before policy filtering.

This PR is the lower-level piece that proves the engine/storage path can be page-sized once the planner has a safe source to use.

## Tests

- `cargo check -p jazz --features testing`
- `cargo test -p jazz one_shot_ --features testing`
- `cargo test -p groove --lib`
- `JAZZ_BENCHMARK_DIRTY_OK=1 JAZZ_OWNER_FILTER_ROWS=100000 JAZZ_OWNER_FILTER_OWNED_ROWS=10000 JAZZ_OWNER_FILTER_RESULT_ROWS=500 cargo bench -p jazz --features testing --bench owner_filter_diagnostic --quiet`

Pre-commit also ran:

- `cargo clippy --package groove --package jazz -- -D warnings`
- `rustfmt`

The sensitive-data guard was not available in this worktree because `jazz-private/dev/gates/no-sensitive-data.sh` was missing.

### DIFF
--- a/crates/groove/src/ivm/op_types.rs
+++ b/crates/groove/src/ivm/op_types.rs
@@ -71,6 +71,10 @@ pub struct IndexSourceOp {
 pub enum StaticScanSpec {
     Point(Vec<LiteralValue>),
     Prefix(Vec<LiteralValue>),
+    PrefixLimit {
+        prefix: Vec<LiteralValue>,
+        limit: u64,
+    },
     Range {
         start: Vec<LiteralValue>,
         end: Vec<LiteralValue>,
--- a/crates/groove/src/ivm/runtime/evaluation_session.rs
+++ b/crates/groove/src/ivm/runtime/evaluation_session.rs
@@ -29,6 +29,12 @@ pub(super) enum StorageRequestKey {
         index: String,
         prefix: Vec<u8>,
     },
+    IndexedRowsPrefixLimit {
+        table: String,
+        index: String,
+        prefix: Vec<u8>,
+        limit: u64,
+    },
     IndexedRowsRange {
         table: String,
         index: String,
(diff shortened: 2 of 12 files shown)


aeplay:
🤖 Superseded by #1709, which preserves the bounded-hydration goal and receipts while replacing the narrow prefix-limit side path with the unified `scan(ScanRequest)` API across storage implementations.

PR #1750, opened 2026-08-23
Establish canonical issuer-scoped author subjects

aeplay:
🤖 Establishes issuer-scoped canonical authorship while keeping external JWT claims raw and making trusted backend authorship explicit.

## Behavior

- `session.iss`, `session.sub`, and application claims remain provider values; `sub` is not rewritten.
- Logical row authorship is the exact canonical JSON encoding of `[iss, sub]`, locally interned without changing its string semantics.
- Stored provenance and `$createdBy`/authorship comparisons use that canonical author.
- Issuer and subject bytes are preserved exactly; missing or ASCII-blank principals are rejected rather than trimmed.
- Local-first, anonymous, external, and `SYSTEM` identities remain distinct and validate their issuer/auth-mode pairing.
- Policy contexts merge application claims first, then restore canonical built-ins so supplied claims cannot spoof authorship.
- Raw NAPI/WASM identity ingress rejects reserved issuers, malformed canonical authors, and legacy 16-byte AuthorIds. There is no compatibility fallback.
- Verified self-signed proof entrypoints derive their author only after proof verification.
- Explicit NAPI backend constructors derive `SYSTEM` for memory and persistent runtimes. Generic opens cannot escalate from a caller-supplied author or arbitrary nonempty credential.
- `backendSecret` remains a transport credential validated by the remote server; local-only explicit backend mode is trusted application code and also authors as `SYSTEM`.
- Old native artifacts lacking the explicit backend ABI fail closed rather than falling back to the generic constructor.

## Worked cases

- External JWT: issuer `https://id.example` and subject `user-7` remain raw session claims; row authorship is `["https://id.example","user-7"]`.
- Exactness: surrounding whitespace in a nonblank provider value is preserved; an issuer containing only ASCII whitespace is rejected.
- Client raw ingress: supplying the reserved `SYSTEM` author, a legacy UUID author, or `backendCredential: "x"` is rejected before persistent storage creation.
- Backend runtime: selecting the explicit backend constructor derives `SYSTEM`; a wrong remote backend secret still fails server admission and grants no trusted connection.
- Self-signed session: a valid matching proof derives its reserved author; a wrong app, mismatched subject, or tampered proof is rejected.

## Verification

- independent adversarial review cleared canonical encoding, reserved-subject ingress, proof verification, issuer exactness, explicit backend authority, old-artifact failure, and no-compat fixture migration
- NAPI Rust suite: 15/15 passed
- focused native-runtime/NAPI/backend TypeScript suites: 181 passed, 2 skipped
- server issuer exact/nonblank receipt passed
- wrong-backend-secret remote admission receipt passed
- planted mutations of backend `SYSTEM` derivation and raw-ingress rejection failed their boundary tests
- TypeScript check, Rust formatting, Clippy, and repository guards passed

The explicit backend decision is also recorded in the PR discussion so later restacks and status sweeps retain the governing intent.


aeplay:
🤖 Roadmap issue: #1758.

aeplay:
🤖 Overnight adversarial review found that the reserved-issuer rejection correctly closes the public JSON forgery path, but the current head also rejects the real high-level trusted backend path.

JazzContext.asBackend() and withAttribution(...) still route SYSTEM reads/writes through the same serialized session/write-context shapes that are now intentionally untrusted. The direct native no-session test does not cover those high-level paths.

The repair in progress will keep the public boundary fail-closed while giving trusted backend reads and writes a genuinely private, non-forgeable internal authority channel. It will add integration coverage through JazzContext.asBackend() for both query and mutation, alongside planted public-forgery rejection tests.

aeplay:
🤖 A second adversarial receipt found that local-first token refresh currently loses its verified reserved-session capability. Db.refreshLocalFirstToken() mints a fresh local-first JWT, but updateAuthToken() sends it back through generic public session resolution, which correctly rejects reserved issuers and reports a principal change.

The same repair will preserve public token updates as untrusted while allowing only the dedicated in-process local-first mint/verification path to construct a renewed trusted session. A regression will cover creating a secret DB, obtaining its identity proof, and refreshing the token without changing its author subject.

aeplay:
🤖 Final adversarial review after the trusted-backend repair found one remaining P1 boundary: public NAPI/WASM canonical-author ingress accepts the exact SYSTEM canonical JSON. This bypasses the TS reserved-session guard through open-config identities and raw author-byte APIs (identity reads/writes, streaming, transactions, claims, subscriptions, and WASM acceptSubscriber).

Server websocket and wire upload boundaries were separately checked and remain safe: authenticated links bind the peer identity, trusted backend/admin paths are credential-gated, and session uploads reject made_by mismatches.

The repair now in progress centralizes an untrusted AuthorSubject parser for every public binding config/raw-byte seam and rejects all Jazz-reserved issuers there. Private SYSTEM continues only through the non-serializable internal capability path. Planted NAPI/WASM forgery tests and ordinary external-author controls will gate the next head.

aeplay:
🤖 The public binding forgery and rebase-drift repair is now pushed and the six-PR stack has been rebased above it.

All NAPI/WASM public author config/raw-byte ingress now shares an untrusted canonical parser that rejects SYSTEM, local-first, static-bearer, and anonymous issuers. Private backend and verified local-first/anonymous flows remain capability-only. The repair also restored bound exclusive-transaction identity, upload retry identity, and oversized-scalar behavior lost during the old-base rebase.

Verification on the committed head: Jazz lib 1,249 passed / 2 ignored; NAPI and WASM reserved-ingress tests pass; focused backend/local-first/anonymous receipts pass; TypeScript, Clippy with warnings denied, formatting, lint, and sensitive-data gates pass. Fresh CI is running on this head and upper BandChat layers now inherit it.

aeplay:
🤖 Final independent adversarial review found no remaining authorization-boundary bypass. The three branch-local CI blockers are repaired at 86c5607ec: canonical-author wire fixtures were regenerated and reviewed, the reserved-session test payload now matches JwtPayload, and both streaming mocks retain strict author-argument assertions with explicit types.

Local verification:
- jazz-tools build:test passed
- focused Vitest: 108 passed, 1 skipped
- all 5 Jazz wire-fixture tests passed
- planted public-binding, reserved-author, exclusive-identity, and trusted-capability regressions each failed the intended tests and were restored
- worktree clean

Fresh stacked CI is now running. Upper branches remain stacked and will be physically rebased after their active local repairs commit, avoiding disruption of in-progress work.

aeplay:
🤖 Canonical auth is now physically rebased onto the reviewed #1870 head, including the policy/provenance layers beneath it.

The final boundary keeps raw JWT `claims.sub` available to application policy/session logic while row authorship uses the canonical issuer-scoped subject. Public ingress rejects reserved SYSTEM/self-signed issuers; trusted internal links are classified by verified credential type rather than subject strings. The per-identity WebSocket cap applies only to verified external sessions, so internal edge/bootstrap links do not evict replication sockets and forged SYSTEM claims remain rejected.

Independent adversarial review found no auth bypass and planted the old unconditional socket-cap behavior, which the trusted-link receipt caught. Rebase integration repairs ported the newer correlated-policy fixture to raw session-sub strings and externalized API fixtures. Exact gates on the combined tree:

- reserved canonical public ingress: 1/1
- WebSocket suite: 32/32
- exact two-edge core propagation: 1/1 in 0.53s
- `jazz-tools` test typecheck/build: passed

The official GitHub stack link will be updated after the existing BandChat upper layers are physically rebased, so stack tooling does not drop or misbase them.

aeplay:
🤖 The backend-capability re-review produced one safe repair and one deliberately deferred native-boundary question.

The safe repair scopes SYSTEM identity to explicit trusted-backend/attribution contexts, preserves that capability through schema views, and proves ordinary `JazzContext.db()` remains host-identity scoped rather than inheriting backend authority.

Verified local-first/anonymous sessions expose a different boundary: their admission marker currently exists only in TypeScript. A public native `*ForAdmittedSession(authorBytes)` surface would be forgeable by direct NAPI callers, so I stopped that implementation rather than weakening the reserved-issuer rejection. The native-verifiable capability design and its complete read/subscription/mutation surface are captured in #1878. Public raw identity ingress continues to reject reserved issuers; no string allowlist or JS-only marker is being substituted.

aeplay:
🤖 CI repair pushed on top of the verified proof-bearing open design.

- The generated NAPI declaration includes the Rustdoc block for `openMemoryWithSelfSignedProof`; the checked-in declaration had the signatures but omitted that generated comment, so release assembly correctly rejected the byte drift. The checked-in surface now byte-matches fresh generation.
- The next TS gate exposed two ABI-test constructors returning a partial mock while claiming `NativeDb`. The fixture is now structurally complete, with every irrelevant native operation failing loudly; production types were not weakened and no cast was introduced.

Focused proof ABI tests, the Jazz-tools test TypeScript project, `build:test`, and the full fresh/provenanced `build:test-artifacts` gate all pass locally. These are test/build-surface fixes only; authorization semantics are unchanged.

aeplay:
🤖 **Confirmed design decision: backend authorship remains `SYSTEM`.**

An explicit backend runtime is trusted application code and intentionally authors local writes as `SYSTEM`, including a local-only backend that never connects upstream. We do not need a separate `runtime-host` author identity.

The security boundary is the constructor/admission path, not whether backend code may ever obtain `SYSTEM`:

- ordinary client/raw NAPI/WASM opens derive authorship from a verified session and must not accept a caller-supplied reserved subject;
- generic open configuration must never treat an arbitrary nonempty `backendCredential` string as proof and silently escalate;
- an explicit, unmistakable backend-open path establishes backend mode and derives `SYSTEM` authorship;
- when syncing remotely, the edge still validates `backendSecret` before granting trusted-backend upload authority;
- local-only backend mode is trusted because the application explicitly selected the backend constructor, not because a raw credential string was locally verifiable.

Implementation and tests should remove the accidental generic-config escalation while preserving this intentional backend behavior. No `runtime-host` compatibility behavior is required.

PR #1828, opened 2026-08-24
🤖 Expose typed large-value operations on the public Db API

aeplay:
🤖 Exposes typed byte-range and UTF-16 text-range reads plus append/splice operations on the public Db facade introduced by the Groove-native large-values work. Keeps internal clients private, preserves truthful WriteHandle<void> semantics, and adds runtime and compile-time coverage.


aeplay:
🤖 Superseded by #1833. The replacement implements the agreed query/update DSL and deliberately avoids exposing these imperative helpers as public TypeScript API.

PR #1830, opened 2026-08-24
Add the EpicDrop example and benchmark

aeplay:
🤖 Starts the EpicDrop example family with a typed file/folder app, streaming browser uploads, range-oriented usage, and a self-contained Rust benchmark covering streamed creation, listing, and partial reads. This is a WIP reference app that will adopt the replacement large-value query DSL as it lands.


aeplay:
🤖 Overnight progress: EpicDrop now scopes its file browser through the indexed folder_id and projects metadata only, so listing a folder does not request contents. The benchmark receipt passes.

Two public contracts remain explicit rather than bypassed with private runtime helpers:

- #1833 supplies bounded byte-range reads/updates.
- #1857 blocks a deterministic remote-hydration topology receipt; the skipped receipt requires metadata-only subscriptions to publish without waiting for a withheld contents frontier.

The branch is pushed as WIP. Browser validation still needs a lane with complete workspace dependencies; this is not being called review-ready yet.

aeplay:
🤖 Review fixes are now pushed: the UI does not construct a UUID filter until a real folder exists, the empty-initial state has a receipt, and the Rust benchmark now mirrors the app's folder filter, metadata projection, and name ordering without contents. The previously misleading #1857 link was removed; the controlled remote-chunk withholding harness gap is now tracked precisely in #1862.

aeplay:
🤖 Final review correction is pushed. EpicDrop now has no #1857 reference; the controlled remote-value withholding gap links to #1862. The independent reviewer confirmed the guarded initial query and benchmark projection/order fixes, and focused schema validation plus Rust benchmark tests pass.

aeplay:
🤖 Superseded by the current, deliberately narrower foundation in #2057. It reconstructs typed streaming upload, metadata-only folder projection, cancel/retry atomicity, and bounded native range workloads with original attribution. Typed preview/download and withheld-chunk/persistent-worker topology remain explicitly tracked by #1833, #1862, and #1978 rather than being represented as complete.

PR #1833, opened 2026-08-24
Add the large-value query and update DSL

aeplay:
🤖 Replaces #1828 with the intended high-level large-value API: lazy byte, UTF-16 or UTF-8 text, and RFC 6901 JSON selections in the typed query DSL, plus page-relative splice and semantic JSON edits in ordinary updates. This first WIP commit locks the specification and documentation contract; implementation and exhaustive binding/runtime coverage are in progress. No low-level imperative large-value helpers become public TS API.


aeplay:
🤖 Superseded by the current-main reconstruction in #2088. The replacement preserves the typed large-value projections and page-relative edits under the agreed `applyDiffs` surface, adapts to the current encoded-write boundary, and explicitly tracks unresolved transaction/branch/session context semantics in #2087 rather than guessing.

PR #1839, opened 2026-08-24
Carry large-value projection demand into Groove

aeplay:
🤖 Stacked on #1833. Carries typed large-value projection demand through Jazz query identity and maintained terminal metadata into Groove demand-aware materialization, so byte, UTF-16, UTF-8, and JSON-pointer selections can request only the required chunks. Runtime terminal routing and full subscription coverage are still in progress.


aeplay:
🤖 Closing this implementation attempt in favor of #2090. The need remains, but this branch keys demand by graph `NodeId`; distinct terminal consumers sharing a root therefore collapse conflicting ranges to `Full`, violating exact per-consumer projection semantics. #2090 records the correct terminal-specific evaluator/cache design and planted coverage, while #2088 carries the public DSL.

PR #1843, opened 2026-08-24
Add the Jamazon Warehouse example

aeplay:
🤖 Starts the Jamazon Warehouse example family as a self-contained TPC-C-shaped application and benchmark. It models warehouse, district, stock, item, customer, order, order-line, payment, and delivery data; prepares indexed transaction/query shapes and deterministic 100/10k profiles. UI, auth, contention, topology, and fault E2E are intentionally continuing in follow-ups.


aeplay:
🤖 Overnight progress: Jamazon Warehouse now has seeded warehouse/district/customer/item/stock data, bounded operational query shapes, and an exclusive purchase transaction updating stock, district sequence, customer balance, order, and order line atomically. Deterministic receipts cover success and insufficient-stock abandonment; the Rust benchmark suite passes.

The branch still lacks an app scenario driver/auth shell, so it does not yet claim authorization or reconnect/topology coverage. Those are being left explicit rather than invented inside the benchmark-only slice. The PR remains draft pending independent review and a later topology layer.

aeplay:
🤖 Independent review found correctness gaps in the new purchase workload:

- customer balance is overwritten with the latest negative amount instead of accumulated;
- the low-stock query sorts by on_hand but never enforces on_hand < reorder_level;
- README-claimed idempotency is absent, so retry creates a second order;
- district-counter observation is not district-scoped/deterministic;
- Rust and TS disagree on payments.order_id optionality;
- total/counter/derived-ID arithmetic is unchecked.

A repair lane is adding two-purchase accumulation, exact monetary/order-line assertions, caller-key idempotency, multiple-district and mixed-stock receipts, schema parity, and overflow atomicity. The PR remains draft.

aeplay:
🤖 Review fixes are pushed. Purchases now use caller request keys with stable retry receipts, accumulate balances, create exact order-line/payment rows, use checked arithmetic with rollback coverage, scope district observation deterministically, and match nullable payment schema across Rust/TS. Expanded fixtures cover two districts plus mixed low/adequate stock.

The low-stock workload now labels its bounded query honestly and applies the final field comparison locally. Declarative column-to-column lowering is tracked in #1864; the example will remove that post-filter when the single query path supports it. Focused check and five workload tests pass.

aeplay:
🤖 Final adversarial findings are fixed and independently re-verified at 551c743a8.

The retry path now uses an exact warehouse + idempotency-key query backed by the actual runtime index set. Low-stock reads remain complete before the temporary local field-to-field predicate, with #1864 documenting the missing scalable lowering. Nullable payment/order foreign-key parity is restored, and the workload now asserts exact order-line/payment amounts and relationships.

Verification:
- jazz-example-jamazon-warehouse-benchmark: 6 tests passed
- benchmark crate check passed
- planted wrong-amount and reversed-low-stock mutations both failed the intended assertions and were restored
- final runtime-schema assertion proves all five required order index columns survive builder lowering

No new semantics were introduced; the branch remains draft while the example-app stack is assembled.

aeplay:

aeplay:
🤖 Restacked all 16 Jamazon commits exactly onto #1940. No cleanup migration was appropriate: its duplicate/drop/retry faults are app-owned envelope scheduler state already drained by the shared harness, not external browser-route state. Concurrent duplicate checkout, atomic rejection footprint, retry convergence, bounded 20-of-21 query, offline edit, reopen, and authority-transfer receipts are unchanged. Schema, TypeScript, browser-config, benchmark/workload, and optimized compile gates pass; exact browser artifacts remain to run.

aeplay:
🤖 Superseded by the current-main reconstruction in #2082. The replacement preserves the useful TPC-C-shaped schema, topology harness, and benchmark while removing obsolete trusted-author stacking, porting stale transaction calls, and accurately distinguishing deterministic test credentials from a real authenticated UI. It remains draft through final app modernization and independent review.

Issue #1844, opened 2026-08-24
Materialize large values before public bindings


aeplay:
🤖 Public full-value reads and subscription bindings can expose Groove Large descriptors instead of materialized logical scalars. Reproduced through real browser/server topology with EpicDrop Bytea and MusicAgent Text: after a streamed value is globally accepted, the receiving binding throws that storage tag 14 crossed a logical boundary. Db::all returns CurrentRows without hydration, and NAPI Subscription::read_all directly encodes added/updated events without a Db/node hydration handle. Hydrate full-value rows and subscription events before every public binding; preserve async suspension/rearm and typed failures. Regressions live on #1830 and #1838, with lower-level peer-edge coverage in lane 6.

aeplay:
🤖 Independent review of the local public-hydration implementation found additional binding seams that must be repaired before a PR is called ready:

- WASM ordinary and transaction query APIs still synchronously block, so remote chunks cannot be requested/flushed.
- NAPI range/JSON reads and append/splice helpers still core_block_on remote chunk work.
- NAPI/WASM transaction reads can encode physical large-value descriptors without hydration.
- Upstream failover immediately reissues a request but leaves its prior retry deadline armed, producing a duplicate request.
- A suspended NAPI subscription drains into an unbounded retained batch while later events continue accumulating.

The implementation lane is converting these through the same retained async machinery, adding a planted failover regression, and bounding subscription retention. The real two-client/server topology receipt remains preserved; direct WASM creation is being repaired separately under #1857.

aeplay:
🤖 The implementation is now visible as draft PR #1863. It remains deliberately unready until #1857 is integrated, the preserved real server topology receipt is enabled, and the final adversarial/full-gate pass completes after restacking.

tobiaslins:
We encountered a potentially related **write-side large-value descriptor failure** in our app. Posting here because the error suggests a physical/logical scalar boundary problem; we have not established that it has the same root cause as the read/binding failures in this issue.

### Runtime

- Jazz alpha.54 preview from commit **`0394059`**, linked in https://github.com/garden-co/jazz/pull/2595#issuecomment-5555516062.
- Matching JS, WASM, NAPI, tooling, and native server artifacts pinned to that revision.
- macOS arm64 / Electron application; the regression uses an independent `createJazzClient` user client in Node (WASM), a real local Jazz server, and an external test JWT.
[...]

PR #1714, opened 2026-08-23
Add BandChat reference app

aeplay:
🤖 Adds BandChat as the small reference app for permissions, attachments, offline writes, and reconnect delivery.

Behavior and boundaries:

- Room owners and admins control membership; users cannot self-admit.
- Message authorship is bound to the authenticated session profile.
- Attachment size and media types are bounded.
- Normal browser use remains local-first with IndexedDB retention.

Receipts:

- Deployed edge-policy tests cover owner invitation, guest messaging, self-admission denial, forged authorship denial, and post-removal denial.
- A canonical controllable browser topology proves offline write → reconnect → delivery into a fresh persistent store.
- Headless tests, browser tests, production build, and the sensitive-data guard pass.

The example uses existing Jazz abstractions and the canonical test harness; it does not introduce a new sync or storage path.

PR #1846, opened 2026-08-24
Add BandChat topology and recovery E2E

aeplay:
🤖 Stacked on #1714. Adds app-local adopter-facing topology and recovery coverage for BandChat: concurrent messages and reactions, inline attachments, idempotent waiters, offline replay, ordered subscription convergence, authorization/revocation receipts, and an explicit #1844 large-value expected failure. Fault/soak expansion remains in progress.


aeplay:
🤖 Overnight progress: this topology PR now consumes the shared deterministic harness and is part of the verified stack after #1852, #1842, #1856, and BandChat itself.

The receipt covers trusted JWT bootstrap/admission, planted unauthorized server-write rejection, two clients exchanging messages/reactions/a bytes attachment, duplicate wait, disconnect/offline write/reconnect, and exactly-once convergence. The scenario is registered in the bounded soak runner and its harness tests pass.

The available browser server helper does not yet expose independent edge/core restart control, so this branch deliberately does not claim a restart assertion. That gap remains visible for later harness work rather than being simulated inaccurately. Full browser/TypeScript validation is still pending on a checkout with complete workspace dependencies, so the PR remains draft.

aeplay:
🤖 Changes requested from the independent topology review:

- The focused receipt is born red on the current stacked ABI (decode open config: Serde Deserialization Error) before workload/fault assertions.
- Test clients send both JWT and adminSecret, so admission chooses trusted admin; the unauthorized/JWT claim is invalid. Admin secret may deploy, but client connections must authenticate only with JWT.
- The scenario overclaims edge/core restart, packet control, seed, and soak behavior that its registered targets do not implement.
- Attachment convergence is never asserted.
- The registry command collects unrelated browser files instead of only this receipt.
- Calling wait twice on one mutation is not a duplicate/retry test; count plus ID uniqueness does not prove the expected rows exactly once.
- Cleanup needs an explicit bounded scenario hook.

A repair lane is addressing these together, including an exact NAPI artifact rebuild and focused topology/permissions gates. The PR remains draft.

aeplay:
🤖 Restacked patch-identically onto the combined core stack through #1903 so this app benefits from the canonical-auth, recursive-policy, browser-command, and disjunctive-proof fixes. Focused app/scenario/topology gates pass. The remaining reaction INSERT failure is unchanged and tracked as #1900; this stack deliberately does not choose new permission semantics overnight.

aeplay:

aeplay:
🤖 Restacked all 26 prior BandChat commits exactly onto the combined core and added one reviewed cleanup migration using #1939's operation-local `defer`. New receipts cover overflow-sensitive projected windows, a causally isolated offline write, same-IndexedDB tier-local rehydration while the route is blocked, separate reconnect convergence, and revocation. Manual route bookkeeping was removed; normal reconnect and DB cleanup remain. Static/type/topology-harness gates pass. The full headless suite is 10/11 with only known #1900; exact browser artifacts are the remaining runtime gate.

aeplay:
🤖 Superseded by #2072, which reconstructs BandChat topology coverage on the current #2056/#2066 stack. The replacement preserves the app-level concurrency, offline replay, reconnect, projection, and revocation intent while adding the missing current-membership/correlated-room reaction boundary and adversarial receipts.

PR #1719, opened 2026-08-23
Add BigLabel reference app

aeplay:
🤖 Draft pending final independent review.

Adds BigLabel, a multi-tenant record-label reference app and benchmark seed. It models organizations, memberships, artists, releases, teams, and assignments with live Jazz queries and explicit mutation policies.

Security behavior:
- bootstrap is restricted to a server-issued biglabel_admin claim
- users cannot self-admit to foreign organizations
- role escalation and assignment forgery are denied
- foreign reads and mutations are denied by the deployed authority

The branch includes a real local-server authority receipt covering those boundaries, a live organization discovery/mutation UI, workspace lockfile integration, and focused scenario tests.

Local verification after rebasing onto main: 3 tests passed, TypeScript typecheck passed, and the sensitive-data guard passed. The final review should still plant an executable policy regression against the deployed receipt.

PR #1848, opened 2026-08-24
Add BigLabel browser topology E2E

aeplay:
🤖 Stacked on #1719. Starts bounded, adopter-facing BigLabel browser topology coverage with real schema and permissions, authenticated writer/reader clients, artist/release/assignment writes, and ordered subscription convergence across the core path. Revocation, offline replay, restart, and deterministic fault expansion are in progress.

### DIFF
--- a/examples/big-label/README.md
+++ b/examples/big-label/README.md
@@ -29,6 +29,11 @@ Run `pnpm --dir examples/big-label test` for the deterministic fixture and isola
 
 - `jazz-sim s1_saas`: tenant cold load, indexed owned-slice filtering, relations/includes, write churn.
 - Policy-graph fixture: organization → membership → team/assignment authorization shape.
+- CB-012: the valid release-owner assignment path is covered by both the
+  deployed authority receipt and browser topology phase. It currently exposes
+  a core correlated-`allowedTo.insert()` lowering fault
+  (`__flat_join_source_0_row_uuid`), rather than an app policy denial; see
+  `tests/correlated-assignment.repro.test.ts`.
 - Future blocker: schema migrations/versioned reconnect are recorded for a shared migration/topology lane. This example does not invent a migration mechanism.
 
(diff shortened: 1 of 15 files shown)


aeplay:
🤖 Overnight adversarial review added three independently checked follow-ups: production now fails closed when backend or Better Auth secrets are missing; personal-organization lookup is part of the exclusive transaction read set so concurrent first requests converge on one graph; and direct policy tests now model the explicit external issuer required by canonical sessions. Focused secret, concurrency, authority, type, and planted-regression receipts pass. The full browser receipt still needs a fresh app-head artifact run after the lower canonical-auth repair is finalized.

aeplay:
🤖 Restacked app and topology branches onto combined core `8fc527b76` with all prior commits patch-identical. The topology branch now additionally exercises the app's bounded release-with-artist query across reconnect, persistent reopen, outsider isolation, and membership revocation; a second release makes the `limit(1)` assertion sensitive. Static/type gates pass, and an independent adversarial review is in progress before the serialized artifact-backed run.

aeplay:
🤖 Restacked all 25 prior BigLabel commits exactly onto the combined core and added one independently reviewed #1939 cleanup migration. The receipts now genuinely reuse the same IndexedDB while physically offline, prove persisted revocation, use overflow-sensitive roster/release bounds, require the latest exact subscription snapshot, and keep ordinary clients structurally JWT-only. App, unit, benchmark, topology-harness, formatting, lint, and sensitive-data gates pass; exact browser artifacts remain to run.

aeplay:
🤖 Superseded by #2055. The live hardening from this stack layer—production secret rejection and serialized first bootstrap with durable retry/re-read behavior—has been reconstructed into the backend-only BigLabel bootstrap. Its obsolete claim-based bootstrap path and old stack mechanics were intentionally omitted.

PR #1850, opened 2026-08-24
Add the BandBinder example

aeplay:
🤖 Starts the BandBinder multifunctional Notion-like example with a self-contained Next/Better Auth variant. The schema covers recursive pages and blocks, tasks, calendar entries, songs, attachments, suggestions, and branches; the benchmark stresses ordered sibling pagination and recursive child-page traversal. Recursive permission inheritance and full topology E2E remain follow-up work.


aeplay:
🤖 Restacked onto the combined core through #1903. The prior indexed-policy bootstrap failure is fixed: owner and stage-manager bootstrap now settles successfully. The deeper receipt then exposed a distinct silent edge-settlement failure for an authorized relational root-block INSERT; bounded diagnostics and exact precondition checks are retained here, and the minimized core bug is tracked in #1914. No policy or timeout workaround is included.

aeplay:

aeplay:
🤖 Restacked all 20 BandBinder commits exactly onto #1940 and added one reviewed #1939 cleanup migration. Recovery receipts now use latest snapshots, prove an owner-side negative before reconnect, reopen the same DB behind a blocked route at tier-local, and accept only a post-revocation empty publication. Strict DB teardown remains. TypeScript, permissions, benchmark, and topology-harness gates pass; browser execution still stops at known #1914 before the new phases, with #1916 untouched.

aeplay:
🤖 Superseded by the current-main reconstruction in #2094. The replacement preserves BandBinder’s document/task/calendar/song workload while removing obsolete trusted-author and `session.user_id` assumptions, using Jazz magic authorship and canonical issuer-scoped memberships, and keeping bootstrap off the initial read path. It remains draft through benchmark alignment, topology receipts, and independent review.

PR #1851, opened 2026-08-24
Add the Wequencer example

aeplay:
🤖 Starts the Wequencer collaborative sequencer example and benchmark. It models sessions, tracks, steps, roles, transport observations, and presence; exercises ordered reads, local subscription delivery, and edit bursts; and documents smoke, hotspot, reconnect, partition, presence, and permission soak profiles. Real multi-topology fault E2E remains follow-up work.


aeplay:
🤖 Overnight progress: Wequencer now has indexed session/member/track/step/transport/presence schema, an owner/collaborator UI, deterministic two-context Playwright coverage for disjoint online edits and bounded offline/reconnect convergence, transport-state and viewer-denial receipts, plus a native benchmark fixture with matching ordered playback queries.

The branch intentionally does not assert same-field concurrent pad outcomes beyond ordinary Jazz merge semantics, and observed_at is only UI ordering data—not a clock-sync or audio-authority claim. Rust benchmark tests and hooks pass. App TypeScript/E2E execution still needs a checkout with installed JS dependencies, so the PR remains draft pending independent review and that gate.

aeplay:
🤖 Wequencer is updated with the final reviewed repairs.

- The benchmark now executes and reads the real `sessions.orderBy("$createdAt", "desc")` query shape and asserts deterministic title order.
- Presence heartbeats are timer-driven rather than subscription-driven: acknowledgement rerenders cannot create a write loop, React Strict Mode cleanup cancels the probe schedule, committed mounts publish on the intended 5-second cadence, and unmount/session changes clean up.
- The edit-burst receipt now asserts all 64 ordered pad states and exact enabled count 27 instead of a permissive upper bound; the browser receipt also checks exact 64-pad convergence on both peers.

Independent review planted dropped edits, shifted positions, cadence changes, and acknowledgement rerenders. The repaired receipts caught them and were restored. Rust benchmark tests (6/6), heartbeat tests (3/3), typecheck, bench compilation, and hooks pass.

The branch is published for visibility but will be physically restacked onto the final canonical-auth/core head before the browser topology is called ready; its current browser run remains blocked by the shared artifact/core stack rather than being skipped.

aeplay:
🤖 Restacked onto the combined core through #1903. The example now uses canonical `session.author` only when comparing Jazz magic authorship columns, while retaining raw `session.user_id` for modeled user IDs. Reverting that distinction reproduces an immediate server denial; with it restored, the full five-phase topology receipt passes on fresh artifacts.

The deeper receipt also uncovered an independent authorized-delete settlement hang, captured with a minimized reproduction in #1913. No workaround for that core bug is included here.

aeplay:
🤖 Restacked onto combined core `8fc527b76` with all prior commits patch-identical. A follow-up fixes the app's focused browser-test command so it uses the app's own browser root/config while retaining Jazz's strict path validation and stale-artifact preflight; the previous command incorrectly rejected every app test as outside `packages/jazz-tools/tests/browser`. Contract tests pass and the corrected command now fails helpfully at the expected stale-artifact boundary pending the serialized fresh-artifact run.

aeplay:
🤖 Restacked all 18 Wequencer commits exactly onto #1940 and added reviewed compensation plus timeout-budget commits. The topology receipt proves current-state subscription transitions, repeated peer-negative reads while the owner edit is offline, same-cache tier-local rehydration, exact projection/bounds, and separate reconnect convergence. The outer 275s budget now accounts for six phases, five faults, compensation, scenario cleanup, scheduler close, and buffer. TypeScript, unit, benchmark, topology-harness, formatting, and lint gates pass; #1913 remains untouched.

aeplay:
🤖 Superseded by the current-main reconstruction in #2097. The replacement preserves Wequencer’s collaborative sequencer, ordered subscription, presence, topology, and benchmark surfaces while removing obsolete trusted-author and raw-provider-identity assumptions, using Jazz Better Auth persistence, canonical authors, and explicit server bootstrap. It remains draft through topology verification and independent review.

PR #1853, opened 2026-08-24
Add the RecordPlayer example

aeplay:
🤖 Adds the first self-contained RecordPlayer example slice.\n\n- Next.js app scaffold for albums, tracks, playlists, and playlist membership\n- Jazz schema and query shapes for catalogue and playlist-window behavior\n- isolated Divan/CodSpeed benchmark fixture\n- workspace and generated Next configuration\n- records large-audio topology coverage as follow-up work pending the public large-value hydration fix\n\nFocused checks pass: Rust benchmark tests, TypeScript, and production Next build.


aeplay:
🤖 The draft is now materially advanced and rebased onto the shared topology/core-policy stack at f1d11ca8b.

Implemented:
- real public streaming audio creation;
- projected, bounded metadata reads that exclude audio bytes;
- exact indexed playlist windows aligned with the benchmark;
- invitation acceptance that preserves playlist, subject, and role, preventing self-promotion/reassignment;
- mounted query-hook UI;
- real browser→edge→core two-session topology receipts for listener/editor/owner boundaries, revocation, rollback, and offline convergence.

The previous Map-based fake scenario was deleted. Independent review planted an allow-all policy mutation and proved why it was inadequate.

The real receipt then found two shared blockers:
- #1866: even an allow-all one-table browser write settles locally but never at tier edge, with no mutation error;
- #1867: stale N-API/WASM artifacts can negotiate different wire protocol versions; the harness now has a structured diagnostic receipt.

Normal app tests/build and Rust benchmark are green. The browser receipt remains intentionally red on #1866; no assertion was skipped or weakened.

aeplay:
🤖 Restacked onto the combined core through #1903 and expanded with a minimized actual-app receipt. Playlist creation, invitation creation, and recipient visibility settle; the first recipient acceptance UPDATE allocates a transaction ID but remains pending at edge without rejection. The exact boundary and likely old/new self-EXISTS policy contradiction are recorded in #1907. The WIP keeps that bounded failing receipt visible and does not weaken the policy, waits, or compiler semantics overnight.

aeplay:

aeplay:
🤖 Restacked all 49 RecordPlayer commits exactly onto #1940 with no cleanup migration required. New reviewed receipts cover a genuinely pull-driven 4×32KiB stream, offline/reconnect causality, exact 128KiB peer materialization, asynchronous failure after partial consumption with atomic absence across local/edge peers, and runtime metadata-projection omission. Unit, TypeScript, selection, benchmark, and optimized compile gates pass; exact browser artifacts remain to run.

aeplay:
🤖 Superseded by the current-main reconstruction in #2081. The replacement preserves the useful RecordPlayer app, playlist/media query and benchmark surfaces while removing obsolete trusted-author, `session.user_id`, provider, and topology assumptions. It remains draft while permissions, auth, focused receipts, and independent adversarial review are completed.

PR #1856, opened 2026-08-24
Fix correlated authorization join carriers

### LINKED ISSUE #1855: Correlated allowedTo insert loses its synthetic join field
🤖 BigLabel exposed a correctness failure when a valid tenant admin inserts a release-owner assignment governed by correlated `allowedTo.insert(releaseId)` authorization plus an `exists` tenant check.

The server rejects the valid insert with:

```
Storage: graph field not found: __flat_join_source_0_row_uuid
```

This is not an app-policy direction problem: trusted backend bootstrap, admin membership, artist/release creation, and roster visibility all succeed first. A focused native receipt fails in about 270 ms, and the browser topology otherwise presents it as an unresolved edge wait.

Precise reproductions:

- `examples/big-label/tests/authority.server.test.ts` — valid-admin positive receipt added on the topology branch
- `examples/big-label/tests/correlated-assignment.repro.test.ts` — existing CB-012 minimized reproduction

Expected: the authorized assignment commits and becomes visible.
Actual: lowering/runtime looks up the synthetic `__flat_join_source_0_row_uuid` graph field and aborts.

The fix should preserve the single access-path/lowering mechanism rather than introduce a second policy execution path.

aeplay:
🤖 Fixes #1855.

Correlated authorization joins produced by `allowedTo.insert(...)` plus an `exists` check were incorrectly treated as public-result occurrences. Lowering therefore requested the synthetic `__flat_join_source_0_row_uuid` field from an authorization-only plan and valid writes failed with a storage error.

This keeps the single existing occurrence/access-path lowering mechanism, but threads the request context through it:

- authorization subplans do not emit public occurrence-carrier fields
- ordinary public joins retain their existing occurrence identities and behavior
- a compiler regression preserves the exact formerly-invalid plan shape
- a NodeState receipt proves an eligible owner assignment commits
- the same owner with foreign-tenant membership still fails with `AuthorizationDenied`

Independent adversarial review confirmed that authorization predicates retain their internal join fields; only public-result occurrence carriers are suppressed. Planted compiler and authorization mutations proved both positive and negative sensitivity.

Focused tests and full CI are green.

### DIFF
--- a/crates/jazz/src/node/query_engine/lowering/collect_layout.rs
+++ b/crates/jazz/src/node/query_engine/lowering/collect_layout.rs
@@ -55,7 +55,7 @@ pub(super) fn collect_layout(
             })
         })
         .collect::<Vec<_>>();
-    let root_occurrence_inputs = root_join_occurrence_fields(plan, resolved_sources)?
+    let root_occurrence_inputs = root_join_occurrence_fields(plan, resolved_sources, request)?
         .into_iter()
         .filter(|(name, _)| available_fields.contains(name))
         .map(|(name, value_type)| {
--- a/crates/jazz/src/node/query_engine/lowering/graph_lowering.rs
+++ b/crates/jazz/src/node/query_engine/lowering/graph_lowering.rs
@@ -1190,6 +1190,12 @@ fn lower_linear_plan_steps_cached(
                     Some(LinearStep::Join { .. })
                 );
                 if *mode == JoinMode::Inner && next_is_join {
+                    // Consecutive joins must retain the right input for the
+                    // next predicate. Public queries also use those values as
+                    // occurrence carriers; authorization subplans keep the
+                    // same single lowering path, but mark them internal so a
+                    // decision terminal never asks for public row identity.
+                    let policy_subplan = omits_public_occurrence_carriers(request);
                     let (_, right_nullable, right_depths, right_fields) = last_join_right
                         .take()
                         .expect("inner join records its right input");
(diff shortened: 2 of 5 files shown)


aeplay:
🤖 Overnight adopter testing found that #1856's isolated authorization-subplan regression is not sufficient for the production BigLabel shape.

The deterministic app path is a normal public Identity terminal whose read-policy expansion creates assignments → releases(join_via:0) → tenant joins. Its output layout legitimately requests __flat_join_source_0_row_uuid and _1, but graph execution reports the first field missing. Both valid owner assignment and cross-tenant denial receipts fail with GraphFieldNotFound after a fresh NAPI rebuild.

This is not a reason to suppress occurrence carriers: they are public-terminal join identity here. The next repair must make the existing graph path actually materialize every carrier requested by root_join_occurrence_fields, with a core regression matching the converted allowedTo.insert + tenant EXISTS shape. #1856 should not be called complete until both BigLabel receipts pass.

aeplay:
🤖 The production BigLabel/PosterShop policy shape exposed two additional carrier losses, now fixed at b0d6303b6.

- correlated authorization inputs are retained under policy-private carrier names and never enter public result occurrence schemas;
- public row-ID occurrence carriers from earlier inner joins survive a trailing semi-join so collectors can still distinguish root occurrences.

Independent adversarial review found no leakage/collision seam. It verified malicious naming isolation, mixed public/policy carriers, and planted regressions for both fixes. Focused graph-lowering suite: 23 passed; rebuilt-NAPI BigLabel authority and correlated-assignment receipts passed. PosterShop’s preserved correlated policy receipt will be restacked onto this head as another production-shaped confirmation.

Fresh CI is running; upper auth/example branches will be rebased onto this fix as their active local work permits.

Issue #1857, opened 2026-08-24
Direct oversized Text insert spins synchronously


aeplay:
🤖 A real native server-convergence topology for #1844 exposed an earlier creation-path failure: inserting a direct Text value larger than 64 KiB through `JazzClient.insert` spins synchronously at 100% CPU before the returned write or its timeout can yield to the JS event loop.

The behavior reproduces with Text alone and occurs before remote hydration is involved. It prevents ordinary large-value creation and also makes an in-process timeout ineffective because the event loop never regains control.

Expected: a direct oversized primitive either lowers through the large-value streaming/staging path asynchronously or returns a bounded, actionable error directing callers to streaming creation.
Actual: synchronous spin/hang.

The ongoing #1844 lane will minimize this in a watchdog-isolated receipt while using the runtime streaming-mutation surface to continue the remote-chunk topology test.

aeplay:
🤖 The public-hydration topology work found a second minimal reproduction that the repair lane is now covering explicitly: runtime insertStreaming can peg the JS worker before its first await even around a 6 KiB Text value. This happens before any edge request or read-side hydration.

The preserved two-client/server Text/Bytea/JSON topology receipt is linked to this issue and skipped rather than weakened; it will be enabled once creation yields correctly. The repair must therefore prove both direct oversized insert and streaming creation with watchdog-isolated regressions before this issue is closed.

aeplay:
🤖 Correction after rebuilding exact artifacts from clean main: direct NAPI, streaming NAPI, and streaming WASM all complete in roughly 0.6s in the watchdog matrix. Only direct WASM oversized insertion blocks the event loop and is OS-killed.

The earlier 6 KiB insertStreaming observation came from a stale/mismatched artifact path and is not the same defect. Streaming already crosses future_to_promise; direct WasmDb insertion still enters async storage-backed large-value lowering through synchronous block_on.

The repair is therefore being scoped to the direct WASM binding while retaining streaming controls for Text, Bytea, and JSON. It also adds bounded construction/yields, failed-upload cleanup, and a malformed-JSON cleanup regression.

PR #1863, opened 2026-08-24
Hydrate large values across public bindings

aeplay:
🤖 Draft overnight stack layer for #1844.

Public NAPI and WASM reads now treat a missing large-value chunk as an asynchronous storage dependency instead of leaking physical descriptors, blocking the JavaScript thread, restarting hydration, or spinning empty pulls.

Behavior covered in this branch:

- flat, relation, transaction, and subscription results hydrate Text, Bytea, and JSON before public encoding;
- one-shot reads, identity reads, transaction reads, partial ranges/JSON pointers, and append/splice use retained futures;
- WASM ordinary reads return Promises rather than synchronously blocking async storage;
- transient missing chunks retain demand and resume after routed transport wake; fatal errors terminate exactly once;
- retry schedules use generations/deadlines, ignore stale carriers, and disarm predecessor deadlines on failover;
- suspended NAPI subscription batches retain at most 256 events and drain later batches without loss;
- docs remove obsolete scalar/no-large-values language.

Focused Rust, NAPI, WASM, and Jazz-tools receipts pass. The full native-runtime file has the exact same six pre-existing NotCovered unhandled rejections as the measured base.

One real two-runtime/server Text/Bytea/JSON topology receipt is intentionally preserved and skipped against #1857. It is parameterized and can be enabled directly once the creation-path repair lands.

This PR is intentionally draft. It still needs:

- the #1857 fix integrated and the real server topology enabled;
- a final adversarial pass over the repaired head;
- actual restacking after #1750 so upper app layers inherit both identity and hydration fixes;
- canonical full gates after that restack.

### DIFF
--- a/crates/jazz-napi/index.d.ts
+++ b/crates/jazz-napi/index.d.ts
@@ -77,11 +77,11 @@ export declare class NapiDb {
   setLargeValueStagingPolicy(incomingBytesPerWindow: number, windowMs: number, maxAgeMs?: number | undefined | null): void
   /** Run one idempotent expiry pass; native hosts normally call this on a timer. */
   evictExpiredStagedLargeValues(): number
-  readValueRange(table: string, rowId: Uint8Array, column: string, start: number, end: number): Uint8Array
-  readTextUtf16Range(table: string, rowId: Uint8Array, column: string, start: number, end: number): string
-  readJsonPointer(table: string, rowId: Uint8Array, column: string, pointer: string): string | null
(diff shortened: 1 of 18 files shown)


aeplay:
🤖 Superseded by #2083, reconstructed on current `main`. The replacement deliberately leaves already-correct WASM/core behavior alone and addresses the remaining NAPI public-read gap with a pollable, thread-affine hydration boundary. It remains draft while relation, transaction, subscription, and independent adversarial coverage are completed.

PR #1868, opened 2026-08-24
Preserve multi-key correlated policy joins

aeplay:
🤖 Fixes multi-key correlated Exists policy lowering without adding a second authorization path.

A denormalized workspace relation must be able to prove both referenced.id = proposed.referenceId and referenced.workspaceId = proposed.workspaceId, then compose that with membership workspace/subject/role constraints. The converter previously dropped outer equalities or replaced the relation root with its first nested join; execution then lost the private row-id carrier.

This branch:
- preserves all safe outer correlations and the relation root;
- lifts exactly one nested equality-key correlation and fails closed for mismatched or deeper erased scopes;
- consumes the private projected id carrier after relation projection;
- proves same-workspace acceptance plus cross-workspace insert/update denial for the same subject in two workspaces;
- covers delete/old-row policy, compiler fail-closed boundaries, and public descriptor/carrier non-leakage.

Independent adversarial review found no P0/P1 issue. Focused new and legacy policy receipts, planted correlation/delete mutations, clippy, formatting, and sensitive-data gates pass.

This draft is based on #1856 and will be added to the existing official stack once the active upper auth branch is rebased.


aeplay:
🤖 The public permissions DSL can now express the same safe nested filtered RHS already covered by the production-shaped Rust authorization receipt, e.g. joining a referenced row to `policy.members.where({workspaceId: outer.workspaceId, subject: session.user_id})` without collapsing those predicates into a trailing filter.

An adversarial review found that the first signature admitted arbitrary relation shapes whose output scope was not defined. The final API therefore deliberately accepts only a filtered single-table RHS. It rejects unfiltered wrappers, same-table alias collisions, nested joins/hops, select, union, and gather early with targeted diagnostics; it does not introduce another lowering path.

Final independent review approved the guarded form. Exact TS/public receipts, Rust conversion, and same-subject cross-workspace authorization negatives pass; planted guard bypasses were caught and restored.

aeplay:
🤖 Superseded by draft #1958, which ports the still-live policy/provenance commits onto current main without duplicating nine patches already represented there. #1958 is still under adversarial review, but it is now the authoritative continuation of this work.

PR #1869, opened 2026-08-24
Retain provenance keys for ordered collectors

aeplay:
🤖 Retains provenance keys used by root and nested collector windows without exposing them unless explicitly selected.

Flat and nested queries such as orderBy("$createdAt") previously translated correctly but lost the provenance key across collector projection/layout, yielding an unsupported window-key capability error and browser edge-coverage timeouts.

This branch uses the existing Rust lowering path to:
- retain $createdAt, $updatedAt, $createdBy, and $updatedBy sort/tie keys internally;
- preserve duplicate-time deterministic row-id ties with offset/limit;
- support nested collector slots;
- keep ShapeDefault/wildcard and nested descriptors free of hidden magic columns;
- expose only explicitly selected provenance fields.

Independent adversarial review approved it. Twenty-seven terminal tests, row-provenance receipts, 11 adapter tests, and planted retention/leak mutations pass.

This draft is stacked on the multi-key policy fix and will be linked into the existing official stack after the active upper auth branch is rebased.

### DIFF
--- a/crates/jazz/src/node/query_engine/lowering/collect_layout.rs
+++ b/crates/jazz/src/node/query_engine/lowering/collect_layout.rs
@@ -495,13 +495,16 @@ pub(super) fn collect_slot_builder(
 ) -> CollectBySlotBuilder {
     CollectBySlotBuilder::new(
         std::iter::once(parent_row_id.to_owned()).chain(route_fields.iter().cloned()),
-        slot.fields.iter().map(|field| {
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Superseded by #2033, which reconstructs the same two attributable commits directly on current `main` and preserves the intended hidden provenance-key behavior without the retired stack dependency. Focused root/nested collector and adapter receipts pass there.

Issue #1865, opened 2026-08-24
Cover oversized WASM update, upsert, and restore


aeplay:
🤖 Add a direct browser-WASM black-box receipt for oversized whole-value update, upsert, and restore under the current generic resident-write contract.

## Current architecture

#1870's proposed large-value-specific async binding APIs are obsolete. Mutation initiation remains synchronous and installs validated resident state; persistence and any chunk dependency progress through `Db::tick`, as specified by `INV-TX-25` and `SPEC/async_ordered_storage_handover.md`. Persistence failure poisons the live runtime rather than retracting already-published resident output. Large-value staging/admission is implemented by #1842 and lifecycle/refcount hardening continues in #2024.

Current browser-WASM coverage exercises streaming insert/update/upsert and malformed or rate-limited rejection, while NAPI has direct oversized-text receipts. The remaining gap is a direct 64 KiB+ browser-WASM receipt across the ordinary whole-value APIs.

## Acceptance evidence

- Direct browser-WASM update, upsert, and restore for Text, Bytea, and JSON above the inline threshold.
- Same-tick resident visibility followed by bounded tick-driven persistence.
- No executor deadlock, partially admitted row, orphan pending upload, or second binding/evaluator path.
- Explicit malformed JSON and injected storage-failure behavior matching the generic resident-write contract.
- JSON retains its logical kind and RFC 6901 pointer behavior.
- Existing streaming, NAPI, and deferred-local-persistence receipts remain green.

PR #1870, opened 2026-08-24
Preserve synchronous direct large-value creation

aeplay:
🤖 Preserves the synchronous direct-insert contract for oversized Text, Bytea, and JSON values while keeping durable blob publication asynchronous.

Direct values become immediately resident in IVM/query/subscription state. Groove journals the exact descriptor and stable receipt identity, publishes chunks and metadata asynchronously, validates authenticated reachability plus logical Text/JSON content, and atomically promotes the owner row. Failed, retracted, or rejected promotions publish inverse deltas and clean their durable intent so the same key can be retried.

This is stacked on #1869 so it also benefits from the combined policy/provenance fixes. It does not add a new public async API or change direct-write semantics.

Verification:

- `cargo check -p groove -p jazz -p jazz-wasm`
- resident rollback suite: 7/7
- failed promotion/same-key retry
- oversized Text/Bytea/JSON: 3/3
- rebuilt NAPI/WASM watchdog: 8/8
- independent adversarial review of descriptor substitution, kind mismatch, unreachable chunks, malformed content, replay, and double activation: approved with no blockers

Related follow-up: #1865 tracks the adjacent direct oversized update/upsert/restore paths.


aeplay:
🤖 The first full CI run exposed three real lifecycle integration regressions plus a test typecheck error; all are repaired in the latest head.

- direct resident chunk-finalization failures retain their typed backend error instead of being flattened;
- encoded chunk bytes are subject to ingress limits and rejected roots are fully released;
- writes remain immediately resident/query-visible, but `wait({tier: "local"})` stays pending until persistence actually settles;
- the native/WASM watchdog explicitly requires the optional capabilities it exercises;
- an authenticated browser receipt now covers local + edge settlement, no mutation error, and peer delivery.

Focused former failures and resident rollback/retry gates pass. Independent adversarial review planted early-local completion, error flattening, and rate-limit bypass; each was caught by the intended receipt and restored. Final source review approved `7ac67abb3`; CI is rerunning on the combined head.

aeplay:
🤖 The branch-local CI repairs are now pushed at `4b0a11eb4` after independent adversarial review.

- Native runtime test doubles now always provide the required tick capability; planted missing and swallowed-error variants fail the intended receipts.
- The Groove blob-stage failure receipt retains the exact typed `ChunkError::Backend` boundary and proves eager row/journal/chunk cleanup plus a successful content-correct retry.
- The earlier `streaming-wasm-json` protocol failure was reproduced only with stale generated artifacts. After a provenance-verified NAPI/WASM rebuild, the exact fixture passes through `start`, `returned`, and `local`.

Review gates: 113 native-runtime tests passed / 1 skipped; the typed Groove receipt and two related cleanup receipts passed; the fresh-artifact streaming fixture passed. The reviewed range changes only three test files and the worktree was clean. Fresh CI is running on the pushed head.

aeplay:
🤖 Superseded by the generic resident-write contract now on `main`, #1842's large-value staging/admission work, and the lifecycle hardening in #2024. The old large-value-specific async binding/overlay path and inverse retraction semantics should not be restored. The remaining direct oversized browser-WASM coverage gap is retained in the updated #1865.

PR #1873, opened 2026-08-24
Fail fast on stale native artifacts

### LINKED ISSUE #1867: Detect stale N-API/WASM protocol artifacts before browser tests
🤖 Browser topology tests can silently load stale checked-in/built native artifacts even after ordinary Rust builds complete.

Observed while running the RecordPlayer topology receipt:
- the freshly rebuilt WASM browser runtime spoke wire protocol 12;
- startLocalJazzServer loaded crates/jazz-napi/jazz-napi.linux-x64-gnu.node at protocol 11;
- the first server response was a structured unsupported_protocol_version error before hello;
- cargo build did not refresh the Node-loaded .node artifact;
- pnpm --filter jazz-napi build:debug repaired the mismatch.

The shared browser harness should compare/report native and WASM protocol versions before starting application phases, with a command that rebuilds the exact loaded artifacts. Focused browser gates should depend on that artifact step so stale binaries cannot masquerade as transport ordering or auth failures.

aeplay:
🤖 Makes native NAPI/WASM artifact publication atomic, provenance-complete, and fail-closed before runtime startup.

The previous checkout could combine stale generated bindings, manifests, or declarations from different builds. This stack now:

- serializes clone-wide native producers through the verified lease from #1876;
- builds NAPI into a private sealed generation and activates one atomic runtime pointer only after every ABI, provenance, and stable-declaration check passes;
- keeps the tracked CommonJS bootstrap immutable and fingerprints it as part of the published ABI, while a distinct ignored pointer selects the local generation;
- derives local and release expectations only from sealed manifests whose files, fingerprints, target/profile/toolchain, and package inputs are complete and mutually consistent;
- stages release expectations from the actually downloaded artifacts, including all four WASM outputs and every NAPI target;
- preserves the previous working generation on every pre-activation failure and removes abandoned stages/generations;
- gives CommonJS and actual ESM named imports the same guarded binding surface.

Failure diagnostics distinguish a source checkout needing an artifact rebuild from installed dependency/version skew and report the actionable mismatch.

Independent adversarial review ran seven iterations. Planted stale/partial generations, missing or malformed manifest fields, fingerprint-only drift, tracked-bootstrap drift, declaration drift, cross-target disagreement, forged leases, concurrent producers, pointer failures, and rollback failures are all rejected by committed receipts. After rebasing on #1876, 38 focused artifact tests pass; the broader reviewed gates include 63 CI-workflow tests, TypeScript, formatting, and diff checks.

Closes #1867. Stacked directly on #1876.


aeplay:
🤖 Superseded by #1981, stacked on #1980. The replacement seals native generations before activation and verifies exact declarations, release state, and the complete normal Cargo path-dependency provenance closure.

PR #1874, opened 2026-08-24
Deny client access to generated Better Auth tables

joeinnes:
## What changed

- Generate explicit deny rules for reading, inserting, updating, and deleting every Better Auth table.
- Support generated table names that are not valid JavaScript identifiers.
- Merge those rules into the official Better Auth example alongside the app's message rules.
- Document the required permission setup and add a patch changeset for `jazz-tools`.

## Why

Jazz closes access one table at a time. Adding rules for an application table does not protect other tables that were left out of the permission set. The generated Better Auth tables were left out, so ordinary clients could access authentication records.

The generated rules now deny all ordinary client access to those tables. The Better Auth adapter still uses its backend connection, so its database operations continue to work.

## Testing

The first commit adds the regression tests and leaves them failing:

- Three Better Auth adapter checks fail because the generated schema has no table rules.
- The example check fails because its final permission set has no rules for the generated auth tables.

The second commit adds the fix:

- Better Auth adapter tests: 30 passed. The suite's existing duplicate-email restart case remains marked as an expected failure.
- Better Auth example browser tests: 3 passed.
- `jazz-tools` TypeScript check passed.
- Pre-commit formatting, lint, documentation type checks, and sensitive-data checks passed.

## Build note

A separate full test-artifact build rebuilt the release NAPI and WebAssembly artifacts, then stopped in an unrelated existing `jazz-tools` test-build step because `Required<Where>` no longer has `mode` and `@vue/server-renderer` is missing. The artifact verifier and the focused tests above passed.

### DIFF
--- a/.changeset/secure-better-auth-tables.md
+++ b/.changeset/secure-better-auth-tables.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Generate deny-all client permissions for Better Auth tables while retaining backend adapter access.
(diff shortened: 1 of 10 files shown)


aeplay:
🤖 Follow-up captured in #1946: revisit a safe, explicitly bounded way to expose Better Auth user rows read-only for local-first application use, while keeping credentials, sessions, and other generated auth data default-deny.

aeplay (approved):

PR #1876, opened 2026-08-24
Publish WASM artifacts atomically

### LINKED ISSUE #1872: WASM optimization can consume a missing build intermediate
🤖 `pnpm build:core` can finish the Rust/WASM release compile and then fail in the wasm optimization step because its declared input intermediate does not exist:

```text
wasm-opt ... crates/jazz-wasm/pkg/jazz_wasm_bg.wasm-opt.wasm
... No such file or directory
```

This was reproduced while rebuilding exact browser artifacts for the Wequencer E2E after its Rust/query receipts passed. It leaves the worker artifact stale, so subsequent browser topology tests time out before they can exercise the application behavior.

Please make the artifact pipeline atomic and self-describing: each stage should consume a uniquely owned output that the preceding stage has successfully produced, fail immediately with the producing command/context if it is absent, and publish/replace the shared artifact only after the complete build succeeds. The focused rebuild path should also be safe when other lanes are compiling concurrently.

This is related to #1867's stale-artifact detection, but addresses the producer-side missing-intermediate failure rather than consumer-side protocol validation.

aeplay:
🤖 Makes every WASM package producer publish one complete, provenance-matched generation under a clone-wide cross-process lease.

WASM compilation writes to an ignored, provenance-excluded, same-filesystem stage. The producer validates regular/nonempty files and hashes, seals the matching manifest inside the stage, then journal-swaps the complete directory into `pkg`. Failures and interrupted rename states recover the previous good generation; unsafe journal paths, symlinks, malformed hashes, forged/dead leases, and external recovery targets are rejected before mutation.

Fast and release child processes serialize through the same verified lease. Root/test build callers no longer rewrite manifests after publication. Turbo caching is disabled specifically for the live `jazz-wasm` package outputs, so a cache restore cannot bypass the atomic publisher; Cargo/sccache remain available.

Receipts cover incomplete and failed stages, every rename/recovery boundary, manifest failure, prepared-before-rename recovery, external symlinks, unsafe journal paths, forged/dead leases, provenance contamination, and real independent fast/release producers. Final independent adversarial review approved the hardened design; `pnpm test:ci-workflow` passes 59 tests, and concurrent real fast producers serialize cleanly.

Closes #1872.


aeplay:
🤖 The CI/lease repair is pushed at `33d0e89e3` after independent adversarial re-review.

The aggregate artifact build now carries one explicit, parent-selected lock capability through strict Turbo environments. `publishWasmPackage` no longer accepts an `alreadyLocked` boolean: inherited publishers must present the exact live token/path capability, which is revalidated before any package or stage mutation; direct publishers acquire normally.

Evidence:

- Real aggregate NAPI + CLI + atomic fast-WASM + jazz-tools build completed with one parent lock acquisition, no nested/split lock, fresh provenance/load checks, and clean release.
- Missing, forged, different-path, dead, and child-overridden capabilities reject before mutation.
- Direct and concurrent publishers serialize.
- Planted removal of publish-time verification and selected-path propagation both fail their contracts.
- Focused atomic/pipeline contracts, full CI-workflow contracts (63), formatting, and cache-input checks pass.

The previous one-off browser HTML-for-WASM response remains tracked separately in #1880; exact remount receipts pass repeatedly and were not weakened. The aggregate also exposed an orthogonal stale NAPI declaration/cache gap, which the stacked #1873 native-generation/provenance work is closing rather than broadening this WASM-only PR.

aeplay:
🤖 The sole trusted TypeScript failure at `68402803c` was a bounded-test false positive, not an artifact-lease regression: `direct-wasm-text` reached creation return, then its fixture performed three additional resident update/delete/restore persistence handoffs and exceeded the five-second child-process budget under full-suite CI contention.

`55ac79a24` keeps the watchdog finite but gives that complete receipt 12 seconds (with a 15-second parent bound), and adds phase markers around the resident handoff so any future timeout identifies whether creation or settlement stalled. Locally the focused fixture completed 12/12 runs inside an independent five-second outer bound, the full 8-mode watchdog suite passed, and TypeScript/format/diff checks are clean. The historical infinite creation spin still times out deterministically.

aeplay:
🤖 Superseded by #1980, which preserves the reviewed intent while publishing WASM artifacts through private staging, verified leases, journaled recovery, and an atomic generation switch.

PR #1877, opened 2026-08-24
Align WebSocket admission with the wire protocol limit

joeinnes:
## Summary

- use the shared 2 MiB wire-frame limit for Axum WebSocket frame and message admission
- add an integration regression that sends a valid 1.5 MiB fragmented batch and verifies the connection remains usable
- keep fragmentation negotiation local to the boundary test so existing WebSocket scenarios retain their original feature set

Previously, the wire decoder accepted frames up to 2 MiB while the HTTP WebSocket layer rejected messages above 1 MiB before decoding them.

## Red/green commits

1. `29cdcaa0b` adds the failing boundary regression
2. `8558823e2` aligns the admission cap with `MAX_WIRE_FRAME_BYTES`

## Verification

- old limit: the static equality test reported 1 MiB vs 2 MiB, and the integration test observed the server closing the socket after a protocol-sized batch
- new limit: `cargo test -p jazz-server server::routes::websocket::tests -- --test-threads=1` (28 passed)
- pre-commit: `cargo clippy --package jazz-server -- -D warnings`, rustfmt, and sensitive-data check passed


aeplay (approved):

PR #1884, opened 2026-08-24
fix(policy): preserve correlated inherits predicates

### LINKED ISSUE #1875: Inherit correlated SELECT policies through allowedTo.read
🤖 `allowedTo.read(foreignKey)` fails closed when the referenced row's SELECT policy contains a correlated `exists`, preventing otherwise-authorized dependent inserts.

Production-shaped BandChat case:

- inserting a reaction uses `allowedTo.read(messageId)`;
- the referenced message SELECT policy requires `roomMembers.exists(roomId = message.roomId, userId = session.user_id)`;
- the public policy compiles to `Inherits { operation: Select, via_column: "messageId" }` over that correlated SELECT policy;
- the guest can already read the message at edge, but the reaction insert is denied.

This reduces to the existing ignored core receipt:

```sh
cargo test -p jazz-testkit --test policies_integration \
  mixed_predicates_claims_exists_and_inherits_fail_closed -- --ignored
```

It currently fails during server-shell conversion with:

```text
core schema policies do not support Inherits { operation: Select, ... } yet
```

Implement this through the existing native policy-composition/lowering path. Preserve correlation to the referenced row and session claims, reject same-user cross-container witnesses, retract authorization when the witness/read policy closes, and keep internal carriers hidden. Do not replace the app policy with a weaker direct membership check; the semantic value of `allowedTo.read` is that it inherits the referenced row's complete SELECT policy.

aeplay:
🤖 Fixes correlated `SELECT` / `INHERITS` policy lowering without introducing a second policy-evaluation path.

Two independent omissions combined in the failing adopter shape:

- public-schema conversion did not recurse through `PolicyExpr::And`, dropping nested correlated predicates;
- compound claims such as arrays were incorrectly considered collector routing keys even though they are valid prepared-predicate inputs but are not orderable collector keys.

The repair keeps all claims as prepared inputs, recursively retains `And`, and uses one public Groove predicate—`collect_by_ordered_scalar`—at validation and both Jazz claim-discovery boundaries. Nullable ordered scalars can route; arrays, tuples, records, and enums remain predicate-only. Existing scalar routing and the terminal-layout boundary remain unchanged.

Verification includes mixed scalar + array + `EXISTS` + nested `INHERITS`, array membership, same-shape scalar routing, recursive terminal facts, BandChat-like correlated lowering, nullable scalars, and hidden-route layout. Independent adversarial review planted unconditional array routing, non-recursive `And`, missing scalar routing, rejected nullable scalars, and publicly exposed hidden routes; every mutation was caught. Full Groove gates pass (481 passed, 2 ignored).

Closes #1875. Stacked on #1870; it will be rebased above the combined native-artifact fixes before the dependent auth/example work is published.

### DIFF
--- a/crates/groove/src/ivm/graph.rs
+++ b/crates/groove/src/ivm/graph.rs
@@ -11,7 +11,7 @@ use std::hash::{BuildHasher, Hash, Hasher};
 
 use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
-use crate::records::{RecordDescriptor, Value, ValueType};
+use crate::records::{RecordDescriptor, Value, ValueType, collect_by_ordered_scalar};
 use thiserror::Error;
 
 use super::op_types::*;
@@ -1964,25 +1964,6 @@ fn expect_arrangement_inputs(inputs: &[NodeOutput]) -> Result<(), GraphValidatio
         Ok(())
     } else {
         Err(GraphValidationError::InvalidNodeOutput)
-    }
-}
-
-fn collect_by_ordered_scalar(value_type: &ValueType) -> bool {
-    match value_type {
-        ValueType::Nullable(inner) => collect_by_ordered_scalar(inner),
-        ValueType::U8
-        | ValueType::U16
-        | ValueType::U32
-        | ValueType::U64
-        | ValueType::I32
-        | ValueType::I64
-        | ValueType::F64
-        | ValueType::Bool
-        | ValueType::String
-        | ValueType::Bytes
-        | ValueType::Uuid
-        | ValueType::EnumTag(_) => true,
-        _ => false,
     }
 }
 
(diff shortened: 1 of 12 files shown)


aeplay:
🤖 Current red is inherited stack state, not a #1884 policy failure. Rust differential/workspace/focused gates pass; the trusted TypeScript failure is the same five-second `direct-wasm-text` watchdog false positive already repaired and green in #1876 (`55ac79a24`). The aggregate `test-rust` check only reflects that trusted-group failure.

This draft is intentionally still based directly on #1870 while #1873 finishes adversarial artifact-provenance review. I’ll rebase #1884 above #1876/#1873 once that lower stack is finalized, which will bring in the watchdog repair without duplicating it here.

aeplay:
🤖 Superseded by #2045, reconstructed directly on current `main` with the still-live correlated-policy behavior, original commit attribution, explicit before/after examples, and fresh adversarial receipts. The obsolete canonical-token portion of this historical stack was intentionally not carried forward.

Issue #1871, opened 2026-08-24
Correlated read-policy witness does not publish its parent row


aeplay:
🤖 A row made readable by a correlated `exists` policy remains invisible even after the client has received the exact witness row at the edge.

Reduced from the BandBinder browser topology on the combined #1868/#1869 stack:

- `workspaces.allowRead`: `ownerSubject = session.user_id OR members.exists(workspaceId = workspace.id, subject = session.user_id)`
- `members.allowRead`: the subject can read its own membership grant, independently of workspace visibility
- the owner inserts a manager membership and waits for edge durability
- the manager deterministically receives that exact membership row at edge
- the manager's workspace query nevertheless remains empty for 15 seconds, without a query error

This isolates the problem from bootstrap timing, subject identity, and transport settlement: the policy witness is locally present, but the correlated read-policy/closure path does not publish the authorized parent row. The analogous production-shaped write-policy lowering is covered by #1868, but this read/closure behavior needs its own deterministic core receipt and semantics-preserving fix.

Do not weaken the app policy or replace the assertion with a trusted/admin read; the user-facing invariant is that arrival of an authorized witness row makes the referenced parent visible.

aeplay:
🤖 The first lower-boundary reduction narrows this further: a direct two-table Jazz/edge receipt on the combined core stack passes the intended lifecycle (self-readable witness admits exactly its workspace, cross-workspace remains denied, deletion retracts, reinsertion readmits). A planted policy-filter bypass is caught.

So the generic authorization-closure stage is not currently contradicted. The remaining lane is comparing BandBinder's emitted public TS policy and browser publication/subscription path against that passing hand-built Rust relation. In particular, the same app also sees `graph field not found: __root_join_row_0` in its headless public-policy path, which may explain the browser divergence before a separate closure fix is warranted.

aeplay:
🤖 The expanded BandBinder adopter receipt sharpens this to the public-policy/server-shell conversion path:

- The app benchmark/core receipts pass 2/2.
- Both app policy tests fail quickly and deterministically with `server tick after CommitUnit: database error: Storage: graph field not found: __root_join_row_0`.
- The real edge/browser topology then times out waiting for the manager’s authorized workspace, even though its membership row arrives.

[...]

PR #1888, opened 2026-08-24
fix(policy): retain shared-union occurrence carriers

aeplay:
🤖 Preserves UNION occurrence carriers when public-policy lowering uses a shared aliased source.

The lowerer previously checked `root_source()` before dispatching on the relation-input shape. A UNION whose arms share an aliased source therefore looked like an ordinary source join: lowering omitted its `(arm, row)` occurrence carriers even though terminal result-membership metadata correctly requested `__root_join_arm_0` / `__root_join_row_0`. That produced the adopter-visible `graph field not found` failure.

The fix dispatches `RelationInputPlan::Union` first. UNION inputs retain their occurrence carriers; ordinary non-UNION sources keep the existing `root_source()` path. No policy semantics or authorization rule changes.

Receipts cover a shared-alias UNION, an ordinary heterogeneous UNION, and an ordinary non-UNION source-column join. Restoring the old dispatch makes the shared-alias receipt fail for exactly the missing carriers. Fresh native artifacts pass BandBinder’s direct stage-manager permission suite (2/2). Independent adversarial review approved the one-commit/two-file patch.

This fixes the missing-carrier crash tracked in #1871. A distinct browser/edge delivery symptom remains under investigation on the same issue: the membership grant arrives and local policy evaluation returns the workspace, but the maintained cross-topology workspace query remains empty.

Stacked directly on #1884.

### DIFF
--- a/crates/jazz/src/node/query_engine/lowering/graph_lowering.rs
+++ b/crates/jazz/src/node/query_engine/lowering/graph_lowering.rs
@@ -1271,7 +1271,29 @@ fn lower_linear_plan_steps_cached(
                         }
                     }
                     if *mode == JoinMode::Inner && !omits_public_occurrence_carriers(request) {
-                        if let Some(right_source_id) = right.root_source() {
+                        // A union's root source is one of its arms, but its
+                        // public occurrence is the `(arm, row)` pair. Check
+                        // the relation shape before inspecting that source so
+                        // an aliased common arm cannot make us drop the union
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Superseded by #2046, reconstructed directly atop #2045 with the still-live UNION occurrence-carrier behavior, original commit attribution, explicit examples, and fresh planted-sensitivity receipts. The obsolete historical stack mechanics were intentionally omitted.

PR #1889, opened 2026-08-24
Make IndexedDB delta batches observe prospective state

joeinnes:
## Summary

- resolve each key's prospective value in caller order before committing an IndexedDB batch
- make later deltas observe preceding set, delete, and delta operations
- serialize mutation and flush boundaries across clones of the same IndexedDB storage instance
- emit one final write per affected key without cloning owned operation values

Previously, every delta read only durable tree state. Same-batch operations were ignored, and overlapping batches could overwrite a newer winner.

## Red/green commits

1. `1d3740579` adds same-batch and overlapping-batch regressions
2. `3d80b5135` implements prospective planning and a clone-shared mutation gate

## Verification

- `cargo test -p groove --lib storage::idb::tests` — 5 passed
- `cargo clippy --package groove -- -D warnings` — passed
- `cargo fmt --all` — passed

### DIFF
--- a/crates/groove/src/storage/idb.rs
+++ b/crates/groove/src/storage/idb.rs
@@ -1,9 +1,10 @@
 //! Async IDBTree adapter for Groove's ordered storage contract.
 
 use std::cell::RefCell;
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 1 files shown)


aeplay (approved):

PR #1890, opened 2026-08-24
Support runtime-comparable permission literals

joeinnes:
## Summary

- compile timestamp, finite double, byte-array, and recursively supported array policy literals into existing runtime values
- reject invalid or pre-epoch timestamps at the TypeScript authoring boundary
- reject non-finite doubles in raw, tagged, legacy-tagged, nested, and core-schema forms with precise paths

Previously, TypeScript accepted these policy values but core schema compilation rejected them later.

## Red/green commits

1. `13d809dab` adds the failing cross-layer regressions
2. `a5f6b6756` implements the consistent TypeScript/core contract

## Verification

- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts src/schema-permissions.test.ts` — 16 passed
- runtime-comparable Rust compiler regression — passed
- non-finite Rust literal rejection regression — passed
- `oxfmt`, `cargo fmt`, and `oxlint` — passed

### DIFF
--- a/.changeset/compile-policy-value-literals.md
+++ b/.changeset/compile-policy-value-literals.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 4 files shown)


aeplay (approved):

PR #1891, opened 2026-08-24
Complete client cleanup after shutdown failures

joeinnes:
## Summary

- run browser, web, and React Native cleanup steps sequentially even after an earlier synchronous throw or rejected promise
- preserve and rethrow the first failure after every remaining cleanup step has been attempted
- retain the existing browser lifecycle ordering: worker flush, client discard, telemetry/base teardown, then worker shutdown

Previously, a failed worker flush or subscription-manager shutdown skipped database, telemetry, client, and worker cleanup.

## Red/green commits

1. `c6788a301` adds failing browser/web/React Native regressions
2. `c75f818c0` implements ordered best-effort teardown

## Verification

- focused browser manager, web client, and React Native client suites — 11 passed

### DIFF
--- a/.changeset/ordered-shutdown-cleanup.md
+++ b/.changeset/ordered-shutdown-cleanup.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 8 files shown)


aeplay (approved):

PR #1892, opened 2026-08-24
Anchor sibling qualified policy joins to the protected row

joeinnes:
## Summary

- retain an explicit left scope on each internal relation join specification
- keep independently inferred qualified joins anchored to the protected row
- preserve chaining for explicit nested relation tails and recursive gather lowering

Previously, lowering advanced one mutable default scope after each join, so a second sibling relation could be joined from the first related table.

## Red/green commits

1. `a45b6c1c3` adds sibling and chained-join regressions
2. `dcd08a93b` preserves each join's selected source scope

## Verification

- `pnpm --dir packages/jazz-tools exec vitest run --config vitest.config.ts src/permissions/index.test.ts` — 47 passed
- `oxfmt` and `oxlint` — passed

### DIFF
--- a/.changeset/anchor-qualified-permission-joins.md
+++ b/.changeset/anchor-qualified-permission-joins.md
@@ -1,0 +1,7 @@
+---
+"jazz-tools": patch
+---
+
+Anchor qualified permission joins to their source relation scope
+
+Sibling qualified filters on directly related tables now join each related table from the protected row, while explicitly chained relation joins continue from the preceding joined table.
--- a/packages/jazz-tools/src/permissions/index.test.ts
+++ b/packages/jazz-tools/src/permissions/index.test.ts
@@ -1136,6 +1136,76 @@ describe("permissions DSL", () => {
     });
   });
 
+  it("keeps chained relation joins anchored to the preceding joined table", () => {
+    let relation: PermissionRelation | undefined;
+    definePermissions(app, ({ policy }) => {
+      relation = policy.teams
+        .where({})
+        .join(policy.user_team_edges, { left: "id", right: "team" })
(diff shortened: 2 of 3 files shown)


aeplay (approved):

aeplay:
example from codex:

> Consider this permission:
> 
> ```ts
> policy.teams.allowRead.where({
>   "user_team_edges.user_id": session.userId,
>   "resource_access_edges.grant_role": "viewer",
> })
> ```
> 
> The intended meaning is:
> 
> ```text
> protected team
> ├── user_team_edges where team = protected team.id
> │   └── user_id = current user
> └── resource_access_edges where team = protected team.id
>     └── grant_role = "viewer"
> ```
> 
> Both qualified predicates are sibling relationships originating from the protected `teams` row.
> 
> Previously, lowering maintained one mutable “current scope.” After joining `user_team_edges`, it could incorrectly treat that joined table as the source of the next join:
> 
> ```text
> protected team
> └── user_team_edges
>     └── resource_access_edges   ← incorrectly joined from user_team_edges
> ```
> 
> That changes the permission’s meaning and can produce false authorization results depending on coincidentally matching columns.
> 
> #1892 records the source scope on every join. Therefore:
> 
> - Implicit sibling relations remain anchored independently to the protected row.
> - Explicit chains still advance normally:
> 
> ```ts
> policy.teams
>   .join(policy.user_team_edges, { left: "id", right: "team" })
>   .join(policy.resource_access_edges, { left: "team", right: "team" })
> ```
> 
> Here the second join intentionally starts from `user_team_edges`.
> 
> So this is a correctness fix to distinguish sibling joins from deliberately chained joins; it does not introduce another query path.

Issue #1886, opened 2026-08-24
Seal trusted backend attribution across the Rust crate boundary


aeplay:
🤖 Backend-attributed mutations need an explicit cross-crate trust boundary between `jazz` and `jazz-napi`.

The current implementation work uses a Db-local capability flag, but `jazz-napi` is a downstream Rust crate. Rust has no friend visibility, so a `#[doc(hidden)] pub` constructor is still callable by every external Rust integrator and is not actually sealed. The immediate semantics-preserving hardening is to expose only an `unsafe` constructor with a precise contract and have the credential-gated NAPI adapter call it in a narrow documented block. This accurately models arbitrary in-process Rust as trusted while making accidental policy bypass conspicuous.

A truly unnameable capability would require an architectural choice, for example:

- invert the dependency so binding support receives a private token from `jazz`;
- move the trusted backend adapter into the `jazz` crate;
- introduce a dedicated internal crate whose public surface is deliberately treated as the process trust boundary.

Before choosing one, define whether hostile/arbitrary in-process Rust is in scope. This is distinct from remote clients: arbitrary code in the server process already has storage and credential authority. The eventual design should keep normal safe `Db::open` APIs unable to select external row authorship and retain exact audit provenance.

Issue #1881, opened 2026-08-24
Backend attribution for exclusive transactions


aeplay:
🤖 Trusted backend attribution now has a clean representation for direct mutations, streaming finalization, and mergeable transactions: policy admission retains the backend/SYSTEM subject while provenance carries the external author separately.

Exclusive transactions do not yet have that split representation. Passing the attributed author as the transaction session would incorrectly evaluate permissions as that user; retaining only SYSTEM would silently lose requested provenance. Until designed, `withAttribution(...).transaction({ mode: "exclusive" })` should fail closed with an actionable error.

Design the exclusive open-transaction state so admission identity and provenance author remain separate through begin, every operation, retry/replay, commit, rollback, and schema views. Reuse the ordinary transaction machinery rather than create an attributed execution path. Receipts should combine a denied external author with trusted backend success and exact row/tombstone `$createdBy`, plus disconnect/retry and rollback cases.

Split from #1879; mergeable transactions do not need this new representation.

PR #1896, opened 2026-08-24
fix(auth): separate backend admission from row attribution

### LINKED ISSUE #1879: Backend attribution across mutation APIs
🤖 `createJazzContext(...).asBackend().withAttribution(author)` promises to evaluate permissions with trusted backend authority while stamping the external author into row provenance. The current native adapter cannot uphold that contract consistently.

Today:

- Core has attributed helpers only for insert, insert-with-id, update, and delete.
- NAPI exposes none of those attributed helpers.
- Upsert, restore/branch restore, streaming large-value mutations, and transactions have no complete attributed path.
- Treating the attributed author as the operation session makes policy evaluation run as that external user, while retaining SYSTEM admission silently records SYSTEM provenance instead.

Complete the existing contract without adding a second mutation implementation: thread admission identity and provenance author as separate parameters through shared mutation internals, then expose credential-gated native operations used only by trusted backend contexts. Cover every mutation family supported by `withAttribution`, schema views, retries, streaming finalization, and transaction operations. Public raw-author methods must continue rejecting Jazz-reserved issuers.

Receipts should combine a policy the attributed external author cannot satisfy with exact `$createdBy`/tombstone provenance, proving authorization remains backend-scoped while provenance is external. This was discovered during the #1750 capability re-review.

aeplay:
🤖 Lets a credentialed trusted backend perform SYSTEM-admitted mutations while recording the exact external issuer-scoped author as row provenance.

This separates two concepts that were previously conflated:

- admission authority: the trusted backend’s SYSTEM capability;
- audit attribution: the external canonical author written to `$createdBy` / `$updatedBy`.

The capability is Db-local and is created only through explicit trusted-backend opens. Ordinary databases—including ordinary SYSTEM databases—cannot forge external attribution. Because `jazz-napi` is a downstream Rust crate, the two core constructors are `pub unsafe` with a precise trusted-process safety contract; the credential-gated NAPI adapter invokes them in narrow documented blocks. The NAPI credential is an in-process opt-in, not cryptographic authentication. A truly sealed cross-crate capability remains tracked in #1886.

Coverage includes direct insert/update/upsert/delete/restore, streaming finalization, registered schema views, and mergeable transactions. Streaming cannot swap the admitted author, transactions reject Alice/Bob and attributed/unattributed mixing, and history-complete opens preserve the authority. Branch and exclusive-transaction attribution currently fail closed instead of silently dropping provenance; exclusive semantics remain explicitly tracked in #1881.

Independent adversarial review planted safe-constructor calls, credential bypass, provenance swaps, view capability loss, transaction mixing, streaming author overrides, and silent branch drops; committed real Rust/NAPI receipts catch them. The final clean integration excludes BandChat files and preserves #1873’s sealed artifact/fingerprint architecture. Focused real NAPI, Rust anti-forgery/streaming, TypeScript, cargo, formatting, and diff gates pass.

Closes #1879. Stacked directly on #1750.


aeplay:
🤖 Superseded by #1971, which reconstructs the agreed backend-admission/row-attribution design on top of merged #1750 without carrying this branch's obsolete stacked history. The replacement preserves the relevant intent, adds fail-closed core/NAPI/TypeScript boundaries, and has completed two adversarial review iterations.

Issue #1763, opened 2026-08-23
Reject cyclic or unbounded policy graphs deterministically


aeplay:
🤖 Make recursive `INHERITS` and reverse-foreign-key authorization graphs terminate under an explicit, testable contract.

## Contract

Cycles are rejected when they make policy evaluation unsafe or ambiguous; legitimate shared subgraphs are allowed and deduplicated rather than mistaken for cycles. Traversal has deterministic depth/work bounds and returns a typed rejection instead of hanging, overflowing a queue, or silently granting access.

## Examples

- A → B → A is a cycle and must fail predictably.
- A → C and B → C is a shared subgraph, not a cycle; C may legitimately grant both paths.
- A very deep acyclic chain must hit the documented bound with an actionable error.

## Pointers

Policy graph construction/evaluation is under `crates/jazz/src/node/policy.rs` and `crates/jazz/src/node/tests/policies_rls/recursive.rs`. Deployed cases live in `crates/jazz-testkit/tests/policies_integration/recursive_inheritance.rs` and `inheritance_validation.rs`. Include planted cycle/shared-subgraph receipts.

PR #1897, opened 2026-08-24
fix(groove): compile recursive policy graphs iteratively

### LINKED ISSUE #1885: Recursive inheritance policy test overflows the server thread stack
🤖 The full `policies_integration` target aborts in `recursive_inheritance::rebac_recursive_inherits_allows_ancestor_access` because the server thread overflows its stack.

Observed while independently reviewing #1884 at `51c686a32`. The exact test fails identically after temporarily restoring the base (`4b0a11eb4`) public-schema converter, so this is not caused by the correlated `SELECT` / `INHERITS` repair.

Reproduction:

```sh
cargo test -p jazz --test policies_integration recursive_inheritance::rebac_recursive_inherits_allows_ancestor_access -- --exact --nocapture
```

Expected: recursive inheritance settles or returns a bounded diagnostic.

Actual: the server thread reaches a stack overflow and aborts the test process.

Investigation should first identify whether recursion is cycling in policy lowering, graph traversal, or runtime evaluation, then add a planted-sensitive bounded regression. Do not merely increase the thread stack: an unbounded recursive path would remain a production correctness/availability risk.

aeplay:
🤖 Compiles deep recursive policy graphs iteratively instead of consuming the server thread’s call stack.

The failing recursive-`INHERITS` graph is semantically valid, but public/routing projections make it roughly 91 graph nodes deep. Groove recursively walked that owned graph during descriptor inference, terminal validation, nested-recursion detection, and compilation; the server shell’s ordinary thread stack overflowed before evaluation. Increasing the stack proved the policy result itself was correct, but would only hide the unbounded implementation.

`GraphBuilder::postorder()` now produces one child-first traversal used by all four phases. Compilation caches child fragments before their parents, retaining declared UNION/JOIN/Recursive ordering, node deduplication, descriptor errors, and nested-recursion rejection without recursive fallback. No inheritance, cycle, or authorization semantics change.

The previously ignored single-table cyclic-`INHERITS` fail-closed integration receipt is active and passes. Independent adversarial review restored recursive inference, parent-first traversal, and missing compiled-memo reuse; each mutation reproduced the exact stack overflow. Full Groove passes (481, 2 documented ignored), full policy integration passes (57, 48 documented ignored), and focused deep recursion, nested rejection, duplicate derivation/node reuse, format, and diff gates are green.

A separate mutual-table recursion overflow occurs earlier in Jazz public-schema conversion and remains tracked under #1763; this PR does not conflate that path.

Closes #1885. Stacked on #1896.

### DIFF
--- a/crates/groove/src/ivm/graph.rs
+++ b/crates/groove/src/ivm/graph.rs
@@ -520,6 +520,53 @@ impl GraphBuilder {
         }
     }
 
+    /// Return all builder fragments in child-before-parent order without
+    /// recursion. Runtime setup uses this for deeply nested valid query and
+    /// policy graphs, which must not consume the owner thread's call stack.
+    pub(crate) fn postorder(&self) -> Vec<&Self> {
+        let mut pending = vec![(self, false)];
+        let mut ordered = Vec::new();
+        while let Some((graph, visited)) = pending.pop() {
+            if visited {
+                ordered.push(graph);
+                continue;
+            }
+            pending.push((graph, true));
+            match graph {
+                Self::Filter { input, .. }
+                | Self::Project { input, .. }
+                | Self::StreamingChecksum { input, .. }
+                | Self::UnwrapNullable { input, .. }
+                | Self::Unnest { input, .. }
+                | Self::VariantProject { input, .. }
+                | Self::ArgMaxBy { input, .. }
+                | Self::ArgMinBy { input, .. }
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Superseded by #1976, which reconstructs the reviewed stack-safety work directly on current `main`, includes the residual recursive-source traversal fix found during adversarial review, and removes the obsolete #1896 dependency.

PR #1902, opened 2026-08-24
test(topology): validate browser helper commands

aeplay:
🤖 Makes shared browser topology commands explicit, typechecked, and runtime-validated.

## Why

Vitest 4 browser command types can resolve through different pnpm peer instances. Ambient augmentation of `vitest/browser` or `vitest/internal/browser` therefore let app code compile while describing commands that its browser project never registered; the failure surfaced only as `commands.<name> is not a function` at runtime.

## Change

- Replace ambient command augmentation in the shared helpers with explicit `JazzServerBrowserCommands` and `JazzTopologyBrowserCommands` adapters.
- Validate the complete configured command surface at runtime and report a direct browser-project configuration error when a handler is absent.
- Add a focused browser-helper TypeScript project so both helper entrypoints are actually compiled by CI/local gates.
- Keep app-specific Vitest projects responsible for registering the handlers they use.

## Verification

- `pnpm --dir packages/jazz-tools run typecheck:browser-helpers`
- topology harness: 15 passed
- Independent adversarial review planted a missing handler and observed the intended configuration diagnostic.
- A real Jamazon browser project exposed and then repaired its missing network handlers through this contract, proving the check is not declaration-only.

This is a test/tooling contract change only; Jazz runtime semantics are unchanged.

### DIFF
--- a/packages/jazz-tools/package.json
+++ b/packages/jazz-tools/package.json
@@ -135,6 +135,7 @@
     "typecheck:react-native": "tsc --project tsconfig.react-native-tests.json",
     "test:browser": "node ../../dev/artifacts/verify-correctness-test-artifacts.mjs && vitest run --config vitest.config.browser.ts",
     "test:browser:focused": "node scripts/test-browser-focused.mjs",
+    "typecheck:browser-helpers": "tsc --project tests/browser/tsconfig.browser-helpers.json",
     "test:browser:contract": "node scripts/test-browser-focused.test.mjs",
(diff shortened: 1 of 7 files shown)


aeplay:
🤖 Superseded by #2048, reconstructed atop #2046 with the still-live typed browser-command contract, original attribution, explicit examples, and a planted runtime proxy receipt. Unrelated historical topology-harness mechanics were intentionally omitted.

PR #1903, opened 2026-08-24
fix(policy): retain complete scans for disjunctive proofs

aeplay:
🤖 Prevents indexed policy proof evaluation from dropping rows authorized by an alternative relational branch.

## Before

Access-path derivation collected equality constraints from an entire normalized authorization program and applied them as if every proof route required them. For a policy such as:

```text
owner_id = session.user_id
OR EXISTS members(chat_id = row.id AND user_id = session.user_id)
```

adding indexes on `owner_id` and membership columns could restrict the protected-table scan to the owner branch. A non-owner member then received its membership row but never the authorized parent row. The identical schema and data passed when indexes were absent.

## After

- Proof programs containing `Union`, `Join`, or `RecursiveRelation` do not advertise a derived source access path unless the planner can prove arm/source coverage.
- Those authorization proofs take the existing complete-scan path and continue through the same lowered evaluator.
- Pure conjunctive/scalar proofs retain their selective indexed access paths.
- Ordinary application query planning, ordering, and bounds are unchanged.

This is deliberately conservative: some relational policies may scan more than necessary, but indexes can no longer change authorization results. A future planner may recover those optimizations by deriving covered access paths per source/union arm.

## Receipts

- Paired indexed and unindexed real-DB membership-grant receipts produce the same authorized parent.
- Edge rehydration delivers the parent after the membership grant arrives.
- External-JWT browser coverage reproduces the same indexed shape.
- A probe metric proves the unsafe disjunctive proof takes zero index probes, while a simple conjunctive policy still indexes.
- Removing the bailout makes the indexed semantic receipt fail with an empty result.

## Verification

- `cargo test -p jazz --lib node::tests::harness:: -- --nocapture`: **450 passed, 0 failed, 2 ignored**
- Focused indexed/unindexed receipts: **2 passed**
- Independent adversarial review approved the conservative boundary and planted the no-bail regression.
- Rust formatting, clippy, sensitive-data guard, and diff checks pass.

The browser receipt will receive its final fresh-artifact run on the combined stack; the equivalent core failure and repair are deterministic.

### DIFF
--- a/crates/jazz/src/db/tests/node_runtime.rs
+++ b/crates/jazz/src/db/tests/node_runtime.rs
@@ -909,8 +909,7 @@ fn db_sync_surface_edge_session_read_policy_filters_private_table_query() {
 /// resource that grant authorizes.  The second subscription must publish a
 /// result membership even when the first subscription already delivered the
 /// resource as policy support.
-#[test]
-fn db_sync_surface_membership_grant_then_parent_query_keeps_disjunctive_read_proof() {
+fn membership_grant_then_parent_query_keeps_disjunctive_read_proof(indexed: bool) {
     let member_exists = public_exists(
         "members",
         [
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Superseded by #2049, reconstructed atop #2048 with the still-live soundness rule for disjunctive/relational authorization access paths, original attribution, indexed/unindexed examples, and planted probe sensitivity. Obsolete historical browser and stack mechanics were intentionally omitted.

PR #1909, opened 2026-08-24
fix: make React client acquisition lifecycle-safe

joeinnes:
## Summary

- keep `JazzProvider` from acquiring a shared client during server rendering
- serialize different-config client lifecycles so replacement waits for the previous release
- prevent a canceled replacement from acquiring after unmount
- preserve same-config Strict Mode reacquisition and release pending clients correctly

## Validation

- focused React provider suites: 2 files, 5 tests passed
- focused browser lifecycle scenario: 1 passed, 17 skipped
- Oxfmt and Oxlint

### DIFF
--- a/.changeset/react-provider-ssr-lifecycle.md
+++ b/.changeset/react-provider-ssr-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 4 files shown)


aeplay (approved):

PR #1910, opened 2026-08-24
fix: authenticate todo REST requests per session

joeinnes:
## Summary

- create a request-scoped Jazz database for every REST and event-stream request
- derive todo ownership from the authenticated session rather than request input
- route all network reads and writes through normal permission evaluation
- retain the exported administrative database handle for trusted setup callers without using it in routes

## Validation

- REST authentication suite: 14 tests passed
- todo integration suite: 11 tests passed
- Oxfmt and Oxlint


aeplay (approved):

PR #1911, opened 2026-08-24
fix(chat): bind message authorship to authenticated users

joeinnes:
## Summary

- require a message's sender profile to belong to the authenticated session
- prevent writers from attributing messages to another user's profile
- leave existing message reads, UI behavior, and already-secure reaction ownership unchanged

## Validation

- chat permission suite: 7 tests passed
- regression reproduces forged authorship before the policy change
- Oxfmt and Oxlint


aeplay (approved):

PR #1912, opened 2026-08-24
fix: stabilize Expo secret loading across remounts

joeinnes:
## Summary

- share one in-flight secure-secret load across concurrent mounts and Strict Mode remounts
- retain a successful secret while allowing failed secret loads to retry
- distinguish secret-store failures with `SecretLoadError`
- rethrow unrelated provider and descendant errors instead of mislabeling them as secret failures

## Validation

- focused secret lifecycle suite: 3 tests passed
- Oxfmt and Oxlint

Device-level Expo execution is not part of the current workspace; the promise and error-routing contract is covered by the focused Node suite.


aeplay (approved):

PR #1920, opened 2026-08-24
Bind persistence receipts to their database instance

joeinnes:
## Summary
- carry the originating database's persistence-order identity in each receipt
- reject foreign receipts before changing metrics, publication state, or the durable frontier
- cover same-numbered resident publications in two independent databases

## Why
Publication IDs are database-local. A receipt for publication 1 in one database could previously settle publication 1 in another database, removing the wrong resident write and advancing the wrong durable frontier.

## Verification
- `cargo test -p groove db::tests::batches` (31 passed)
- `cargo clippy -p groove --lib -- -D warnings`
- pre-commit hooks passed


aeplay (approved):

PR #1921, opened 2026-08-24
Backpressure native WebSocket outbound frames

joeinnes:
## Summary
- enforce an 8 MiB byte budget across frames queued for the native WebSocket pump
- return `TransportError::Backpressure` synchronously when the budget is exhausted
- release reservations only after queued frames are dropped and wake the owner after pump progress
- serialize bounded frame wrappers without copying them into a second frame vector

## Why
The native transport used an unbounded outbound channel. A slow or stalled socket could let synchronous producers retain an unlimited number of full-size wire frames.

## Verification
- `cargo test -p jazz-native-transport` (18 passed)
- `cargo clippy -p jazz-native-transport --all-targets -- -D warnings`
- pre-commit hooks passed

### DIFF
--- a/crates/jazz-native-transport/src/lib.rs
+++ b/crates/jazz-native-transport/src/lib.rs
@@ -777,4 +777,33 @@ mod tests {
                 .contains("plaintext ws:// bootstrap")
         );
     }
+
+    #[tokio::test]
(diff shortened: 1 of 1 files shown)


aeplay (approved):

PR #1923, opened 2026-08-24
Block index registration while publications are resident

joeinnes:
## Summary
- reject new live index registration while any publication remains resident
- preserve idempotent registration and invalid-field error precedence
- verify registration succeeds after the resident publication becomes durable and backfills the row

## Why
Live index backfill reads durable storage. Registering an index while applied writes are resident could publish an index definition whose initial contents omit those writes.

## Verification
- `cargo test -p groove db::tests::indices` (25 passed)
- `cargo clippy -p groove --lib -- -D warnings`
- pre-commit hooks passed


aeplay (approved):

PR #1924, opened 2026-08-24
Bound loopback HTTP request bodies

joeinnes:
## Summary
- cap loopback HTTP request bodies at 8 MiB before reading the advertised body
- use checked body-offset and body-length arithmetic
- reject both a one-byte-over-limit request and `usize::MAX` without waiting for body bytes

## Why
The loopback parser trusted `Content-Length` and kept extending its request buffer until that many bytes arrived. A local client could force unbounded memory growth, and unchecked offset arithmetic could overflow.

## Verification
- `cargo test -p jazz-server loopback::http::tests` (5 passed)
- `cargo clippy -p jazz-server --lib -- -D warnings`
- pre-commit hooks passed

### DIFF
--- a/crates/jazz-server/src/loopback/http.rs
+++ b/crates/jazz-server/src/loopback/http.rs
@@ -1006,6 +1006,31 @@ mod tests {
         assert_eq!(shell.runtime_write_schema_revision(), 0);
     }
 
+    #[test]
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Superseded by #1951, which preserves the request-body bound on the real Axum serving path and removes the historical custom loopback HTTP parser.

PR #1927, opened 2026-08-24
Compile effective DELETE policy fallback

joeinnes:
## Summary
- compile UPDATE USING as DELETE USING when no explicit delete predicate is declared
- preserve explicit DELETE USING precedence and accurate conversion error paths
- cover both direct delete policy output and inherited DELETE authorization metadata

## Why
The public schema contract documents DELETE USING as falling back to UPDATE USING. Conversion previously omitted that fallback even though inherited-policy admission already reasoned from the effective predicate, denying legitimate deletes and leaving inherited DELETE filters without a compiled parent predicate.

## Verification
- public schema conversion tests: 40 passed
- focused direct and inherited fallback regression passed
- `cargo clippy -p jazz --lib -- -D warnings`
- pre-commit hooks passed

### DIFF
--- a/crates/jazz/src/tools/public_schema_convert.rs
+++ b/crates/jazz/src/tools/public_schema_convert.rs
@@ -548,6 +548,24 @@ fn convert_table(
         "policies.select.using",
         table.policies.select.using.as_ref(),
     )?;
+    let update_using = convert_optional_policy(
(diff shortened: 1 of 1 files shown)


aeplay (approved):

PR #1931, opened 2026-08-24
Clear Inspector mutations on table navigation

joeinnes:
## Summary
- clear queued edits, inserts, deletes, save errors, and pending row scroll state when the routed table changes
- retain staged state across sort, filter, pagination, and column-layout changes within the same table
- cover navigation with a staged insert and prove no database mutation is issued

## Why
React Router reuses the data-grid component across table routes. Staged changes from table A could survive into table B and then be applied through table B's proxy and schema.

## Verification
- `TableDataGrid.test.tsx` (39 passed)
- Inspector TypeScript build passed
- changed files pass oxfmt and oxlint
- pre-commit hooks passed


aeplay (approved):

Issue #1933, opened 2026-08-24
Worker restart drops persisted mutation rejections


aeplay:
🤖 Canonical author identity exposed a missing rejection-routing boundary in the persistent browser worker.

## Exact regression

On the direct pre-stack base `b0d6303b`, all four worker-restart rejection receipts pass. On the combined candidate they time out while rollback still succeeds; the rest of the browser suite is unaffected:

- an authority rejection received after the persistent worker restarts;
- rejected insert after restart;
- rejected update after restart;
- rejected delete after restart.

Removing proof-backed canonical identity happens to restore one callback but breaks seeded update/delete attribution, so identity is exposing the seam rather than being reverted.

## Required semantics

Durable state correction and foreground notification are deliberately separate:

1. The persistent worker must durably remove a rejected optimistic transaction from ordinary visible state.
2. When the worker processes the rejection, it forwards the disposition to every foreground runtime currently attached to that auth scope.
3. An active `.wait()` consumes the rejection for its runtime and suppresses that runtime's `onMutationError` fallback.
4. A transaction may have been authored by an earlier foreground runtime. If a successor runtime is attached when a reopened worker later receives its fate, that attached runtime still observes the rejection.
5. If no foreground runtime is attached when the worker processes the fate, Jazz does not retain a durable notification for a future app launch. A later runtime rehydrates corrected state without receiving a historical callback.
6. INV-TX-9 remains strict: the worker must not retain foreign rejected row payload outside `tx_id.node`.

This intentionally does **not** provide a durable user-notification inbox. For example, an app can commit offline, close completely, and remain closed while its worker later reconnects and learns that the commit was rejected. On the next launch the optimistic data is absent, but Jazz does not synthesize a delayed error toast. Apps needing that continuity should model a durable operation/activity row and query its eventual outcome.

## Required receipts

- restarted worker receives rejection while a successor foreground runtime is attached: rollback occurs and that runtime observes exactly one rejection;
- rejected insert, update, and delete all follow that path;
- active `.wait()` rejects and suppresses fallback callback;
- attached runtimes receive the rejection without requiring author equality with the original tab node;
- no foreign rejected retry payload is retained;
- a runtime attaching only after rejection processing receives corrected state but no historical callback;
- planted removal of current-runtime routing or rollback makes the corresponding receipt fail.

The topology contract and deliberately unsupported app-less interval are documented in `crates/jazz/SPEC/9_topology_edge.md`.

PR #1934, opened 2026-08-24
fix(napi): export generated JsonValue type

### LINKED ISSUE #1932: jazz-napi declaration references undeclared JsonValue
🤖 `crates/jazz-napi/index.d.ts` refers to `JsonValue` throughout the generated public declaration but does not declare or import that type. Consumers which typecheck the declaration directly fail with `Cannot find name 'JsonValue'`; PosterShop's app typecheck reproduces this on both its old and final-core heads.

The native artifact build only checks that generated declarations byte-match the checked-in file, so it currently preserves this invalid surface rather than detecting it. Fix the NAPI-RS declaration source/generation boundary so `JsonValue` is emitted as a self-contained public type, and add a consumer-style TypeScript receipt which imports/typechecks the built package without relying on an ambient declaration.

aeplay:
🤖 Fixes #1932.

## Before

The generated `jazz-napi` declaration used `JsonValue` throughout its public methods without declaring or importing that name. Direct package consumers therefore failed typechecking, even though the checked-in declaration still byte-matched fresh generation.

## After

- napi-rs owns and emits the public `JsonValue` alias from `#[napi] pub type JsonValue = serde_json::Value`.
- The generated package declaration is self-contained. At this raw FFI boundary napi-rs represents `serde_json::Value` as `any`; higher Jazz APIs retain their typed JSON surfaces.
- A package-name consumer receipt imports both `NapiDb` and `JsonValue`, and normal Turbo `test` selection runs it.
- `jazz-napi` explicitly owns the Node declarations needed by its generated `Buffer` references, rather than relying on root hoisting.

## Verification

- fresh NAPI generation byte-matches the checked-in declaration;
- isolated package consumer TypeScript passes after a frozen install;
- Turbo executes `jazz-napi:test` successfully;
- `cargo test -p jazz-napi --lib`: 15/15;
- NAPI artifact contract: 7/7;
- planted removal of either the Rust export or checked alias makes the relevant generation/consumer receipt fail.

This is stacked on #1903 and changes no runtime behavior.

### DIFF
--- a/crates/jazz-napi/index.d.ts
+++ b/crates/jazz-napi/index.d.ts
@@ -178,6 +178,17 @@ export declare class Write {
   close(): boolean
 }
 
+/**
+ * Any JSON-compatible value crossing the native JavaScript boundary.
+ *
+ * Keep this alias exported through napi-rs rather than relying on its Rust
+ * import name: exported methods use `JsonValue` throughout their generated
+ * declarations, so the package must define that name for TypeScript
+ * consumers.
+ */
+export type JsonValue =
+  any
+
 export declare function mintLocalFirstToken(seedB64: string, audience: string, ttlSeconds: number): string
 
 /** Exact build/ABI fingerprint for the generated native artifact. */
--- a/crates/jazz-napi/package.json
+++ b/crates/jazz-napi/package.json
@@ -41,7 +41,8 @@
     "typecheck:consumer": "tsc --project tests/types/tsconfig.json"
   },
   "devDependencies": {
-    "@napi-rs/cli": "^3"
+    "@napi-rs/cli": "^3",
+    "@types/node": "catalog:default"
   },
   "napi": {
     "binaryName": "jazz-napi",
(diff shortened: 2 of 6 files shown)


aeplay:
🤖 CI confirms this PR's own declaration boundaries are green: lint, Rust workspace, Rust differential, and the package consumer path all pass. The aggregate TypeScript job is red only on two independently isolated regressions inherited from the canonical-author stack below:

- #1933: four persisted mutation-rejection callbacks disappear after worker restart, although rollback succeeds (259 other browser tests pass);
- #1938: Todo Server's trusted backend open is rejected as a forged reserved system issuer, causing 9/11 integration failures.

Both direct pre-stack controls are green and both repairs are active on separate lanes. This PR remains draft until those fixes are stacked above it and the combined gate is rerun.

aeplay:
🤖 Superseded by #2050, reconstructed atop #2049 with the still-live generated `JsonValue` export, original intent, a current branch-API consumer example, and planted declaration sensitivity. The obsolete historical branch-specific insert call was intentionally omitted.

PR #1937, opened 2026-08-24
fix: generate row UUIDv7 IDs in Rust

nrainhart:
## Description

Rows created through the TS client without an explicit id were using a non-uuidv7 id generated on the TS side. This meant that rows were not sorted by creation time when using the default query ordering (id asc).

This PR also fixes the proliferation of CRUD methods across the TS<>Rust boundary, instead keeping one insert, update, delete and restore method with an options param.

### DIFF
--- a/crates/jazz-cli/tests/cli_dry_run.rs
+++ b/crates/jazz-cli/tests/cli_dry_run.rs
@@ -806,10 +806,13 @@ fn server_command_loads_published_schema_and_persists_ws_data_across_restart() {
         subject,
         identity_for_subject(0xc1, subject),
     );
-    let write = block_on(writer.db.insert_with_id(
+    let write = block_on(writer.db.insert(
         "todos",
-        RowUuid::from_bytes([0x41; 16]),
         todo_cells("durable cli row", true),
+        jazz::db::InsertOptions {
+            row_id: Some(RowUuid::from_bytes([0x41; 16])),
+            ..Default::default()
+        },
     ))
     .expect("write todo through client db");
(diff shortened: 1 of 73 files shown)


aeplay (changes requested):
🤖 Changes requested: the UUIDv7 direction looks sound, but the CRUD consolidation currently introduces two correctness regressions.

1. **Exclusive operations discard `updatedAtMs`.** The unified NAPI/WASM surface accepts and forwards the option, but exclusive insert, upsert, delete, and restore reach staging helpers that unconditionally call `next_now_ms()`. Exclusive update already threads the supplied timestamp correctly. Please carry the option through every exclusive staging path so the consolidated API does not silently change caller-controlled ordering/provenance.

2. **Exclusive upsert does not preserve physical large-value cells.** It still merges through the public/materialized read path and then stages an insert. Updating an unrelated scalar column can therefore rematerialize an unchanged large value under a new locator. The adjacent exclusive-update fix correctly merges against `tx_read_in_schema` physical cells; upsert needs the same treatment. A planted replacement of update with upsert in `exclusive_transactions_lower_oversized_scalars_before_publication` failed its stable-locator assertion, then was restored.

Please add direct receipts for the affected timestamp operations and for exclusive upsert preserving an unchanged large-value locator.

aeplay (approved):
🤖 Approved by Anselm after re-review at `2fa3140fa9a96bec96179e482a4e75a029a8b1db`. Both requested correctness fixes are present: every exclusive CRUD path preserves explicit `updatedAtMs`, and exclusive upsert preserves unchanged physical large-value cells. Focused receipts and planted negatives passed with no new findings.

PR #1939, opened 2026-08-24
test(topology): guarantee compensating fault cleanup

### LINKED ISSUE #1936: Topology faults need guaranteed compensating cleanup
🤖 `runTopologyScenario` skips later faults after a phase failure. Several example scenarios model disconnect/reconnect by blocking Playwright HTTP/WebSocket routes during one fault and relying on a later fault to unblock them. If an assertion fails between those faults, the browser context remains blocked and can poison subsequent tests with misleading cascades.

BandChat, BigLabel, PosterShop, and WorldTour each had to add local blocked-state tracking and best-effort cleanup. Move this guarantee into the shared topology/fault abstraction: a fault that acquires external state should register an idempotent compensating cleanup which the runner executes from `finally`, preserving the original phase error while reporting cleanup diagnostics separately. Add a harness receipt that deliberately fails between block and unblock and proves the next scenario is unaffected.

aeplay:
🤖 Fixes #1936.

## Before

Topology scenarios acquired external fault state (for example, blocked browser routes) in one fault and relied on a later fault to release it. A failed intermediate phase skipped that release and could poison later tests. Each example had begun hand-rolling its own recovery flags and `afterEach` fallbacks.

## After

- Every phase and fault can register a named, idempotent compensation with `defer(...)` while that operation is active.
- The runner closes each operation-local registrar when the operation settles or times out, rejecting stale asynchronous registration.
- Remaining inverses run in LIFO order from `finally`, before scenario cleanup.
- A timed-out inverse is aborted diagnostically and then drained before earlier cleanup or the next scenario starts; the runner never claims isolation while abandoned cleanup can still mutate external state.
- Cleanup failures receive deterministic receipt rows, do not skip other inverses, and preserve the original scenario error as the cause.
- Scenario cleanup uses the same abort-and-drain ownership rule and receives a cleanup-only context, so it cannot acquire new obligations.

An inverse that ignores abort forever intentionally keeps the runner pending: bounded return and guaranteed absence of post-return mutation cannot both be promised.

## Verification

- focused topology typecheck and fixture suite: 21/21;
- planted stale phase/fault registrar regressions fail their receipts;
- planted compensation-execution removal fails four receipts;
- replacing either compensation or scenario cleanup drain with an ordinary timeout fails exact late-mutation isolation receipts;
- repeated clean runs and full Jazz Tools TypeScript check pass.

This is a test-harness robustness change stacked on #1934; it does not change Jazz runtime semantics.


aeplay:
🤖 Superseded by #2053, reconstructed atop #2050 with the still-live operation-scoped compensation and cleanup-drain guarantees, original attribution, explicit normal/failure/timeout examples, and planted registrar sensitivity. Obsolete historical topology-scheduler mechanics were intentionally omitted.

PR #1940, opened 2026-08-24
fix(backend): separate trusted host authority from sync credentials

### LINKED ISSUE #1938: Todo Server backend open is rejected as reserved system issuer
🤖 The canonical-author stack makes the Todo Server TypeScript integration suite fail 9/11 because ordinary CRUD/bootstrap requests return HTTP 500 with `author issuer is reserved: urn:jazz:system`.

Exact evidence on PR #1934 / run 32739947919:

- create/list/get/update/delete and 404 handling all return 500;
- the session-context owner filter and cold-restart receipt also return 500;
- SSE consequently times out;
- only two unrelated tests pass.

The direct pre-canonical-author base #1856 is green. The trusted backend capability must remain distinct from public/client attempts to forge a reserved author, while the example's supported backend-open path must construct or prove its system identity through the intended trusted API. Fix the actual consumer or binding boundary and retain the reserved-issuer rejection; do not weaken validation globally.

aeplay:
🤖 Fixes #1938.

## Decision: backend capability is not a credential

`isBackend` identifies an application-controlled trusted runtime that may perform backend/SYSTEM operations and evaluate request-scoped identities. It does **not** mean that the runtime is connected upstream, possesses credentials, or acts as an edge/core fate authority.

`backendSecret` has one separate purpose: authenticating that backend when it connects to an upstream server. Possessing a credential must never confer local backend capability.

The distinction is observable for a disconnected backend: it still needs local SYSTEM/request-scoped behavior even though it has no upstream and therefore no `backendSecret`. Connected backends normally have both properties, but they remain independent security boundaries.

## Before

`createJazzContext().asBackend()` correctly selected Jazz's private SYSTEM session, but native runtime construction accidentally used presence of `backendSecret` as its trusted-runtime selector. A local-only backend has no upstream secret, so SYSTEM was serialized through public identity ingress and rejected as a forged reserved issuer. Todo Server consequently returned 500 for ordinary CRUD and passed only 2/11 integration tests.

## After

- `createJazzContext` internally constructs its runtime with `isBackend: true`; this is not exposed through public client configuration.
- `NativeRuntimeAdapter` never accepts an upstream credential or infers local capability from one. `isBackend` selects only the private `jazz:backend-runtime` native marker.
- Schema views inherit their owner's backend status and cannot elevate an untrusted runtime.
- The real `backendSecret` flows only into JazzClient transport authentication.
- Empty or whitespace-only upstream secrets fail configuration rather than producing divergent TS/native behavior.
- Public reserved-issuer parsing and rejection are unchanged.

## Verification

- Todo Server integration: 11/11.
- A disconnected secretless persistent backend writes with exact `$createdBy` provenance `["urn:jazz:system","system"]`.
- A public session forging that reserved issuer is rejected.
- Backend unit receipts prove native construction receives `isBackend` but no transport credential, while JazzClient receives the real secret.
- Native schema-view attribution and reserved-issuer suites pass.
- Focused backend context and native adapter suite: 122 passed / 1 skipped.
- Real NAPI core suite: 36 passed / 1 skipped.
- Formatting, lint, diff, and sensitive-data gates pass.

This is stacked on #1939 and preserves the existing backend-versus-public-session semantics.

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -3464,9 +3464,10 @@ where
     if config.history_complete {
         let db = if trusted_backend {
             // SAFETY: `trusted_backend` is true only after this NAPI open
-            // configuration carried a non-empty backend credential. This is
-            // the binding's intentionally narrow trusted in-process boundary;
-            // it is not inferred from SYSTEM or a caller-supplied author.
+            // configuration carried the binding's private backend-runtime
+            // marker. This is the intentionally narrow trusted in-process
+            // boundary; it is not inferred from an upstream credential,
+            // SYSTEM identity, or a caller-supplied author.
             unsafe {
                 core_block_on(CoreDb::open_history_complete_with_backend_attribution(
                     db_config,
@@ -3481,8 +3482,8 @@ where
         Ok(db)
     } else {
         let db = if trusted_backend {
-            // SAFETY: see the history-complete branch above. The credential
-            // gate is checked before this unsafe core opt-in.
+            // SAFETY: see the history-complete branch above. The private
+            // backend-runtime marker is checked before this unsafe core opt-in.
             unsafe { core_block_on(CoreDb::open_with_backend_attribution(db_config)) }
         } else {
             core_block_on(CoreDb::open(db_config))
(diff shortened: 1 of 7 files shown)


aeplay:
🤖 Superseded by merged #1971 and its follow-up backend admission/attribution work. Current `main` deliberately uses explicit `openMemoryAsBackend` / `openPersistentAsBackend` capability-minting APIs, keeps `backendSecret` transport-only, and separates backend permission identity from stored authorship. Restoring this PR’s `isBackend` flag/private-marker design would regress that boundary.

The only unmatched behavior—optional eager rejection of whitespace-only transport secrets—is tracked separately in #2051 rather than being treated as part of backend authority.

PR #1953, opened 2026-08-24
Document secure invite links

aeplay:
🤖 Replaces #927 with a server-backed, capability-safe invite recipe.

Invite codes live only in a private table and URL fragments; redemption runs on an authenticated backend with canonical session attribution. One exclusive transaction covers membership existence, invite validation, membership insertion, and optional single-use consumption, preventing duplicate and double-redemption races. Multi-use links remain explicitly replayable.

Docs, client snippets, and standalone server snippets typecheck. The full server-doc example retains unrelated pre-existing API-drift errors. The authorization and concurrency boundary was independently adversarially reviewed.


aeplay:
🤖 Follow-up #1954 tracks moving this secure invite flow into BandChat and exercising it end to end.

PR #1956, opened 2026-08-24
Fix native provenance timestamp scaling

joeinnes:
## Summary
- scale provenance timestamps to the public microsecond representation in every native decode path
- keep ordinary application timestamp columns in milliseconds
- share one timestamp decoder between direct queries, terminal rows, and subscriptions
- add a patch changeset for jazz-tools

## Before and after

The Rust boundary supplies provenance times as physical milliseconds. The public row representation uses microseconds, and the final row transformer converts those microseconds back to a JavaScript Date.

Consider a row written at 1,700,000,000,000 milliseconds.

Before this change, direct queries multiplied the value by 1,000, but terminal subscription rows did not.

- direct query: 1,700,000,000,000 ms -> 1,700,000,000,000,000 microseconds -> 2023-11-14T22:13:20.000Z
- subscription: 1,700,000,000,000 was treated as microseconds -> 1,700,000,000 ms -> 1970-01-20T16:13:20.000Z

After this change, both paths use the same conversion.

- direct query and subscription: 1,700,000,000,000 ms -> 1,700,000,000,000,000 microseconds -> 2023-11-14T22:13:20.000Z

An ordinary application Timestamp remains unchanged: a stored value of 1,700,000,000,000 still represents that same millisecond timestamp. Only the provenance columns ($createdAt and $updatedAt) are converted to the public microsecond representation.

## Verification
- focused native runtime and subscription tests: 146 passed, 1 skipped
- targeted oxfmt and oxlint checks passed
- pre-commit hooks passed

The package TypeScript check reaches the checkout's existing linked native-artifact mismatch: NapiDb predates insertEncoded/deleteEncoded. CI rebuilds that artifact from this branch.


aeplay:
🤖 Superseded by #1989, which adopts the agreed milliseconds-only public and wire contract, keeps packed HLC state internal, expands the logical counter, removes compatibility conversions, and includes the reviewed boundary/overflow/protocol receipts.

PR #1997, opened 2026-08-25
Bound retained auxiliary chunk relay obligations

joeinnes:
## Stack dependency

This PR is stacked on #1985, which keeps browser large-value relays live across facade rebuilds. It preserves that PR's local-reader retargeting and bounded diagnostic tracing while adding retained-state admission and detach safety.

## Summary

- cap pending remote relay waiters and queued relay responses under one 4,096-entry budget
- retain request correlation for admitted relay work
- release capacity only when a response drains or its connection disconnects
- prevent storage lookups that finish after detach from recreating relay work or responses
- stop a disconnected native upstream pump instead of repeatedly resolving its readiness future
- cover both the auxiliary pump and direct `PeerConnection::tick` subscriber paths

## Before and after

Before, the 4,096 limit counted unique missing chunks. Repeated requests for one missing chunk could append unlimited waiter IDs. Completing those waiters moved them to an unbounded response queue, so repeated fill-and-complete cycles could still grow memory indefinitely.

A detached native upstream pump also treated its cancellation wake as ordinary outbound readiness. Its local executor loop stayed continuously ready, preventing the shell owner from receiving its storage-shutdown command and leaving the catalogue RocksDB lock held.

After, one counter covers the complete retained relay obligation: pending remote waiters plus queued responses. Moving a waiter into the response queue transfers its reservation instead of releasing it. A cancellation wake now terminates the native pump, allowing the owner thread to close storage deterministically.

## Governing invariants

- `relay_chunk_obligations = pending remote waiters + queued relay responses` in stable state.
- The total never exceeds 4,096.
- Local in-process chunk consumers do not consume remote relay capacity.
- Completion transfers a reservation from waiter to response.
- Draining or disconnecting releases reservations.
- Restoring an unsent response batch reacquires exactly the slots released by its drain.
- When already saturated, a new overload response may be dropped rather than retained without bound.
- Disconnect readiness terminates native upstream pumping; it is not another work notification.

### Coalesced demand

Two downstream requests ask for the same missing chunk with different request IDs. Jazz sends one upstream chunk request, retains two bounded waiter obligations, and returns one correctly correlated response for each admitted request.

### Undrained responses

A peer fills 4,096 waiter slots and the upstream completes them. The resulting queued responses continue to occupy all 4,096 slots. Further requests cannot grow retained state until responses drain or the connection disconnects.

### Detach during storage lookup

A downstream connection closes while its local chunk lookup is awaiting storage. Whether storage later reports found or missing, Jazz drops the outcome and does not recreate a relay waiter or send a response to the detached connection. The native owner pump exits on that disconnect, so server shutdown can join the owner and release RocksDB.

## Unchanged behaviour

Wire batch limits, request IDs, local chunk reads, upstream coalescing, retry delay values, PR #1985's facade retargeting, and its redacted trace format remain unchanged.

## Red/green commits

1. `82baad3e5` adds the admission, response-backlog, accounting, and detach-race regressions.
2. `7f0fadad2` implements total-obligation accounting and detach fencing on top of #1985.
3. `b747a0fe3` makes the native owner pump observe that detach as terminal.

## Verification

- `cargo test -p jazz --no-default-features --features testing,transport-compression-zstd --lib db::tests::chunk_io_pump` — 14 passed
- `cargo nextest run --profile jazz-ci -p jazz-testkit --test catalogue_sync_integration` — 26 passed
- pre-commit `cargo clippy --package jazz -- -D warnings` — passed

### DIFF
--- a/.changeset/bound-chunk-relay-obligations.md
+++ b/.changeset/bound-chunk-relay-obligations.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Bound retained auxiliary chunk relay obligations and stop the native upstream pump cleanly after disconnection.
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -144,6 +144,9 @@ fn storage_error_kind(error: &groove::chunks::ChunkStorageError) -> &'static str
         groove::chunks::ChunkStorageError::Backend(_) => "backend",
     }
 }
+// Relay correlation entries and their queued responses share one global
+// budget while in-process consumers remain outside peer admission.
+const MAX_RELAY_CHUNK_OBLIGATIONS: usize = MAX_PENDING_CHUNK_DEMANDS;
 
 enum ChunkDemandWaiter {
     Local {
@@ -166,6 +169,7 @@ struct PendingChunkDemand {
 struct ChunkDemandState {
     next_request_id: u64,
     next_waiter_id: u64,
+    relay_chunk_obligations: usize,
     pending_by_chunk: BTreeMap<groove::chunks::ChunkRequest, PendingChunkDemand>,
     chunk_by_upstream_id: BTreeMap<u64, groove::chunks::ChunkRequest>,
     outbound: VecDeque<ChunkRequestEntry>,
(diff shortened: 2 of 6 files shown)


joeinnes:
CI follow-up: this branch already contains #2004, and the rerun still failed in `trusted / test-ts`. Rebasing after #2004 lands will therefore not fix this failure by itself; the distinct WASM failure needs separate investigation.

joeinnes:
Follow-up: #2014 addresses rejected process-wide WASM loads remaining cached and poisoning later worker connections. Once #2014 lands, rebase this PR onto the updated stack and rerun CI. That should prevent the seven-test cascade; if the initial HTML-served-as-WASM failure remains, it will still need separate investigation.

aeplay:
🤖 Superseded by #2068, which reconstructs this work on the current #1985 base without rewriting the stale branch. The replacement restores both auxiliary-pump and direct-peer detach fencing, bounds pending and response obligations under one budget, includes the lost found/missing direct-tick regression, and has passed an independent adversarial re-review.

PR #1998, opened 2026-08-25
Emit legacy rename lenses in the forward direction

joeinnes:
## Summary

- correct both legacy migration collectors to emit source-to-destination rename operations
- align legacy `migrate(... col.renameFrom(...))` output with canonical `defineMigration`
- add regressions for the singular and plural collection APIs

## Before and after

For `migrate("users", { emailAddress: col.renameFrom("email") })`, the legacy collectors previously emitted:

```ts
{ type: "rename", column: "emailAddress", value: "email" }
```

That describes the rename backwards. They now emit:

```ts
{ type: "rename", column: "email", value: "emailAddress" }
```

Here, `column` is the old/source name and `value` is the new/destination name.

## Governing invariants

- Forward rename lenses always encode old name to new name.
- Reverse projection continues to invert the same canonical operation.
- Both legacy collectors produce the same orientation.
- Canonical `defineMigration` output is unchanged.

### Forward migration

A schema changes `email` to `emailAddress`. The generated forward lens reads existing `email` values and exposes them as `emailAddress` in the new schema.

### Reverse projection

Projecting the new schema back to the old version inverts the operation, mapping `emailAddress` back to `email`.

## Unchanged behaviour

Add, drop, backwards defaults, table collection, state reset, canonical migrations, and already-published lens documents are unchanged. Consumers that deliberately compensated for the old reversed output should remove that workaround.

## Red/green commits

1. `0580026c6cfe99922b4563a8b5b7b93a9bcab9cc` adds regressions for both legacy collectors.
2. `2429337362ccda2c9c67eaa102062c56b423f720` corrects the emitted orientation.

## Verification

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/dsl.test.ts -t "legacy migration collection"`
- 2 passed


aeplay:
🤖 Closing rather than repairing this deprecated compatibility API. A repository-wide search found no consumers of the legacy `migrate()` collector outside its own tests; current migrations use `s.defineMigration(...)`. We require no backwards compatibility here, so the legacy collector and its remaining traces will be removed instead.

PR #2005, opened 2026-08-25
Fail startup when the durable catalogue cannot be read

joeinnes:
🤖 Fails server construction when the durable catalogue cannot be read or interpreted safely.

## Before

Catalogue scan and decoding failures could be replaced with an empty in-memory index. A stored schema whose payload decoded but whose `published_at` metadata was missing or nonnumeric was also accepted and then omitted by `latest_published_schema()`. A dynamic server could therefore restart without its stored schemas, migrations, permission heads, or recovered shell while appearing healthy.

## After

- Scan/backend failures fail server construction.
- Recognized malformed, trailing, or metadata-incomplete catalogue records fail closed.
- Errors identify the catalogue operation, durable object/type, and underlying corruption.
- Failed startup leaves durable state untouched and releases the backend so it can be repaired and reopened.
- Unknown catalogue kinds remain intentionally forward-compatible.

## Examples

- A truncated stored schema payload produces a startup error instead of an empty catalogue.
- A valid schema payload with missing or nonnumeric `published_at` produces a contextual corruption error instead of silently disappearing from the recovered published schema.
- After repairing the corrupt durable record, the same RocksDB store can be reopened successfully.

## Verification

- `cargo test -p jazz-server server::builder::tests -- --nocapture`: 18 passed.
- `cargo test -p jazz-server server::catalogue_payload_codec::tests -- --nocapture`: 27 passed.
- `cargo fmt --check` and Clippy with warnings denied passed.
- Planted restoration of permissive scan and publication-metadata paths made their receipts fail.
- Merged and verified against current `main` without semantic conflicts.


joeinnes:
CI follow-up: once #2004 lands, rebase this PR onto the updated `main` and rerun CI. #2004 addresses the persistence replay timing failure seen here, so that rebase should clear the inherited `trusted / test-ts` failure.

PR #2007, opened 2026-08-25
Flush pending RocksDB WAL writes on close

joeinnes:
🤖 Makes RocksDB close perform the existing synchronous WAL durability boundary.

## Before

RocksDB inherited the storage interface's no-op close implementation. A clean Jazz shutdown could return successfully while the final acknowledged write batches were still waiting for the next cadence boundary.

## After

- `close()` delegates to the existing `flush_write_boundary()`.
- That boundary calls `flush_wal(true)` and propagates RocksDB failures as the existing backend error.
- Partial cadence state resets only after a successful synchronous flush.
- Repeated close/retry remains safe and idempotent.

## Examples

- If shutdown happens after fewer writes than the normal cadence threshold, close still synchronizes those pending WAL writes before releasing storage.
- If the WAL flush fails, close returns the backend failure and retains pending cadence state so a retry does not falsely treat the writes as synchronized.

## Unchanged

Normal write cadence and batch semantics do not change. This receipt proves delegation and pending-state behavior; physical power-loss guarantees remain RocksDB's `flush_wal(true)` contract.

## Verification

- `cargo test -p jazz-storage-rocksdb`: 13 unit tests and 4 receipts passed; 1 existing receipt ignored.
- Planted replacement of `close()` with a no-op made the partial-cadence close receipt fail.
- `cargo fmt --all -- --check`, workspace pre-commit checks, and sensitive-data guard passed.
- Merged current `main` while preserving its v2/conditional-merge tests; no semantic conflict.


joeinnes:

PR #2009, opened 2026-08-25
Avoid copying uncompressed inbound wire payloads

joeinnes:
## Summary
- Borrow uncompressed payload bytes during inbound semantic decoding instead of copying the complete message.
- Keep compressed payloads owned and preserve the existing public Rust decoder interface.

## Before
Fragment reassembly produced one contiguous payload, then the uncompressed decoder copied the whole buffer before semantic decoding. A legal 256 MiB message could therefore hold roughly 512 MiB of byte buffers before decoded allocations.

## After
The inbound transport uses a crate-private borrowed-or-owned decode path. Uncompressed messages borrow the assembled bytes, compressed messages own their decompressed bytes, and external callers still receive the original owned `Vec<u8>` API.

## Test plan
- [x] Run the 30 `wire::tests` cases with the maintained Jazz feature set.

### DIFF
--- a/.changeset/borrow-uncompressed-wire-payloads.md
+++ b/.changeset/borrow-uncompressed-wire-payloads.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reduce peak memory use when decoding uncompressed inbound wire payloads by borrowing the assembled message bytes.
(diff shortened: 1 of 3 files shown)


joeinnes:
CI follow-up: once #2004 lands, rebase this PR onto the updated `main` and rerun CI. That should remove the known persistence-test instability, but this rerun still failed with the distinct topology-cleanup timeout, so #2004 should not be assumed to fix that failure.

PR #2010, opened 2026-08-25
Require admin claims for announcement mutations

joeinnes:
🤖 Requires an externally authenticated admin session to mutate announcement messages and binds update authorization to one predicate across both old and new row state.

## Before

Any externally authenticated user could mutate announcements. An initial attempt expressed old-row and new-row update constraints as separate rules; because those compile as independently ORed `using` and `with_check` predicates, the creator of a General message could move it into Announcements without being an admin.

## After

- Announcement insert and delete require `claims.role == "admin"`.
- Update uses one room/role predicate applied to both old and new state.
- A non-admin may update an ordinary message only while both states remain within the ordinary-chat rule.
- Generic chat permissions otherwise remain unchanged.

## Examples

- An admin can create, edit, move, or delete an announcement.
- A regular member cannot create or edit an announcement.
- A member who created a General message cannot bypass the announcement rule by changing that row's `chat_id` to the Announcements room.

## Verification

- Focused policy tests and compiled-policy parity guard passed.
- Added a public move-denial regression.
- Planted restoration of the independently ORed predicates made the parity guard fail.
- Formatting and current-main merge checks passed.

The browser mutation fixture remains unchanged; local execution encountered a separately tracked generated-artifact codec mismatch before the authorization assertions, so CI is being watched as the canonical fresh-artifact receipt.


joeinnes:
CI follow-up: once #2004 lands, rebase this PR onto the updated `main` and rerun CI. That should remove the known persistence-test instability, but this rerun still failed with the distinct NAPI panic, so #2004 should not be assumed to fix that failure.

PR #2011, opened 2026-08-25
Enforce anonymous sessions as read-only

joeinnes:
🤖 Enforces anonymous sessions as read-only at core fate admission, including authority-local exclusive commits.

## Before

Anonymous sessions were documented as read-only, but a raw-wire client could write a policy-free table or an operation allowed by `.always()`. Authority-finalized local exclusive commits also bypassed the ordinary write-policy gate, allowing an anonymous identity to commit through that path.

## After

- Insert, update, and delete are rejected when the effective permission subject is anonymous, regardless of table policy.
- Authority-local exclusive commits run the same write-policy admission before becoming accepted.
- Public anonymous reads remain available.
- Trusted backend writes are judged using their effective permission subject; storing anonymous `made_by` provenance does not turn the trusted backend itself into an anonymous caller.
- Existing SYSTEM/admin behavior remains unchanged.

## Examples

- An anonymous WebSocket client can read a public table but cannot insert, update, or delete even when its policy is `.always()`.
- An anonymous identity cannot bypass that rule by opening an authority-local exclusive transaction.
- A trusted backend may still import a row whose historical attribution is anonymous because admission authority and stored authorship remain distinct.

## Verification

- Authority-local anonymous exclusive-write regression passed.
- Raw-WebSocket anonymous read-only receipt passed.
- Trusted-backend write with anonymous provenance passed.
- Existing exclusive conflict receipt passed.
- Planted removal of the exclusive admission gate made the anonymous write succeed and the new receipt fail.
- `cargo fmt --check`, Clippy, sensitive-data guard, and current-main merge checks passed.


joeinnes:
CI follow-up: once #2004 lands, rebase this PR onto the updated `main` and rerun CI. #2004 addresses the persistence replay timing failure seen here, so that rebase should clear the inherited `trusted / test-ts` failure.

Issue #1762, opened 2026-08-23
Require read visibility for update and upsert


aeplay:
🤖 Require session-authored `update` and `upsert` operations to observe the target row through that session's effective read policy before staging the mutation.

## Chosen semantics

- Every session `update` requires target-row read visibility, including full-row replacements, partial patches, transaction updates, and branch-view updates.
- Every session `upsert` requires read visibility when the target exists. A hidden existing row must not be treated as absent or overwritten.
- Failure is explicit and must happen before hidden cells, existence, or policy reasons are disclosed.
- Trusted/internal mutation paths may inspect authoritative hidden row state as policy evidence. That is an implementation privilege used to evaluate authorization; it does not make a read-hidden row updateable by the session.
- Insert and row-id delete semantics are outside this issue and remain governed by their existing invariants.

## Why

Updating a specific existing row inherently depends on the row the caller is modifying. Permitting a write-only update either copies hidden values into a new version or, as attempted in #2012, authors only the visible patch and silently drops omitted hidden cells. Both outcomes are surprising. Requiring read visibility makes the operation and its merge basis coherent.

## Required work

- Remove the branch-update exception proposed by #2012 and any `MaterializedPatch` machinery supporting it.
- Remove the existing full-row write-only update exception from `INV-RLS-20` and its tests.
- Apply the same rule consistently across root, branch, transaction, native, and binding entry points.
- Preserve the authority's ability to evaluate `UPDATE USING`/`WITH CHECK`, inherited grants, and policy dependencies against raw internal evidence without leaking that evidence.
- Add planted-sensitive coverage for hidden targets across full/partial update, upsert, branch views, and transactions.

## Pointers

The current invariant is `INV-RLS-20` in `crates/jazz/SPEC/7_authorization.md` and `crates/jazz/SPEC/INVARIANTS.md`. Existing write-only receipts are in `crates/jazz/tests/authorization_scope_reentry.rs`; branch behavior is exercised by #2012. Policy-evidence evaluation remains covered under `INV-RLS-21` and the policy authorization suites.

PR #2012, opened 2026-08-25
Avoid copying read-hidden cells during branch updates

joeinnes:
## Summary
- Build branch update candidates from cells visible to the writing identity rather than raw branch storage.
- Allow write-only patches without copying omitted hidden values into a new version.

## Before
Identity-scoped branch updates merged the caller's patch with raw branch cells. A caller allowed to write but not read could unknowingly copy hidden values into the authored result.

## After
The update path resolves existing cells through the caller's read policy and authors only the explicit patch when the source is hidden. Write-authorised updates remain possible without disclosing or reproducing hidden data.

## Test plan
- [x] Run the focused Jazz library regression — 1 passed.


aeplay:
🤖 Superseded by #2034 under the clarified decision in #1762. Session-authored update and upsert now require target-row read visibility across root, branch, and transaction paths; trusted policy evaluation may still inspect hidden evidence internally. This replaces the patch-only branch exception here, which would otherwise silently drop omitted hidden cells.

PR #2013, opened 2026-08-25
Expire incomplete fragmented messages

joeinnes:
## Summary
- Reclaim incomplete messages after 30 seconds without novel progress and after a five-minute absolute lifetime.
- Expire stale state before admission, receive and every send attempt, including outbound backpressure and failure paths.

## Before
A peer could send most of a fragmented message and withhold the final extent indefinitely. The staged bytes and message ID remained allocated for the lifetime of the connection.

## After
Indexed idle and maximum-age deadlines reclaim stale bytes and IDs on every transport turn. Duplicate fragments do not extend the deadline, while legal messages making real progress still complete normally.

## Test plan
- [x] Run the 26 `db::tests::wire_transport` tests.


joeinnes:

PR #2014, opened 2026-08-25
Retry failed broker-worker context initialisation

joeinnes:
## Summary
- Evict failed runtime contexts and rejected process-wide WASM loads so later connections can retry.
- Close every partially acquired database, page-store and telemetry resource exactly once.

## Before
A failed worker initialisation could leave a rejected promise in the process-wide cache or retain a failed context by key. Later tabs received the same permanent failure, and resources acquired before the error could leak.

## After
Failure removes only the unowned cached state and cleans resources in reverse ownership order. A later connection can retry the same key, while concurrent successful connections still share one initialisation.

## Test plan
- [x] Run the focused worker suite — 6 passed.
- [x] Run `pnpm --filter jazz-tools check`.
- [x] Run targeted `oxfmt --check`.

Includes a patch changeset for `jazz-tools`.


joeinnes:

Issue #1965, opened 2026-08-24
perf thesis: indexed ORDER BY LIMIT is bounded by requested results


aeplay:
## Thesis

An indexed `ORDER BY … LIMIT k` query should scale with `k` and index traversal, not total matching cardinality.

## Workload

Use BigLabel catalogue and release-time ordering. At 10k, 100k, and 1M releases measure:
- first page by indexed order;
- middle navigation using keyset/cursor pagination;
- offset pagination as an explicitly separate conventional baseline;
- limits 10, 25, 100, and 1,000;
- narrow and broad predicates.

SQLite is the first reference. Query/index definitions and returned ordering must be equivalent.

## Correctness gate

Assert exact ordered IDs, boundaries, no duplicates, and correct page transitions through the public API.

## Measurements

- latency and rows/second;
- index entries visited, rows materialized, and rows returned;
- planning versus execution;
- keyset versus offset behavior;
- allocations and decoding/materialization phases.

## Acceptance / rejection

Supported if keyset/bounded indexed reads track requested result size rather than total matching rows.

Rejected if bounded indexed execution materializes, sorts, or scans the complete matching set. Offset costs should be characterized rather than mistaken for keyset behavior.

## Non-goals

No claim that arbitrarily deep offset pagination is constant-time, and no initial claim about live subscription maintenance.

## Evidence log

Record all benchmark runs, profiles, changes, and conclusions as comments.

Issue #1962, opened 2026-08-24
perf thesis: indexed bounded reads scale with result size, not table size


aeplay:
## Thesis

On a BigLabel-derived dataset, a warm indexed query returning a fixed number of rows should remain approximately independent of total table size from 10k through 1M current rows.

## Workload

Use the BigLabel schema and deterministic release distribution. Measure indexed equality and bounded ordered reads at 10k, 100k, and 1M releases, holding returned cardinality fixed at 1, 10, 100, and 1,000 rows.

Cover the app-level catalogue load and derived mechanisms separately:
- indexed equality lookup;
- indexed ordered range with `LIMIT`;
- prepared-query reuse;
- in-memory and persistent storage where supported.

## Correctness gate

Before timing, assert exact row identities, cardinality, filtering, and ordering through the public API.

## Measurements

- latency and throughput;
- rows examined/materialized/returned;
- index and query execution phase timings;
- allocations/bytes where available;
- comparison against an equivalent indexed SQLite query.

All retained evidence must record commit, scale, storage/durability profile, fixture seed/distribution, and benchmark URI.

## Acceptance / rejection

Supported if fixed-cardinality warm-read latency has no material dependence on total table size across 10k–1M rows, apart from expected index traversal effects, and phase counters confirm bounded work.

Rejected if latency or work grows with total table size. A rejection should identify the responsible phase before proposing an optimization.

## Non-goals

This thesis does not claim cold-open performance, full-scan performance, subscription maintenance performance, or semantic equivalence with server databases.

## Evidence log

Add every relevant benchmark run, profile, code/benchmark change, and conclusion as a comment.

aeplay:
Fresh local receipt on current main (`607057bc8`) from the maintained native `selective_global_hydration` harness: **this thesis holds for the tested Global indexed one-shot shape.**

Fixed selection: 100 index candidates, limit/result 50, RocksDB reopen before query.

| table rows | query wall | index reads | current-row reads |
|---:|---:|---:|---:|
| 1,000 | 8.675 ms | 100 | 100 |
| 10,000 | 8.779 ms | 100 | 100 |

A 10x table-size increase produced 1.0x storage work and essentially flat wall time. This is a useful positive control against #2127: settled Global indexed hydration is already bounded by selected candidates; the pathological history scaling was specifically Local/Edge ahead-current candidate retention.

[...]

PR #2027, opened 2026-08-25
perf: bound W1 local one-shot read work

### LINKED ISSUE #2026: perf thesis: W1 local one-shot reads avoid unbounded hydration and no-op materialization
## Thesis

For Local-tier one-shot equality reads, Jazz currently performs unbounded ephemeral IVM hydration because physical index access is restricted to Global reads. Whole-record decode/re-encode during indirect-field materialization amplifies that scan even when inspected fields cannot contain indirect values. Removing the no-op materialization first, then enabling bounded Local access paths without weakening read-your-writes semantics, should make W1 reads scale with candidates/results rather than total table size.

## Workload

Preserve W1 profile S as the adopter-comparable compatibility workload: 3,000 tasks, 12,000 comments, 9,000 activity events. Derive isolated measurements for:
- Local `query_board`: `project_id = X ORDER BY updated_at DESC LIMIT 200`;
- Local comments and activity legs of `query_task_detail`: `task_id = X ORDER BY created_at DESC LIMIT 200`;
- corresponding Global-tier controls;
- fixed-result scaling pairs at smaller and profile-S table sizes;
- the complete weighted W1 mix once individual operations fit hosted timing budgets.

Use the anonymized public fixture only. Seed and prepare outside measured closures.

## Predicted mechanisms and checkpoints

1. Descriptor-proven no-op indirect-field materialization avoids `to_values()` and `RecordDescriptor::create()` when requested fields cannot contain indirect values. Based on the adopter bisect, predict roughly 5–7x improvement, restoring the pre-large-value range while retaining O(table) scaling.
2. A Local one-shot source combines bounded indexed settled/global candidates with the local-ahead overlay while preserving immediate local visibility. Predict a further 10–30x on selective task-detail reads, with cost tracking candidates/results instead of total rows.

## Correctness gates

Through public APIs, assert exact row identities, ordering, limit behavior, read-your-writes visibility, update/delete behavior, and Local versus Global tier semantics. Large values and fields capable of nesting indirect values must still block/materialize exactly as before.

## Measurements

- per-operation wall time and throughput;
- scaling with table rows at fixed returned cardinality;
- rows examined/materialized/returned where exposed;
- CodSpeed wall-time flamegraphs, instructions, and memory statistics;
- simulation-mode mechanism receipts where affordable;
- adopter hardware checkpoint results.

Every receipt records commit, fixture/profile, tier, storage backend, and benchmark URI.

## Acceptance / rejection

Supported when profile-S task-detail falls below 20 ms, board below 10 ms, W1 exceeds 200 ops/s as an initial milestone, and fixed-cardinality work no longer scales materially with table size.

Rejected or narrowed if no-op materialization does not recover most of the post-merge regression, or if Local indexed execution cannot preserve immediate local overlay semantics without unbounded work.

## Non-goals

This thesis does not alter query results, permission evaluation, durability semantics, subscription maintenance semantics, or large-value authentication/materialization rules. It does not initially claim parity with the old core or server databases.

## Evidence log

Record every benchmark run, flamegraph finding, checkpoint, adopter result, code change, and conclusion as a comment.

Related: #1962, #1965.

aeplay:
## Thesis

Closes #2026. Related to #1962 and #1965.

W1-style Local equality reads were hydrating unbounded ephemeral IVM graphs. Two independent amplifiers were visible: no-op large-value materialization on structurally scalar fields, and rejection/bypass of binding-specific Local physical access paths.

## Before / after

Before:

- requested fields were decoded and recreated even when their types could not contain an indirect stored scalar;
- Local access paths were discarded;
- `Db::read(&PreparedQuery)` reused a generic graph that could not embed the current binding's index prefix;
- physical ahead current tables lacked the corresponding user/FK indices.

After:

- structurally impossible large-value requests return the original record deltas;
- eligible Local equality reads compile a binding-specialized one-shot graph;
- both settled-global and local-ahead candidates are index-scanned, merged by the existing winner rule, and passed through the unchanged predicate/order/limit pipeline;
- generic maintained, Edge, and unsupported query programs retain their established full-source graphs.

The W1-derived in-memory 12k-comments receipt improved locally from 146.4ms to 1.557ms median (94.0×). Tenfold row growth now costs 1.57× rather than 15.1×. CodSpeed confirmation is running.

## Governing invariants

- Query rows, residual predicates, ordering, limits, permissions, and durability semantics do not change.
- Local reads retain immediate read-your-writes visibility by probing both physical global and ahead sources.
- Winner selection remains `arg_max_by(row_uuid, tx_time, tx_node_id)`.
- Any selected scalar or nested composite that may contain an indirect stored scalar still follows the existing interruptible materialization path.
- Invalid field indices continue to fail.
- Edge and maintained programs do not receive binding-specific source bounds.

## Worked examples

Normal path: a prepared `comments WHERE task = $task ORDER BY created_at DESC LIMIT 200` Local read now probes the task FK index in both global and ahead physical current tables, merges candidates, applies deletion visibility, then performs the unchanged ordering and limit.

Local-ahead edge: if a matching comment exists only in the local-ahead table, its ahead-index entry supplies it. If global and ahead contain the same row UUID, the existing transaction timestamp/node winner calculation selects the visible version.

Residual-filter edge: `owner = $owner AND status = 'open'` may use the owner index, but `status = 'open'` remains in the graph and still decides final membership.

Unsupported case: an unindexed or Edge query retains the ordinary source graph; no speculative bound is applied.

## Non-goals

This does not change large-value encoding/authentication, subscription maintenance, Edge coverage, query semantics, or deletion-register candidate selection. Visible-query source limits remain suppressed because deletion filtering can make a pre-anti-join limit unsound.

## Testing

- `cargo test -p groove`
- 77 query-evaluation tests
- exact Local primary-key, declared-index, and residual-filter receipts via `dev/t`
- W1 benchmark target check
- pre-commit clippy, rustfmt, and sensitive-data guard
- CodSpeed W1 walltime receipt pending

Draft while broader CI and CodSpeed complete.


aeplay:
CI diagnosis: the two repeatable browser failures were a real regression from maintained index access paths, not timeout noise. Local focused A/B with rebuilt release NAPI + fast WASM artifacts:

- enabled: filtered empty-prefix subscription timed out; Global settled snapshot timed out;
- disabled: both passed (1.04s and 3.95s test bodies respectively).

`de1663cdd` reverts `18dad5fdd` in full. One-shot index intersections remain. The separate maintained-subscription thesis is tracked in #2077. The Rust shadow failures are unrelated runner setup failures before tests (`git rev-parse HEAD^{tree}` failed).

aeplay:
## Corrected hosted baseline after maintained-index revert

CodSpeed wall-time completed at `de1663cdd`, the adopter-safe checkpoint after reverting the asynchronous-storage-unsafe maintained-index experiment.

| W1 receipt | memory median | RocksDB median |
|---|---:|---:|
| board, profile S | 34.51 ms | 38.49 ms |
| two-equality activity page, 9k rows / limit 50 | 38.01 ms | 42.98 ms |
| same page, 30k rows / limit 50 | 93.67 ms | 106.5 ms |
| comments point-filter, 1.2k table | 6.555 ms | 6.249 ms |
| comments point-filter, 12k table | 6.29 ms | 7.666 ms |
| composite task detail | 13.86 ms | 16.70 ms |
| indexed update, no subscription | 8.555 ms | 10.59 ms |
| maintained subscription delta | 42.01 ms | 47.01 ms |

Interpretation:

- The key positive invariant remains: the fixed-result comments query is table-size-flat from 1.2k to 12k, including RocksDB. The one-shot Local access path survives the correctness revert.
- The bounded two-equality page still scales with candidate/index-bucket size: 9k to 30k rows costs 2.46-2.48x for the same limit. This is the ordered/bounded-read thesis now tracked in #2079.
- RocksDB is within roughly 0-24% of memory on these shapes. Remaining cost is overwhelmingly engine/lifecycle/materialization work, not ordinary storage I/O.
- Against the adopter's old-core W1 receipt (different machine, directional only), RocksDB task detail is ~13x slower (16.7 vs 1.25 ms), board ~29x (38.5 vs 1.35 ms), update ~62x (10.6 vs 0.17 ms), and subscription delta ~71x (47 vs 0.66 ms). These ratios are not publishable cross-machine comparisons, but they rank the remaining work.
- The subscription delta is ~4.4x the no-subscription update on the same hosted runner. That makes maintained evaluation/publication the next ordinary-path target after correcting for update cost.

Adopter instruction: `de1663cdd` remains the checkpoint to test. This result does not introduce a newer read checkpoint; it establishes the safe baseline for the next subscription/cold-lifecycle work.

aeplay:
🤖 Changes requested: **P1 — the Local indexed one-shot path can return a stale global row when a newer ahead winner changes the indexed predicate column.**

Concrete receipt: settle `docs[row]` with `owner=A` globally; commit a newer pending Local update for the same row with `owner=B`; then run Local `WHERE owner=A`. The branch currently returns the old global row instead of an empty result.

The global index scan finds old `A`, while the ahead scan uses the same `A` prefix and therefore cannot see the newer `B` version. The downstream winner aggregation only sees the stale global candidate and cannot suppress it. Make the access path dominance-safe (or decline the optimization where that cannot be proven), and retain this exact predicate-changing ahead-winner receipt. This is required by current Local winner semantics / INV-READ-7.

Issue #2030, opened 2026-08-25
Perf thesis: multi-table ingest has bounded row cost


aeplay:
## Performance thesis

Bulk fixture ingestion should have bounded amortized per-row cost as a table grows. In a realistic multi-table backend workload, successive fixed-size insert batches currently get materially slower, consistent with cumulative O(n²) setup.

Initial code/profile lead: the mergeable insert path may recreate and hash logical record descriptors while evaluating the transaction against existing current rows. If descriptor/schema work or query-transaction hydration is proportional to current table cardinality, moving it out of the per-row/current-row loop should flatten each fixed-size batch.

This is a thesis, not yet a conclusion. The first benchmark must distinguish descriptor construction, IVM evaluation, storage writes, index maintenance, and durability.

## Public anonymized reproduction

Create a new, neutral synthetic example benchmark. It must not map names or vocabulary from any private application:

- tenant rows and tenant-scoped grouping rows;
- roughly 1k resource rows with optional group relation, enums, timestamps, numeric fields, and several selective indices;
- tens of thousands of tenant-scoped job rows with a status enum and optional resource relation;
- a few thousand fully indexed session rows.

Use generated UUIDs and synthetic tenant identifiers. Do not copy private schemas, permissions, literals, domains, comments, IDs, or customer terminology. Preserve only performance-relevant column types, nullability, index density, relation fanout, cardinality class, and load/query distribution.

The ingest receipt should:

1. seed prerequisite tables outside the measured phase;
2. insert the large table in fixed 1k batches;
3. report every batch independently plus cumulative throughput;
4. run at 1k/5k/10k initially, extending toward 25k only after runtime is bounded;
5. provide both in-memory CPU-isolation and RocksDB walltime lanes;
6. attribute descriptor/schema, query/IVM, index, storage, and durability phases.

## Acceptance evidence

- Fixed-size batch latency is approximately flat with existing-row count, with an explicit tolerance established from the first stable baseline.
- Total ingest scales approximately linearly from 1k to 10k.
- Query semantics, indices, permissions, transaction atomicity, and durability are unchanged.
- CodSpeed tracks the bounded-size lanes; the native realistic workflow retains the larger end-to-end receipt.

## Expected first investigation

Trace and measure:

- `node::currency::query_transaction`;
- logical-to-physical descriptor construction and `ValueType` hashing;
- whether descriptors are rebuilt per inserted row, per existing row, or per tick;
- graph hydration/current-row scans triggered by each insert;
[...]

PR #2032, opened 2026-08-25
bench: isolate anonymized ingest depth scaling

aeplay:
## Thesis

Tracks #2030.

A fixed-size insert batch should have bounded amortized cost as the table and pending-upload outbox grow. The anonymized fixture first falsified ordinary core/schema/policy/storage causes, then reproduced the adopter curve through host-driven upload maintenance.

## Before

Every host tick could perform multiple O(outbox depth) operations:

1. prune every pending upload by re-reading transaction fate, even with no peer or with a connected peer that delivered no remote input;
2. linearly scan the ordered outbox to suppress duplicate enqueue;
3. for each upstream link, scan the full outbox to compute `outbox - uploaded`;
4. scan it again for each candidate to retrieve an optional staged commit unit.

A host that ticks after each local insert therefore accumulated O(writes²) work while disconnected or connected to a slow/silent upstream.

## After

- Fate pruning consumes exact successfully applied Global/rejected FateUpdate IDs; it no longer re-reads every pending transaction.
- Released oldest uploads pop from a deque in O(1), with stable fallback for out-of-order fates; every link forgets the same released IDs.
- The ordered upload vector has companion O(1) transaction-ID membership.
- A link visits only the contiguous unuploaded suffix when length accounting proves that suffix contains every missing ID, carrying each staged unit directly.
- Fate cleanup or authority handoff can create holes; the proof fails and preserves the original full set-difference fallback.

The benchmark primes the legitimate one-time upload of an existing backlog outside the measured closure, then measures steady next-1k host-tick maintenance while the transport accepts uploads but returns no fates.

## Receipts

Original connected-silent reproduction, next 1k at 0/1k/3k/5k pending depth:

- before remote-input gate: 7.224 / 12.22 / 20.44 / 30.86 s
- after remote-input gate: 1.243 / 1.249 / 1.657 / 1.847 s (the first measured tick still included backlog catch-up)

Controlled steady-state receipt after priming that catch-up:

| pending depth | membership-only | final suffix/unit fast path |
|---:|---:|---:|
| 0 | 412.7 ms | 396.0 ms |
| 1,000 | 439.6 ms | 396.5 ms |
| 3,000 | 499.2 ms | 407.3 ms |
| 5,000 | 544.4 ms | 407.3 ms |

The final 0→5k penalty is ~11 ms; steady accepting-but-silent ingest is approximately depth-flat.

For a slow active authority returning one Global fate per tick, exact fate-addressed pruning improves next-1k from 6.106/11.36 s at 0/1k depth (deeper stopped) to 0.738/0.749/0.752/0.797 s through 5k.

## Governing invariants

- The vector remains the source of upload order; membership does not reorder transactions.
- Exact duplicate tx IDs remain suppressed across local publication, replay, subscriber ingest, reconnect, and authority handoff.
- A connection marks an upload sent only after transport acceptance; backpressure retries the same unit.
- Rate-limited or failed large-value uploads retain their established retry behavior.
- Rejected/global-settled entries remain until durable fate is observed and are then removed from both vector and membership together.
- When unuploaded entries are not a contiguous suffix, discovery falls back to the prior general set difference.

## Worked examples

Disconnected seed: 5,000 writes remain queued. Per-write host ticks do not re-read 5,000 unchanged fates; reconnect later still sees the complete ordered outbox.

Silent connected seed: the transport accepts each new commit but returns no fate. The link checks only the new suffix entry, while the pending vector remains available for eventual fate/reconnect handling.

Backpressure: sending the suffix entry fails once. It is not inserted into `uploaded`; the next tick finds and retries it.

Authority handoff: removing an older tx from the successor link's `uploaded` set creates a hole. Suffix cardinality no longer equals the missing count, so discovery performs the original full scan and redrives that tx.

## Non-goals

- Initial upload of an N-entry backlog remains intentionally O(N).
- Initial upload of an N-entry backlog and genuinely out-of-order fate removal remain O(N); steady ordered fates are O(released IDs).
- No transaction, fate, durability, retry, or upload-order semantics change.

## Validation

- anonymized memory/RocksDB/core/public-client differential benchmark API
- exact backpressure retry
- exact authority-handoff redrive
- missing-upload fatal and detach coverage
- iteration-tier incremental-delivery canaries and low-seed differential oracle running
- clippy, rustfmt, sensitive-data guard

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2110,6 +2110,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "jazz-example-multi-table-ingest-benchmark"
+version = "0.1.0"
+dependencies = [
+ "codspeed-divan-compat",
+ "jazz",
+ "jazz-storage-rocksdb",
+ "tempfile",
+]
+
+[[package]]
 name = "jazz-example-music-agent-benchmark"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ members = [
   "examples/band-chat/benchmarks",
   "examples/music-agent/benchmarks",
   "examples/benchmarks/smoke",
+  "examples/benchmarks/multi-table-ingest",
   "dev/benchmarks/storage/bench-core",
 ]
 exclude = ["examples/todo-server-rs", "examples/docs/todo-server-rs"]
--- a/crates/groove/src/large_values.rs
+++ b/crates/groove/src/large_values.rs
@@ -1,6 +1,7 @@
 //! Canonical indirect representation for large logical scalar values.
 
 use std::collections::{BTreeMap, BTreeSet};
+use std::sync::OnceLock;
 
 use serde::{Deserialize, Serialize};
 
@@ -2164,6 +2165,18 @@ pub fn inline_scalar_bytes(kind: LargeValueKind, encoded: &[u8]) -> Result<&[u8]
 }
 
 fn stored_scalar_schema(kind: LargeValueKind) -> EnumSchema {
+    static BYTES: OnceLock<EnumSchema> = OnceLock::new();
+    static STRING: OnceLock<EnumSchema> = OnceLock::new();
+    static JSON: OnceLock<EnumSchema> = OnceLock::new();
+    match kind {
+        LargeValueKind::Bytes => BYTES.get_or_init(|| build_stored_scalar_schema(kind)),
+        LargeValueKind::String => STRING.get_or_init(|| build_stored_scalar_schema(kind)),
+        LargeValueKind::Json => JSON.get_or_init(|| build_stored_scalar_schema(kind)),
+    }
+    .clone()
+}
+
+fn build_stored_scalar_schema(kind: LargeValueKind) -> EnumSchema {
     let primitive = RecordDescriptor::new([(
         "value",
         match kind {
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -1970,8 +1970,12 @@ impl UploadOutbox {
         true
     }
 
-    fn iter(&self) -> impl Iterator<Item = &PendingUpload> {
+    fn iter(&self) -> impl DoubleEndedIterator<Item = &PendingUpload> + ExactSizeIterator {
         self.entries.iter()
+    }
+
+    fn len(&self) -> usize {
+        self.entries.len()
     }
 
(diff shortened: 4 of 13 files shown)


aeplay:
Added a connected-but-stalled upstream ingest differential and fixed unconditional outbox fate rescans in `43d74abc6`.

Result for next fixed 1k Memory inserts across 0/1k/3k/5k existing pending rows: `7.224/12.22/20.44/30.86s` before, `1.243/1.249/1.657/1.847s` after (5.8x–16.7x). See #2030 for mechanism, proof scope, and adopter instruction.

aeplay:
Rebased onto current main and added checkpoint `7ad0bc28f`: the shared ordered outbox now has O(1) duplicate membership. Connected-silent next-1k improves another 1.9–2.5x across 0–5k pending depth. Full mechanics and adopter instruction are on #2030.

aeplay:
Checkpoint `1bcc9549b` bounds steady accepting-but-silent upstream maintenance. Controlled next-1k is 396.0/396.5/407.3/407.3 ms across 0/1k/3k/5k pending depth after moving legitimate initial backlog upload outside measurement. Full mechanics and adopter instruction are on #2030.

aeplay:
Checkpoint `ca95c3c09` makes active-fate pruning exact and state-change-addressed. One-fate-per-tick next-1k improves from 6.106/11.36 s at 0/1k depth to 0.738/0.749/0.752/0.797 s through 5k. Full mechanics, gates, and adopter instruction are on #2030.

aeplay:
Checkpoint `e804e05e1`: cache canonical stored-scalar schemas.

Active-fate next-1k curve changes from 738/749/752/797 ms (0/1k/3k/5k prior curve; immediate depth-0 rerun 714 ms) to 458/458/472/499 ms. All 38 Groove large-value tests and pre-commit gates pass. Adopter has been asked to test this cumulative SHA.

aeplay:
Checkpoint `6e6941ee5`: remove unused current-source descriptor construction. Active-fate next-1k improves another 9-13%; cumulative over the immediate post-fate depth-0 baseline is 714.2 -> 399.8 ms (~44%, 1.79x). Focused physical-current and write-scale tests plus hooks pass. Adopter test request now points at this cumulative SHA.

aeplay:
🤖 Superseded by #2170, which reconstructs the still-valid outbox optimizations directly on the current core with bounded discovery and a planted rejection/reconnect no-replay receipt. The old-core branch also contains obsolete benchmark and architecture history, so it will not be mechanically restacked.

PR #2064, opened 2026-08-26
Apply rejection cascades iteratively

aeplay:
🤖 Reconstructs the still-live rejection-cascade stack-safety fix isolated while retiring #1678.

## Before

Applying a rejected transaction recursively settled every causal descendant. A sufficiently deep but valid causal chain overflowed the Rust stack on both client and authority paths.

## After

Jazz collects transitive cascade descendants iteratively once, then settles each descendant with further discovery disabled. Client fate application and authority cascade rejection use the same bounded-stack path while retaining deterministic causal order and pending-edge cleanup.

## Examples

- Rejecting the root of a 1,024-transaction chain settles the whole chain without recursion.
- A cycle in pending-edge bookkeeping terminates through the seen set.
- A nested descendant is still discovered and settled; disabling nested discovery leaves it pending and fails the regression.

## Invariants and non-goals

Every causally dependent transaction is rejected exactly once in deterministic transaction-id order. Terminal permission-subject and inbound/outbound pending-edge cleanup remain per transaction. This does not change rejection policy or introduce a global scan.

## Verification

- 1,024-deep cascade regression passes in about four seconds.
- Re-enabling recursion makes the regression stack-overflow and abort.
- Disabling nested discovery makes the later descendant remain pending.
- Existing local, authority, causality-payload, root-cause, and recovery cleanup receipts pass.
- Independent adversarial review found no correctness issue.
- Formatting, clippy, diff, and sensitive-data gates pass.

### DIFF
--- a/crates/jazz/src/node/ingest/fates.rs
+++ b/crates/jazz/src/node/ingest/fates.rs
@@ -10,6 +10,18 @@ where
         global_time: Option<GlobalTime>,
         durability: Option<DurabilityTier>,
     ) -> Result<(), Error> {
+        self.apply_fate_update_with_cascade(tx_id, fate, global_time, durability, true)
+            .await
+    }
+
+    async fn apply_fate_update_with_cascade(
+        &mut self,
+        tx_id: TxId,
+        fate: Fate,
+        global_time: Option<GlobalTime>,
+        durability: Option<DurabilityTier>,
+        cascade_descendants: bool,
+    ) -> Result<(), Error> {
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Follow-up design question captured in #2070: determine when a causal descendant of a rejected transaction remains semantically valid, whether read/write sets or observed versions are a better dependency signal than parentage alone, and whether recovery should be automatic or application-controlled. This PR remains narrowly responsible for deterministic, stack-safe rejection cleanup.

PR #2068, opened 2026-08-26
Bound retained auxiliary chunk relay obligations

aeplay:
🤖 ## Stack dependency

This replacement is stacked on #1985 and supersedes #1997. It reconstructs the bounded relay work on the current browser-relay base without rewriting the stale branch.

## Before

Repeated requests for the same missing chunk could retain unbounded waiter IDs even though only unique chunks were capped. Completing waiters could then move them into an unbounded response queue. A disconnect during an awaited lookup could also recreate a waiter or send a found response after the requesting peer had detached.

## After

One 4,096-entry budget covers pending remote waiters plus queued and in-flight responses. Completion transfers a reservation rather than releasing it early; drain or disconnect releases it. Both the auxiliary pump and direct `PeerConnection::tick` paths fence before and after async lookup and before sending. Native owner pumps treat disconnect as terminal.

## Examples

- Two request IDs for one missing chunk share one upstream request but retain two bounded response obligations.
- If 4,096 waiters complete before responses drain, the response backlog continues to occupy all 4,096 slots and further work cannot grow retained state.
- If a peer detaches during a local lookup, neither a missing result nor a found result may recreate work or emit a response.

## Invariants and non-goals

- Stable retained obligations equal pending remote waiters plus queued/in-flight responses and never exceed 4,096.
- Local in-process readers do not consume the remote relay budget.
- Failed sends restore exactly their drained reservations.
- Wire limits, retry delays, request IDs, upstream coalescing, and #1985 facade retargeting are unchanged.

## Verification

- Focused chunk I/O suite: 14/14
- Server runtime: 4/4
- `cargo check -p jazz-napi -p jazz-wasm`
- Clippy/rustfmt/sensitive-data gates
- Independent adversarial review planted reservation and detach-fence regressions; each failed at the intended assertion and was restored cleanly.

### DIFF
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -92,6 +92,10 @@ pub(crate) type SharedNodeState<S> = Rc<LocalMutex<NodeState<S>>>;
 
 const DEFAULT_CHUNK_FORWARD_HOPS: u8 = 8;
 const MAX_PENDING_CHUNK_DEMANDS: usize = 4096;
+// Downstream relay waiters and their response hand-offs share one budget. A
+// response remains reserved while a binding is deciding whether its send
+// succeeded, so a failed send can be restored without racing fresh admission.
+const MAX_RELAY_CHUNK_OBLIGATIONS: usize = MAX_PENDING_CHUNK_DEMANDS;
 // Auxiliary chunk routing crosses several independent transports in browser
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 GitHub automatically closed this stacked PR when merged #1985's base branch was deleted. The unchanged reviewed head is republished directly against `main` as #2071; no code or review findings were lost.

PR #2072, opened 2026-08-26
Restore BandChat topology correctness coverage

aeplay:
🤖 ## Stack

This reconstructs and supersedes #1846 on top of #2056 and the merged deterministic fault scheduler from #2066.

## Before

BandChat had unit-level permission coverage, but its stale topology draft no longer ran against the current runtime. It did not reliably prove concurrent app activity, offline replay, bounded reconnect behavior, or authorization after membership revocation. Its reaction policy also allowed a revoked member to mutate reactions on previously learned message IDs.

## After

The maintained browser/core topology runs real BandChat query and mutation shapes through concurrent messages, reactions, an attachment, offline replay, bounded projection across reconnect, and membership revocation. Reaction insertion and deletion require self-authorship, a referenced message in the same room, and current membership; updates remain denied.

## Examples

- Two members concurrently send messages and reactions while an attachment is published; the final projected room state contains each admitted operation exactly once.
- A disconnected client writes locally, reconnects, and receives settlement through the real edge/core topology.
- After membership removal, the former member cannot add a new reaction or delete an existing reaction, even when it retained the message ID.
- A client cannot forge another author's reaction.

## Invariants and non-goals

- Current membership, not merely prior knowledge of a row ID, gates future room authority.
- Child references remain correlated to the same room.
- Projection bounds survive reconnect; the test does not require shipping the complete child collection.
- This PR adds topology receipts; it does not introduce new application semantics or a new fault-scheduler path.

## Verification

- Canonical test-artifact build and clean-tree freshness preflight
- BandChat typecheck
- Permission receipt: 1/1
- Focused Chromium topology: 2/2
- Independent adversarial review planted forged-author and post-revocation authorization regressions; both failed at the intended assertions and were restored cleanly.


aeplay:
🤖 Superseded by #2155. GitHub automatically closed this stacked PR when its merged base branch was deleted; the reviewed head is unchanged and #2155 retargets it to `main`.

Issue #2069, opened 2026-08-26
perf thesis: policy-aware point reads scale with policy result, not table size


aeplay:
## Thesis

A policy-aware Local point read bounded by row UUID should scale with the target row and the bounded policy evidence needed to authorize it, not with the outer table size or unrelated policy-relation rows.

For indexed `exists` policies, reusable policy graphs should use available equality/relationship access paths at Local durability while preserving the normalized final predicate and session binding semantics.

## Motivation

An adopter-shaped persistent RocksDB harness observed a 10-80x session tax on point reads with owner and membership-`exists` policies, while some list/filter reads were faster through the session path. Current mechanism tests also deliberately record Global policy index probes while expecting Local policy evaluation to fall back to full scans. That suggests a durability-specific access-path gap rather than unavoidable policy evaluation cost.

## Anonymized workload

Add a public microbenchmark derived from W1-style schemas, without customer names or identifiers:

- `documents`: UUID primary key, indexed owner/ref/status fields;
- `memberships`: indexed principal and container/ref fields;
- direct owner policy: `owner = session.user_id`;
- relationship policy: `exists memberships where membership.container = document.container AND membership.principal = session.user_id`;
- backend and session controls;
- fixed one-row result at 10k and 100k outer rows;
- independently vary membership-table cardinality and selectivity;
- memory and RocksDB receipts where practical.

## Correctness invariants

- Hidden rows remain hidden; a primary-key candidate must not bypass policy evaluation.
- Session claims retain their declared/coerced types, including nullable references.
- Reusable policy graphs remain valid across repeated probes and relevant row/policy-evidence transitions.
- Local/Ahead winner semantics remain unchanged.

## Measurements

- wall time for backend versus session point reads;
- source full scans, index probes, candidate rows, and returned rows;
- cold graph construction separated from repeated graph reuse;
- wall-time flamegraph/instruction/memory data on CodSpeed when the benchmark shape is stable.

## Acceptance / rejection

Supported if warm Local session point-read latency and examined candidates are approximately independent of unrelated outer-table and membership-table size, apart from index traversal and actual policy evidence.

Rejected if either source hydrates unrelated rows or rebuilds a table-sized policy graph per point read. A rejection should identify the exact lowering/runtime phase before optimization.

## Relationship to existing theses

- #1962 covers ordinary indexed bounded reads.
- #1965 covers indexed `ORDER BY ... LIMIT`.
[...]

PR #2073, opened 2026-08-26
bench: expose policy point-read amplification

aeplay:
## Before

`owner_filter_diagnostic` used a test-only seed/finalize plus close/reopen lifecycle and only broad Global owner-filter cases. On current runtime semantics that lifecycle produced no current rows, so it could not measure the adopter-reported policy tax on a one-row query.

## After

The diagnostic now:

- seeds through the ordinary public batched transaction API;
- installs an explicit provider `user_id` claim rather than conflating provenance identity with application identity;
- keeps setup outside measured phases;
- compares an indexed one-row Local read under System authority with the identical identity-policy read;
- retains broad owner, ordered, and limited controls;
- exposes authorized-set cardinality independently from total table size.

At 10k outer rows, the one-row System control performs 2 logical reads in ~0.9 ms. With 500 authorized owner rows, the identical point result performs 1,002 logical reads in ~10 ms. With 5,000 authorized rows it performs 10,002 logical reads in ~156 ms. This rejects the thesis that policy-aware point reads are bounded by the requested candidate.

## Invariants and non-goals

- Exact row counts are asserted before emitting a receipt.
- The final owner policy remains authoritative; this commit changes only the diagnostic.
- No performance fix or adopter-testable checkpoint is claimed yet.
- Correlated membership/`exists` policy coverage is follow-up work in #2069.

## Testing

- `dev/gates/benchmark-smoke.sh jazz owner_filter_diagnostic`
- pre-commit clippy, sensitive-data guard, and rustfmt

Tracks #2069. Stacked on #2027 because it measures access paths introduced and corrected there.


aeplay:
🤖 Closing as superseded historical diagnostic work. Thirteen of this branch's sixteen commits were carried into the later performance integration branch, and the policy point-read proof it tracked was addressed by #2120 / closed issue #2069. The remaining commits are benchmark diagnostics rather than a current implementation dependency; if the regression reappears, port the receipt afresh onto `main` rather than reviving this stale stack.

PR #2075, opened 2026-08-26
Serialize concurrent streaming publication

aeplay:
🤖 ## Before

Two streamed writes could ingest their sources concurrently and then call the shared WASM runtime's `upload.finish()` concurrently. On the single-threaded WASM evaluator this wedged both publications and later shutdown indefinitely. The equivalent NAPI path completed, which let weaker or stale native-only coverage hide the browser/default-runtime bug.

MusicAgent also lacked current public-app receipts for streaming transcript/tool rows, exact binary attachment contents, cancellation, and concurrent streamed turns.

## After

Source ingestion remains concurrent. Only the shared native finalization/publication boundary is FIFO-serialized per owner runtime, including across schema views. Every queue slot releases in `finally` on success or failure. Ordinary mutations and NAPI behavior are unchanged.

MusicAgent now runs the same two concurrent streamed turns through the public store API and verifies persisted transcript, tool output, exact Bytea payload, and cancellation without publication.

## Examples

- Two assistants stream turns concurrently. Both producers can yield chunks together; publication finishes first turn then second turn, and both rows become locally durable.
- If a producer throws, its staged upload aborts and releases its publication slot; a following stream can finish.
- Cancelling an attachment stream rejects and publishes no attachment row.

## Invariants and non-goals

- Serialization is scoped to one owner runtime, not global across databases.
- Registered schema views share the same publication queue.
- Only `upload.finish()` is queued; source reads and pushes remain concurrent.
- This does not change ordinary insert/update semantics or the NAPI ABI.

## Verification

- Concurrent child watchdogs: NAPI and WASM 2/2
- Native runtime streaming/cancellation suite: 12/12
- MusicAgent Jazz/WASM suite: 5/5
- MusicAgent TypeScript build and Jazz Tools runtime/RN typecheck
- Canonical NAPI and WASM artifact provenance: fresh
- Independent adversarial review confirmed owner scoping, failure release, cancellation, and unchanged NAPI behavior; removing the finalization gate restores the deterministic WASM timeout.


aeplay:
🤖 Superseded by #2109. #2075 accumulated unrelated stack work; the replacement retains only the owner-runtime streaming-finalization behavior and focused sensitivity-tested receipts on current `main`.

Issue #2087, opened 2026-08-26
Define applyDiffs semantics in transactions, branches, and attributed sessions


aeplay:
🤖 `Db.applyDiffs` currently accepts only root/database identity and explicitly rejects transaction, branch-view, and attributed/session contexts.

The first public partial-value DSL can safely expose root-context page-relative text/byte splices and JSON edits, but extending it requires a deliberate answer to three coupled questions:

- Within a transaction, should staged diffs compose against the transaction view and commit atomically with ordinary row updates?
- On a branch/view, which version/root is the splice coordinate space relative to, and how should consolidation preserve branch semantics?
- For attributed/session writes, which author/admission identity must flow through the low-level partial-update boundary?

The implementation must not silently fall back to root/SYSTEM identity or bypass the caller’s branch/transaction view. Design this as a follow-up to the root-context DSL reconstruction; until then those contexts should fail explicitly and be documented as unsupported.

Relevant work: #1833 (old draft), its current-main replacement (to be linked after creation), and #1839 for demand propagation.

aeplay:
🤖 The current-main root-context implementation is now visible in #2088. This issue is the explicit follow-up boundary; #2088 must continue failing unsupported contexts rather than silently changing identity/view semantics.

PR #2088, opened 2026-08-26
Add the large-value query and applyDiffs DSL

aeplay:
🤖 Reconstructs and supersedes #1833 on the current stack. It exposes one high-level TypeScript API for reading and editing part of a large value without restoring removed low-level binding helpers.

## Example: read only the needed part

```ts
const textPage = { from: 4, to: 124 }; // UTF-16 offsets, matching JavaScript strings

const [document] = await app.documents
  .where("id", documentId)
  .select({
    title: true,
    body: textPage,
    attachment: { from: 1_000_000, to: 2_000_000 },
    metadata: { at: "/chapters/11/title" },
  });
```

The returned fields are ordinary JavaScript values: a string page, a byte range, and the JSON value at the RFC 6901 pointer. The native runtime emits exactly one bounded projection per selected field rather than also materializing the complete value.

## Example: replace and diff columns atomically

```ts
await db.update(
  app.documents,
  documentId,
  {
    title: "Revised document", // complete primitive replacement
  },
  {
    applyDiffs: {
      body: {
        within: textPage,
        splices: [{ at: 3, delete: 1, insert: "new text" }],
      },
      attachment: {
        within: { from: 1_000_000, to: 2_000_000 },
        splices: [{ at: 4, delete: 3, insert: replacementBytes }],
      },
      metadata: {
        edits: [{ op: "set", at: "/chapters/11/title", value: "Finale" }],
      },
    },
  },
);
```

The replacement and every diff commit as one atomic update. A column may appear in either the replacement map or `applyDiffs`, never both. Typed tables reject overlap in TypeScript; the runtime also rejects it universally.

For a diff-only update, pass an empty replacement map:

```ts
await db.update(app.documents, documentId, {}, {
  applyDiffs: {
    body: {
      within: textPage,
      splices: [{ at: 3, delete: 1, insert: "new text" }],
    },
  },
});
```

There is deliberately no separate public `Db.applyDiffs` method. Whole-column replacement and partial edits are two representations within the ordinary update operation.

Splice positions are relative to the selected page. Text uses UTF-16 offsets by default so positions match JavaScript string indexing; explicit UTF-8 variants are also supported.

## Runtime boundary

TypeScript validates and lowers the typed selection or update. NAPI/WASM dispatch it to Jazz, which validates the schema-derived value kind and commits all replacements and diffs atomically. UTF-16 conversion descends through the tree with stored metrics and applies the edit tail without materializing or scanning the entire prefix.

Session writes require the runtime's exact canonical author. Transaction, branch, differently attributed, or different-author contexts currently fail closed; their desired semantics are tracked in #2087.

Exact per-terminal chunk-demand propagation remains #2090. The API works now; that issue is an evaluation-efficiency follow-up.

## Evidence

- runtime declarations and React Native typechecks;
- focused adapter, in-memory, native-runtime, and typed-fixture tests;
- compile-time and runtime receipts for replacement/diff overlap;
- Rust receipts for non-BMP UTF-16 offsets, edit-tail traversal, schema-kind rejection, and mixed-update rollback;
- canonical NAPI/WASM artifact verification;
- planted regressions for dispatch, duplicate projections, TypeScript inference, author binding, UTF-16 boundaries, and schema-kind checks;
- independent adversarial review after repair cycles.


aeplay:
🤖 Approved by Anselm.

Issue #2077, opened 2026-08-26
Perf thesis: maintained indexed subscriptions must preserve async-storage delivery


aeplay:
## Problem

The first attachment of a maintained filtered subscription still scans the
complete root table even when the corresponding one-shot authority query uses
an equality index and is tenant-cardinality bounded.

This dominates edge/global latency for the main multi-tenant query:

```text
WHERE ownerId = session.user
ORDER BY createdAt DESC
LIMIT N
```

## PR #2361 evidence

PR #2361 head: `9ad56bd8f`. The route benchmark held one binding, its 100-row
result, and retained maintained memory fixed while adding unrelated source rows:

| Total source rows | First subscribe | Retained heap |
| ---: | ---: | ---: |
| 2,000 | 63.0 ms | 78.9 KB |
| 10,000 | 159.7 ms | 78.9 KB |
| 100,000 | 1.224 s | 78.9 KB |

The fixed footprint confirms that source-wide witness retention is no longer
the cause in this prototype. The residual first-attach slope is the maintained
program's full physical-current source.

By contrast, the one-shot protected owner query already uses the owner index:

| Total rows | Owned/result rows | Logical reads | Query time |
| ---: | ---: | ---: | ---: |
| 10,000 | 500 | 1,000 | 42.804 ms |
| 100,000 | 500 | 1,000 | 42.796 ms |

The maintained path is therefore table-size dependent while the authority
one-shot path is not. Extrapolating the measured maintained slope puts a
1-million-row first attachment around 12 seconds on this native harness.

## Why the obvious index switch was reverted

PR #2027 briefly enabled the one-shot equality-index access path for maintained
programs. Memory tests passed, but IndexedDB/WASM lost delivery in two cases:

- an initially empty indexed prefix did not deliver its first matching insert;
- a fresh Global subscription did not deliver its settled synced snapshot.

The experiment was reverted in `de1663cdd`. Index and current-row storage can
settle at different frontiers, so maintained source selection needs an explicit
completeness/delivery proof.

## Implementation options

1. Make maintained equality-index sources frontier-aware and incrementally
   complete across Memory, RocksDB, IndexedDB/WASM, and async worker delivery.
2. Add a direct authority one-shot query RPC for initial selective hydration,
   then attach a maintained continuation from a compatible frontier.

Option 1 is the desired unified model. Option 2 may be a useful staged design if
the snapshot/continuation boundary can be made race-free.

## Acceptance criteria

[...]

PR #2093, opened 2026-08-26
perf(jazz): bound flat subscription delta bookkeeping

### LINKED ISSUE #2086: Perf thesis: subscription deltas must scale with changed rows, not retained results
## Thesis

A flat maintained subscription delta should cost proportional to the changed rows, not the retained result cardinality. On current main, updating one row with an unfiltered 1,000-row memory subscription is about 69 ms versus 0.91 ms for the identical update without an observer (~76x).

## Grounded mechanism

The ordinary maintained flat-result branch in `refresh_subscriptions_in` clones both the complete `RelationSnapshot` and complete `RelationSnapshotIndex` before applying every incremental update. Those previous copies are only consumed when `authoritative_membership_changed` is true, to rebuild an ordered suffix after authority reconciliation. For normal local one-row deltas that flag is false, so the O(result rows) copies are dead work.

The current benchmark `observer_write_path` already gives a clean reproduction: fixed 1,000-row table, identical content-only update, with and without `subscribe(Query::from("documents"))`. Current `[profile.perf]` receipt:

- no observer: 0.906 ms median
- observe all: 69.15 ms median

## Proposed proof and change

1. Extend the benchmark over multiple result cardinalities and attribute maintained evaluation versus snapshot/index/event work.
2. Move the previous snapshot/index copies inside the rare authoritative-reconciliation branch, preserving them before mutation only when that branch needs them.
3. Require exact subscription add/update/remove and async-storage delivery canaries; this optimization must not revive the #2077 opening/delta correctness failure.
4. If residual cost still grows with result size, inspect `relation_snapshot_has_materialized_required_cells` and the maintained reducer/index update independently.

Expected first-lever improvement: 3-20x for 1k-row ordinary deltas, with the multiplier growing with retained result size. The lower bound allows Groove maintenance and write publication to remain significant after the dead copies disappear.

Adopter instruction: **do not test yet**. This issue records a locally reproduced thesis; a checkpoint and exact receipt will be posted when correctness gates pass.

aeplay:
## Summary

Stacked on #2027. This removes two result-cardinality-sized bookkeeping operations from ordinary flat maintained subscription deltas and extends the existing observer benchmark across 100, 1,000, and 5,000 retained rows.

Closes #2086.

## Before / after

Before, every maintained flat delta cloned the complete public `RelationSnapshot` and `RelationSnapshotIndex`, even though the copies are only needed for an authoritative membership reconciliation. Removal handling then walked every root and reverse-looked-up its occurrence identity, including updates with no actual removal.

After:

- the previous snapshot/index are retained only for authoritative membership changes;
- removals iterate the delta's removed occurrence ids and use the existing occurrence-to-position index;
- normal add/update/remove event semantics and the retained snapshot remain unchanged.

Example: a content update to one row in an ordinary `subscribe(all documents)` stream updates that one occurrence without copying or reverse-scanning every retained root. An actual removal still removes the indexed occurrence and shifts later positions exactly as before. Authority reconciliation still captures the complete pre-change snapshot before mutation and uses it to publish the ordered suffix.

## Performance receipt

Local memory `[profile.perf]`, same branch/harness:

- 1,000 retained rows: ~69.15 ms median before -> ~41.15 ms after (1.68x)
- 5,000 retained rows after: ~215.65 ms
- no-observer update control remains ~0.74-0.91 ms

This is a useful first checkpoint, not the end-state. Tick diagnostics show exactly 40 processed delta records at both 1k and 5k, while latency and retained evaluator payload remain linear (7.3 MB -> 36.5 MB). The next thesis is O(retained-state) runtime bookkeeping around an otherwise bounded incremental tick; CodSpeed wall-time should identify the exact owner.

## Correctness receipts

- flat subscription update with nullable sort payload: pass
- offset window after leading-row deletion: pass
- authoritative reset occurrence/order/count rebuild: pass
- benchmark target compile and pre-commit clippy/rustfmt/sensitive-data guard: pass

Browser async-storage delivery gates are required before landing because #2077 showed that a semantically plausible maintained-path optimization can miss asynchronous first-match delivery.

## Non-goals

- no maintained query planning or index-selection change;
- no wire protocol or subscription event shape change;
- no change to structured/terminal-row materialization;
- no claim yet that delta latency is result-size-independent.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -89,7 +89,10 @@ jobs:
     if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'benchmark')
     name: W1 read wall time (memory + RocksDB)
     runs-on: codspeed-macro
-    timeout-minutes: 20
+    # A cold macro runner currently spends ~14 minutes rebuilding RocksDB and
+    # another ~11 minutes measuring both backends. Keep enough headroom for the
+    # complete pair so a successful memory suite cannot starve RocksDB.
+    timeout-minutes: 35
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
--- a/crates/jazz/benches/observer_write_path.rs
+++ b/crates/jazz/benches/observer_write_path.rs
@@ -100,7 +100,7 @@ fn all_documents_query(db: &BenchDb) -> jazz::db::PreparedQuery {
 fn update_write_path_with_and_without_observer(c: &mut Criterion) {
     let mut group = c.benchmark_group("observer_write_path/update_content");
 
-    for scale in [1_000usize] {
+    for scale in [100usize, 1_000, 5_000] {
         group.throughput(Throughput::Elements(1));
 
         group.bench_with_input(
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -4234,37 +4234,29 @@ fn apply_maintained_update_to_snapshot(
         }
     }
 
-    let mut index = 0;
-    while index < snapshot.root_count {
-        let occurrence_id = snapshot_index
-            .roots
-            .iter()
-            .find_map(|(occurrence, position)| (*position == index).then(|| occurrence.clone()))
-            .expect("every maintained root row has an occurrence index");
-        if update_removed.contains(&occurrence_id)
-            && !update_added
-                .iter()
-                .any(|(added, _)| *added == occurrence_id)
-        {
-            let row = snapshot.rows.remove(index);
-            snapshot.root_count -= 1;
-            snapshot_index.roots.remove(&occurrence_id);
-            for position in snapshot_index.roots.values_mut() {
-                if *position > index {
-                    *position -= 1;
-                }
-            }
-            for position in snapshot_index.related.values_mut() {
+    for occurrence_id in &update_removed {
+        if update_added.iter().any(|(added, _)| added == occurrence_id) {
+            continue;
+        }
+        let Some(index) = snapshot_index.roots.get(occurrence_id).copied() else {
+            continue;
+        };
+        let row = snapshot.rows.remove(index);
+        snapshot.root_count -= 1;
+        snapshot_index.roots.remove(occurrence_id);
+        for position in snapshot_index.roots.values_mut() {
+            if *position > index {
                 *position -= 1;
             }
-            removed.push(RemovedRow {
-                table: row.table().to_owned(),
-                row_uuid: row.row_uuid(),
-                occurrence_id,
-            });
-        } else {
-            index += 1;
         }
+        for position in snapshot_index.related.values_mut() {
+            *position -= 1;
+        }
+        removed.push(RemovedRow {
+            table: row.table().to_owned(),
+            row_uuid: row.row_uuid(),
+            occurrence_id: occurrence_id.clone(),
+        });
     }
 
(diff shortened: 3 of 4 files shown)


aeplay:
## Checkpoint 2: retained ordered TopBy state

Pushed `20c3c19e1` to #2093.

Instead of rebuilding and sorting complete before/after groups, `TopBy` now owns an ordered per-group multiset. Each tick:

- applies only changed encoded records to their `(sort key, record bytes)` entries;
- reads the old/new offset+limit windows from ordered state;
- diffs only those windows and preserves root-ordering operations;
- retains weighted bag semantics and the complete-record final tie-breaker;
- rebuilds the same state from full snapshot input during hydration.

Local memory `[profile.perf]` receipt:

| retained rows | before this PR | checkpoint 1 | checkpoint 2 |
|---:|---:|---:|---:|
| 100 | ~15.51 ms | 7.62 ms | **1.417 ms** |
| 1,000 | 69.15 ms | 41.15 ms | **2.001 ms** |
| 5,000 | not separately captured | 215.65 ms | **7.697 ms** |

At 1k this is **34.6x faster than the original baseline** and 20.6x faster than checkpoint 1. At 5k it is **28.0x faster than checkpoint 1**. The identical no-observer update control is ~0.74-0.91 ms, so the remaining subscription premium is ~1.1 ms at 1k and ~7 ms at 5k locally.

Correctness receipts:

- complete Groove gate: 535 passed, 2 ignored; all integration/doc suites green
- Jazz flat ordered update canary: pass
- Jazz offset-window deletion canary: pass
- Jazz authoritative reset order/count/occurrence canary: pass
- clippy/rustfmt/sensitive-data hook: pass

The branch has been pushed and hosted/browser CI restarted. CodSpeed is expected to add the authoritative wall-time/flamegraph receipt; the current W1 subscription shape should move sharply while its no-subscription update control stays flat.

Adopter instruction: **not yet**—please wait for the browser async-storage gate on this exact SHA. If it passes, this will be the first checkpoint I recommend testing for the 28-30 ms `subscribeAll` delta receipt.

aeplay:
Correctness update for `20c3c19e1`: the exact hosted trusted suite is green, including **test-ts/browser**, Rust workspace, differential, lint, and benchmark API smoke. This clears the async-storage concern from #2077 for this checkpoint. W1/BigLabel timing jobs are still running; they are performance receipts, not correctness blockers.

Adopter instruction: **ready to test `20c3c19e1` now** for the `subscribeAll` delta shape. It is stacked on the adopter-safe #2027 head and includes both generic delta-bookkeeping and incremental ordered-TopBy improvements. Please report the same p50/p95 subscription-delta row and whether initial subscription order/cardinality remain unchanged. One-shot reads, ingest, and update semantics are intended to remain unchanged.

aeplay:
Final hosted receipt: subscription delta is 14.4 ms memory / 16.3 ms RocksDB versus ~42 / 47 ms on the #2027 baseline (2.9x both backends). All correctness checks and the extended W1 job pass.

The full wall-time flamegraph confirms TopBy is no longer the residual bottleneck. The leading costs are general update read-back/query compilation (7.3 ms, 41.5%; compile/access-path lowering 5.5 ms, 31.3%) and commit (6.2 ms, 35.2%). Follow-up should optimize update-by-known-row rather than add more subscription-specific machinery.

aeplay:
🤖 Superseded by #2179, which reconstructs the same generic flat-delta optimization directly on current `main` without rewriting this PR's divergent history. The replacement preserves the intended single bookkeeping path, reuses the existing scale benchmarks already on `main`, and adds fresh focused and planted-regression evidence.

Issue #2091, opened 2026-08-26
Isolate generated test artifacts per worktree


aeplay:
🤖 Generated WASM/NAPI test artifacts are published into repository-shared paths, so independently isolated worktrees can invalidate each other between build and test.

Observed repeatedly while reconstructing #2081, #2082, and reviewing #2085:

- A lane runs the repository artifact build and its local JS expects fingerprint A.
- Another worktree later publishes fingerprint B into shared `crates/jazz-wasm/pkg` / generated locations.
- The first lane’s focused browser receipt now fails before app setup with an explicit expected-A/got-B ABI mismatch.
- The publisher serializes each individual publish, but it does not preserve a worktree-local JS/WASM pair after the lock is released.

This defeats worktree isolation and forces an exclusive global build/test window for every browser receipt. It also creates false-red feedback unrelated to the branch under test.

Give each worktree/build invocation an isolated artifact output (or immutable fingerprint-addressed artifacts), then make test runners resolve the artifact paired with their own generated bindings. Keep the explicit ABI mismatch check; remove the shared mutable publication dependency.

Acceptance receipt: two worktrees build different commits concurrently/sequentially, then each runs its browser tests successfully against its own matching artifact without rebuilding after the other publishes.

PR #2094, opened 2026-08-26
Reconstruct the BandBinder example

aeplay:
🤖 ## Summary

Reconstructs BandBinder, superseding #1850, as a self-contained Next.js + Better Auth example of a permissioned collaborative workspace.

## What changed

- Uses Jazz magic authorship and issuer-scoped canonical session authors for membership and policy checks.
- Keeps account/workspace bootstrap explicitly user-triggered and resumable, rather than creating data on the initial read path.
- Covers interruption and retry while bootstrap is partway through creating workspace, members, pages, and blocks.
- Covers same-subject/different-issuer isolation.
- Uses explicit final-authority receipts for allowed and denied mutations.
- Includes a representative benchmark with canonical authors and exact `$createdAt` ordering/projection behavior.
- Adds the missing root workspace lockfile entry so frozen installs include this application.

## Verification

- BandBinder typecheck and seven focused application tests pass.
- Frozen workspace install for BandBinder passes.
- Independent adversarial review passed for bootstrap, issuer isolation, authorization receipts, and benchmark semantics.

## Follow-up

Browser-topology receipts are deferred by #2091, which isolates generated test artifacts per worktree. The current focused tests exercise the application behavior without claiming that missing receipt.


aeplay:
🤖 Superseded by #2171, which preserves the BandBinder app commits on an explicit #2096 stack, migrates removed `session.author` usage to canonical `session.user`, and runs current artifact/topology receipts instead of retaining obsolete core history.

PR #2096, opened 2026-08-26
Expose canonical user identity on sessions

### LINKED ISSUE #2092: Expose canonical author on public sessions
🤖 The permission DSL exposes canonical `session.author`, but public React `useSession(): Session` currently exposes only `issuer` and `user_id`. Apps that persist an invitee/member author therefore have to reproduce the internal canonical JSON encoding themselves.

Example: BandBinder needs to store an invited principal in a membership row whose policy compares it with `session.author`. The schema/policy can use the canonical author, but the UI cannot obtain the same logical value through the public session API.

Expose the canonical author string through public session/binding APIs (or one shared canonical constructor used by every binding) so applications never hand-roll encoding. Preserve the established identity split:

- JWT `sub` remains provider-controlled.
- Logical author is the canonical encoding of issuer+subject.
- Local interning remains an implementation detail.
- SYSTEM remains the single special system author.

Add cross-binding receipts proving React/TS and native sessions produce exactly the value the permission DSL sees, including two issuers with the same subject.

aeplay:
🤖 ## Before

Session identity and provider claims shared a flat policy namespace. The permission DSL exposed `session.author`, public binding sessions did not expose the same canonical value, and an arbitrary provider claim could collide with a reserved Jazz field. Apps persisting memberships had to reproduce Jazz’s internal identity encoding.

## After

- `session.user` is the sole top-level application identity in public sessions and permission policies.
- Its value is an opaque canonical JSON encoding of the exact JWT `[iss, sub]` pair. It is not a user-row reference, display name, or provider `sub` alone.
- Provider claims are nested and accessed explicitly as `session.claims[key]`; arbitrary flat session claims are removed.
- A provider claim named `user` remains available as `session.claims["user"]` and cannot shadow or spoof `session.user`.
- `userIdentity(issuer, subject)` canonicalizes identities supplied outside an active session, such as invitations.
- `$createdBy` retains its authorship name and uses the same canonical identity representation.
- Raw JWT `iss` and `sub` remain present under `session.claims`. SYSTEM and private local intern handles remain unchanged.
- There are no compatibility aliases for `session.author`, `sessionAuthor`, or flat provider claims.

## Examples

- Two providers issue the same `sub`: their `session.user` values remain distinct.
- A payload contains `{ user: "forged" }`: policies see it only through `session.claims["user"]`; `session.user` is still derived from `[iss, sub]`.
- An app stores `session.user` in its membership row and compares that field with `$createdBy` or `session.user`.
- Attempting `session.someProviderClaim` fails with guidance to use `session.claims["someProviderClaim"]`.

## Verification

- Full Jazz library: 1,376 passed, 2 ignored.
- Exact provider-user collision/spoof fixture: passed.
- Focused TypeScript permission/session/type tests: 68 passed.
- TypeScript typecheck, Clippy, formatting, lint, and sensitive-data guard passed.
- BandChat’s current reaction policy was migrated to `session.user` while preserving its membership and self-authorship rules.

Closes #2092.


aeplay:

PR #2097, opened 2026-08-26
Reconstruct the Wequencer example

aeplay:
🤖 ## Summary

Reconstructs Wequencer on current `main`, superseding #1851, as a self-contained collaborative sequencer example with a matching native workload.

## Before

The old draft used in-memory Better Auth, persisted raw provider user IDs as application identity, hid profile creation in a UI read/create flow, and described membership and presence more strongly than its policies and storage guaranteed.

## After

- Persists Better Auth in Jazz while denying its generated tables to ordinary clients.
- Uses canonical issuer-scoped session authors for profiles, invites, and membership identity, including a same-subject/different-issuer receipt.
- Keeps profile bootstrap explicit and server-side.
- States the implemented creator-bound membership authority; transferable administration remains a design question in #2100.
- Uses bounded ordered queries in both the app and native benchmark, including actual playhead windows and latest-transport `limit(1)` reads.
- Describes presence as stale-capable cached observations without inventing expiry or clock-synchronization semantics.
- Uses one documented application origin across Better Auth, JWKS, bootstrap, and canonical author derivation.

## Verification

- TypeScript typecheck and five focused application tests pass.
- Native benchmark tests pass 6/6.
- Frozen dependency installation and formatting gates pass.
- Independent adversarial review passed, including planted issuer-collision and Better Auth visibility regressions.

Browser-topology execution is deferred by #2091's shared generated-artifact isolation problem; this PR does not claim that missing receipt.


aeplay:
🤖 Superseded by #2208, which reconstructs the same example as an app-only diff on the current stack, preserves the useful commits and tests, ports identity/creator semantics to the current API, and now passes the full browser → Edge → Core topology after #2206. Closing this older branch so the open queue reflects the implementation we will actually integrate.

PR #2099, opened 2026-08-26
fix(jazz): make graceful shutdown a real storage boundary

### LINKED ISSUE #2098: Perf thesis: graceful RocksDB reopen must not replay the full recent WAL
## Thesis

A graceful persistent-client reopen should not replay the application's full recent write history before the first bounded read. Current W4 profile S (3k tasks / 12k comments / 9k activity plus related rows) takes **1.58-1.72 s** for connect + one indexed `project_id`, ordered, `limit(200)` read on this machine.

## Profile receipt

The steady one-cycle native profile, using a pre-seeded persistent store and timing only `connect_client` plus first query, is dominated before query evaluation by:

`RocksDbStorage::open_with_durability` -> `DB::Open` -> `DBImpl::Recover` -> `RecoverLogFiles` / `ProcessLogFile` -> `WriteBatch::Iterate` / memtable insertion.

Other visible reopen work (`NodeState::new_with_options_inner`, `restore_pending_uploads`, current/index scans) is secondary. This is distinct from the already-fixed one-shot query hydration issue.

## Grounded mechanism

Jazz's documented graceful `Db::close` drains subscription/runtime state and calls storage `close`. The RocksDB adapter's `close` delegates to `flush_write_boundary`, which calls `flush_wal(true)` only. That makes the WAL durable but does not flush memtables to SSTs. With the shared 256 MB write-buffer budget, a sizable recently seeded store can close cleanly while retaining a large recoverable WAL; the next open synchronously replays it.

## Experiment and tradeoff

Measure the same seeded store across:

1. current WAL-sync-only close;
2. explicit RocksDB memtable flush on graceful close;
3. a lower/periodic background flush threshold that spreads work across the write phase.

Every receipt must report seed/write time, graceful shutdown time, reopened `DB::Open`, first-query time, WAL/SST bytes, and total close+reopen time. A faster reopen that merely hides a larger unreported shutdown is not a valid win.

Expected outcomes:

- explicit close flush: **5-20x faster RocksDB open**, but shutdown may absorb hundreds of milliseconds and write amplification;
- background threshold: smaller reopen tail with lower shutdown cliff, at some steady-write throughput cost;
- bounded query time after open should remain in the tens of milliseconds established by #2027.

The decision is use-case-sensitive: server processes may value uninterrupted ingest and tolerate restart recovery, while desktop/mobile adopters directly expose reopen latency. We may need a lifecycle policy or tuned default rather than one universal setting.

Adopter instruction: informational only; **do not test a new SHA yet**. The current profile explains the existing cold-reopen row but contains no change.

aeplay:
Closes #2098.

## Before / after

Before, `JazzClient::shutdown(self)` only disconnected its upstream peer and returned. It did not reach `Db::close()`, so it did not drain subscription finalization, write the node clean-close marker, or ask storage to establish its close durability boundary. W4 therefore called an API named “shutdown” but produced a crash-style WAL-only store.

After, shutdown detaches upstream and awaits the existing core `Db::close()` lifecycle. The existing RocksDB behavior remains WAL-sync only; a tested experiment adding synchronous all-CF memtable flush was rejected because it added roughly 4.6 s to profile-S shutdown and did not improve the warm-cache reopen.

W4 now also reports `cold_reopen_open` and `cold_reopen_first_query` alongside end-to-end `cold_reopen`. On profile S with the current read stack, the accepted WAL-only graceful boundary measured 7.40 s seed+shutdown and 696 ms reopen: 688 ms open plus 8.15 ms first query.

## Invariants

- Public graceful shutdown is an awaited local database/storage lifecycle boundary, not merely a network disconnect.
- Upstream is detached before local close begins.
- Existing node-close ordering remains authoritative: close subscription-finalization admission, drain finalizations, flush node-local maintenance, persist the clean-close marker, then establish the storage durability boundary.
- RocksDB close policy remains WAL sync; ordinary shutdown does not synchronously flush all memtables.
- Any close error is returned to the caller.

## Worked cases

Normal offline path: write rows, await local durability, call `shutdown()`. Jazz persists its clean-close marker and synchronizes the WAL before returning. A later context can reopen and read the rows.

Connected path: shutdown first detaches upstream, then closes the same local state. No new sync work is admitted during local close.

Failure path: a local close/storage error is returned rather than reporting a successful shutdown.

Crash/kill path: unchanged. Ungraceful termination cannot call `shutdown()` and continues to use normal WAL/crash recovery.

Clone edge: `shutdown()` remains context-wide in effect. Consuming one `JazzClient` handle closes its shared core even if another facade clone exists; callers must not use clones after shutdown.

## Tests and receipts

- public client persistent rehydrate test now uses `shutdown()` rather than `drop()`, exercising the public boundary before reopen
- focused test passed via `dev/t`
- pre-commit clippy, rustfmt, and sensitive-data guard passed
- realistic runner compile passed

## Non-goals

- no synchronous memtable flush
- no change to write durability tiers
- no optimization yet for the remaining O(rows) recovery scans
- no change to crash recovery

### DIFF
--- a/.changeset/quiet-streams-close.md
+++ b/.changeset/quiet-streams-close.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Ensure Rust client shutdown cancels retained subscription forwarding before closing persistent local storage, allowing the same client directory to reopen immediately.
--- a/crates/jazz-native-transport/examples/realistic_bench.rs
+++ b/crates/jazz-native-transport/examples/realistic_bench.rs
@@ -893,6 +893,8 @@ async fn run_w4_cold_start(
     let hot_project = seed.projects[0];
 
     let mut latencies = Vec::with_capacity(cycles);
+    let mut open_latencies = Vec::with_capacity(cycles);
+    let mut first_query_latencies = Vec::with_capacity(cycles);
     progress(format!("W4 start cycles={}", cycles));
 
     for i in 0..cycles {
@@ -902,12 +904,17 @@ async fn run_w4_cold_start(
         copy_dir_recursive(&data_dir, &cycle_data_dir)?;
         let t0 = Instant::now();
         let client = connect_client(app_id, cycle_data_dir, None).await?;
+        let open_elapsed_ms = t0.elapsed().as_secs_f64() * 1000.0;
         let query = jazz::query::Query::from("tasks")
             .filter(eq(col("project_id"), lit(*hot_project.uuid())))
             .order_by("updated_at", OrderDirection::Desc)
             .limit(200);
+        let query_started = Instant::now();
         let _ = client.query(query, None).await?;
+        let first_query_elapsed_ms = query_started.elapsed().as_secs_f64() * 1000.0;
         let elapsed_ms = t0.elapsed().as_secs_f64() * 1000.0;
+        open_latencies.push(open_elapsed_ms);
+        first_query_latencies.push(first_query_elapsed_ms);
         latencies.push(elapsed_ms);
         client.shutdown().await?;
         drop(cycle_temp);
@@ -916,6 +923,14 @@ async fn run_w4_cold_start(
     let wall_time_ms = latencies.iter().sum::<f64>();
     let mut operation_summaries = BTreeMap::new();
     operation_summaries.insert("cold_reopen".to_string(), summarize_op(&latencies));
+    operation_summaries.insert(
+        "cold_reopen_open".to_string(),
+        summarize_op(&open_latencies),
+    );
+    operation_summaries.insert(
+        "cold_reopen_first_query".to_string(),
+        summarize_op(&first_query_latencies),
+    );
 
     Ok(BenchResult {
         scenario_id: scenario.id.clone(),
--- a/crates/jazz-storage-rocksdb/src/lib.rs
+++ b/crates/jazz-storage-rocksdb/src/lib.rs
@@ -15,8 +15,8 @@ use std::path::{Path, PathBuf};
 
 use rocksdb::{
     BlockBasedOptions, Cache, ColumnFamilyDescriptor, DB, DBCompactionStyle, DBCompressionType,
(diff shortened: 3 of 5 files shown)


aeplay:
Receipt correction: the W4 2.1–2.3x claim in the current description is retracted. `JazzClient::shutdown()` does not call this storage `close()`, and the measured fixture produced no SST flush. I am keeping this draft open while wiring and testing the real graceful-shutdown path; do not review the performance claim yet.

aeplay:
Updated decision: synchronous memtable flush is rejected by the controlled tradeoff (adds ~4.6 s shutdown and does not improve warm reopen). I will remove that behavior from this draft. The PR will retain only the missing public graceful-shutdown wiring plus W4 phase attribution; residual recovery-scan optimization belongs in a follow-up.

aeplay:
🤖 Changes requested from adversarial review:

- **P0 — shutdown is not context-wide with retained `JazzClient` clones.** `ClientDb::close` invokes core close but leaves the shared backend installed. A retained clone keeps the RocksDB handle/lock; reopening the same store fails with `LOCK: No locks available`, and the clone can still reach a supposedly closed context. Shutdown needs shared terminal state that removes/drops the backend and makes every retained facade reject further work.
- **P1 — the new shutdown test does not prove the close boundary.** Replacing `ClientDb::close` with detach-and-`Ok(())` leaves it green because consuming the sole client handle eventually drops storage anyway. Add a retained-clone reopen receipt and an observable/controllable close-boundary receipt proving finalization/clean-close/storage close actually ran.

aeplay:
🤖 Changes requested: context-wide shutdown still leaks persistent storage through retained public subscriptions. A planted black-box receipt retained a public subscription, called shutdown, then reopened the same RocksDB directory; reopen deterministically failed with LOCK held by the current process. Each public subscription forwarding task owns the core stream, while shutdown currently removes only ClientDbInner.db and does not cancel/drain those tasks before core close. Please close facade subscription admission, cancel and await forwarding tasks before Db::close, race-proof a subscribe already in flight, and add retained-subscription reopen plus stream-terminality receipts. Cancellation/error terminality should also receive direct tests.

PR #2081, opened 2026-08-26
Reconstruct the RecordPlayer example

aeplay:
🤖 ## Work in progress

This reconstructs and will supersede #1853 on current `main`. The first checkpoint preserves the self-contained RecordPlayer app, playlist/media query shapes, topology receipts, and matching benchmark so the ongoing modernization is visible.

## Intended before/after

- Before: the old draft is stacked on obsolete trusted-author infrastructure and uses stale `session.user_id` authorization and provider assumptions.
- After: a current-core Next.js/Better Auth example with server-side account bootstrap, canonical issuer-scoped authorship, current query hooks, faithful playlist/media benchmarks, and app-user topology/fault receipts.

## Current checkpoint

- Restored the app and benchmark foundation on current `main`.
- Modernizing permissions, auth/provider structure, and topology assumptions is still in progress.
- Focused receipts and independent adversarial review are required before this leaves draft.

PR #2103, opened 2026-08-26
Harden the RecordPlayer example

aeplay:
🤖 ## Stack

This replaces #2081 on top of #2096 because RecordPlayer now uses the public canonical session-author API instead of hand-rolled JWT decoding. Merge #2096 first.

## Before

The old reconstruction’s app and benchmark used different indices; the benchmark omitted the headline audio metadata/projection workload; maintained tests omitted provider/selection/security receipts; invite authors were manually decoded from JWT payloads; token failures could leave the UI stuck; and UI copy claimed playback behavior not yet mounted.

## After

- App, topology and Rust benchmark share one access-path manifest with parity receipts.
- The benchmark models nullable 64KiB `audio_bytes`, metadata-only projection, and exact CoverFlow/track/playlist bounds, order and projected fields.
- Maintained package tests include unit, permissions, provider, and selection receipts; topology remains an explicit gate deferred only by #2091.
- Better Auth tables have real client read/write denial receipts.
- Invitations use public `session.author`; same subject under two issuers remains isolated.
- Provider token rejection/null states surface errors and support retry.
- UI/docs claim only implemented metadata-first behavior.

## Verification

- TypeScript typecheck; unit 7, provider 5, selection 2.
- Rust benchmark workloads: 2/2; bench compile passes.
- Planted CoverFlow-bound, Better Auth allow-read, and issuer-collision regressions each failed at the intended boundary and were restored.
- Exact deferred topology command: `pnpm --filter record-player-next-betterauth test:topology` after #2091 provides worktree-isolated artifacts.

Independent final re-review remains before readiness.


aeplay:
🤖 Superseded by draft #2209, which reconstructs RecordPlayer as an app-only diff on the current stack and preserves the relevant example, benchmark, and topology work without the obsolete copied core/session changes. The replacement remains draft while its final recipient-visibility correctness failure is diagnosed, so progress is visible without keeping two competing branches open.

PR #2105, opened 2026-08-26
perf(storage): bound RocksDB recovery debt during writes

### LINKED ISSUE #2104: Perf thesis: bound RocksDB recovery debt during ordinary writes
## Thesis

Jazz's desktop RocksDB profile uses a 256 MiB shared write-buffer budget. Realistic local-only datasets well below that size can remain entirely in memtables/WAL until process exit, so every reopen replays the whole recent working set. Bounding the shared budget should convert recovery debt to small background flushes during ordinary writes, improving cold reopen without a synchronous shutdown cliff.

## Profile-S evidence

Composed with the current read, descriptor-cache, and graceful-shutdown work; fresh stores; 3k tasks / 12k comments / 9k activity rows:

| shared WBM | seed + WAL-sync shutdown | reopen | open | first query | tradeoff |
|---:|---:|---:|---:|---:|---|
| 256 MiB | 7.15 s | 525 ms | 517 ms | 8.17 ms | 55.9 MB remains in one WAL |
| 32 MiB | 7.21 s (+0.8%) | 387 ms (-26.4%) | 378 ms | 8.09 ms | background SSTs begin |
| 16 MiB | 7.30 s (+2.1%) | 292 ms (-44.4%) | 284 ms | 8.03 ms | proposed knee |
| 8 MiB | 7.47 s (+4.5%) | 251 ms (-52.2%) | 243 ms | 8.14 ms | many tiny SSTs / file churn |

A synchronous all-CF flush at shutdown was also tested and rejected: ~12.0 s seed+shutdown and ~722 ms warm reopen in the original comparison. It moves too much work to lifecycle latency.

## Prediction

Set the default shared WBM to 16 MiB. Expected on this fixture: ~44% faster cold reopen for ~2% ingest/seed cost. Larger/write-heavy server deployments may prefer a different profile later; this issue is explicitly choosing a desktop/local-first default tradeoff.

## Required receipts

- focused RocksDB storage suite and full core correctness
- realistic W4 phase receipt
- BigLabel ingest wall-time / adopter seed receipt to ensure the write cost remains acceptable
- store metrics/file count to detect small-SST or compaction amplification

## Non-goals

- no synchronous memtable flush on shutdown
- no change to WAL durability/fsync semantics
- no claim that 16 MiB is optimal for server-only bulk ingest

aeplay:
Closes #2104.

## Before / after

Before, Jazz shared a 256 MiB RocksDB write-buffer budget across all column families. A realistic local dataset could stay entirely in memtables and a ~56 MB WAL through shutdown, forcing the next open to replay the recent working set.

After, the shared budget is 16 MiB. RocksDB emits incremental background SSTs during sustained writes, bounding recovery debt without a synchronous shutdown flush.

Profile-S W4 on the current read/descriptor/graceful-close stack:

- 256 MiB: 7.15 s seed+shutdown; 525 ms reopen (517 ms open, 8.17 ms query)
- 16 MiB: 7.30 s seed+shutdown; 292 ms reopen (284 ms open, 8.03 ms query)
- result: **44.4% faster reopen for 2.1% seed cost**

The tested 32 MiB point was 387 ms reopen / 7.21 s seed. An aggressive 8 MiB point reached 251 ms / 7.47 s but created substantial small-SST/file-number churn, so 16 MiB is the chosen knee.

## Invariants

- WAL remains enabled with the same durability and sync policy.
- Write-buffer memory remains shared across column families.
- No foreground all-CF flush is added to writes or shutdown.
- Stored bytes, compaction styles, compression, block cache, and target file sizes are unchanged.
- Backpressure/flush decisions remain RocksDB's; Jazz only lowers the global memory budget that triggers them.

## Worked cases

Desktop/local sustained writes: once aggregate memtables approach the budget, RocksDB flushes in the background. A later reopen has less WAL state to recover.

Small dataset: if writes remain below the budget, behavior is effectively unchanged.

Bulk ingest: more background flush work may overlap writes; profile S measured ~2.1% additional seed wall time. Server-only ingest may eventually warrant an explicit storage profile rather than weakening the desktop default.

Crash: unchanged durability. WAL still contains every write not represented in a durable SST and normal recovery replays it.

## Tests and receipts

- full `jazz-storage-rocksdb` suite: 17 passed, 1 manual receipt ignored
- pre-commit clippy, rustfmt, sensitive-data guard
- local 256/32/16/8 MiB W4 curve documented in #2104

## Non-goals

- no synchronous shutdown flush
- no server-specific bulk-ingest tuning in this PR
- no change to Jazz's remaining O(pending rows) recovery/index reconstruction


aeplay:
🤖 Review note: the 16 MiB write-buffer manager appears durability-safe, but its performance contract currently has no checked-in sensitivity. Reversibly restoring 256 MiB leaves all 17 focused nonignored storage tests green.

This is also a repository-wide default, including server workloads, while the documented receipt is desktop/reopen-oriented. RocksDB may trade the lower recovery debt for more small-L0 flush/compaction work under sustained ingest; the shared budget is a soft flush trigger, not a hard memory cap.

Before merge, please either (a) make the budget storage-profile-specific, or (b) add a hosted/configuration receipt that distinguishes 16 MiB from 256 MiB and covers both reopen debt and sustained-ingest cost, with the global-default tradeoff explicit in the PR.

aeplay:
🤖 Approved with the current shared 16 MiB default. Follow-up configurability—potentially including a CLI/storage-profile option—is tracked in #2165.

PR #2108, opened 2026-08-26
perf(jazz): slim ahead-current recovery index

aeplay:
## Before

Every node open scans the durable ahead-current overlay and decodes five fields per record to rebuild three process-local ordered indexes. Only the exact replay-idempotency membership index has an external consumer; the other two indexes are maintained but never read.

At 27,053 ahead-current entries, the composed W4 fixture attributes about 74.2 ms to this rebuild.

## After

- retain only the exact replay-idempotency index
- key it by `(physical table id, content/deletion layer, encoded primary key)` directly from Groove's raw scan
- represent it as `FxHashSet`, matching exact contains/insert/remove use
- remove record-field decoding, logical schema/table resolution, and two unread derived indexes

Five local reopens measured the rebuilt index at 14.3–15.5 ms, about 80% faster for this phase. Composed open measured 223.7–229.3 ms versus roughly 284 ms before this lane (~20–21% faster); stable hosted attribution remains follow-up.

## Governing invariants

- Replaying the exact same physical `(table, layer, primary key)` remains a no-op before insertion.
- The encoded primary key is Groove's canonical branch/row/tx-time/node identity, not a lossy fingerprint.
- Settling or rejecting a version removes that exact key after scheduling durable deletion.
- Opening an existing store reconstructs every durable ahead-current exact key before accepting work.
- Hash iteration order is irrelevant: no semantic path iterates the set.

## Worked examples

Normal replay: a peer resends a pending content version already present in its physical ahead-current table. The rebuilt hash set contains that table id, content layer, and exact encoded primary key, so insertion returns early exactly as before.

Fate cleanup: a pending version receives global fate. Cleanup deletes its physical ahead-current row and removes that exact encoded key. A distinct newer version for the same logical row has a different primary key and remains present.

Cold reopen: persisted content and deletion overlay keys are copied directly from Groove's raw scan into the exact set before the node is returned. Record values do not need decoding to prove physical-key presence.

## Non-goals

- no storage format or query-result change
- no change to winner selection or local visibility
- recovery remains O(ahead-current entries), now with much smaller per-entry work

Tracks #2104.

## Validation

- `cargo check -p jazz --no-default-features --features testing,transport-compression-zstd`
- exact replay-after-reopen test
- exact fate-cleanup test
- exact identical receiver replay/conflict test
- rustfmt, clippy, sensitive-data pre-commit gates


aeplay:
🤖 Review gap: the encoded ahead-current replay key appears complete, but this behavior-changing recovery rewrite has no sensitive permanent receipt for pending same-version replay after reopen.

Please add a protocol/public-style test that stages pending content and deletion-register entries across two physical tables for the same logical row, reopens, replays the exact units without duplicating physical rows, accepts a distinct same-table encoded key, and proves fate/rejection removes only the intended key. Existing recovery tests cover visibility or already-global versions; a planted idempotency bypass leaves them green, so they do not protect this boundary.

PR #2110, opened 2026-08-26
Isolate generated correctness artifacts per worktree

### LINKED ISSUE #2091: Isolate generated test artifacts per worktree

aeplay:
🤖 ## Before

Correctness builds published fast-WASM and release-NAPI artifacts into repository-shared mutable package paths. A second worktree could replace those artifacts after the first worktree built them, leaving the first lane's generated bindings paired with a different binary fingerprint. Its browser or NAPI receipt then failed before app setup, despite isolated source checkouts.

## After

- After the ordinary producers seal provenance, correctness builds snapshot the exact WASM/NAPI pair into an immutable fingerprint-addressed store under that worktree's Git metadata.
- Each checkout atomically records a private pointer to its matching snapshot.
- Browser bundling, runtime test helpers, and native loading resolve the private correctness snapshot; release/package assembly continues to use the ordinary package outputs.
- ABI/provenance verification remains fail-closed, while local snapshot pointers are excluded from source provenance and packages.
- Snapshot identities are exact contained fingerprint pairs; source and stored symlinks, path traversal, incomplete inventories, and tampered reuse fail closed.
- Mutable NAPI is ABI-preflighted (and repaired/revalidated when necessary) before an immutable snapshot can be published.

## Verification

- Two distinct worktrees build different fingerprints, one replaces the mutable package outputs, and both still dynamically load their own WASM and NAPI snapshots.
- Full test-artifact build and correctness verification pass.
- Provenance suite passes 15/15.
- Artifact-pipeline, browser contract 8/8, and Jazz Tools typecheck pass.
- A live Vite receipt serves the worktree-private snapshot while rejecting paths outside the worktree.
- Adversarial traversal, symlink, tamper, poisoned-generation, and stale-fingerprint receipts pass.

Fixes #2091.

Snapshot retention/GC is tracked separately in #2111.


aeplay:
🤖 Follow-up cleanup once subagents run in isolated VMs: https://github.com/garden-co/jazz/issues/2156

PR #2115, opened 2026-08-26
perf: skip policy-free update authorization query

### LINKED ISSUE #2114: Perf thesis: policy-free partial updates should not compile a serving query
## Thesis

A partial root update currently calls `merge_existing_cells_for_client_identity`, which evaluates `local_row_for_client_identity` by preparing, lowering, binding, and executing a policy-capable physical-row serving query before reading the same row's physical cells. For `SYSTEM` and tables without a read policy, that authorization query cannot filter the row and is therefore redundant.

The W1 update benchmark is exactly this case: `SYSTEM` identity, no read policies. Its latest walltime flamegraph attributes roughly 7.3 ms / 41.5% of the update+subscription operation to `local_row_for_client_identity -> query_rows_for_client_physical_row`, while the no-subscription update is still about 8.2 ms in memory and 10.2 ms on RocksDB.

## Prediction

Load the current physical cells and content winner directly when the writer is `SYSTEM` or the table has no read policy; retain the existing authorization-query path for policy-bearing client identities. This should remove most of the point-query segment and plausibly improve policy-free W1 update latency by 2-5x, with a larger fraction of the remaining cost attributable to commit/durability and maintained-view work.

## Governing invariants

- Partial updates may merge existing cells only when the writer is allowed to observe the row.
- `SYSTEM` bypass and an absent table read policy are unconditional visibility cases.
- Policy-bearing client identities continue through the existing query evaluator, including claim binding and relation-policy evaluation.
- Deletion winner checks, content winner parent selection, merge semantics, and authored-column tracking remain unchanged.

## Experiment

1. Add a direct current-row helper for unconditional-visibility writes.
2. Exercise both no-policy/SYSTEM and policy-bearing partial updates through public APIs.
3. Measure W1 memory and RocksDB update walltime before/after, then inspect the walltime flamegraph.
4. Only consider reusable/cached policy point graphs after the unconditional case is removed from the profile.

aeplay:
Closes #2114. Stacked on #2027.

## Before / after

Before, every partial client-local root update prepared and executed a policy-capable physical-row serving query before loading the row's physical cells. That proof was performed even for `SYSTEM` and tables without a read policy, where visibility is unconditional. The physical cells and their causal-parent winner were then resolved in two separate lookups.

After, those unconditional cases load the current physical cells and exact content winner together in one lookup. Policy-bearing client identities retain the existing #2027 bounded authorization query.

Exact local W1 profile-S A/B on the same revision and machine:

| backend | #2027 median | skip-query checkpoint | fused-winner head | head improvement |
| memory | 2.482 ms | 0.525 ms | 0.345 ms | 7.2x faster |
| RocksDB WAL-no-sync | 1.702 ms | 0.560 ms | 0.397 ms | 4.3x faster |

Setup remains outside the measured closure. Both receipts toggle one indexed predicate on a 9,000-row activity table and wait for local durability, without an installed subscription.

## Invariants

- Partial updates inherit omitted physical cells only after read visibility is established.
- `SYSTEM` and an absent table read policy are unconditional visibility cases.
- Policy-bearing client identities still execute the existing claim/relation-aware authorization query.
- Deletion checks, physical descriptor preservation, winner-parent selection, authored-column tracking, and commit behavior are unchanged.
- The causal parent returned with the physical cells is the same exact winning `VersionRow` from which those cells were decoded; the parent and preimage cannot diverge between duplicate lookups.

## Worked cases

Normal path: `SYSTEM` patches `activity.kind` on a policy-free table. Jazz reads the current physical winner once, decodes its cells, derives its transaction ID as the parent, merges `kind`, and commits without constructing a serving query.

Policy edge: a write-only principal patches one field on a table whose read policy hides the row. Jazz does not take the fast path; the existing authorization probe denies the update before omitted cells can be inherited.

Missing/deleted edge: the existing deletion check and missing-current-cell error remain authoritative; the fast path does not synthesize a row or turn an update into an upsert.

## Tests / receipts

- `dev/t --test authorization_scope_reentry write_only_full_row_update_succeeds_but_partial_update_and_upsert_are_denied`
- all three `incremental_delivery_canary` mechanism tests
- `cargo fmt --check`
- `cargo check -p jazz`
- `cargo clippy --package jazz -- -D warnings` (pre-commit hook)
- sensitive-data guard (pre-commit hook)
- focused W1 memory and RocksDB walltime A/B above


aeplay:
Checkpoint `d02b2a7b5` adds a public-API policy-bearing W1 case and removes the redundant serving query after trusted session read authorization succeeds.

Receipt at 9,000 activity rows: 301.3 -> 151.0 ms (~2.0x). Correctness: both write-only denial and authorized scope-reentry update tests pass.

The benchmark also exposes the next bottleneck rather than declaring victory: 900 rows = 16.53 ms, 9,000 rows = 150.3 ms. That new point-policy scaling thesis is isolated in #2117; #2115 does not attempt the graph-representation fix.

PR #2116, opened 2026-08-26
Bound relation delivery canary setup

aeplay:
🤖 ## Before

The INV-INC-1 relation-delivery canary spent roughly 2.5 minutes single-core constructing and initially hydrating 1,000 versus 20,000 relation rows before its measured one-child update. The setup alone nearly exhausted CI's 180-second per-test watchdog; the actual measured update was fast. This made unrelated branches red without detecting an incremental-delivery regression.

## After

- Seed rows are inserted in bounded ordinary 500-row transactions.
- Fixture sizes are 250 versus 5,000, preserving the same 20× scale contrast.
- The measured allocation invariant and one-child delivery behavior are unchanged.

Focused runtime is about 22 seconds rather than roughly 150 seconds.

## Verification

- Exact canary passes with the bounded fixture.
- A planted full-rebuild allocation in the measured window fails the existing byte-ratio bound (`14.25 > 3`), then passes after restoration.
- Formatting, Clippy, diff, and sensitive-data checks pass.

This repairs the workspace CI failure exposed while refreshing #2072. It is independent of #2113's deep same-row fate replay fix.

### DIFF
--- a/crates/jazz/tests/incremental_delivery_canary.rs
+++ b/crates/jazz/tests/incremental_delivery_canary.rs
@@ -214,20 +214,34 @@ fn seed_relation_fixture(db: &Db<TestStorage>, child_rows: usize) -> RowUuid {
     ))
     .expect("insert parent");
 
-    for index in 0..child_rows {
-        block_on(db.insert(
-            "children",
-            BTreeMap::from([
-                ("parent_id".to_owned(), Value::Uuid(parent.0)),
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Follow-up performance investigation: #2154 tracks the unexpectedly slow large-fixture seeding and initial hydration exposed while repairing this canary.

Issue #2122, opened 2026-08-26
Perf thesis: parent validation should scale with touched rows, not transaction width


aeplay:
## Thesis

Validating a mergeable write's explicit parents should cost O(parents × touched rows), not O(all versions in each parent transaction).

This matters for realistic batch ingest: a later update/delete commonly names a parent transaction that inserted thousands of unrelated rows. The validation invariant only asks whether a parent version for the same physical table and row belongs to a different branch, but the current implementation calls `query_versions_for_tx`, clones the entire cached transaction-version vector, sorts it, and then filters one row.

## W1 receipt

The anonymized W1 fixture seeds tasks, comments, and activity in one transaction, then deletes one task. Varying only unrelated activity rows exposes parent-transaction width:

- 900 activity rows: delete 1.15 ms
- 9,000 activity rows: delete 3.92 ms
- point preflight itself: ~0.23–0.29 ms and flat
- `commit_mergeable_in_schema`: 0.93 → 4.26 ms
- cached `query_versions_for_tx`: 0.29 → 1.61 ms (clone/sort of the whole seed transaction)

A prototype using the existing `(tx, table, row)` targeted lookup for branch-local parent validation produces:

- 900 activity rows: 0.64 ms
- 9,000 activity rows: 0.52 ms

That is ~7.6× faster at the large profile and removes the slope. Restore is already flat at ~0.43/0.52 ms.

## Proposed invariant-preserving change

For each explicit parent, query only both layers for the commit's `(table,row)` at the parent's `(time,node alias)`, then retain the existing physical-table and branch-key mismatch rejection. Do not weaken validation, trust claimed parent locality, or change pending-parent behavior.

## Evidence to add

- W1 wall-time scaling benchmark where unrelated rows widen the seed transaction.
- Black-box/public behavior test proving a parent from the same row but a different branch is still rejected, including a parent transaction containing unrelated rows.
- Existing transaction, branch, sync, and incremental-delivery canaries.

## Expected impact

Large batch-created rows should see constant-time parent validation on subsequent update/delete/restore. This is distinct from the parked canonical-large-value work and should benefit both client-local and backend authored writes.

aeplay:
Implementation refinement after adversarial correctness testing:

- A direct `(tx, table, row)` helper was **not** safe because the existing helper assumes the default branch. `version_parents_cannot_cross_branch_keys` caught that it would fail to discover/reject a parent under another branch. That prototype was discarded.
[...]

PR #2123, opened 2026-08-26
perf(jazz): scope parent validation by table

aeplay:
## Summary

Stacks on #2121. Fixes the first checkpoint of #2122.

### Before

A local mergeable update/delete validates each explicit parent by calling `query_versions_for_tx`. When the parent transaction is cached, that clones and sorts every version in the transaction before filtering for the one touched physical table and row. A realistic W1 seed transaction contains tasks, comments, and activity, so deleting one task became slower as unrelated activity rows grew.

### After

Parent validation asks for versions from the touched physical table only:

- cache hit: map the physical table across known schema aliases, reject unrelated versions by interned logical name before reading the schema tag, and clone only matching-table versions;
- cache miss: scan `by_tx` only in that physical content-history table plus shared deletion history, filtering the latter by physical table id.

The existing validation then checks the same row and rejects any parent found solely under a different branch.

## Governing invariants

- A causal parent for the same physical table/row under another branch remains invalid.
- Schema-renamed logical tables that map to the same physical table remain visible to validation.
- Deletion parents remain visible through shared deletion history.
- Missing/pending parents retain the existing behavior.

## Non-goals / residual

- This does not add a `(tx,row)` storage index. On a cache hit it still performs cheap interned-name comparisons across the transaction, so the algorithm is not strictly O(1) in total transaction width even though the W1 cross-table measurement is flat.
- Same-table transaction width is not yet bounded by the touched row.
- No subscription, query-result, fate, or persistence semantics change.

## Worked examples

Normal path: one transaction inserts 300 tasks, 1,200 comments, and 9,000 activity rows. A later delete of one task now filters the cached parent transaction by the task physical table before cloning records.

Branch edge: a task version authored under branch A is named as a parent of a version under branch B. Table-scoped lookup still discovers the branch-A version, and the existing `version_parents_cannot_cross_branch_keys` rejection fires.

Cache-miss edge: after tx-version cache eviction, the lookup reads the requested physical history table and shared deletion history through their transaction indices; it does not assume the default branch.

## Performance receipt (local MemoryStorage, W1)

| sibling activity rows in seed transaction | before | after |
|---:|---:|---:|
| 900 | 1.15 ms | 0.68 ms |
| 9,000 | 3.92 ms | 0.64 ms |

~6.1x faster at 9k; observed cross-table slope removed. Restore control remains flat (~0.43–0.52 ms).

The benchmark returns its consumed fixture from the timed Divan closure so whole-database destruction is excluded.

## Tests

- `dev/t version_parents_cannot_cross_branch_keys`
- all three `incremental_delivery_canary` tests
- `JAZZ_SEED_COUNT=30 dev/t --exact node::tests::harness::m3_maintained_one_shot_differential_oracle -- --ignored`
- `cargo clippy -p jazz --no-default-features --features testing,transport-compression-zstd --lib -- -D warnings`
- `cargo check -p jazz-example-w1-benchmark --bench reads_memory_walltime`
- pre-commit rustfmt, Clippy, and sensitive-data guard

### DIFF
--- a/crates/jazz/src/node/currency.rs
+++ b/crates/jazz/src/node/currency.rs
@@ -542,6 +542,78 @@ where
         Ok(versions)
     }
 
+    /// Return the versions authored by one transaction for one logical table.
+    ///
+    /// Unlike [`Self::query_versions_for_tx`], this deliberately visits only
+    /// the requested table's physical history sources. It still uses the
+    /// transaction index rather than a branch-local primary key so callers
+    /// validating a causal parent can observe versions from every branch.
+    pub(super) async fn query_versions_for_tx_table(
+        &mut self,
+        tx_id: TxId,
+        schema_version: SchemaVersionId,
+        table: &str,
+    ) -> Result<Vec<VersionRow>, Error> {
+        let requested_table_id = self.physical_table_id_for_schema(schema_version, table)?;
+        if let Some(cached) = self.query.tx_versions_cache.get(&tx_id) {
+            let table_aliases = self
+                .catalogue
+                .physical_mappings
+                .iter()
+                .flat_map(|(schema_version, mapping)| {
+                    let schema_alias = self.catalogue.schema_version_aliases.get(schema_version);
+                    mapping.tables.iter().filter_map(move |(table, mapping)| {
+                        (mapping.table_id == requested_table_id)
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Superseded by #2228, reconstructed directly on current `main` with the original commits preserved and the parent lookup tightened from physical-table scope to the exact physical row.

Issue #2125, opened 2026-08-26
Investigate large single-transaction byte-wire sync stalling before W1 resume setup


aeplay:
## Finding

While translating `realistic_phase1/r11_byte_wire_resume` to the W1 example workload, a single seed transaction containing roughly 8,000 W1 rows never became visible on the history-complete server (zero task rows), while 2,400- and 4,000-row transactions completed.

The harness:

- negotiates current sync payload, session, structured-error, and message-fragmentation features;
- repeatedly drives both semantic `Db::tick()` directions;
- explicitly routes both `PeerIoPump` auxiliary directions;
- observes all-or-nothing visibility (the same topology succeeds below the size boundary);
- receives no surfaced error from either tick.

This blocks using a 10k+ monolithic seed transaction as resume setup and may indicate a large logical-message/chunk handoff or progress/error-surfacing problem. It is distinct from resume latency and from #2030's local ingest cost.

## Next investigation

- Minimize/bisect by encoded transaction bytes and write count.
- Record wire fragment counts, auxiliary request/response counts, staged inbound state, and scheduler progress.
- Compare one 8k-write transaction with equivalent smaller transactions.
- Verify behavior through the production serving shell/binding pump before claiming a core protocol bug.

Parked for now so the resume thesis can proceed with a known-good 2.4k/4k matrix.

Issue #2124, opened 2026-08-26
Perf thesis: byte-wire resume should not clone a current result set the cursor already covers


aeplay:
## Thesis

A byte-wire reconnect with a valid membership cursor and one changed row should scale with the catch-up delta, not with the full subscribed result set.

## Existing benchmark gap

`realistic_phase1/r11_byte_wire_resume` is a correctness canary, not a resume timing measurement: every Criterion iteration opens writer/server/client databases, seeds all data, performs initial sync and subscription hydration, writes the disconnected update, reconnects, and resumes. It also uses a pre-fragmentation direct transport-driving assumption. Its reported duration cannot support a resume throughput claim.

## W1-derived probe

A prototype W1 Divan benchmark prepares the three-node byte-wire topology and disconnected one-task update in `with_inputs`, then times only reconnect + server catch-up + client application/materialization. Initial load and full snapshot are asserted outside the timer.

Local MemoryStorage receipt (one changed task):

| W1 rows | subscribed tasks | resume |
|---:|---:|---:|
| 2,400 | 300 | 15.05 ms |
| 4,000 | 500 | 20.91 ms |

A 1.67x larger result set produces 1.39x resume latency despite a fixed one-row delta.

## Grounded mechanism

`PeerState::rehydrate_query_for_subscription_from_maintained_subscription` currently calls `PeerSubscriptionState::member_result_set()`, which clones the complete `BTreeSet<ResultMemberEntry>`, and immediately collects it into another full `Vec`. Only after those O(result rows) allocations does it compare the known membership cursor to canonical settlement and clear the vector when no reset/snapshot resend is required.

The existing maintained view and downstream known state are already retained across `ResumeCursor`; this work is avoidable on the common cursor-covered path.

## Proposed change

- Borrow the canonical result-member set for duplicate-view/removal checks.
- Resolve authorization and membership-cursor coverage first.
- Materialize `result_member_adds` only for reset or post-cursor members that actually need delivery.
- Preserve authorization mismatch, partial-position, removal, program-fact, and relay behavior unchanged.

## Expected result

For an unchanged authorization scope and a cursor at/after canonical settlement, eliminate the full set clone/vector allocation. Expect the 300→500 task slope to become approximately flat and resume latency to improve materially (initial estimate 20–40%, pending flamegraph/instruction receipt). Payload bytes must remain delta-sized and byte-identical in meaning.

## Required evidence

[...]

PR #2126, opened 2026-08-26
perf(jazz): avoid covered resume member clones

aeplay:
## Summary

Stacks on #2123. Tracks #2124; setup-size follow-up is #2125.

This PR first translates the old `realistic_phase1/r11_byte_wire_resume` correctness scenario into an official W1 example-derived Divan/CodSpeed benchmark whose setup is outside the timed closure. It then removes an eager full result-member clone on cursor-covered resume.

## Before / after behavior

Before, resumed initial delivery cloned the complete retained `BTreeSet<ResultMemberEntry>` and collected a full vector before checking whether the receiver's membership cursor already covered canonical settlement. The vector was then cleared on the common covered path.

After, the retained set is borrowed for duplicate/removal decisions. Cursor coverage is resolved first; result members are cloned only for a reset or for members whose settlement position is newer than a partial cursor.

Wire behavior, result membership, settlement, authorization, and resume bytes are unchanged.

## Governing invariants

- Authorization mismatch still forces reset semantics and a complete visible result set.
- A partial cursor still receives exactly members newer than its position.
- Cursor-at/after-canonical settlement sends no redundant result-member additions.
- Flat removals, program facts, relay behavior, and maintained facts are unchanged.
- Resume setup asserts the full initial task count; the measured path asserts exactly one changed row and a resume payload smaller than the initial snapshot.

## Benchmark design

The old r11 benchmark times database open, all seed writes, initial sync, initial full snapshot, disconnected update, and resume in one Criterion closure. The new W1 benchmark prepares writer/server/client, initial subscription, cursor, and disconnected one-row update in `with_inputs`; only reconnect, catch-up, client application/materialization, and delta observation are timed.

It uses current wire feature negotiation and explicitly drives auxiliary pumps. Profiles are currently bounded to known-good 2.4k/4k row transactions because the 8k single-transaction setup exposed #2125.

## Local receipt

| W1 rows / subscribed tasks | benchmark-only baseline `7adb84d54` | optimized `a2e156768` |
|---|---:|---:|
| 2,400 / 300 | 15.05 ms | 14.70 ms |
| 4,000 / 500 | 20.91 ms | 19.39 ms |

The ~7% large-case gain disproves the initial hypothesis that cloning was dominant. It remains a valid allocation cleanup; hosted wall-time instruction/memory flamegraphs should identify the next dominant lever in maintained transition draining or payload reconstruction.

## Tests

- `db::tests::node_runtime::subscriber_connection_serves_current_rows_and_resumes_from_cursor`
- `db::tests::node_runtime::byte_wire_subscriber_connection_serves_current_rows_and_resumes_from_cursor`
- W1 benchmark compile and local wall-time run
- pre-commit Clippy, rustfmt, sensitive-data guard

### DIFF
--- a/crates/jazz/src/peer/publication.rs
+++ b/crates/jazz/src/peer/publication.rs
@@ -1437,7 +1437,7 @@ impl PeerState {
             .ok_or(Error::InvalidStoredValue(
                 "coverage group subscription is missing peer state",
             ))?;
-        let current_result_member_set = canonical_state.member_result_set();
+        let current_result_member_set = &canonical_state.result_member_set;
         let can_forward_flat_removals = client_link
             && ordinary_flat_row_duplicate_view(
                 shape,
@@ -1455,7 +1455,6 @@ impl PeerState {
         } else {
             Vec::new()
         };
-        let mut result_member_adds = current_result_member_set.into_iter().collect::<Vec<_>>();
         let tier = self
             .publication_states
             .get(&maintained_subscription)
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Superseded by #2231, reconstructed above #2228 on the current core with the benchmark and allocation cleanup preserved. The replacement explicitly tracks the missing maintained-publisher cursor-sensitivity receipt before readiness.

Issue #2127, opened 2026-08-26
Perf thesis: connected default-tier current reads must be bounded by current rows, not version depth


aeplay:
## Adopter receipt

A periodic-status workload holds row count fixed at 200 and performs 50 rounds of updating every row (10,000 updates). On the current new-core read stack (`20c3c19e1`), local persistent driver:

Connected to the matching server:

| history | per-update | local-tier 200-row list | default-tier 200-row list | store |
|---:|---:|---:|---:|---:|
| baseline | 0.41 ms | 9.8 ms | 0 ms | 504 KB |
| 2,000 updates | 1.83 ms | 9.8 ms | 4,822 ms | 9,020 KB |
| 10,000 updates | 1.92 ms | 10.7 ms | 7,011 ms | 39,740 KB |

Disconnected, the same current list degrades 9.7 → 41.4 ms and the store reaches 17,356 KB. Old core list reads remain ~2.5 ms and its store reaches 10,024 KB after 10k updates.

Source receipt: #2026 adopter comment following #2122 corroboration.

## Thesis

Reading the current 200-row result at Edge/default tier should be bounded by current rows plus the requested result, not by historical versions per row. Explicit Local being flat while connected default-tier reaches seconds isolates the dominant problem to settlement/authority-view/current-Global delivery rather than ordinary app-row decoding.

## Initial code hypotheses to discriminate

1. Edge/default one-shot hydration rebuilds a maintained authority view by replaying historical/global-change facts instead of seeking compact current winners.
2. Complete-transaction/view-scoped witness construction repeatedly walks or validates prior versions for each current member.
3. Receiver authority-view application or post-apply `refresh_subscriptions_in` rebuilds structures proportional to retained history.
4. Global-current compaction is correct in memory but persistent history/global-change tables are unbounded and leak into the read plan.

These may interact with #2124's client apply finding, but this issue is history-depth scaling at fixed current cardinality and should have its own benchmark/thesis.

## Benchmark plan

Create an anonymized example-derived periodic-status fixture:

- 200 current rows;
- history depths 0, 10, 50, and eventually 500 updates/row;
- explicit Local vs default Edge vs Global read/subscribe;
- connected and disconnected controls;
- fixed result digest/cardinality;
- phase attribution for server hydration, wire bytes, client apply, and local subscription refresh;
- persisted bytes by physical class.

[...]

PR #2128, opened 2026-08-26
bench: expose ahead-current history scaling

aeplay:
## Integration-ready benchmark checkpoint

This PR now contains only the benchmark that proves Local/Edge current reads grow with retained ahead-current candidate depth.

The earlier runtime experiment is deliberately removed after integration testing against #2108. Its decoded logical row/latest caches would reintroduce state that #2108 removes to make reopen cheaper. The two changes were mechanically mergeable but performance-incompatible.

## Thesis and receipt

A current-state read should scale with current rows, not retained Local/Edge candidates for those rows.

At one current row with 100 / 1k / 10k retained candidate versions:

- Local: 2.743 / 23.533 / 391.662 ms
- Edge: 5.428 / 49.033 / 566.348 ms
- storage attribution: exactly 100 / 1k / 10k ahead-current row reads

The deterministic smoke uses depth 3 at both tiers.

## Next implementation constraint

The fix must compose with #2108's compact encoded replay-idempotency index. It should use an encoded-key ordered candidate structure or bounded storage seek, rather than restoring decoded table/branch/row caches during open.

## Invariants

- Candidate versions remain available for rejection and concurrency fallback.
- Local and Edge visibility semantics remain unchanged.
- Global settled reads remain independent of superseded history.
- This checkpoint changes no runtime behavior.

Tracks #2127.


aeplay:
🤖 Changes requested for benchmark integration:

- **P1:** This receipt is not run by CodSpeed; the CodSpeed workflow currently runs example benchmark packages only. The PR’s hosted report is therefore unrelated to this benchmark.
- **P1:** Claimed ahead-current reads are emitted as generic `core_storage_read_other_reads`; no dedicated metric or assertion distinguishes them from unrelated storage work.
- **P2:** The smoke asserts only one visible row, not that the newest seeded candidate wins or that retained candidate depth/read count equals the configured depth.
- **P2:** Legacy CI leaves `JAZZ_HOT_PENDING_DEPTHS` at `100,1000,10000` for both Local and Edge, silently adding 22,200 persisted commits to an existing benchmark run.

Please put this receipt in the W1 example/CodSpeed path, add explicit ahead-current read attribution, and make the smoke assert winner identity plus exact depth/read behavior with bounded CI defaults.

Issue #2129, opened 2026-08-26
Perf thesis: same-shape subscription bindings must share source witness state


aeplay:
## Problem

Maintained subscriptions currently retain source-wide version and replacement
witnesses independently for every binding of the same query shape. For a
multi-tenant owner query, memory and attach work therefore grow as:

```text
source rows x active tenant bindings
```

This is unnecessary for simple root-table queries whose result payload can be
loaded exactly when a row enters the delivered result.

## Reproduction

`crates/jazz/benches/route_subscription_curve.rs` on PR #2361 head
`9ad56bd8f`: 2,000 documents, one parameterized `team = $team` shape,
`ORDER BY updated_at DESC LIMIT 100`, and 100 retained bindings.

| 100 bindings | PR #2361 baseline | Storage-backed prototype | Improvement |
| --- | ---: | ---: | ---: |
| Total attach | 12.103 s | 321.6 ms | 37.6x |
| Median attach | 120.40 ms | 2.71 ms | 44.4x |
| Maintained/control heap | 683.39 MB | 447.42 KB | 1,527x |
| Version identities | 200,000 | 0 | eliminated |
| Replacement entries | 200,000 | 0 | eliminated |
| Matching write | 17.45 ms | 3.07 ms | 5.7x |
| Unrelated write | 14.31 ms | 2.27 ms | 6.3x |

The baseline footprint is exact multiplication: each binding stores witnesses
for all 2,000 source rows even though only 100 rows, or often one row, are in
that binding's result.

## Proven implementation direction

A local prototype stacked on PR #2361 adds a bounded
`StorageBackedMaintainedView` output mode:

- retain result membership per binding;
- do not request source-wide version/replacement witness terminals;
- load the exact immutable content version when a row enters the result;
- load only the current deletion-register winner needed for delete/restore
  delivery;
- on removal, never load a newer content winner, so policy revocation cannot
  disclose newly unauthorized content;
- keep the existing self-contained witness path for unsupported query shapes.

The initial eligibility proof is deliberately narrow: default current read,
unbranched root table without references, no joins/includes/reachable/inherits,
no array subqueries or aggregate. Direct predicate-only read policies remain
eligible, which covers the common `ownerId == session.user` query.

Verification passed all 68 maintained-view focused tests, owner-policy
revocation, ordered offset/limit churn, delete/restore delivery, the three
incremental-delivery canaries, benchmark smoke, and the maintained-vs-one-shot
differential oracle at 30 seeds.

[...]

PR #2135, opened 2026-08-26
perf(jazz): release flat subscription app-row state

### LINKED ISSUE #2130: Perf harness: maintained rehydrate scaling is blocked by stale flat-query structured collector state
## Observation

The maintained rehydrate scaling benchmark is born-red on current `main` at its first 100-row rung. After priming a flat filtered subscription with 50 active rows and changing one inactive row to active:

- maintained and freshly rehydrated public result sets both contain 51 rows;
- the maintained wire update correctly contains one add;
- all result/version/replacement index cardinalities match;
- but the maintained `structured_app_rows` collector still contains 50 roots, while fresh rehydrate contains 51;
- retained heap differs by exactly the missing encoded collector root (240 bytes at this fixture size).

The public metrics currently hide `structured_app_rows(_bytes)` while including its bytes in `total_heap_bytes`, so the benchmark panic initially appeared as an unexplained 240-byte total mismatch.

## Code-grounded thesis

`apply_local_maintained_view_transitions_inner` only consumes `structured_app_row_changes` when `array_subqueries` makes `structured_output` true. A flat query publishes its correct membership/materialized row through the ordinary result path, so this stale collector does not alter its current wire result. However, initial hydration still populates the structured collector and exact maintained-vs-rehydrate footprint equality therefore fails.

We should establish whether the structured terminal is intentionally irrelevant for flat queries:

1. If yes, do not retain/populate it for flat queries and make the benchmark compare every semantically live footprint component. This removes dead memory and unblocks the scaling receipt.
2. If no (for example reusable policy-graph state consumes it), repair incremental structured terminal delivery and add a black-box maintained-vs-rehydrate canary before measuring.

## Success criteria

- The benchmark reaches 100 / 1k / 10k and emits attributed timing/read/heap receipts.
- Maintained and fresh rehydrate have equivalent semantically live state.
- Flat subscriptions do not retain an unused table-sized collector.
- A public-API regression test covers whichever invariant is selected.

No adopter test yet: this is a harness/state diagnosis, not a candidate performance fix.

aeplay:
Closes #2130 and advances #2147.

## Before

Opening every maintained peer query populated and retained Jazz's recursive `app_rows` collector, including flat queries. Flat peer subscriptions do not read that Jazz-side duplicate after the reset: subsequent row deltas are owned by result membership and version/replacement witnesses. This retained a second encoded row representation proportional to result size.

Separately, Jazz's flat snapshot bridge treated a content-version replacement as a potentially real removal because its transition contains remove+add for the same stable occurrence. It scanned every current root before discovering that the removal must be coalesced into an update.

## After

After opening a flat peer-maintained query, Jazz clears its duplicate app-row collector and descriptor. The Groove terminal remains subscribed and public terminal operations/layouts are unchanged, including for unordered raw NAPI subscriptions.

Before scanning roots for removals, the snapshot bridge asks whether any removed occurrence is not re-added in the same transition. Pure version replacements skip the scan and remain one-row updates.

The maintained-view footprint reports structured app-row cardinality and bytes separately.

## Governing invariants

- A flat reset is formed before Jazz releases its duplicate collector.
- Groove's terminal remains live: unordered raw subscribers still receive terminal inserts/layouts.
- Remove+add of one stable occurrence is published as an update, not a transient removal.
- Ordered roots and nested arrays retain terminal update/move/structural semantics.
- Maintained and freshly rehydrated subscriptions retain equivalent live index state.

## Worked examples

- Flat peer `documents.where(status = active)`: opening returns all active documents, then Jazz's retained structured-app-row cardinality is zero; later membership/version witnesses publish rows normally.
- One content-version replacement with the same occurrence skips the whole-root removal walk.
- Raw NAPI insert into an unordered flat query still emits one terminal `Insert` plus its registered layout.
- An actual membership removal still executes root removal.
- Array subqueries retain recursive app-row state.

## Non-goals

- This does not retire Groove's public terminal or fix its result-sized update work; #2150 tracks that bottleneck.
- This does not reduce version/replacement witness indexes tracked in #2129.
- This does not change cold hydration or wire protocol.
- Point-update preimage lookup is separately fixed by #2146.

## Receipt

`maintained_rehydrate_scaling` on local RocksDB:

| source rows | maintained delta | reads | cold rehydrate | reads |
|---:|---:|---:|---:|---:|
| 100 | 0.222 ms | 2 | 30.8 ms | 252 |
| 1,000 | 0.364 ms | 2 | 310 ms | 2,502 |
| 10,000 | 4.69 ms | 2 | 3.43 s | 25,002 |

The coalesced-removal guard reduces the Jazz bridge portion of a 10k same-occurrence replacement from ~49 ms to sub-millisecond. The separate Groove terminal remains dominant until #2150.

## Checks

- `dev/t db::tests::subscriptions::structured::flat_subscription_updates_with_nullable_sort_payload`
- `dev/t db::tests::subscriptions::structured::array_subquery_subscription_reflects_child_mutations_and_parent_removal`
- `dev/t peer::tests::maintained_subscription_view_hit_metrics_and_footprint_update`
- raw NAPI terminal-layout TS regression rerun in trusted CI
- pre-commit clippy/rustfmt/sensitive-data guard

Tooling friction: machine-readable phase artifacts would avoid recompiling profiled benches to recover attribution after a long tool yield.

### DIFF
--- a/crates/jazz/src/node/maintained_subscription_view.rs
+++ b/crates/jazz/src/node/maintained_subscription_view.rs
@@ -48,7 +48,7 @@ struct VersionDecodePlan {
     authored_columns_idx: usize,
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub(crate) struct MaintainedSubscriptionView {
     result_weights: BTreeMap<ResultMemberEntry, i64>,
     result_payloads: BTreeMap<ResultMemberEntry, ResultMemberPayloadEntry>,
@@ -57,8 +57,26 @@ pub(crate) struct MaintainedSubscriptionView {
     /// touching the rendered trees for other roots.
     structured_app_rows: BTreeMap<RowUuid, BTreeMap<Vec<u8>, i64>>,
     structured_app_row_descriptor: Option<RecordDescriptor>,
+    /// Whether this maintained subscription retains the recursive app-row
+    /// collector. Flat unordered subscriptions release it after their reset;
+    /// subsequent terminal deltas must not rebuild the duplicate state.
+    retains_structured_app_rows: bool,
     versions: WeightedVersionIndex,
     replacements: ReplacementIndex,
+}
+
+impl Default for MaintainedSubscriptionView {
+    fn default() -> Self {
+        Self {
+            result_weights: BTreeMap::new(),
+            result_payloads: BTreeMap::new(),
+            structured_app_rows: BTreeMap::new(),
+            structured_app_row_descriptor: None,
+            retains_structured_app_rows: true,
+            versions: WeightedVersionIndex::default(),
+            replacements: ReplacementIndex::default(),
+        }
+    }
 }
 
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
@@ -388,7 +406,9 @@ impl MaintainedSubscriptionView {
                     }
                 }
                 NetEvent::StructuredAppRow(root, record) => {
-                    self.apply_structured_app_row_delta(root, record, weight);
+                    if self.retains_structured_app_rows {
+                        self.apply_structured_app_row_delta(root, record, weight);
+                    }
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


aeplay:

Issue #2136, opened 2026-08-26
Perf thesis: caught-up fast resume must not rehydrate the full maintained view


aeplay:
## Thesis

A caught-up `FastCurrentMembership` resume currently has constant wire and storage I/O but still pays CPU proportional to the entire current view because `rehydrate_query_maintained_subscription_view` opens and hydrates a fresh Groove maintained graph before applying the cursor to publication output.

For a reusable query/policy graph, attaching a caught-up subscriber should scale with the missed delta (zero here), not current table cardinality. The likely architectural lever is to share already-hydrated binding/source state and retain only peer-specific membership/publication state, rather than constructing one fully hydrated graph per reconnect.

## Local RocksDB receipt (`known_state_scaling`, current main)

Caught-up fast declaration, no changes since the declared settlement position:

| current rows | wall | storage reads/ranges | emitted members/bundles | total exchange |
|---:|---:|---:|---:|---:|
| 100 | 15.1 ms | 0 / 0 | 0 / 0 | 76 B |
| 1,000 | 149 ms | 0 / 0 | 0 / 0 | 76 B |
| 10,000 | 1.52 s | 0 / 0 | 0 / 0 | 76 B |

The exact-version fallback is nearly identical in CPU (15.6 ms / 158 ms / 1.58 s), despite doing extra point reads and emitting full membership. This isolates graph hydration/evaluation from serialization and storage bundle loading.

## Code mechanics

`PeerState::rehydrate_query_maintained_subscription_view`:

1. calls `open_seeded_maintained_subscription_view`, fully opening/hydrating the graph;
2. only afterward reads `downstream_known_states` and computes `known_membership_position`;
3. clears caught-up membership additions and reset semantics;
4. builds an empty 68-byte update.

The cursor therefore suppresses publication, not computation. A naive early return is insufficient because the subscriber still needs a correctly seeded maintained graph for future deltas. The repair must preserve the handoff invariant between the declared position and live incremental delivery.

## Candidate design

- Own hydrated query/policy graph state at a reusable binding/read-view/identity scope.
- Attach per-peer consumers at a declared settlement frontier.
- Replay only deltas after that frontier, then transition atomically to live delivery.
- Keep peer-specific authorization generation and result membership isolated; never trust a client cursor across a changed authorization generation.

This overlaps the ownership mechanism in #2129 (same-shape subscriptions currently duplicate source witness state), but tracks a distinct externally visible latency claim.

## Success criteria

[...]

PR #2137, opened 2026-08-26
bench(jazz): measure caught-up fast resume

aeplay:
Tracks #2136.

## Before

`known_state_scaling` measured only `ExactVersionSet` through `current_rows_update`. That path proves payload-body suppression and repair coverage, but it still performs an authoritative membership publication. It did not measure the normal caught-up `FastCurrentMembership` resume that can suppress both payload and membership.

## After

The benchmark retains the exact-version coverage curve and adds an explicit caught-up fast-resume lane through the usage-site rehydrate API. Each receipt labels its declaration kind and attributes wall time, storage reads/ranges, membership/bundles, declaration bytes, update bytes, and total exchange.

## Governing invariants

- Exact declarations continue to emit complete membership and only omit payload bodies explicitly held by the receiver.
- A caught-up fast cursor emits no membership changes or bundles and does not reset the receiver.
- The fast declaration position is taken from the fixture's observed settlement watermark, not reconstructed from authored timestamps.

## Worked examples

At 1,000 current rows and 100% known state:

- exact fallback: 1,000 membership additions, 0 bundles, 1,000 point reads, 149 KB exchange;
- caught-up fast resume: 0 additions, 0 bundles, 0 reads/ranges, 76 B exchange.

Despite that, fast resume is 149 ms because the maintained graph is fully hydrated before the cursor suppresses publication. The 100 / 1k / 10k curve is 15.1 ms / 149 ms / 1.52 s, establishing #2136.

## Non-goals

This PR changes no runtime behavior and does not attempt the shared hydrated-graph design needed to make warm resume scale with missed deltas.

## Checks

- `dev/gates/benchmark-smoke.sh jazz known_state_scaling`
- local 100 / 1k / 10k RocksDB timing receipts
- pre-commit hooks

Tooling friction: a shared release artifact cache across worktrees would avoid rebuilding RocksDB for each independent performance thesis.

### DIFF
--- a/crates/jazz/benches/known_state_scaling.rs
+++ b/crates/jazz/benches/known_state_scaling.rs
@@ -11,8 +11,10 @@ use jazz::ids::{NodeUuid, RowUuid};
 use jazz::node::{MergeableCommit, NodeState, SKEW_TOLERANCE_MS};
 use jazz::peer::PeerState;
 use jazz::protocol::{
-    KnownStateDeclaration, RowVersionRef, SubscriptionKey, SyncMessage, expand_version_carriers,
+    KnownStateCompleteness, KnownStateDeclaration, RegisterShapeOptions, RowVersionRef,
+    SubscriptionKey, SyncMessage, expand_version_carriers,
 };
+use jazz::query::Query;
 use jazz::schema::JazzSchema;
 use jazz::tools::{ColumnType, SchemaBuilder, TableSchemaBuilder};
 use jazz::tx::{Fate, TxId};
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Superseded by #2168, which preserves the caught-up fast-resume thesis while moving it into BandChat’s self-contained Divan/CodSpeed benchmark suite. The replacement derives its cursor from a real settled publication, asserts zero replay output at 100/1,000/10,000-row histories, and avoids extending the legacy `crates/jazz/benches` path.

PR #2138, opened 2026-08-26
Reject oversized bytes in nested payloads

joeinnes:
## Summary
- validate byte sizes recursively through enum payloads, arrays and nested rows before encoding
- return the existing `ByteaTooLarge` error for oversized nested values instead of reaching a validated-only panic path
- preserve the existing encoding and error behaviour for valid values

Before, only top-level byte values were checked. After, the same size limit applies at every supported nesting level.

## Testing
- `cargo test -p jazz encode_row_rejects_oversized_bytea_nested_in_enum_payload` — passed

### DIFF
--- a/.changeset/validate-nested-enum-payloads.md
+++ b/.changeset/validate-nested-enum-payloads.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reject oversized byte values nested inside payload enums with the normal encoding error instead of panicking during encoding.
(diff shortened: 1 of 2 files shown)


joeinnes:
Closing as superseded by e6cd2fbe0 on main. That commit now performs recursive size validation for arrays, rows, and enum payloads, and its regression test covers oversized bytea values in direct enum payloads, nested rows, arrays, and the direct encoder. Keeping this branch would duplicate the match arm and fail clippy as unreachable code.

PR #2143, opened 2026-08-26
Preserve parent-aware deltas during RocksDB compaction

joeinnes:
## Summary
- classify current-winner and conditional deltas as dependent on the durable base value
- defer those operands to RocksDB’s full merge instead of reducing them during partial merge
- preserve operand order so a lower-ordered child can still replace its recorded parent

This keeps more base-dependent operands until full merge, trading some eager compaction for correct parent-aware semantics.

## Testing
- `cargo test -p jazz-storage-rocksdb compaction_preserves_parent_sensitive_delta_order` — passed

### DIFF
--- a/.changeset/steady-stones-merge.md
+++ b/.changeset/steady-stones-merge.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Preserve parent-aware current-winner semantics across RocksDB compaction.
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Independent review found no blocker in this PR. Pre-existing RocksDB/MemoryStorage malformed-delta admission parity is tracked separately in #2159.

aeplay:
🤖 Superseded by #2162, which removes semantic storage deltas and RocksDB merge operators entirely while retaining only generic atomic `put_if_absent` and `compare_and_delete` primitives across backends.

PR #2149, opened 2026-08-26
feat: replace `Db.subscribeAll` with `Db.subscribe`

nrainhart:
## Description

This PR does a cleanup pass around subscriptions:

- Replaces `Db.subscribeAll` with `Db.subscribe`, which exposes only the full query result instead of the delta
- `Db.subscribeDelta` is kept as an internal API used by UI bindings
	- this lets us iterate on the delta format if needed, we can expose this later if needed for advance end-user use cases
- Centralizes subscription delta processing in Rust
- Removes a standalone `subscriptionAll` function we were exporting as part of the public API
- Removes a lot of unused machinery that we kept in internal subscription delta representations
- Removes redundant delta re-encoding and decoding we were doing on the JS side

These are not all of the changes I want to make on subscriptions (in particular, I think we can still simplify the IRs for subscription deltas), but want to get a first round of changes out.


aeplay (changes requested):
🤖 Changes requested from the exact-head adversarial review:

- **P1 — complete the public API migration.** After rebuilding `jazz-tools` from this branch, `pnpm --filter ts-localfirst build` fails because the shipped starter still imports `SubscriptionDelta`, declares `subscribeAll`, and calls it. Stale public uses also remain in the BetterAuth/hybrid starters, docs quickstart, realistic benchmark, local-first README, and `SPEC/INVARIANTS.md`. Please migrate every public consumer to `Db.subscribe(query, rows => ...)` and update the invariant pointer.
- **P2 — add release intent.** This removes/adds public `jazz-tools` API but has no changeset.

The focused subscription suites and planted callback-result regression receipt otherwise passed.

aeplay:
🤖 Additional repaired-head blocker: **P0 — the committed native artifact expectations are stale after the binding changes.** A fresh core/runtime build produces different WASM and NAPI fingerprints, so browser subscription and shared-binding tests abort before executing with a Jazz WASM artifact ABI mismatch. Starter bundles compile but would reject their bundled runtime at initialization.

Please regenerate/stage both committed fingerprint expectation files after the final merge/conflict resolution, then prove a real browser subscription/shared-binding runtime receipt against those fresh artifacts. The public export migration, starter compilation, reducer tests, and fixed-group changeset otherwise passed the independent review.

aeplay (approved):
🤖 Approved by Anselm after the final exact-head review and green CI. The requested public-consumer migration, changeset, artifact refresh, and terminating native-queue fixture are all present.

PR #2151, opened 2026-08-26
perf(groove): maintain ordered TopBy windows incrementally

### LINKED ISSUE #2150: Perf thesis: required subscription terminals must be delta-sized
## Thesis

After initial hydration, an ordered flat subscription should apply a one-row payload/sort-key update in time proportional to the changed occurrence and the ordering structure (ideally O(log n)), not proportional to the full result cardinality.

## Current receipt

On current main plus the policy-free point-update fix and unordered-terminal work from #2146/#2135, changing one row's payload and `created_at` under `ORDER BY created_at ASC` measures on local MemoryStorage:

| result rows | update + delivered terminal delta |
|---:|---:|
| 100 | 5.22 ms |
| 1,000 | 41.26 ms |
| 10,000 | projected ~400-500 ms from the already-attributed terminal curve |

The same unordered subscription is ~1-1.6 ms through 10k after #2135. Ordering therefore keeps the exact whole-terminal amplification that unordered subscriptions can now retire.

## Code-grounded mechanism

`Db::subscribe` marks ordered roots as terminal-backed because a sort-key mutation can require `Move`, and payload-only updates must retain terminal layout. The maintained query consequently keeps `JAZZ_APP_ROWS_SINK`. A one-row physical publication reaches Groove as only 6 table-delta records / ~40 processed operator records, but the app-row terminal values are result-sized; `tick_resident_staged` previously measured ~404 ms at 10k. Retiring that sink makes Groove ~0.8 ms, but is incorrect for ordered queries: existing nullable-sort coverage loses its required terminal `Update`.

This is value-size amplification across a small operator delta, not an evaluator table scan.

## Candidate directions

1. Split ordered root patch ownership from structured whole-row assembly: retain a lightweight identity/order/payload terminal for flat ordered queries, reserving recursive app-row assembly for array subqueries.
2. Teach the ordered terminal to emit `Update`/`Move` directly from changed root occurrence plus maintained `TopBy` positions, without re-encoding every root.
3. If Groove's terminal node already owns a positional index, expose its before/after occurrence position and row patch as the Jazz bridge contract rather than materializing a result-sized record.

## Proof target

At 100/1k/10k ordered roots after hydration:

- payload-only update emits `Update`, no `Move`;
- sort-key update emits the correct minimal `Move`/replacement sequence;
- latency at 10k is <=5x the 100-row rung;
- nested arrays, nullable sort keys, ascending/descending ties, limits/offsets, and authority resets remain correct;
- Groove processed records and terminal encoded bytes are delta-bounded.

## Benchmark plan

Extend `observer_write_path` with separate unordered-flat and ordered-flat variants at 100/1k/10k. Keep nested/structured scaling separate because it has different output-size semantics.

## Adopter guidance

DO NOT TEST yet. #2135 improves unordered flat subscriptions; ordered dashboards are intentionally excluded until this thesis has an implementation checkpoint.

aeplay:
🤖 This makes sorted query windows update incrementally instead of re-sorting every matching row after each change.

For example, consider a chat query that shows the newest 50 messages out of a 10,000-message room. If one visible message receives a reaction but its timestamp does not change, Jazz currently copies and sorts all 10,000 matching messages twice. After this change, it updates the two affected internal entries and publishes the changed message without sorting unrelated rows.

`TopBy` is Groove's internal operator for `orderBy` queries with an optional offset and limit. This remains its single implementation path; the PR changes the state it maintains, not query semantics.

## Before

Every sorted-window update cloned and sorted the complete matching group before and after applying a delta. Payload-only replacements therefore scaled with retained group size even when membership, complete order key, and duplicate count were unchanged.

On the combined adopter-performance stack, an ordered payload update measured:

| shape | 100 rows | 1,000 rows | 10,000 rows |
|---|---:|---:|---:|
| finite `LIMIT 50`, before | 6.57 ms | 41.50 ms | 403.56 ms |
| unbounded stable-key, after | 1.35 ms | 1.43 ms | 2.02 ms |

## After

Groove retains each group's records in sorted order. An update mutates that ordered state and derives only the affected before/after window rather than cloning and sorting the complete group.

The same representation handles bounded and unbounded queries, payload replacements, insertions, removals, and order-key changes.

## Correctness boundaries

- Initial and historical reads still construct the same complete sorted result.
- Full sort keys, including tie-breaking fields, determine ordering.
- Duplicate rows retain their count rather than being collapsed.
- Offset/limit boundaries emit the same insert, remove, move, and update effects as before.
- Empty groups release their retained state.
- Query semantics and public wire layouts are unchanged.

Boundary example: if a timestamp change moves a message across the 50-message window boundary, the displaced message leaves and the next eligible message enters. That takes the full ordered-state path; it does not use the stable-order shortcut.

## Benchmarks and tests

- Extends `observer_write_path` with stable ordered payload updates at 100/1,000/10,000 rows.
- Covers unbounded replacement and bounded in-window/outside-window replacement through Groove's public database API.
- `dev/gates/benchmark-smoke.sh jazz observer_write_path` passes.
- Full `cargo test -p groove`: 538 passed, 2 ignored.

This supersedes #2153 and closes #2150 and #2152.

### DIFF
--- a/crates/groove/src/db/tests/queries/windows.rs
+++ b/crates/groove/src/db/tests/queries/windows.rs
@@ -235,6 +235,67 @@ async fn top_by_hydrates_limit_two() {
 }
 
 #[futures_test::test]
+async fn unbounded_top_by_propagates_payload_replacement_with_stable_order_key() {
+    let storage = MemoryStorage::new(&["history", "rows", "blockers"]);
+    let mut database = Database::new(history_schema(), storage).await.unwrap();
+
+    let mut batch = database.open_batch();
+    batch.insert("history", history_values(1, 10, 1, "before"));
+    batch.insert("history", history_values(1, 20, 1, "second"));
+    database.commit_batch(batch).await.unwrap();
+
+    let subscription = database
+        .subscribe_one_sink(history_top_by_stamp_asc_unbounded())
+        .await
+        .unwrap();
+    let _initial = subscription.recv().unwrap();
+
+    let mut batch = database.open_batch();
+    batch.update("history", history_values(1, 10, 1, "after"));
+    database.commit_batch(batch).await.unwrap();
+    assert_eq!(
+        subscription.recv().unwrap().to_values().unwrap(),
+        [
+            (history_values(1, 10, 1, "before"), -1),
+            (history_values(1, 10, 1, "after"), 1),
+        ]
+    );
+}
+
+#[futures_test::test]
+async fn finite_top_by_applies_stable_payload_replacement_only_inside_window() {
+    let storage = MemoryStorage::new(&["history", "rows", "blockers"]);
+    let mut database = Database::new(history_schema(), storage).await.unwrap();
+
+    let mut batch = database.open_batch();
+    batch.insert("history", history_values(1, 10, 1, "inside"));
+    batch.insert("history", history_values(1, 20, 1, "outside"));
+    database.commit_batch(batch).await.unwrap();
+
+    let subscription = database
+        .subscribe_one_sink(history_top_by_stamp_asc(1))
+        .await
+        .unwrap();
+    let _initial = subscription.recv().unwrap();
+
+    let mut batch = database.open_batch();
+    batch.update("history", history_values(1, 10, 1, "inside-after"));
+    database.commit_batch(batch).await.unwrap();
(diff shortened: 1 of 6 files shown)


aeplay:

PR #2153, opened 2026-08-26
perf: keep finite TopBy payload updates window-sized

### LINKED ISSUE #2152: Perf thesis: finite TopBy payload updates must be window-sized, not group-sized
## Thesis

After hydration, a payload-only replacement inside or outside a finite ordered window should cost O(delta log window), not O(group rows log group rows), when group/order/tie fields are unchanged.

## Receipt

On the contract-preserving stack of #2135 + #2146 + #2151, local MemoryStorage, one payload update to the first row of `ORDER BY created_at ASC LIMIT 50`:

| table rows | latency | throughput |
|---:|---:|---:|
| 100 | 6.57 ms | 152/s |
| 1,000 | 41.50 ms | 24.1/s |
| 10,000 | 403.56 ms | 2.48/s |

The matching unbounded stable-sort-key subscription after #2151 is 1.35/1.43/2.02 ms. The finite 10k path is therefore ~200x slower despite exposing only 50 rows.

## Code-grounded mechanism

`TickEvaluator::update_top_by` excludes finite limits from #2151's balanced sort-key proof. For every touched group it then:

1. calls `ArrangementState::records_for_key`, cloning every positive record in the group;
2. calls `top_by_window_from_records` for the complete after state;
3. reconstructs the complete before state with `top_by_window_before_from_deltas`;
4. computes sort keys and sorts both complete groups;
5. only then applies offset/limit and diffs the two <=50-row windows.

The arrangement is keyed by group only; its bucket is ordered by complete encoded record bytes, not the `TopBy` sort key, so it cannot seek the boundary or answer selected membership cheaply.

## Candidate first step

Retain each finite `TopBy` node's selected before/after window per group as copy-on-write operator state.

For a delta batch whose multiplicities balance by `(group, complete sort key including ties)`:

- if the removed occurrence is in the retained window, replace it in place and emit the canonical remove/add pair;
- if it is outside, update the base arrangement but emit nothing;
- root positions are unchanged because the complete sort key is unchanged.

Any insert/remove, group change, order/tie change, missing state, finite multiplicity ambiguity, or hydration/replace tick falls back to the current full reconstruction. A later ordered arrangement can address boundary-changing updates.

## Proof target

At 100/1k/10k group rows with limit 50:

- stable-key update inside window emits one Update and no Move;
- stable-key update outside window emits no terminal event;
- sort-key update crossing the boundary retains exact Insert/Remove/Move behavior;
- nullable keys, ties, offset, multiplicity, rollback/snapshot isolation, reusable graphs, and multiple groups remain correct;
- 10k inside-window latency is <=5x the 100-row rung.

## Benchmark

Promote the scout as `observer_write_path/update_finite_ordered_payload/limit_50_stable_sort_key/{100,1000,10000}` in the implementation PR.

## Adopter guidance

DO NOT TEST this thesis yet. The combined checkpoint on #2026 improves unbounded stable-key subscriptions; finite LIMIT windows remain on the full-group path measured above.

aeplay:
## WIP / stacked benchmark checkpoint

This draft starts with the official benchmark for #2152. It is intentionally stacked on the contract-preserving #2135 + #2146 + #2151 adopter checkpoint while the retained finite-window implementation is developed.

## Before

A payload-only replacement inside `ORDER BY created_at ASC LIMIT 50` clones and sorts the complete group twice:

| group rows | latency | throughput |
|---:|---:|---:|
| 100 | 6.57 ms | 152/s |
| 1,000 | 41.50 ms | 24.1/s |
| 10,000 | 403.56 ms | 2.48/s |

The matching unbounded stable-key case is 1.35/1.43/2.02 ms.

## Intended after

Retain the selected finite window by node/group. When incoming multiplicities balance by complete group+sort key, replace a selected payload in place or suppress an outside-window replacement. Arrangement state still advances.

## Invariants / non-goals

- Stable group/order/tie fields only.
- Inserts, removes, sort/tie changes, hydration, missing state, and ambiguous multiplicity fall back.
- Boundary-changing updates are unchanged in this first step; an ordered arrangement is a later architectural lever.
- Observable Update/Move/Insert/Remove and raw terminal layouts remain unchanged.

The benchmark-only first commit compiles through `dev/gates/benchmark-smoke.sh jazz observer_write_path`.

Closes #2152.

### DIFF
--- a/crates/groove/src/db/tests/queries/windows.rs
+++ b/crates/groove/src/db/tests/queries/windows.rs
@@ -235,6 +235,34 @@ async fn top_by_hydrates_limit_two() {
 }
 
 #[futures_test::test]
+async fn unbounded_top_by_propagates_payload_replacement_with_stable_order_key() {
+    let storage = MemoryStorage::new(&["history", "rows", "blockers"]);
+    let mut database = Database::new(history_schema(), storage).await.unwrap();
+
+    let mut batch = database.open_batch();
+    batch.insert("history", history_values(1, 10, 1, "before"));
+    batch.insert("history", history_values(1, 20, 1, "second"));
+    database.commit_batch(batch).await.unwrap();
+
+    let subscription = database
+        .subscribe_one_sink(history_top_by_stamp_asc_unbounded())
+        .await
+        .unwrap();
+    let _initial = subscription.recv().unwrap();
+
+    let mut batch = database.open_batch();
+    batch.update("history", history_values(1, 10, 1, "after"));
+    database.commit_batch(batch).await.unwrap();
+    assert_eq!(
+        subscription.recv().unwrap().to_values().unwrap(),
+        [
+            (history_values(1, 10, 1, "before"), -1),
+            (history_values(1, 10, 1, "after"), 1),
+        ]
+    );
+}
+
+#[futures_test::test]
 async fn top_by_finite_zero_stays_empty() {
     let storage = MemoryStorage::new(&["history", "rows", "blockers"]);
     let mut database = Database::new(history_schema(), storage).await.unwrap();
--- a/crates/groove/src/db/tests/support.rs
+++ b/crates/groove/src/db/tests/support.rs
@@ -346,6 +346,17 @@ pub(super) fn history_top_by_stamp_asc(limit: u64) -> GraphBuilder {
     )
 }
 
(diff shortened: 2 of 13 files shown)


aeplay:
Superseded by #2151. The canonical PR now contains the persistent ordered-state implementation for both finite and unbounded TopBy windows, plus the finite-window benchmark and black-box regressions. Closing this WIP so there is one implementation and review surface.

Issue #2166, opened 2026-08-27
Route React Native clients through the native relay


aeplay:
🤖 Native relay routing implementation completed in the merged RN stack through #2552.

The capability-gated JSI foreground routes through the shared native adapter into a scope-owned Rust relay. SQLite remains native-owned; platform adapters do not implement a second database/query engine. Android and iOS expose private-session admission and preserve ABI, lifecycle and scope boundaries. Unsupported modes/transaction-read cases fail explicitly.

This closes the routing implementation. It does not claim complete installed-device Edge/Core acceptance: #2291 retains the iOS private-session topology and run-bound upstream-observation gap. Broader device scenarios remain under #1974; simultaneous physical foreground roots are explicitly deferred to #2532 in alpha.55.

Issue #1974, opened 2026-08-25
Verify the native relay in real Android and iOS device scenarios


aeplay:
🤖 Track remaining React Native device acceptance for alpha.54.

## Implemented and exercised

Installed Android/iOS apps exercise local writes/subscriptions, fresh foreground readback, process restart/readback, scope isolation and revocation. The one-foreground-root Edge/Core topology and direct Core observation are complete; #2291 and #2567 are closed.

Merged #2604 adds stronger reconnect acceptance: preserve the original RN client and subscription, stop and restart the real Edge endpoint, then require both that subscription and a local read to observe a nonce-bound marker authored by a separate Core writer after recovery. This prevents a local echo from satisfying the proof. Installed iOS and Android jobs passed in run33999303476 on functional commit c9c3f38528c544846e4e9cb83e08c4c22b1359ac; the final additional commit only formats harness scripts.

Final preview03940590 also passed packaged native consumer builds for both platforms in run34000111613. These packaged consumers use Expo-generated native hosts; successful linking alone is not device runtime acceptance.

## Additional acceptance still not demonstrated

- bounded transport/backpressure scenarios;
- corrupted/foreign SQLite-store failure in the installed device journey;
- a separately verified vanilla React Native host, in addition to the Expo-generated hosts already exercised.

The coordinated external-package app-building session will add realistic Android usage, with a separate user-run iOS simulator task when final preview packages are ready. Prefer vanilla hosts where practical and record actual coverage. Do not report unimplemented scenarios as passing or silently move their release disposition out of alpha.54.

The public RN terminal-failure surface currently reports LifecycleFailure while retaining native diagnostic detail internally; keep this diagnostic limitation visible during app acceptance.

Multiple independent physical foreground runtimes remain separately scoped in #2532, explicitly deferred to alpha.55. They do not block the single-root acceptance above.

aeplay:
🤖 Release-readiness reconciliation at `a848ea51ef2fbe5e697e17e17fe13fd8dee4a2c7`.

#2291's one-root installed acceptance is implemented, including run-bound Core observation, scope isolation/revocation and an offline fresh-process reopen. The final metadata candidate `193f5894d1d1f305fd2045b4a149c0305a93ab13` is running Android emulator and iOS simulator acceptance: https://github.com/garden-co/jazz/actions/runs/33998136838. Scaffold checks are not device acceptance, and this run is not a physical-iPhone receipt.

[...]

Issue #1972, opened 2026-08-25
Ship React Native relay artifacts in the published package


aeplay:
🤖 Artifact-pipeline implementation completed by #2167, #2545 and the merged RN integration #2552.

Published preview a0708063b1 contains the Android native libraries and iOS XCFramework with architecture verification, matching V1 header/ABI, autolinking/staging contracts and explicit missing/incompatible artifact errors. Native-source/toolchain fingerprints govern production and reuse.

Exact-head isolated packaged consumers passed:
- Android: https://github.com/garden-co/jazz/actions/runs/33912712717/job/101156043531
- iOS: https://github.com/garden-co/jazz/actions/runs/33912712717/job/101156043517

These receipts use Expo-generated native hosts. A separately verified vanilla React Native host and broader installed-device journeys remain acceptance work under #1974/#2291; this closure does not claim those receipts or final npm release publication. The package-production/linkage implementation is complete.

PR #1832, opened 2026-08-24
Restore React Native foundations on the new core

aeplay:
🤖 Replaces #1720 with the preserved React Native restack on the current `main`/new-core foundation. It restores the native relay, SQLite storage, RN/Expo packaging and code generation, and focused native ABI coverage. Published early as a draft so remaining device work, CI, and review stay visible.

## Architecture

The central design choice is that React Native does **not** get a second Jazz implementation. Queries, mutations, permissions, subscriptions, and synchronization remain ordinary Jazz/Groove behavior. Native code provides a durable process-local relay; every foreground UI runtime remains a normal in-memory Jazz client connected to that relay through the existing peer protocol.

```text
RN UI runtime A ─ in-memory Jazz Db ┐
RN UI runtime B ─ in-memory Jazz Db ├─ native relay Db ─ upstream edge/core
                                    └─ SQLite ordered-KV storage
```

### Responsibility split

| Layer | Technology / boundary | Responsibility |
| --- | --- | --- |
| App and Jazz TypeScript | The normal high-level Jazz API in each RN JavaScript runtime | Schema/query/update/subscription APIs and UI lifecycle. Each foreground instance owns only an in-memory client `Db`; it does not query SQLite or implement sync itself. |
| React Native package | New-Architecture Codegen/TurboModule, with thin Kotlin/JNI and Objective-C++ adapters | Autolinking, Expo/bare-RN installation, ABI negotiation, and forwarding opaque binary commands and peer frames. It deliberately exposes no object-per-row native API. |
| Stable native seam | Versioned C ABI, currently ABI 3; opaque host handles and postcard-encoded commands/responses | Keeps Rust independent of RN and makes the same core reusable from future Swift/Kotlin packages. ABI negotiation fails early when an OTA JS bundle expects a different embedded native library. |
| Native relay host | Rust `jazz-native-relay`, one owner thread and one relay per `{app namespace, storage namespace, auth scope}` | Owns the durable ordinary Jazz `Db`, multiplexes UI peers, bounds queues/pump work, connects upstream through the normal peer protocol, and serializes access to executor-local core values. |
| Native persistence | Rust `jazz-storage-sqlite` implementing Groove's async `OrderedKvStorage` contract | Portable ordered KV semantics, atomic batches, reopen/format validation, and SQLite WAL durability. No binding may bypass the relay `Db` and answer application queries from SQLite directly. |

[...]

PR #2167, opened 2026-08-27
Ship React Native relay artifacts in published packages

aeplay:
🤖 Implements the artifact-packaging portion of #1972 on top of the native relay foundation from #1832.

## Example

A developer installs `jazz-rn` from an `rn-preview-release` preview and creates an Expo development build.

Before this PR, React Native autolinking found the package, but the published tarball did not contain the Android relay libraries or iOS XCFramework. The native module therefore reported ABI 0 and every persistent-relay command failed before an app could connect.

After this PR, that opt-in preview contains:

- four Android architecture slices;
- the iOS XCFramework;
- the generated native header;
- a manifest recording the relay ABI and exact source commit; and
- SHA-256 checksums and a sorted inventory for every staged file.

Package assembly rejects missing, stale, extra, or mismatched artifacts before publication. The label-gated CI then builds a small Expo host through Gradle and CocoaPods/Xcode, proving the packaged libraries actually link.

## Labels and cost

- `preview-build` remains fast and deliberately omits expensive React Native artifacts.
- `rn-preview-release` builds and publishes `jazz-rn` with its native artifacts.
- `react-native` runs RN validation/link CI on an ordinary PR without publishing a preview.

The reusable package workflow defaults RN off. Alpha and full releases opt in explicitly and build artifacts from the exact package source commit.

## Scope boundary

This PR makes the package complete; it does not yet make persistent React Native clients usable end to end. Trusted platform admission is #2169, the foreground-client relay bridge is #2166, and device scenarios are #1974.

Expo Go remains unsupported because it cannot load Jazz's custom native module. Bare React Native and Expo development builds are the intended hosts.


aeplay:
🤖 Approved by Anselm.

PR #2168, opened 2026-08-27
bench(examples): add BandChat caught-up fast-resume receipt

aeplay:
🤖 Adds a performance/correctness receipt for reconnecting a fully caught-up BandChat client.

## Example

A room has accumulated 10,000 messages. A client has already received all of them, briefly disconnects, and reconnects with the exact cursor from its last settled update.

Jazz should confirm that the client is current without replaying the room: no reset, added/removed membership, row versions, or commit bundles. The amount of transmitted result data should therefore remain zero whether the history contains 100, 1,000, or 10,000 messages.

## Before

The same thesis was measured by a bespoke RocksDB/JSON timing driver under `crates/jazz/benches`. It was disconnected from the example-app benchmark workflow and mixed protocol behavior with native-storage setup.

## After

BandChat's benchmark package owns a minimal one-table fixture built through public Node/Peer protocol APIs. It first performs a real warm publication to obtain a settled `FastCurrentMembership` cursor, then reconnects using that cursor and asserts that Jazz emits no replay payload.

Divan and CodSpeed run the receipt at histories of 100, 1,000, and 10,000 messages. This is a diagnostic receipt, not a runtime optimization and not a promise of constant wall-clock time. Its invariant is specifically zero replay output for an already-current cursor.

The fixture intentionally isolates that protocol invariant. It does not claim to reproduce BandChat's room filter and ordering; those application-level query shapes have separate coverage.

## Evidence

- BandChat benchmark tests: 2/2
- benchmark listing and test-mode execution at all three history sizes
- Rust-throughput workflow contract: 39/39
- formatting, diff check, and scoped pre-commit checks

Replaces #2137 in the current example/Divan/CodSpeed benchmark model.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -45,7 +45,7 @@ jobs:
         run: cargo install cargo-codspeed --version 5.0.1 --locked
 
       - name: Build example benchmark suites
-        run: cargo codspeed build --package jazz-example-benchmark-smoke --package jazz-example-benchmark-w1 --package jazz-example-big-label-benchmark --package jazz-example-band-chat-benchmark --package jazz-example-world-tour-benchmark --bench fixture --bench ahead_current --bench loads --bench queries
+        run: cargo codspeed build --package jazz-example-benchmark-smoke --package jazz-example-benchmark-w1 --package jazz-example-big-label-benchmark --package jazz-example-band-chat-benchmark --package jazz-example-world-tour-benchmark --bench fixture --bench ahead_current --bench loads --bench queries --bench fast_resume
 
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 The current CodSpeed regression compares this stacked head against main because no successful run existed on immediate base #2174. The measured slowdown comes from #2187 intentionally restoring deletion liveness for policy-scoped maintained point roots, not from this fast-resume layer. I opened #2189 to recover that performance without weakening correctness, and triggered a benchmark run on #2174 so this PR can be re-evaluated against its actual base.

aeplay:

Issue #1882, opened 2026-08-24
BandBinder Better Auth variant is not integrated


aeplay:
🤖 `examples/band-binder/apps/next-betterauth` currently has Better Auth only as a package dependency. The application bootstraps and authenticates with `useLocalFirstAuth()`; there is no Better Auth server/client integration or account bootstrap route.

That makes the variant name and example catalogue misleading to adopters. Either complete it as the self-contained Next.js + Better Auth variant (external session, trusted server-side account bootstrap, issuer-scoped identity mapping, logout/refresh, and relevant E2E) or rename it to the local-first variant until that work exists. Preserve the example’s current schema/query/permission coverage and avoid shared cross-example auth helpers.

Add an exact smoke that proves the declared auth adapter is actually exercised rather than merely installed.

PR #2171, opened 2026-08-27
Reconstruct BandBinder on canonical session identity

### LINKED ISSUE #2091: Isolate generated test artifacts per worktree

aeplay:
🤖 Reconstructs BandBinder from #2094 as an explicit stack on canonical session identity #2096, preserving the app commits without replaying obsolete core history.

Before, the stale branch used removed `session.author` policy/member values, carried hundreds of old-core conflicts, and claimed browser topology was blocked by now-closed #2091.

After:

- Jazz ownership and membership policies use canonical `session.user`;
- provider-boundary fixtures retain raw JWT input while stored/query identity uses canonical issuer+subject JSON;
- app typecheck, permission tests, production Next build, benchmarks, and lane-local artifact provenance pass;
- browser topology now runs against exact local artifacts and records mutation errors rather than hiding them behind a timeout.

Known WIP blocker: manager page and nested-page writes settle, but a block write using nested `policy.exists(page)` remains pending at the edge with no rejection. This aligns with #2096’s currently repairing real-NAPI `session.user` claim injection, so this PR intentionally contains no timeout relaxation or app workaround. It must be restacked and rerun after #2096’s repaired head lands.

Despite its historical folder name, this app currently uses Jazz local-first auth rather than an actual Better Auth UI session; this PR does not fabricate such an integration.


aeplay:
🤖 Closing this stale reconstruction rather than carrying its old core/session history forward. The branch is not actually wired to Better Auth, bootstrap can strand a partially created workspace, and its large topology receipt bypasses the app's React/auth surfaces. Current PosterShop already demonstrates the canonical `session.user`, Better Auth, idempotent server bootstrap, and real browser topology shape. The remaining BandBinder curriculum is tracked by #1882; any revival should be a fresh app-only reconstruction with one explicit auth model, retry-safe bootstrap, issuer-isolation coverage, component-level receipts, and real `$createdAt` ordering assertions.

PR #2174, opened 2026-08-27
fix(groove): reconcile chunk metadata after staged install failure

### LINKED ISSUE #2039: Fault-test durable commit and recovery boundaries
🤖 Enumerate and fault-test every durable transition in `node::state::commit`, node recovery/consistency markers, Groove batch persistence, and `OrderedKvStorage::{write_many,reopen,flush_write_boundary}`.

Cover crashes between history, current, index, catalogue, fate, marker, and large-value metadata/blob writes. For each boundary, specify whether reopen rebuilds, resumes, retries, or fails closed; no marker may expose partially durable derived state as settled. Add deterministic fault-injection receipts on persistent backends.

Relates to #1774, #1779, and recovery/settlement work such as #1995.

aeplay:
🤖 Fixes a crash-recovery hole when Jazz downloads part of a large value. It closes #2039.

## User-visible failure

Suppose a client opens a document whose next chunk is not local:

1. Jazz downloads and durably stores the chunk bytes.
2. Before recording how that chunk connects to the large-value tree, the app crashes or the metadata write fails.
3. On retry, Jazz sees that the bytes are already local and previously treated the download as complete.

The document could then remain unreadable, and Jazz could lack the child links and reference counts needed to traverse or eventually reclaim that chunk.

## New recovery rule

Before staging downloaded bytes, Groove writes a small durable `install/<chunk>` marker. It clears that marker only in the same durable metadata update that records the chunk's tree links and reference counts.

After a crash, a locally present chunk with a marker resumes the missing metadata step. A normal, fully installed local chunk does not re-enter installation logic.

If writing bytes returns an ambiguous error, Jazz keeps the marker because the bytes may nevertheless have reached storage. If the remote request fails before any byte write is attempted, Jazz creates no marker.

The byte/blob store remains unaware of permissions and tree metadata; Groove owns this recovery bookkeeping.

## Why the narrower trigger matters

An earlier repair retried metadata installation on every local chunk hit. That made ordinary reads re-enter the metadata lifecycle and deadlocked publication tests. This version reconciles only chunks that still carry the durable marker.

The extra asynchronous metadata step also exposed a scheduler bug: runtimes could stop polling once the raw bytes arrived even though Groove still needed to finish installation. The repair adds a one-shot readiness wake-up across Client, Server, NAPI, and WASM. Actual storage progress schedules one more owner turn; there is no timer or polling loop.

## Failure and concurrency coverage

- metadata failure after bytes are staged, followed by reopen and successful recovery
- crash points before and after the marker, byte stage, and metadata commit
- ambiguous byte-store failure retaining the recovery marker
- ordinary local reads never invoking the installer
- concurrent readers installing metadata once
- provisional same-process cache entries not hiding durable recovery markers
- one-shot scheduler wake-up in Client, Server, NAPI, and WASM paths
- planted missing-marker, install-on-every-hit, and no-op-waker regressions

Verification includes `cargo test -p groove --lib` (534 passed, 2 ignored) and the CI-profile Groove suite (602 passed, 4 skipped).

The current CodSpeed comparison incorrectly falls back to an older `main` because the actual stacked base has no baseline receipt; its reported policy-query regression therefore includes unrelated lower-stack changes and must be remeasured after the lower stack lands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2354,6 +2354,7 @@ dependencies = [
  "base64 0.22.1",
  "console_error_panic_hook",
  "ed25519-dalek",
+ "futures-channel",
  "futures-util",
  "getrandom 0.2.17",
  "getrandom 0.3.4",
--- a/crates/groove/src/chunks.rs
+++ b/crates/groove/src/chunks.rs
@@ -727,13 +727,23 @@ where
                     Ok(bytes)
                 }
                 Err(ChunkStorageError::Unavailable) => {
+                    let bytes = self.resolver.resolve(request.clone()).await?;
+                    // Resolver failure has no byte-store side effect, so do
+                    // not create a durable recovery obligation until it has
+                    // supplied mechanically valid bytes. This keeps a
+                    // disconnected/cancelled fetch from leaking journal
+                    // entries for chunks that were never staged.
+                    if bytes.len() > crate::large_values::MAX_ENCODED_NODE_BYTES
+                        || object_hash(&bytes) != node_ref.object_hash
+                    {
+                        return Err(ChunkError::Integrity);
+                    }
                     // This durable marker is deliberately written before the
                     // independent byte store.  If either staging or metadata
                     // installation is interrupted, a later exact read knows
                     // that it must finish the install rather than treating
                     // resident bytes as a complete mapping.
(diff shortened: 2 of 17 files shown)


aeplay:

PR #2179, opened 2026-08-27
perf(jazz): bound flat subscription delta bookkeeping

aeplay:
🤖 Supersedes #2093 with the same generic optimization reconstructed directly on current `main`, preserving the original PR's history without force-pushing it.

## Before

Every maintained flat delta cloned the complete prior snapshot and index, even when authoritative membership and ordering were unchanged. Root removals also rediscovered occurrence positions by repeatedly scanning the index.

For example, updating one nullable field in a 10,000-row subscription paid for full snapshot/index clones even though it could apply the delta in place.

## After

Prior snapshots are retained only when authoritative membership changes require reordering. Root removals use the maintained occurrence-to-position index directly. This remains the single generic flat-delta path; it does not introduce a second TopBy or query-specific implementation.

## Correctness evidence

- nullable flat update
- offset-window leading deletion
- authoritative reset rebuilding occurrence sidecars, order, and count
- persistent IndexedDB first-match delivery
- observer write-path benchmark smoke
- planted removal suppression fails the offset-window receipt, then passes after restoration

The existing 100/1,000/10,000-row benchmark scales are already present on current `main`, so this reconstruction does not duplicate them.

### DIFF
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -4461,30 +4461,66 @@ fn apply_maintained_membership_update_to_snapshot(
         }
     }
 
-    for occurrence_id in &update_removed {
-        if update_added.iter().any(|(added, _)| added == occurrence_id) {
-            continue;
-        }
-        let Some(index) = snapshot_index.roots.get(occurrence_id).copied() else {
-            continue;
-        };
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2191, opened 2026-08-27
Fix crate alpha publish allowlist

joeinnes:
🤖 Fixes the alpha Rust-crate release list—the explicit list of packages that the workflow is allowed to publish—so a stale package name cannot cause a partially completed release.

## Example

The workflow still listed `jazz-tools` as a Cargo package, but `jazz-tools` is now only a binary produced by the `jazz-cli` package.

Previously, an alpha release could:

1. publish an earlier crate successfully;
2. reach the stale `jazz-tools` entry;
3. fail because no Cargo package had that name; and
4. leave the release only partly published.

The old dry run did not catch this because it merely compiled packages rather than asking Cargo to assemble and validate the publishable crate.

After this PR, the workflow validates every listed package's exact name and version from Cargo metadata before performing any registry operation. A missing or renamed package therefore stops the release before the first crate is published.

Dry runs now execute `cargo publish --dry-run`, exercising Cargo packaging rather than only compilation. Real retries remain idempotent: if the exact package version already exists on crates.io, the workflow skips it and continues.

## Evidence

- `dev/scripts/publish-crates-alpha.test.sh`
- maintained local tooling gate
- live `cargo publish --allow-dirty --dry-run -p jazz-wasm-tracing`
- exact-head CI previously passed lint, TypeScript, Rust, workspace, and differential jobs
- independent review found no correctness issue

### DIFF
--- a/dev/gates/local-ci-equivalent.mjs
+++ b/dev/gates/local-ci-equivalent.mjs
@@ -86,6 +86,9 @@ export const ciPartitions = Object.freeze({
   lint: Object.freeze([
     command("format check", "pnpm", ["format:check"]),
     command("lint", "pnpm", ["lint"]),
+    command("crate publication selector contract", "bash", [
+      "dev/scripts/publish-crates-alpha.test.sh",
+    ]),
     command("CI workflow contracts", "pnpm", ["test:ci-workflow"]),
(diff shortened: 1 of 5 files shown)


aeplay (approved):
🤖 Approved by Anselm.

Issue #2040, opened 2026-08-25
Freeze backend format manifests for RocksDB, SQLite, and IDB


aeplay:
🤖 Produce an explicit backend-format manifest and compatibility review for `jazz-storage-rocksdb`, `jazz-storage-sqlite`, and `groove::storage::key_codec`.

Cover SQLite meta/version rows, column-family IDs, WAL/synchronous settings and schema validation; RocksDB CF names, private value tags, merge operands, WAL durability and format markers; and IDB's CF-prefixed ordered-key representation. Add backend-owned valid/invalid fixture stores and a matrix for fresh open, additive CF reopen, unknown/old/new markers, malformed metadata, missing CFs, interrupted batches, and prefix/range ordering.

This complements the portable semantic contract in #1774 and should account for the SQLite consumers tracked by #1756.

aeplay:
🤖 Baseline update: #2025 merged as `00f8f78ff`. The freeze review should treat its RocksDB v2 tagged-value envelope, format marker, empty-store initialization, and non-mutating rejection of nonempty unmarked stores as the current format—not any of the intermediate raw-sentinel or migration designs discussed during review.

aeplay:
🤖 Audit refinement: the current physical envelopes intentionally differ—SQLite has its v1 metadata contract, RocksDB a raw-v3 marker/column-family contract, and IDB a v1 CF-key codec plus separately versioned/checksummed pages. Backend fixtures should prove equivalent ordered-KV semantics while explicitly testing each backend’s distinct unknown/old/new marker and corruption policy; format-number equality across backends should not be implied.

aeplay:
🤖 Final storage-format audit refinement: reserve and validate engine-owned column-family namespaces (including `__groove_*` / `__groove_large_values`) and freeze one consistent maximum logical column-family name length across backends. Today generic storage key packing admits a `u32` length while the IndexedDB codec bounds it to `u16`; this is not an observed corruption path, but it should become an explicit cross-backend format contract rather than an incidental backend difference.

aeplay:
[...]

PR #2192, opened 2026-08-27
Reject column family namespace collisions

joeinnes:
## Summary
- reject duplicate and internally reserved logical column-family names before runtime or storage-layout I/O
- enforce JazzClassV1-specific physical reservations while preserving Identity-layout compatibility
- apply the same ownership rules to dynamic table and index registration

## Before
Application tables and direct stores could share one physical family or claim Groove metadata, index, and JazzClassV1 engine families. Backends could silently deduplicate those names, allowing incompatible record formats or internal metadata to share storage.

## After
Database admission assigns one owner to every logical and selected-layout physical family. Conflicts return structured owner diagnostics before runtime, layout-marker, backfill, or storage writes. Physical mappings and valid persisted stores are unchanged, so valid stores require no migration; valid distinct families, conditional index ownership, and Identity-layout names remain supported.

## Test plan
- [x] Run `cargo test -p groove`
- [x] Verify initial and dynamic namespace failures perform no storage operations


aeplay:
🤖 Superseded by #2193, which preserves the useful storage-contract specification and pre-mutation admission receipts while adopting the storage-freeze decision from #2040: application names reserve `__groove_*`, `indices`, and `default` globally, independent of active layout or index state. #2193 additionally closes the raw backend ingress surfaces that this branch intentionally did not cover—open/reopen validation, persisted RocksDB/SQLite catalogues, IndexedDB framing, Memory snapshot import, NUL rejection, and the portable UTF-8 byte bound. Joe’s spec/invariant contribution was adapted onto #2193 in commit `02ed01ef0`; the conditional `indices` and Identity-layout allowances were intentionally not retained because they would make schema legality change as features/layouts evolve.

PR #2193, opened 2026-08-27
fix(storage): reserve portable physical namespaces

### LINKED ISSUE #2040: Freeze backend format manifests for RocksDB, SQLite, and IDB

aeplay:
🤖 Closes #2040.

## Before

- Application tables and direct stores could collide with Groove-owned physical families such as `__groove_large_values`, generated class families, `indices`, or RocksDB `default`.
- Backend name constraints differed: IndexedDB framed names as `u16`, RocksDB could panic on embedded NUL, and persisted/discovered catalogues or memory snapshots could bypass later schema admission.
- Live `register_table` admission did not apply the same namespace checks as initial database construction.

## After

- One application-schema contract rejects reserved engine namespaces, duplicate table/direct-store names, embedded NUL, and names beyond the portable `u16::MAX` UTF-8 byte limit before Groove mutation.
- One lower-level physical-family contract validates every backend open/reopen path, persisted/discovered RocksDB and SQLite catalogues, IndexedDB names, and Memory snapshot import.
- `MemoryStorage::new` is now honestly fallible; callers explicitly handle construction failure. No compatibility path retains unchecked public construction.

## Behavioral examples

- A table named `__groove_class_meta`, `indices`, or `default` fails schema admission instead of aliasing engine metadata.
- `rows\0evil` returns a typed error before RocksDB reaches its C-string boundary; it cannot panic or create a store.
- A legacy SQLite catalogue, RocksDB family list, or imported Memory snapshot containing an invalid family is rejected without replacing live state.
- A valid application name that merely differs in case remains valid because physical namespaces are case-sensitive.

## Verification

- Groove library: 550 passing (2 ignored on the implementation pass).
- RocksDB library: 14 passing.
- SQLite library/integration: 9 passing.
- Jazz realistic Phase 1 benchmark compiles; Jazz doctests and targeted lifecycle coverage pass.
- Independent adversarial review planted late registration, Rocks FFI, persisted-catalogue, snapshot-import, and reopen-fast-path regressions and confirmed the new receipts fail.

This is stacked on #2088 and should be reviewed before #2167.


aeplay:
🤖 Approved by Anselm.

Issue #2095, opened 2026-08-26
Local NAPI edge does not relay writes to configured upstream core


aeplay:
🤖 A locally started NAPI edge configured with an upstream core accepts and serves edge-local writes, but those writes do not reach the upstream core and global settlement never completes.

Minimal topology observed while repairing #2085:

- Start core with `startLocalJazzServer({ inMemory: true, ...shared app/auth configuration })`.
- Start edge with the same app ID, admin/backend secrets and JWKS, plus `upstreamUrl: core.url`.
- Deploy/initialize core; connect a browser client to edge and an admin-secret observer directly to core.
- An edge write settles at edge tier and remains queryable through the edge.
- The same write never appears for the direct core observer (zero rows after 20 seconds).
- Waiting globally through the edge hangs rather than settling or rejecting.

Expected: `upstreamUrl` establishes the application relay session; accepted edge writes flow upstream and global waits settle once core admits them.

This is distinct from the truthful browser→serving-core example now in #2085. Investigate the local server’s NAPI edge composition/session bootstrap and add a minimal edge→core relay receipt covering positive propagation, global settlement, rejection, reconnect, and upstream unavailability. Do not make example topology labels imply an edge when only one server exists.

PR #2195, opened 2026-08-27
Expose native transport terminal completion

joeinnes:
## Summary

- Give each connected native transport a terminal future that resolves when its WebSocket closes or fails, even while the database is otherwise idle.
- Separate bootstrap failure from post-connect termination so callers can retry connection setup and reconnect an established session for the right reasons.
- Preserve bounded outbound backpressure and wake blocked producers when the transport pump terminates.

## Before and after

Before, an idle peer closure was only discovered when another semantic frame or outbound retry touched the transport. An Edge could therefore continue reporting a dead upstream as connected.

After, the transport driver publishes one terminal outcome independently of database traffic. The owner can react immediately without polling or manufacturing work.

## Testing

- `cargo test -p jazz-native-transport --test edge_websocket` — 11 passed

Part of #2095.


aeplay:
🤖 Superseded by #2227, reconstructed directly on current `main` while preserving Joe’s original commit and authorship. The replacement isolates the terminal-completion primitive from the obsolete integration-stack ancestry so its dependent reconnect work can stack cleanly.

PR #2198, opened 2026-08-27
Authorize writes relayed through an Edge

### LINKED ISSUE #2095: Local NAPI edge does not relay writes to configured upstream core

joeinnes:
## Summary

- Preserve the `Admin` credential classification used to authenticate an Edge control-plane connection while assigning ordinary `SYSTEM` relay traffic `TrustedBackend` policy trust.
- Reserve `TrustedAdmin` for the explicit catalogue-bootstrap path.
- Keep the original transaction permission subject for application-policy evaluation while retaining `made_by` as provenance.

## Before and after

Before, the admin secret used to authenticate an Edge-to-Core connection also assigned `TrustedAdmin`, allowing ordinary relayed application writes to bypass row-level write policy.

After, ordinary relay traffic is authenticated but policy-checked as `TrustedBackend`. Only explicit catalogue bootstrap receives the administrative policy bypass.

For example, a user write accepted by an Edge now reaches Core and either settles globally under that user’s policy subject or returns Core’s rejection. Restart and temporary upstream unavailability retain the same policy boundary.

## Testing

- `cargo test -p jazz-server ws_system_admin_data_link_preserves_relay_write_authorization` — passed
- `cargo test -p jazz-testkit --test edge_server_mode edge_to_core_relay_` — 3 passed
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/dev/dev-server.test.ts -t "makes a backend write through Edge globally visible from Core"` — passed

Closes #2095.


aeplay:
🤖 Superseded by #2229, which reconstructs the isolated Edge-relay authorization fix directly on current `main` while preserving Joe’s authorship and excluding obsolete stacked changes.

PR #2199, opened 2026-08-27
fix(browser): coordinate shared namespace offline state

aeplay:
🤖 Fixes explicit-offline state and transport ownership for multiple tabs sharing one persistent browser namespace.

## Before

- The worker owns one server connection per persistent namespace, but `disconnect()` only marked the initiating tab offline while tearing down the namespace-wide transport.
- A remaining or late tab could classify `remote-if-possible` as remote even though no upstream carrier existed, and edge waits could remain pending.
- Concurrent disconnect/reconnect messages from different ports could interleave and publish stale state; closed peers could remain parked in offline waiters.

## After

- Explicit offline state is namespace-wide, durable in the worker context, and broadcast to current and late-attaching tabs.
- Only a newly attaching browser follower waits for its initial authoritative transport-state handshake before choosing a tier. Established/direct reads preserve choose-once semantics.
- Namespace transport transitions are serialized; any peer can reconnect the shared namespace in order.
- Offline waiters are peer-owned and aborted on tab close, preventing stale posts and retained closures.

## Behavioral examples

- Tab A disconnects a shared namespace; tab B immediately gets local fallback for `remote-if-possible`, while an edge fate remains pending.
- A tab opened after the disconnect learns the offline state before its first one-shot/subscription; reconnecting from that tab upgrades the subscription and settles the pending fate.
- A reconnect racing a slow disconnect cannot start a replacement carrier and then be overwritten by the older transition.
- Closing a tab while explicitly offline removes its parked server waiter.

## Verification

- 32/32 focused read-tier, BrowserConnectionManager, and worker-initialization tests pass.
- Independent review reproduced the late-tab failure on the first candidate and planted timing/race regressions; the final candidate closes them.
- TypeScript and runtime builds pass. The browser E2E is included for CI; local replay was blocked by stale generated artifact provenance, not a test failure.

Stacked after #2191.

### DIFF
--- a/docs/content/docs/reference/internals.mdx
+++ b/docs/content/docs/reference/internals.mdx
@@ -226,6 +226,12 @@ database namespace. The worker owns the IndexedDB page store and the server conn
 communicates with that runtime through a message port. There is no tab-leader election, Web Locks
 dependency, or per-tab durable database.
 
+Calling `db.disconnect()` in one tab explicitly takes that **whole persistent namespace** offline:
+the worker has one upstream connection, so every attached tab uses local-first fallback for
+`"remote-if-possible"` reads and remote reads wait until any tab calls `db.reconnect()`. This is
+intentional; disconnecting only one tab while continuing to sync its writes through the shared
+worker would give that tab an incoherent offline contract.
+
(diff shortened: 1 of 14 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2200, opened 2026-08-27
fix(server): serialize permission head publication

joeinnes:
## Summary

- make permission-head parent validation and publication one in-process critical section
- ensure only one concurrent publisher can consume a given expected parent
- keep the stored catalogue head and active runtime policy aligned with the winning publication

## Before

Two administrators could publish different permission bundles against the same expected parent.

Each request validated the parent while holding the catalogue index lock, then released that lock before writing to storage. Both requests could therefore validate the same parent, both return `201 Created`, and the later write could silently replace the first. Runtime policy updates could also diverge from the winning stored head.

## After

Permission publication now holds the storage and index locks through parent validation, bundle-first/head-second storage, and in-memory head advancement.

Exactly one concurrent request can consume a parent:

- the winner receives `201 Created`;
- the loser receives `409 Conflict`;
- the stored bundle and head match the winner;
- the active runtime policy matches the same winner.

This guarantee applies to callers sharing one in-process `StoredCatalogue`. It is not a cross-process compare-and-swap or a crash-atomic two-entry storage transaction. Bundle-first ordering means a crash may leave an unreferenced bundle, but cannot publish a head whose bundle was never written.

## Testing

- `cargo test -p jazz-server server::catalogue::tests::concurrent_permissions_publications_compare_and_swap_the_parent -- --exact --nocapture` — passed
- `cargo test -p jazz-server server::routes::tests::concurrent_permissions_publications_install_only_the_winning_runtime_head -- --exact --nocapture` — passed

### DIFF
--- a/.changeset/atomic-permission-heads.md
+++ b/.changeset/atomic-permission-heads.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Make permission publications sharing one in-process catalogue store compare and swap the current parent while preserving bundle-first, head-second storage.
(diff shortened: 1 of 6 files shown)


aeplay (approved):
🤖 Approved by Anselm.

PR #2201, opened 2026-08-27
Retain read view during authoritative reconcile

joeinnes:
🤖 Keeps a maintained subscription on the exact historical/branch view the caller requested when the server later has to reconcile its membership.

## Example

A client subscribes to a branch view defined by a branch head plus its base. Row A belongs to that view because it is still visible through the base.

Later, a deletion witness makes the server re-check which rows belong in the subscription. Previously, the cached peer plan remembered only the requested durability tier. Reconciliation therefore reopened a default current-state query, where Row A might be absent, and incorrectly removed it from the branch subscription.

After this PR, the cached plan retains the complete immutable `ReadViewSpec`—the object identifying the requested current, historical, or branch view—and reuses it for retries, hydration, and authoritative reconciliation. Row A is removed only if it is absent from that actual branch view.

Ordinary current-state subscriptions continue using the default current view and retain their existing incremental behavior.

## Scope

This changes only in-memory query-plan context. It does not change the wire protocol, storage format, or public query API.

## Evidence

- `maintained_branch_view_reconcile_retains_undeleted_base_members`
- `default_current_subscription_reconciles_deletion_witness_without_reset`
- peer-level receipts cover cached-plan reuse and real served-connection reconciliation

### DIFF
--- a/crates/jazz/SPEC/16_maintained_subscription_views.md
+++ b/crates/jazz/SPEC/16_maintained_subscription_views.md
@@ -43,6 +43,14 @@ are coverage facts for the maintained subscription view only; they do not become
 complete transaction payload refs. The peer state machine MUST NOT answer a live
 subscription by running an independent semantic scan.
 
+A maintained peer publication MUST retain the complete semantic read view and
+tier resolved at admission. Initial hydration, suspended retries, incremental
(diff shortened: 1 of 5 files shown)


aeplay (approved):
🤖 Approved by Anselm.

PR #2202, opened 2026-08-27
Reject conflicting subscription key reuse

joeinnes:
## Summary
- make one active subscription key retain one canonical coverage owner
- allow same-coverage replays to refresh known state without duplicating maintained resources
- prevent ordinary and current-row producers from claiming the same wire handle

## Before
A peer could reuse one active `SubscriptionKey` with a different shape or binding. Jazz overwrote the reverse lookup but left the key in its old coverage group, allowing multiple groups to publish unrelated rows to one wire handle and leaving the old group unreachable during unsubscribe. Ordinary and current-row subscription paths could also claim the same key.

## After
The first active owner remains authoritative. Same-key/same-coverage replay refreshes delivery through one coverage group, while changed coverage, shape mismatch, and cross-producer collisions are rejected before subscription, known-state, relay, or publication side effects. The original stream remains live and unsubscribe cleans its sole resources. No wire or storage format changes.

## Test plan
- [x] Run the six focused active-key and current-row ownership regressions
- [ ] Run the full Jazz suite once the current-main structured-subscription stack overflow is resolved


joeinnes:
CI note: `trusted / test-rust-workspace` hit the known 180-second timeout in `maintained_physical_point_subscriptions_keep_policy_scopes_live`. The subscription can miss the deletion register and wait for the test deadline. #2187 targets `main` directly with the most applicable fix; #2177 addresses the same regression on the structured-rehydrate stack. Please rerun this PR’s CI after the applicable fix lands. The lint, TypeScript, and Rust differential partitions passed on the current head.

aeplay (approved):
🤖 Approved by Anselm.

PR #2204, opened 2026-08-27
fix(auth): coalesce backend JWKS refreshes

joeinnes:
## Summary

- coalesce concurrent JWKS downloads by URL so one provider request supplies every waiting authentication
- start the 30-second forced-refresh cooldown only after a successful download
- clear failed in-flight downloads so a later request can retry immediately

## Before

Concurrent requests for an uncached JWKS URL could each download the same document.

An invalid JWT could also trigger another provider request immediately after a successful cold download. This amplified traffic without providing newer keys.

## After

Requests for the same JWKS URL now share one in-flight download.

A successful download installs the document and starts a 30-second forced-refresh cooldown. Invalid JWTs are rejected against that document during the cooldown instead of causing another immediate download.

A failed download is shared by its current waiters but is not cached. Once those waiters receive the failure, a later request can retry immediately.

The coordination is process-local. It does not coordinate downloads across server processes, apply provider cache headers, or add outage backoff.

## Testing

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/backend/request-auth.test.ts` — 13 passed


joeinnes:
CI is currently blocked by the known, unrelated Rust subscription-liveness timeout. Please rerun CI after #2187 (or the overlapping #2177) lands.

aeplay (approved):
🤖 Approved by Anselm.

PR #2205, opened 2026-08-27
fix(auth): expire cached backend JWKS keys

joeinnes:
## Summary

- limit cached backend JWKS documents to five minutes of trust
- coalesce concurrent refreshes when a document reaches its expiry
- fail closed after expiry when the identity provider cannot supply a fresh document

## Before

A cached JWKS document had no maximum lifetime.

If a JWT remained valid against a removed signing key, signature verification continued to succeed and therefore never triggered the forced-refresh path. The removed key could remain trusted for the life of the process.

## After

A cached JWKS document expires after five minutes. The first authentication at or after expiry fetches a new document before verifying the JWT, and concurrent authentications share that fetch.

A successful refresh installs the replacement keys and starts the forced-refresh cooldown. Tokens signed by a retired key are then rejected without causing another immediate download.

If the provider is unavailable after expiry, authentication fails closed. The expired document is not used as a stale fallback. A later request retries immediately because failed downloads are not cached.

This deliberately favours prompt key revocation over authentication availability during a JWKS provider outage.

## Testing

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/backend/request-auth.test.ts` — 16 passed


joeinnes:
CI is currently blocked by the known, unrelated Rust subscription-liveness timeout. Please rerun CI after #2187 (or the overlapping #2177) lands. This run also hit the unrelated WASM artifact-lock concurrency failure; a fresh run should verify both.

aeplay (approved):
🤖 Approved by Anselm.

PR #2206, opened 2026-08-27
Retain union-arm identity in prepared memberships

aeplay:
🤖 Closes the prepared-membership schema mismatch discovered by the Wequencer cross-topology receipt.

## Before

Prepared result-membership lowering retained hidden joined-row identifiers such as `__root_join_row_0`, but dropped their paired union-arm discriminator such as `__root_join_arm_0`. When two authorization branches contributed occurrences for the same maintained view, the published record no longer matched its declared terminal schema. Direct runtimes reported a protocol error; the persistent browser worker could leave the affected write pending until timeout.

## After

Result-membership public fields retain every union-arm discriminator alongside the corresponding occurrence row identifier. The emitted prepared terminal record and `ResultMembershipSchema` now describe the same occurrence address.

## Example

For an owner-or-editor policy, an occurrence addressed as `(joined row = R, arm = editor)` now carries both `__root_join_row_0 = R` and `__root_join_arm_0 = editor`; previously only the row component survived projection.

## Verification

- Focused Rust field-contract receipt, including a planted removal that fails with the missing arm.
- Full Wequencer browser Edge/Core topology, including offline edit and deterministic transport retry: all phases pass.
- `cargo clippy -p jazz -- -D warnings`.

### DIFF
--- a/crates/jazz/src/node/query_eval/lowering.rs
+++ b/crates/jazz/src/node/query_eval/lowering.rs
@@ -202,6 +202,11 @@ pub(super) fn fact_public_fields(
                     .filter(|field| **field != schema.row_field)
                     .cloned(),
             );
+            // A union arm is part of a result occurrence address, just like
+            // its paired joined-row id. Retaining only the id causes prepared
+            // maintained subscriptions to publish a descriptor that cannot be
+            // decoded against the membership schema.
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2208, opened 2026-08-27
Reconstruct the Wequencer example on the current core

aeplay:
🤖 Reconstructs Wequencer as an app-only example on the current linear core stack, preserving the useful work from #2097 without its obsolete copied core/session changes.

## Before

The earlier reconstruction was based on a much older core and carried stale identity and runtime assumptions. Its browser topology also exposed a real prepared-membership lowering bug: after an owner had written a transport observation and disconnected, an editor write could remain pending because the hidden union-arm discriminator was projected away.

## After

- Uses current `session.user` identity and Jazz magic authorship columns.
- Keeps creator administration bound to immutable session `$createdBy`, not a mutable membership row; deleting or downgrading the creator membership does not transfer creator authority.
- Keeps Better Auth data server-owned and profile bootstrap explicit/off the initial read path.
- Exercises bounded transport queries and native benchmark shapes.
- Includes a full browser → Edge → Core topology covering creator authority, collaborative edits, offline local edits, deterministic transport retry, offline reopen, and revocation.
- Includes the track-restoration sensitivity check after the creator-policy phase.

## Worked behavior

The creator may remain the session administrator after their ordinary membership row is removed. An editor can then edit while locally disconnected and, after transport recovery, the write settles at the Edge and converges at Core. #2206 fixes the internal prepared-membership occurrence schema required by that path.

## Verification

- `pnpm exec tsc --noEmit`
- app unit suite: 5/5
- native benchmark workload suite: 6/6
- full browser Edge/Core topology: 1/1, all phases
- clean generated-artifact and worktree state after the receipts


aeplay:
🤖 Approved by Anselm.

PR #2209, opened 2026-08-27
Reconstruct the RecordPlayer example on the current core

aeplay:
🤖 Reconstructs RecordPlayer as an app-only example on the current core, preserving useful work from #2103 without obsolete core/session history.

## Before

The earlier branch targeted an older core and mixed the example with stale identity/runtime work. Its browser → persistent worker → Edge → Core topology exposed several defects:

- a detached bounded Edge window could be mistaken for fresh authority coverage, so a later local subpage applied its offset twice;
- after invitation revocation, a policy-scoped exact-ID Edge read could resurrect a cached child row;
- the first exact-ID repair opened a seeded Edge child while waiting for its not-yet-created Global source, deadlocking browser row hydration;
- after every foreground tab closed, the persistent browser worker could retain an accepted row but fail to publish settled authority to a newly opened tab;
- one large finite Rust receipt exceeded the platform’s 2 MiB test-thread stack.

## After

- Uses current `session.user` identity while retaining `$createdBy` authorship.
- Keeps Better Auth tables server-owned, bootstrap explicit/off the initial read path, and provider errors retryable.
- Restores indexed catalogue and playlist-window queries, nullable 64 KiB audio coverage, streamed audio, sharing, offline edits, and concurrent invitation acceptance.
- Retains detached nonzero-offset Edge membership only for local materialization; it is not settled authority coverage, and a fresh Edge request clears/reacquires it.
- A policy-scoped exact-ID Edge read acquires and hydrates from matching Global authority coverage. A broad Edge query still owns public result membership, and no Local cache fallback is used. Standalone exact-ID and internal row hydration therefore avoid the circular seeded-child wait.
- A browser persistent worker is explicitly marked as the relay-authority owner. It publishes a distinct `RelayAuthoritySession` source that survives foreground-tab detach/reopen; ordinary local relays keep their normal Edge→Global authority lift.
- A revoked cached row is absent remotely, while an explicit Local read may still see its cached body.
- The one deliberately large Rust topology receipt gets a scoped 6 MiB test thread; production/global settings are unchanged.

## Examples

A broad Edge playlist query returns two visible rows. Hydrating each row’s policy-scoped exact ID attaches the outer membership at Edge, acquires each exact authority source at Global, then returns through the existing Edge membership. It never attaches Local and no longer waits for a child receipt that cannot be emitted until after attachment.

After invitation revocation, the exact Global source becomes empty and the broad Edge membership removes the child. A separate Local read can still expose the cached body without claiming remote authorization.

When tab 1 receives an accepted row and closes, the worker retains the row and relay-authority source. Tab 2 can later attach to the same worker and receive a settled Edge handoff without pretending that every ordinary relay is a persistent-worker authority.

## Verification

- RecordPlayer topology: 9/9 phases.
- Todo persistent-browser canary: 11/11 after the exact-source ordering repair.
- Browser relay durability suite: 23/23, including a real close/detach/reopen handoff.
- Worker initialization tests: 12/12, plus TypeScript checks.
- Direct exact-ID receipt asserts Global-only attach/read.
- Broad Edge receipt asserts Edge outer membership plus Global exact hydration, never Local.
- Planted missing-source scheduling, Global→Local fallback, and relay-authority-key mutations each fail.
- Forced clean NAPI/WASM rebuild precedes canonical artifact sealing, avoiding stale planted binaries.
- Complete unit/provider/selection/browser-relay, formatting, Clippy, and artifact checks pass.


aeplay:
🤖 Approved by Anselm in the orchestration thread on exact head 33e72a029; merge when required CI is green.

PR #2212, opened 2026-08-27
Preserve bounded query membership across relay tiers

aeplay:
🤖 Replacement for #2210. GitHub automatically marked #2210 merged when its head became an ancestor of draft #2209 during a stack-base correction; that merge was only into the feature branch and nothing reached `main`. This PR restores the same independently reviewable core diff at its intended position directly above #2208.

## Before

A Core could correctly select playlist-entry positions 8–23 for `offset(8).limit(16)`. The persistent relay then materialized only those 16 selected rows and evaluated the original offset again, publishing positions 16–23. The foreground Local view could apply the window yet again and become empty. Running an unbounded query first appeared to fix the problem only because it hydrated all raw rows.

## After

- A relay Edge child records and consumes the settled Global source membership selected by its upstream authority.
- Client-local evaluation removes only the already-applied outer slice; filters, joins, order, and query identity remain intact.
- A Local subwindow may reuse an Edge receipt only when that same query window fully covers it.
- Transition reconciliation reads the Global source binding rather than stale child Edge membership.
- One-shot browser scope release preserves bounded receipt membership needed downstream.

## Worked example

For rows positioned 0–32, upstream `offset(8).limit(16)` selects 8–23. Edge and Local now both observe 8–23. A distinct `offset(10).limit(4)` remains its own shape and returns 10–13. Inserting before both windows invalidates and shifts both memberships independently.

## Verification

- New three-hop native relay receipt covering Edge/Local agreement, independent subwindows, nested bounded coverage, and membership-shift invalidation.
- `browser_relay_durability`: 21/21.
- Planted disabling of Local receipt routing fails with 8 trailing rows instead of 16, then restores cleanly.
- RecordPlayer advances through its previously empty recipient window.

### DIFF
--- a/crates/jazz/src/db/subscriptions.rs
+++ b/crates/jazz/src/db/subscriptions.rs
@@ -336,6 +336,16 @@ where
         )
     }
 
+    #[cfg(any(test, feature = "testing"))]
+    /// Internal receipt-lifetime coverage needs to inspect state that has no
+    /// public equivalent: detached Local overlays are intentionally best-effort.
+    pub fn settled_authoritative_receipt_counts_for_test(&self) -> (usize, usize) {
+        self.node
+            .node
+            .borrow()
+            .settled_authoritative_receipt_counts_for_test()
+    }
+
     /// Detach a one-shot query coverage request.
     pub fn detach_query(&self, attachment: QueryAttachment) {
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2214, opened 2026-08-27
fix: rebuild previews after workflow changes

joeinnes:
## Summary
- Stop release publishing from reusing preview artifacts when the release source tree differs from the tree validated by the preview build.
- Remove the workflow-only drift exception so packages are always produced by the publishing workflow version they are released with.

## Before and after
Before, a manual release could reuse preview artifacts when every change since the preview was under `.github/workflows/`. After this change, any tree difference disables reuse and rebuilds the release artifacts.

## Testing
- `node --test dev/artifacts/release-staging-contract.test.mjs` — passed (6 tests)


aeplay (approved):
🤖 Approved by Anselm.

PR #2223, opened 2026-08-27
Rename write batch IDs to transaction IDs

aeplay:
🤖 Renames the public write identity from batch terminology to transaction terminology. This is a runtime/binding API cleanup; it does not change Jazz’s persisted or wire transaction encoding.

## Before

```ts
const write = db.insert(app.todos, { title: "Ship it" });
const id: BatchId = await write.batchId;
const open: OpenBatchId = createOpenBatchId();
```

A committed write is already one Jazz transaction, while “batch” was leftover terminology. The handle also exposed a deprecated `transactionId` alias pointing back to `batchId`, obscuring the intended API.

## After

```ts
const write = db.insert(app.todos, { title: "Ship it" });
const id: TxId = await write.txId;
const open: OpenTransactionId = createOpenTransactionId();
```

There are no compatibility aliases: `batchId`, `BatchId`, `OpenBatchId`, and the deprecated handle `transactionId` are removed from the public API.

The rename is complete across TypeScript, Rust public APIs, NAPI/WASM bindings and generated declarations, runtime JSON, examples, documentation, diagnostics, and tests. Internal uses of “batch” remain only where they describe actual frame or query-engine batching.

## Generated-artifact boundary

NAPI/WASM artifacts are generated per checkout rather than committed. Their verifier now explicitly rejects a generated `WasmWrite.batchId`, `transactionId`, or `wasmwrite_batchId`, and requires `txId` / `wasmwrite_txId`.

Artifact source fingerprints exclude only the known generated target-manifest filenames. Near-miss, attacker-named, and malformed manifest files remain fingerprint inputs. This prevents stale ignored files in a reused worktree from changing a supposedly source-only fingerprint while preserving tamper sensitivity.

## Storage boundary

The underlying committed identity remains the existing Rust `TxId`/`TransactionId`. No row-history, transaction, wire, or persisted storage layout changes are intended.

## Verification

- fresh clean-checkout NAPI and WASM generation
- archive and worktree source fingerprints match committed expectations
- generated NAPI/WASM APIs are `txId`-only
- public API typecheck and Rust transaction-ID receipts
- NAPI/WASM runtime binding suites
- planted stale generated-output and provenance near-miss receipts
- independent adversarial review and exact stack-ancestry audit


aeplay:

PR #2228, opened 2026-08-27
perf(jazz): scope parent validation to the physical row

aeplay:
🤖 Reconstructs #2123 directly on current `main`, preserving its commits while completing the exact physical-row performance and correctness boundary.

## Before

Parent validation looked up every version from the parent transaction. A W1-shaped transaction containing tasks, comments, and activity therefore made validating one task update scale with all sibling rows. An intermediate table-scoped optimization still iterated every transaction version and cloned every same-table sibling before filtering by row.

## After

The transaction cache is addressable by `(schema alias, logical table, row_uuid)`. Parent lookup narrows to the resolved physical table and exact row before cloning on cache hits and before decoding content/shared-deletion candidates on cache misses. Schema aliases, branch/layer validation, deletion history, and missing/pending-parent behavior remain intact.

## Example

If one transaction inserts task A, 128 other tasks, and thousands of activity rows, a later update to task A materializes exactly one candidate parent version. Task B cannot satisfy the parent merely because it shares the transaction and table, and its version is never cloned for task A validation.

## Governing boundary

Parent lookup is physically `(PhysicalTableId, RowUuid, TxId)`; branch/layer/schema validation remains explicit after lookup. This implements #2122’s O(parents × touched rows) target rather than merely reporting wider transaction cost in a benchmark.

## Verification

- Cache-hit and forced-cache-miss receipts each seed 128 same-table siblings and assert exactly one materialized `VersionRow`.
- Removing/inverting the row predicate makes the exact receipt fail.
- Cross-branch, deletion, alias, and unknown-parent receipts remain green.
- W1 benchmark target compiles.
- Invariant registry, formatting, diff checks, and library/package Clippy pass.

Supersedes #2123 and advances #2122.

### DIFF
--- a/crates/jazz/SPEC/11_branch_views_time_travel.md
+++ b/crates/jazz/SPEC/11_branch_views_time_travel.md
@@ -159,6 +159,13 @@ that coordinate, and a version parent must belong to the same branch key
 (`INV-BVIEW-3`, `INV-BVIEW-4`). An application move is an explicit atomic write to
 the source and destination branch-local rows, not a cross-branch-key parent edge.
 
+Validation addresses an explicit version parent by its `(PhysicalTableId,
+RowUuid, TxId)` across both content and deletion history before comparing its
+`BranchKey`. Schema aliases resolve to the same physical table identity, and a
+missing parent remains a pending causal dependency. A cached parent lookup must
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Anselm approved this after the independent adversarial review and replicated-path repair.

PR #2229, opened 2026-08-27
Authorize writes relayed through an Edge

### LINKED ISSUE #2095: Local NAPI edge does not relay writes to configured upstream core

aeplay:
🤖 Reconstructs the unique authorization fix from Joe’s #2198 directly on current `main`, preserving his authorship while excluding stale stacked durability, claims, specification, and release-history changes.

## Before

An Edge authenticated its Core control-plane connection with the admin secret. Ordinary application writes relayed over that connection consequently inherited `TrustedAdmin` and could bypass the transaction author’s row-level write policy.

## After

The credential still authenticates the Edge connection, but ordinary relayed application writes are admitted as `TrustedBackend`. Their original permission subject remains authoritative for policy evaluation. `TrustedAdmin` is reserved for the explicit Core catalogue-bootstrap path.

## Example

A user submits a write that its nearby Edge accepts locally. When the Edge forwards it to Core, the infrastructure credential proves the sender is a trusted relay; it does not turn the user write into an administrative write. Core evaluates the transaction’s original permission subject.

## Verification

- The route admission canary proves ordinary relay links become `TrustedBackend`; a planted `TrustedAdmin` bypass makes it fail.
- The established Edge-offline-then-Core-promotion receipt separately proves real Edge→Core transaction traversal.
- An earlier combined “Core rejection” receipt was removed because the Edge could correctly reject locally, making that test insensitive to Core admission.
- Formatting, diff checks, focused tests, and canonical pre-commit Clippy pass.

Supersedes #2198. Closes #2095.

### DIFF
--- a/crates/jazz-server/src/server/routes/websocket.rs
+++ b/crates/jazz-server/src/server/routes/websocket.rs
@@ -204,10 +204,22 @@ async fn ws_admission(
     if let Some(admin_secret) = auth.admin_secret.as_deref() {
         crate::middleware::auth::validate_admin_secret(Some(admin_secret), &state.auth_config)
             .map_err(|(_, message)| message.to_owned())?;
+        // An admin credential authenticates Edge's control plane, but ordinary
+        // relay commits must retain their transaction permission subject for
+        // application-policy evaluation. Catalogue bootstrap is the sole
+        // privileged admission, and is further checked at the protocol edge.
+        let trust = if prelude.bootstrap_catalogue
+            && peer_identity == AuthorSubject::SYSTEM
+            && state.topology == crate::server::ServerTopology::Core
+        {
+            CommitUnitTrust::TrustedAdmin
+        } else {
+            CommitUnitTrust::TrustedBackend
+        };
         return Ok(WebSocketAdmission {
             identity: peer_identity,
             claims: BTreeMap::new(),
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Approved by Anselm after the repaired security boundary and independent adversarial re-review.

PR #2197, opened 2026-08-27
Harden Edge write durability completion

joeinnes:
## Summary

- Add the invariant that a Global write wait completes only after `Fate::Accepted`, Global durability and an authority-assigned `GlobalTime` are all present.
- Release an Edge outbox entry only for a direct receipt from its currently admitted authority link, and let an exact inbound canonical commit unit replace a missing-body or reconstructed same-transaction entry.
- Require fresh exact-`SubscriptionKey` coverage for reattached one-shot remote reads and select canonical physical history on history-complete nodes.

## Before and after

Before, transport-level Global progress could satisfy a wait without proving authority acceptance, and a stale or reconstructed same-transaction entry could survive in the outbox. A detached subscription receipt could also cover a later equal-shape attachment.

After, acceptance, durability and authority time form one completion boundary. Outbox release is bound to the admitted authority session, canonical payloads win, and each reattachment proves its own remote coverage.

## Scope and trade-offs

This PR builds on #2178 and #2181. Their action-proof, admission, shared transaction/history/outbox, reconnect and backpressure mechanics remain authoritative; this PR adds only the completion and freshness constraints above.

`ReadTier::Remote`, `ReadTier::RemoteIfPossible` and deprecated read APIs keep their existing lowering. Only write-wait `EdgeServer` lowering changes. Links without negotiated authority context retain the existing direct-receipt behaviour for compatibility.

## Testing

- `cargo test -p jazz --lib pending_global_state_does_not_complete_remote_wait_or_prune_upload` — passed
- `cargo test -p jazz --lib edge_write_before_upstream_admission_binds_and_redrives_fate_route` — passed
- `cargo test -p jazz --lib stale_same_authority_session_cannot_settle_or_forward_a_routed_fate` — passed
- `cargo test -p jazz --lib late_detached_view_update_does_not_cover_equal_shape_reattachment` — passed
- `cargo test -p jazz --features runtime,testing --lib read_tier_lowers_without_changing_write_durability` — passed
- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/native-runtime/runtime.test.ts` — 112 passed, 1 skipped
- `pnpm --filter jazz-tools check` — passed

Part of #2095.

PR #2230, opened 2026-08-27
Harden Edge write durability completion

### LINKED ISSUE #2225: Nonterminal Global fate receipts prune the Edge outbox
🤖 The Edge write-durability stack currently releases a canonical upload after any eligible current-authority `FateUpdate` with Global durability, even when the fate is still `Pending` or an accepted fate lacks `global_time`. Those receipts do not satisfy a Global wait, so pruning leaves the waiter pending and removes the upload needed for reconnect replay.

Release an Edge outbox entry only after an eligible current authority reports either a terminal rejection or `Accepted` with Global durability and a concrete `global_time`.

Add a peer-level planted receipt for `Pending / Global / None`, followed by reconnect, proving the outbox is retained and retransmitted. Also prove `Accepted / Global / None` remains replayable until the time-bearing terminal acceptance arrives.

aeplay:
🤖 Reconstructs Joe’s Edge write-durability work directly on current `main`, preserving his original commit and adding the missing nonterminal-fate and authority boundaries from #2225.

## Before

An Edge could prune a queued transaction after seeing Global durability even when the authority had not accepted it or assigned `global_time`. A routed write could also be released by a featureless upstream with no authenticated authority context. During reconstruction, backend relay links were additionally mistaken for Edge authority sessions, causing offline Edge-local writes to defer on a scope receipt that could not exist until Core returned.

## After

A routed outbox entry is released only by an exact receipt from its currently selected admitted authority, and only when it is either:

- an explicit rejection, or
- accepted with Global-or-stronger durability and an authority-assigned `global_time`.

Pending, Local/Edge-only, stale-session, featureless/unnegotiated, and accepted-but-timeless receipts retain the upload. Only Session links participate in Edge authority-scope selection; `TrustedBackend` relay admission never impersonates an end-user Edge authority.

## Examples

An Edge receives `Accepted + Global` without time and retains the commit. A featureless transport then supplies time-bearing acceptance; it still cannot release the routed upload. The matching admitted Core authority can release it with time-bearing acceptance or explicit rejection.

Separately, Bob writes through an Edge while Core is down. The trusted backend link authenticates infrastructure but does not become the row’s Edge authority. Bob’s write settles locally, remains queued, and promotes when Core reconnects instead of deadlocking on an unavailable scope receipt.

## Verification

- Upload-retention, featureless-upstream, and offline-then-promote receipts pass.
- All 37 peer-admission/fate tests pass.
- Planted Global-only, `is_none_or`, and broad role-based authority predicates fail or reproduce the timeout.
- Exact offline receipt passes repeatedly after restoring Session-only selection.
- Invariant registry, formatting, diff checks, library Clippy, and hooks pass.

Supersedes #2197 and fixes #2225. Part of #2095.


aeplay:
🤖 Anselm approved this after the final adversarial review and offline trusted-backend repair.

PR #2231, opened 2026-08-27
perf(jazz): avoid covered resume member clones

aeplay:
🤖 Reconstructs #2126 above replacement #2228, preserving the W1 byte-wire resume benchmark and the covered-resume allocation cleanup on the current core.

## Before

Resumed maintained delivery cloned the complete retained `BTreeSet<ResultMemberEntry>` and collected a vector before checking whether the receiver’s cursor already covered canonical settlement. On the covered path, that allocation was immediately discarded.

## After

The retained set is borrowed for duplicate/removal decisions. Cursor coverage is resolved first; members are cloned only for reset delivery or when their settlement position is newer than a partial cursor. Wire membership, settlement, authorization, and resume semantics are unchanged.

## Example

A client reconnects with a cursor at canonical settlement after subscribing to 500 tasks. The publisher returns no redundant additions/removals and does not first clone all 500 members. With a partial cursor, only members strictly newer than the cursor are emitted. An authorization mismatch still forces the existing complete reset.

## Benchmark

The W1 Divan/CodSpeed receipt prepares database setup, seed writes, initial snapshot, cursor, and a disconnected one-row update outside the timed closure. The measured path is reconnect, catch-up, client application/materialization, and delta observation. The original local receipt showed roughly a 7% improvement at the 4,000-row/500-task profile, identifying this as a useful allocation cleanup rather than the dominant resume cost.

The separate `maintained_rehydrate_scaling` benchmark has a valid timed boundary and equivalence checks, but does not exercise this fast-cursor branch and is not claimed as evidence for this allocation win.

## Verification

- Direct maintained-publisher receipts prove:
  - a fully covered cursor emits no additions/removals and no reset;
  - a partial cursor emits exactly the post-cursor member;
  - authorization mismatch retains reset behavior.
- Changing the strict `settled > cursor` boundary to `>=` makes the covered receipt fail.
- Existing direct and byte-wire resume receipts pass.
- Rust formatting, library Clippy, benchmark compilation, and diff checks pass.

Supersedes #2126. Tracks #2124; setup-size work remains #2125.

### DIFF
--- a/crates/jazz/SPEC/11_branch_views_time_travel.md
+++ b/crates/jazz/SPEC/11_branch_views_time_travel.md
@@ -159,6 +159,13 @@ that coordinate, and a version parent must belong to the same branch key
 (`INV-BVIEW-3`, `INV-BVIEW-4`). An application move is an explicit atomic write to
 the source and destination branch-local rows, not a cross-branch-key parent edge.
 
+Validation addresses an explicit version parent by its `(PhysicalTableId,
+RowUuid, TxId)` across both content and deletion history before comparing its
+`BranchKey`. Schema aliases resolve to the same physical table identity, and a
+missing parent remains a pending causal dependency. A cached parent lookup must
+be row-addressable: unrelated rows from the same transaction, including
+same-table siblings, must not be materialized merely to validate a parent.
+
 #### Storage and indices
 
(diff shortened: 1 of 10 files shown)


aeplay:
🤖 Anselm approved this after the independent adversarial review.

PR #2227, opened 2026-08-27
Expose native transport terminal completion

aeplay:
🤖 Reconstructs Joe’s #2195 directly on current `main`, preserving his original commit and authorship while removing obsolete integration-stack ancestry and repairing the terminal ownership race.

## Before

An established native WebSocket could close while idle without its owner learning that the pump stopped. The initial reconstruction exposed a terminal future, but `Drop` aborted the only task holding its sender: owner drop became a synthetic failure, and a peer-close/drop race could lose the real outcome. `Closed(String)` also conflated local ownership with remote closure.

## After

Every connected native transport exposes one exactly-once terminal future with typed outcomes:

- `OwnerDropped`: the transport owner deliberately released it;
- `PeerClosed(reason)`: the remote peer/socket closed cleanly;
- `Failed(error)`: invalid input or I/O/pump failure.

A shared take-once publisher outlives the abortable pump. `Drop` publishes `OwnerDropped` before aborting; a concurrent peer outcome can win only once. Bootstrap failure remains a separate connection result.

## Behavioral boundary

This PR establishes terminal completion and classification only. Reconnect/backpressure policy is stacked separately above it.

## Verification

- Exact idle peer-close and explicit owner-drop receipts pass.
- A planted missing terminal publication yields synthetic failure and makes the strict peer-close receipt fail.
- Full native-transport tests, formatting, and Clippy pass.
- Bounded producer wake behavior remains covered.

Supersedes #2195. Part of #2095.

PR #2196, opened 2026-08-27
Own the Edge upstream reconnect lifecycle

joeinnes:
## Summary

- Make the server builder own one cancellable Edge upstream connector across bootstrap, connection, terminal observation and reconnect.
- Classify retryable setup failures, reconnectable established-session failures and fatal protocol failures, with bounded exponential backoff and visible health state.
- Preserve authenticated-session replay state, relay subscription ownership and queued writes across a valid replacement connection; shutdown cancels pending bootstrap and reconnect work.

## Before and after

Before, the connector returned after the first successful attachment. An established upstream closure could leave the Edge running against a dead session, while reconnect and shutdown ownership were split across tasks.

After, one lifecycle owns the session until shutdown. Retryable failures reconnect, fatal protocol failures remain visible, and replacement connections receive only state belonging to the authenticated destination.

## Testing

- `cargo test -p jazz-server idle_upstream_terminal_reconnects_and_reaches_connected_again` — passed
- `cargo test -p jazz --lib current_session_claims_reach_late_and_reconnected_upstreams` — passed
- `cargo test -p jazz --lib scope_intent_retries_after_upstream_reconnect` — passed
- `cargo test -p jazz --lib reconnect_replays_live_scope_waiters_once_and_drops_cancelled_ones` — passed
- `cargo test -p jazz --lib rejected_upload_is_not_replayed_after_reconnect` — passed
- `cargo test -p jazz --features testing --test browser_relay_durability worker_relay_` — 6 passed

Part of #2095.

PR #2232, opened 2026-08-27
Own Edge upstream reconnect and retry backpressure

### LINKED ISSUE #2224: Native upstream backpressure drops queued semantic and chunk frames
🤖 The native upstream driver introduced in the #2195–#2198 Edge stack removes ordinary semantic frames and auxiliary chunk traffic from their queues before calling the bounded wire transport. If `send_frame` returns `Backpressure`, the driver treats that as terminal and reconnects without restoring either obligation.

This violates the existing pump ownership contract: a failed send must retain or restore the exact semantic obligation until delivery succeeds or ownership is explicitly transferred. Losing the frame can drop writes, chunk requests, or chunk responses precisely when the transport is temporarily full.

Repair the reconstructed reconnect PR so backpressure retries on the same live transport. Preserve ordinary frames at the front of `WireQueues`, and give auxiliary traffic a reservation/commit-or-restore handoff. Add planted receipts for first-send backpressure on ordinary frames, chunk requests, and chunk responses, proving exactly-once delivery and correct later reconnect behavior.

aeplay:
🤖 Reconstructs Joe’s #2196 above #2227, preserving his lifecycle commit while repairing ordinary and auxiliary frame loss on backpressure.

## Before

Edge upstream bootstrap, terminal observation, reconnect, and shutdown were split across tasks. An idle closure could leave a dead session attached. In addition, a `Backpressure` result destructively popped the frame before retry: ordinary semantic frames could disappear, and auxiliary chunk request/response frames could lose their only queued copy.

## After

One cancellable lifecycle owns the selected Edge upstream from bootstrap through terminal observation and reconnect:

- `PeerClosed` and retryable `Failed` outcomes reconnect with bounded backoff;
- `OwnerDropped`/cancellation stop cleanly rather than pretending the peer failed;
- authenticated-session replay, live scopes, queued writes, and shutdown ownership stay with the lifecycle;
- ordinary backpressure pushes the frame back to the front and retries it on the same live connection;
- auxiliary frames use an RAII reservation, committed only after successful send and automatically restored on backpressure, failed send, or dropped attempt.

## Examples

If the socket queue is temporarily full while sending a commit, the same connection is retried with that exact commit still first in line; reconnect is not manufactured merely for flow control.

If a chunk response is reserved and the send backpressures, dropping the attempt restores the response. A later pump can still serve it exactly once.

## Verification

- Idle peer-close reconnect and owner-versus-peer terminal mapping receipts pass.
- Ordinary and auxiliary backpressure retry receipts pass.
- Chunk request/response lifecycle receipts pass.
- Planted ordinary-frame drop and suppressed RAII restore each make their focused receipt fail.
- Formatting, Jazz/server/native-transport Clippy, and commit hooks pass.

Supersedes #2196. Part of #2095; fixes #2224.

### DIFF
--- a/.changeset/edge-upstream-reconnect-lifecycle.md
+++ b/.changeset/edge-upstream-reconnect-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Own the edge server's upstream connector across bootstrap, retry, established disconnect, and shutdown. Reconnect transient failures with bounded exponential backoff, expose fatal lifecycle health, detach the exact upstream peer before retry, and cancel and join connector work before storage closes.
--- a/crates/jazz-server/src/lib.rs
+++ b/crates/jazz-server/src/lib.rs
@@ -6,8 +6,8 @@ pub mod server;
 
 pub use middleware::AuthConfig;
 pub use server::{
-    BuiltServer, ServerBuilder, ServerState, ServerTopology, ShutdownController, ShutdownPhase,
-    StorageBackend,
+    BuiltServer, EdgeUpstreamHealth, ServerBuilder, ServerState, ServerTopology,
+    ShutdownController, ShutdownPhase, StorageBackend,
 };
 #[cfg(feature = "embedded-server")]
 pub use server::{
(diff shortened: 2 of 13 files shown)


aeplay:
🤖 Superseded by #2234, which carries the same independently reviewed feature commits directly on main after GitHub auto-closed this stacked PR when its parent branch was deleted.

PR #2233, opened 2026-08-27
Wake browser waits on terminal worker transport errors

### LINKED ISSUE #2207: Persistent browser worker masks terminal mutation errors
🤖 A server-side protocol or mutation error can be converted into an indefinitely `Pending/Local` write by the persistent browser-worker path instead of being surfaced to the initiating runtime.

This was observed while diagnosing the Wequencer Edge/Core topology: a malformed maintained-subscription terminal record produced `Protocol: maintained view terminal missing required field __root_join_arm_0` in a direct-memory reproduction, while the persistent worker left the corresponding edge fate wait pending until the test timeout. #2206 fixes that particular descriptor mismatch, but the transport/error propagation boundary remains unsafe for future failures.

Expected contract: terminal server/protocol rejection must deterministically settle or reject the corresponding client mutation and wake all relevant waiters. A worker reconnect must not silently convert a known failure into an unbounded liveness wait.

Add a focused worker receipt that injects a terminal server error, proves the caller observes it, and verifies no pending fate/waiter survives reconnect or peer teardown.

aeplay:
🤖 Fixes the persistent-worker liveness boundary from #2207 without turning transport failure into transaction rejection or durable notification history.

## Before

A direct runtime records a terminal upstream server/protocol error and wakes active Edge/Global waits. In the persistent browser topology, that direct runtime lives inside the worker. The worker had no error path to already-attached tabs, so a tab could remain forever inside `WriteHandle.wait(Edge|Global)` after its upstream had terminally failed. A later reconnect did not reliably wake that existing wait.

## After

The worker sends a distinct transient `transport-error` event only to currently initialized peers. The follower records it in its tab-side `NativeRuntimeAdapter` before disposal, which rejects active Edge/Global waits and remote subscriptions using the same terminal-error machinery as a direct runtime.

## Behavioral boundaries

- The error is not `Fate::Rejected`, does not roll back state, and does not invoke `onMutationError`.
- A locally durable write and Local wait remain valid.
- A real authoritative rejection still uses `PersistedWriteRejectedError`.
- A runtime attached later receives reconciled durable state but no replayed historical toast/error.
- Ordinary offline/disconnected state is not treated as terminal protocol failure.

## Example

A tab writes locally through its persistent worker and begins waiting for Edge settlement. Core then returns a terminal protocol error before fate. The worker notifies that attached tab; its active Edge wait rejects promptly with the transport error. If the tab was closed and a new tab opens later, the new tab sees the worker’s corrected durable state but does not receive an old error notification.

## Verification

- Worker broadcast, follower bridge, active wait/subscription, Local preservation, fate-rejection distinction, reconnect race, and no-replay receipts pass (32 focused tests).
- Planted missing worker/follower propagation makes the focused receipts fail.
- Jazz Tools typecheck passes.
- Normative topology/API note records foreground notification versus durable reconciliation.

Fixes #2207.

### DIFF
--- a/crates/jazz/SPEC/13_db_api.md
+++ b/crates/jazz/SPEC/13_db_api.md
@@ -55,6 +55,13 @@ Invariant digest:
   does not weaken the exact node-and-author definition above outside the paired
   browser client/worker boundary.
 
+  A terminal server or protocol transport failure is not an authority fate. A
+  durable browser worker MUST relay it only to currently initialized foreground
+  peers so their active Edge/Global waits and remote subscriptions reject with
+  that transport error; Local durability remains valid. The worker MUST NOT
+  fabricate `Rejected`, roll back local data, invoke `onMutationError`, or
+  replay that transient foreground error to a peer attached later.
+
 ## Details
 
 ### 13.1 Two audiences
--- a/packages/jazz-tools/src/runtime/native-runtime/browser-follower-connection.test.ts
+++ b/packages/jazz-tools/src/runtime/native-runtime/browser-follower-connection.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it, vi } from "vitest";
+import { MessagePortBrowserFollowerConnection } from "./browser-follower-connection.js";
+import type {
+  BrowserFollowerPortEvent,
+  BrowserFollowerPortRequest,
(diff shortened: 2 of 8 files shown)


aeplay:
🤖 Anselm approved this after the final independent review.

PR #2234, opened 2026-08-27
Own Edge upstream reconnect and retry backpressure

### LINKED ISSUE #2224: Native upstream backpressure drops queued semantic and chunk frames

aeplay:
🤖 Restacks the independently reviewed #2232 directly on `main` after merging its parent #2227. GitHub auto-closed the original child when the deleted parent branch disappeared; this replacement contains the same reviewed feature commits plus the merge of current `main`.

## Before

Edge upstream bootstrap, terminal observation, reconnect, and shutdown were split across tasks. An idle closure could leave a dead session attached. A `Backpressure` result also destructively popped the frame before retry: ordinary frames could disappear, and auxiliary chunk request/response frames could lose their only queued copy.

## After

One cancellable lifecycle owns the selected Edge upstream from bootstrap through terminal observation and reconnect:

- `PeerClosed` and retryable `Failed` reconnect with bounded backoff;
- `OwnerDropped`/cancellation publish `Stopped` health and stop without reconnect;
- authenticated-session replay, live scopes, queued writes, and shutdown stay with the lifecycle;
- ordinary backpressure restores the frame to the front and retries on the same live connection;
- auxiliary frames use RAII reservation, committed only after successful send and restored on backpressure, failed send, or dropped attempt.

## Examples

A temporarily full socket queue keeps the exact commit first in line and retries the same connection. A reserved chunk response that backpressures is restored so a later pump can serve it exactly once.

## Verification

- Idle peer-close, owner-drop, cancellation/Stopped health, and reconnect receipts pass.
- Ordinary and auxiliary backpressure receipts pass.
- Planted ordinary-frame loss, suppressed auxiliary restore, and stale Connected health each fail.
- Final independent adversarial re-review found no remaining issue.
- Formatting, Jazz/server/native-transport Clippy, and hooks pass.

Supersedes auto-closed #2232 and original #2196. Part of #2095; fixes #2224.

### DIFF
--- a/.changeset/edge-upstream-reconnect-lifecycle.md
+++ b/.changeset/edge-upstream-reconnect-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Own the edge server's upstream connector across bootstrap, retry, established disconnect, and shutdown. Reconnect transient failures with bounded exponential backoff, expose fatal lifecycle health, detach the exact upstream peer before retry, and cancel and join connector work before storage closes.
--- a/crates/jazz-server/src/lib.rs
+++ b/crates/jazz-server/src/lib.rs
@@ -6,8 +6,8 @@ pub mod server;
 
 pub use middleware::AuthConfig;
 pub use server::{
-    BuiltServer, ServerBuilder, ServerState, ServerTopology, ShutdownController, ShutdownPhase,
-    StorageBackend,
+    BuiltServer, EdgeUpstreamHealth, ServerBuilder, ServerState, ServerTopology,
+    ShutdownController, ShutdownPhase, StorageBackend,
 };
 #[cfg(feature = "embedded-server")]
 pub use server::{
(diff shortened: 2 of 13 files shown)


aeplay:
🤖 Anselm approved the reviewed #2232. Carrying that approval onto this exact replacement/restack; no feature behavior changed.

PR #2235, opened 2026-08-27
Replace postcard branch-key storage encoding

nrainhart:
## Problem

BranchKey encoded each BranchColumnValue through postcard over the general-purpose Groove Value enum, then postcard-encoded the surrounding key again. Those bytes are persisted in rows and used as physical key prefixes, so the durable ordering and identity contract depended on Rust and serde enum layout.

## Solution

Replace that representation with an explicitly versioned canonical branch-key codec. Permanent numeric tags identify branch-key components, while each value is encoded according to its declared Groove column type. Rust and TypeScript now produce the same format and reject type mismatches or non-canonical encodings.

Branch selection semantics are unchanged. This is an intentional pre-v1 storage cut: the old postcard representation is not accepted as an alternate spelling.

### DIFF
--- a/crates/jazz/SPEC/11_branch_views_time_travel.md
+++ b/crates/jazz/SPEC/11_branch_views_time_travel.md
@@ -138,6 +138,43 @@ branch-specific declaration or binding abstraction.
 The empty branch key denotes shared data. It is not a privileged root branch key and
 has no lifecycle semantics (`INV-BVIEW-7`).
 
+#### Canonical branch-coordinate codec
+
+Branch coordinates use an engine-owned binary codec, not the serde layout of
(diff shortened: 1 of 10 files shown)


aeplay:
🤖 We completed the authorized repair directly on this branch with additive commits only. Persisted branch-key decoding is now schema-scoped across ordinary/shared history, current/deletion, and maintained-witness recovery. New Memory/Rocks receipts cover malformed envelopes, wrong scalar/enum domains, local ahead projections, explicit Global promotion, and reopen. A planted schema-validation bypass makes the corruption fixture fail; final independent rereview found no remaining issue. Fresh stack CI is now running.

aeplay (approved):
🤖 Approved by Anselm after the repaired branch-key and physical-column-ID changes passed independent adversarial review and fresh CI.

PR #2236, opened 2026-08-27
Store authored columns as physical column IDs

nrainhart:
## Problem

Every applicable history, current, and ahead-current row stored authored_columns as JSON containing logical column names. This repeated strings across rows and coupled local provenance metadata to logical names that can change across schema versions and lenses.

## Solution

Declare authored_columns as a nullable Groove array of physical U64 column IDs. IDs are stored in strictly increasing order, resolved back to logical names only at schema and wire boundaries, and propagated consistently through ingest, projection, merge, and materialization paths.

An absent value retains the existing legacy or lens fallback in which present cells are treated as authored. The former JSON-in-Bytes representation is deliberately not accepted after this pre-v1 storage change.


aeplay:
🤖 Authorized repair is now pushed additively, including the repaired #2235 parent. Persisted authored-column IDs are validated against the exact row schema before ahead/global derived writes; zero/unknown corruption fails before mutation. New receipts cover raw reopen corruption, cross-node remapping, and a title-to-name lens rename while retaining physical identity. A planted validation bypass fails the corruption receipt; final independent rereview found no remaining issue. Fresh stack CI is running.

aeplay (approved):
🤖 Approved by Anselm after the repaired branch-key and physical-column-ID changes passed independent adversarial review and fresh CI.

PR #2237, opened 2026-08-27
Store contribution provenance as Groove records

nrainhart:
## Problem

The transaction table stored contribution_merge as postcard bytes containing nested Rust structs and enums. That made durable contribution provenance depend on serde field order and Rust enum layout even though its complete physical shape is known to the schema.

## Solution

Declare contribution_merge as a nullable Groove record composed of typed nested records, arrays, transaction IDs, and a payload enum with permanent component tags. Encoding and decoding now use the centralized schema descriptors and reject malformed or non-canonical provenance.

Merge behavior is unchanged; only the durable representation changes. The old postcard byte payload is not retained as a second storage spelling.


aeplay (approved):
🤖 Approved by Anselm after the physical table/column identity repairs, complete byte golden, independent adversarial review, and fully green CI.

PR #2238, opened 2026-08-27
Encode contribution operation identities with Groove types

nrainhart:
## Problem

After contribution provenance became a Groove record, operation components still carried opaque postcard identities. In particular, G-Set operations postcard-encoded a Rust tuple of column name and element value, leaving an identity-bearing field coupled to serde layout and obscuring its declared column type.

## Solution

Give operation components explicit column and identity fields. Counter operations identify their column directly, while G-Set element identities are encoded and decoded through a Groove record derived from the declared array element type. Canonical re-encoding rejects alternate or malformed identities.

Contribution and merge semantics remain unchanged; this only replaces the operation identity representation.


aeplay (approved):
🤖 Approved by Anselm after centralized durable-ingress admission repairs, planted-sensitive receipts, independent adversarial re-review, and fully green CI.

PR #2239, opened 2026-08-27
Replace postcard large-value metadata encoding

nrainhart:
## Problem

Large-value NodeRef identities under the root, node, reclaim, and install key namespaces were postcard-encoded, and the corresponding reference, staging, and pending-upload metadata values also used serde and postcard. Both physical keys and recovery state therefore depended on Rust layout.

## Solution

Reuse the canonical Groove NodeRef record already used inside LargeValueRef for all metadata key namespaces. Encode metadata values with an explicit GLVM version envelope, permanent record-type tags, and declared Groove descriptors for root references, node references, staged values, and pending uploads. Decoders validate fixed-width identities, canonical child ordering, record shape, and canonical re-encoding.

These two commits form one storage-contract layer: the first centralizes NodeRef key encoding and the second migrates the associated metadata values. Legacy postcard metadata is intentionally not an alternate encoding.


aeplay (approved):
🤖 Approved by Anselm after standard Groove record conversion, physical permanent-slot repairs across metadata and nested large-value records, reclaim key/value binding, independent adversarial re-review, and fully green CI.

PR #2240, opened 2026-08-27
Store merge heads as Groove transaction ID arrays

nrainhart:
🤖 ## Before

`jazz_merge_heads.heads` was an opaque postcard-encoded Rust `Vec<TxId>`. Its durable key was not explicitly frozen as the complete content-row coordinate.

For example, concurrent heads `(10, node-a)` and `(10, node-b)` were hidden inside bytes whose interpretation depended on Rust/serde layout.

## After

- `heads` is a normal Groove `Array<Tuple<U64, Uuid>>`.
- Values must be strictly increasing and duplicate-free.
- The derived content-frontier key is exactly `(PhysicalTableId, canonical BranchKey, RowUuid)`; deletion remains a separate register layer.
- Two non-default branches using the same table and row retain two independent frontier rows across reopen.
- The storage-epoch manifest rejects former postcard roots before any ambiguous 24-byte payload can be decoded as a valid transaction ID.
- The format and receipts are recorded as `INV-HIST-18`.

Example: `[(10, node-a), (10, node-b)]` is stored directly in canonical order; replaying `node-a` does not append a duplicate. Malformed, duplicate, out-of-order, or epoch-incompatible state fails closed before it affects merging.

## Unchanged

The frontier remains node-local derived metadata. Immutable content history is authoritative and can rebuild it; transaction identity and merge semantics are unchanged.


aeplay (approved):

PR #2241, opened 2026-08-27
Replace postcard IDB page encoding

nrainhart:
🤖 ## Before

IDBTree pages used a versioned/checksummed envelope around postcard payloads. Page tags and fields followed serde/Rust layout, overflow lengths were host-sized, malformed cross-page graphs could loop, and point updates could overwrite pages reachable from a previously published root.

A failed oversized write could also advance the allocation frontier and leave dirty orphan pages that a later flush persisted.

## After

- Canonical IDBTree v2 bytes use permanent page/value tags, fixed-width little-endian integers, bounded lengths, checksum validation, exact Rust/TypeScript fixtures, and no postcard.
- Overflow logical lengths are fixed `u64`; materialization grows only from validated bytes rather than preallocating from an untrusted durable length.
- Traversal rejects structural and overflow cycles, shared subgraphs, oversized pages, duplicate children, trailing bytes, invalid IDs, and malformed metadata.
- Writes publish immutable leaf-to-root COW closures. A reader retaining the old root continues to see the old tree until the new root is atomically published.
- Browser commits validate the complete page/metadata batch before any IndexedDB mutation.
- Failed `put` and `write_many` operations restore the root, allocation frontier, dirty pages, and deletion state exactly; later flush/reopen cannot expose orphan pages.
- Old closures are not eagerly reclaimed in this PR.

## Worked examples

- Updating `key = old` to `key = new` allocates a new path. Before publication, a reopened reader sees `old`; after atomic publication it sees `new`.
- A checksum-valid overflow page pointing to itself returns a typed corruption error instead of spinning.
- An oversized key/value that cannot fit returns `PageTooLarge`; `next_page_id` and the durable store remain unchanged.
- A `write_many` batch rolls back earlier successful operations if a later operation fails.

## Compatibility

This intentionally establishes the zero-backcompat v2 browser storage format. Unsupported earlier roots fail closed via the storage epoch/page-format boundary.


aeplay (approved):

PR #2242, opened 2026-08-27
Replace postcard result-member key encoding

nrainhart:
## Problem

Settled ResultMemberEntry values were postcard-encoded and embedded as Bytes in physical keys. Their identity includes member variants, row layers, recursive read sources, occurrences, transaction IDs, and optional provenance, so a Rust or serde layout change could silently change persisted key identity or make recovery fail.

## Solution

Add explicit JRME and JRSE v1 envelopes backed by declared Groove records and payload enums. Member, source, and layer variants have permanent tags; nested transaction and occurrence identities reuse canonical Groove representations; recursive sources are bounded and versioned; and decoding requires exact canonical re-encoding.

Settled-key writes and durable recovery now share this codec. Protocol wire serialization and in-memory footprint estimates remain unchanged. Golden fingerprints cover every member tag, all source variants round-trip, and malformed, legacy, wrong-version, and trailing encodings fail closed.


aeplay (approved):
🤖 Approved by Anselm after canonical result-member/source encoding review, non-default read-view and permanent-tag freeze receipts, planted sensitivity, and fully green CI.

Issue #2160, opened 2026-08-26
Complete the storage-epoch-1 compatibility corpus audit


nrainhart:
🤖 Final alpha.54 storage source/corpus audit completed at `a848ea51ef2fbe5e697e17e17fe13fd8dee4a2c7`, with release metadata candidate `193f5894d1d1f305fd2045b4a149c0305a93ab13`.

## Settled contract

Storage epoch 1 starts at the selected settlement baseline; superseded unreleased alpha stores are unsupported. Authoritative encodings within the settled epoch are immutable. Future incompatibility requires an explicit epoch/migration boundary, never fallback guessing. Storage epoch, IndexedDB schema generation and wire/binding versions remain distinct.

## Final integrated audit

Typed physical identities (#2558), inactive result-persistence removal (#2580), dense JPFK V1 tags (#2596), V1 normalization (#2588), complete browser historical-root coverage (#2546/#2561), and canonical CI partition coverage (#2564/#2586) are merged. #2548 is abandoned.

The reviewed corpus and producers/consumers cover Groove record/scalar/ordered-key and physical profiles; SQLite/RocksDB manifests and complete native Jazz roots; IndexedDB manifest/pages and complete browser Jazz roots; catalogue/kernel/schema/activation/lineage/physical-mapping records; branch/current/history/version/transaction/provenance/merge-head state; large-value objects; retained source/program-fact keys; and local identity-secret codecs. Exact-byte and malformed/unknown-format cases remain part of the existing compatibility gates. Historical fixtures were not regenerated merely to hide encoder drift.

Wire and binding contracts are recorded separately: recursive typed Postcard Query/Shape and native-direct query carriers (#2597/#2601), preserved native row/cell encodings and ResultKey V1, and the deliberately retained JSON WebSocket admission prelude (#2599). The explicit inventory and retain/replace decisions are in #2592.

## Verification

- Final release metadata candidate: all CI partitions passed, including native/browser storage compatibility, exhaustive Rust workspace, differential, TypeScript and RN bridge: https://github.com/garden-co/jazz/actions/runs/33998110029.
- Main settlement source storage-compat partition passed: https://github.com/garden-co/jazz/actions/runs/33998090245/job/101392015921.
- Final query integration and documentation CI passed: https://github.com/garden-co/jazz/actions/runs/33997140296 and https://github.com/garden-co/jazz/actions/runs/33997672379.
[...]

PR #2243, opened 2026-08-27
Freeze native storage epoch manifests

aeplay:
🤖 Freezes the epoch-one storage manifest boundary and pins the native RocksDB/SQLite physical formats before we settle storage compatibility.

## Before

- Existing stores had backend-local markers, but no shared epoch/codec manifest or explicit adjacent-migration contract.
- SQLite could adopt a table-free foreign file and overwrite its header identity.
- RocksDB inspection failures could be collapsed into a fresh-store path.
- Future migration registration could ambiguously define multiple transitions from one epoch.

## After

- A canonical JSM1 manifest pins epoch, adapter identity/version, codec IDs, and format parameters.
- Unknown, corrupt, missing, or inconsistent manifests fail before ordinary data is mutated.
- Migrations are explicit and adjacent-only, with one unambiguous transition per source epoch.
- SQLite pins application_id, user_version, exact DDL/meta identity, and rejects either independently foreign header field before WAL setup.
- RocksDB pins its reserved namespace/raw marker/comparator contract and rejects malformed or uninspectable existing roots before create/open.

## Examples

- A table-free SQLite file with only a foreign application_id is rejected byte-for-byte unchanged; the same is true when only user_version is foreign.
- A RocksDB directory with a malformed CURRENT file is rejected without creating families or markers.
- Registering two epoch 1 → 2 migrations is rejected instead of depending on registration order.

## Verification

- Final independent adversarial review found no remaining issues.
- Planted weakening of each SQLite header guard, Rocks inspection boundary, and migration uniqueness made its receipt fail.
- Groove manifest tests: 6 passed.
- SQLite ordered-KV/storage-format tests: 11 passed.
- RocksDB storage-format tests: 17 passed.

Part of #2160 and #2040.


aeplay:
🤖 Anselm approved this after the final adversarial review and green CI.

Issue #2043, opened 2026-08-25
Freeze HLC, UUID, and alias persistence semantics


aeplay:
🤖 Review `crates/jazz/src/time.rs`, ID types, alias records in `node::codec`, allocation/recovery paths, and every ordering/key conversion crossing HLC, UUID, schema/node alias, or public timestamp domains.

Pin fixture bytes and boundary behavior for UUID order, millisecond HLC packing/counters/overflow, public provenance reconstruction, alias collision and restart, aliases retained by old history, and UUIDv7 or caller-supplied IDs. Add cross-node replay tests where local alias assignments differ.

Relate any row-version CAS/freshness work such as #1990 to these frozen durable coordinates.

aeplay:
🤖 Audit refinement: durable aliases are represented as `u64`, but schema aliases later narrow into Groove `u32` variant tags. The fallible conversion exists; freeze coverage should include allocation exhaustion, reopen/recovery at the boundary, corrupt/out-of-range aliases, and proof that no panic-only path remains. HLC fixtures should also pin the current 46-bit millisecond / 18-bit counter packing at both boundaries.

aeplay:
🤖 Storage-settlement decision: authoritative cross-node references use canonical global identities, never storage-local aliases. UUIDs are canonical 16-byte values whose lexicographic byte order is their physical order. Hashes and locators are fixed 256-bit domain-separated digests. HLCs have one canonical millisecond-plus-counter representation and ordering; counter exhaustion advances the logical millisecond by one rather than blocking or widening the format. Node/schema/table/column aliases are strictly local compression: their mapping must be durably established atomically before dependent bytes, retained while any durable reference can reach it, and may differ across replicas. Aliases never enter sync, hashes, semantic equality, or public provenance. Missing, malformed, or colliding mappings fail closed before decode or mutation; implementations never guess or reassign.

aeplay:
[...]

PR #2244, opened 2026-08-27
Freeze storage identity and alias invariants

aeplay:
🤖 Freezes canonical UUID/HLC persistence and makes local alias mappings fail closed before dependent history can use them.

## Before

- Alias state could become resident before its durable catalogue row succeeded, so a recoverable storage failure could leave an alias usable by later history.
- Recovery did not validate schema-version alias mappings bijectively.
- HLC tests asserted Rust numeric ordering rather than the actual ordered-key bytes.

## After

- Node and schema aliases become resident only after ordered persistence succeeds.
- Reopen rejects duplicate, conflicting, out-of-range, or malformed alias mappings before decoding dependent payloads.
- UUID, TxTime, and GlobalTime boundaries and exact ordered primary-key bytes are pinned by fixtures.
- Aliases remain node-local implementation details and do not affect wire identity, equality, provenance, or ordering.

## Examples

- An injected alias persistence failure leaves no resident alias and no dependent durable history; reopen is clean and a later retry succeeds.
- Two schemas claiming one alias, or one schema claiming two aliases, fail closed on reopen.
- GlobalTime exhaustion returns a typed error instead of wrapping.

## Verification

- Final independent adversarial review found no remaining storage-identity issue.
- Planted pre-persistence resident mutation made the new failure receipt fail.
- Focused identity, recovery, invariant, clippy, formatting, and sensitive-data checks passed before restack.
- The current main lib-test target has an unrelated cfg mismatch in settled_authoritative_receipt_counts_for_test; that repair is already active on #2209 and presently blocks rerunning the focused lib tests after restack.

Addresses the storage-freeze portion of #2043.


aeplay:
🤖 Approved by Anselm after independent adversarial review and green CI.

PR #2245, opened 2026-08-27
Compile receipt helpers in library tests

aeplay:
🤖 Aligns the test-only receipt helper cfg boundary so the Jazz library test target compiles on current main.

## Before

Db exposed settled_authoritative_receipt_counts_for_test under cfg(any(test, feature = testing)), but the delegated NodeState method existed only under feature = testing. Plain cargo test -p jazz --lib therefore failed to compile.

## After

Both layers use the same cfg(any(test, feature = testing)) boundary.

## Verification

- cargo test -p jazz --lib --no-run passes.

This is isolated from the larger #2209 repair so current CI and storage settlement branches can use a compilable main immediately.

### DIFF
--- a/crates/jazz/src/node/query_eval/subscriptions.rs
+++ b/crates/jazz/src/node/query_eval/subscriptions.rs
@@ -223,7 +223,7 @@ where
             .sum()
     }
 
-    #[cfg(feature = "testing")]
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved this urgent main test-compile unblocker.

Issue #2175, opened 2026-08-27
Freeze local-first auth secret and recovery phrase formats


aeplay:
🤖 Local-first authentication persists a long-lived identity root outside Groove storage, but its format is not represented by the current storage-freeze issues.

Today:
- `generateAuthSecret()` stores 32 CSPRNG bytes as unpadded base64url;
- browser storage keys are derived as `jazz-auth-secret` plus URI-encoded app/user/session scope segments;
- Expo SecureStore uses the same default key but currently carries a separate implementation;
- recovery phrases map the 32-byte secret through the English BIP-39 wordlist into 24 normalized words.

These bytes and names determine whether an existing installation or human backup can recover the same author identity. Accidental changes to padding acceptance, byte length, scope-key construction, wordlist/version, Unicode normalization, or platform-specific encoding would silently create a new identity or make backups unrecoverable.

Before freezing the format:
- specify the canonical secret bytes/string and accepted decoder forms;
- freeze browser and native storage-key derivation, including escaping and empty segments;
- pin recovery phrase wordlist, normalization, checksum, and secret round-trip behavior;
- share one codec across browser/Expo/native implementations;
- add hard-coded cross-platform fixtures plus historical load/restore tests;
- define explicit migration behavior for any future version rather than changing the stored string in place.

Pointers: `packages/jazz-tools/src/runtime/auth-secret-store.ts`, `packages/jazz-tools/src/runtime/recovery-phrase.ts`, and the Expo auth-secret store.

aeplay:
[...]

PR #2246, opened 2026-08-27
Freeze local-first auth secret formats

aeplay:
🤖 Freezes the local-first identity secret and recovery-phrase contract, removes legacy seed acceptance, and prevents backend admission credentials from crossing client runtime boundaries.

## Before

- Arbitrary unprefixed strings could still mint authors through a legacy fallback.
- Scope-key normalization could collapse distinct app/profile identifiers.
- Empty persisted values could be treated as missing and silently regenerated.
- Browser and React Native client config could retain or serialize an untyped backendSecret.
- React swallowed malformed stored-secret failures into an anonymous-looking state.

## After

- Local-first roots are exactly 32 random bytes encoded as jazz-auth-v1 plus 43 unpadded base64url characters; every present value is strictly parsed.
- BIP-39 English 24-word recovery normalizes phrase input only; app/profile scope identifiers remain byte-exact canonical JSON before hashing.
- Browser/Expo stores reject malformed present values without overwriting them.
- Client runtime factories centrally reject backendSecret before config retention; the explicit trusted-backend context remains supported.
- React exposes typed format failures; RN requires an injected native secure store instead of falling back to browser localStorage.

## Examples

- An empty or padded stored value fails closed and remains byte-for-byte untouched.
- App IDs app and space-app-space select distinct scopes.
- An untyped RN object spread containing backendSecret is rejected before runtime load.
- The same 32 bytes round-trip through canonical root and 24-word recovery phrase.

## Verification

- Final independent adversarial review found no remaining issue.
- Planted truthiness regressions in Browser and Expo make the no-regeneration receipts fail.
- Planted client credential-guard bypass makes the RN boundary receipt fail.
- Focused Browser/Expo auth tests: 28 passed after restack.
- Full TS/RN local checks are presently blocked by the known stale generated WASM ABI artifact on main; fresh CI rebuilds authoritative artifacts.

Fixes the storage-freeze contract in #2175.


aeplay:
🤖 Approved by Anselm after independent adversarial review and fully green CI.

Issue #2038, opened 2026-08-25
Freeze branch, history, deletion, and index coordinates


aeplay:
🤖 Review `protocol::{BranchSelector,BranchColumnValue,BranchKey}`, `BranchKey::canonical_bytes`, `node::codec::history_primary_key`, and `node::physical::{shared_deletion_history_primary_key,version_storage_*}` against the branch/time-travel spec.

Create frozen fixtures and adversarial reopen receipts for selector canonicalization, branch-column evolution/defaults, content versus deletion history, shared deletion routing, head/base fallback, index prefixes, copy-on-write, and provenance. Prove that the same `RowUuid` cannot alias rows, parents, indices, or tombstones across branches.

aeplay:
🤖 Final storage-format audit refinement: branch/history coordinates are semantically durable but currently composed through postcard in crates/jazz/src/protocol.rs. Freeze their canonical byte representation and ordering here; track the broader dependency on opaque serde/postcard under #2158. Include malformed and non-canonical decode rejection fixtures.

aeplay:
🤖 Storage-settlement decision: authoritative cross-node references use canonical global identities, never storage-local aliases. UUIDs are canonical 16-byte values whose lexicographic byte order is their physical order. Hashes and locators are fixed 256-bit domain-separated digests. HLCs have one canonical millisecond-plus-counter representation and ordering; counter exhaustion advances the logical millisecond by one rather than blocking or widening the format. Node/schema/table/column aliases are strictly local compression: their mapping must be durably established atomically before dependent bytes, retained while any durable reference can reach it, and may differ across replicas. Aliases never enter sync, hashes, semantic equality, or public provenance. Missing, malformed, or colliding mappings fail closed before decode or mutation; implementations never guess or reassign.

aeplay:
[...]

PR #2248, opened 2026-08-28
Freeze exact row-version history coordinates

aeplay:
🤖 Addresses #2038.

## Before

Durable history keys and version-parent validation did not consistently freeze the complete physical coordinate. A parent `TxId` could be resolved without proving that it belonged to the same physical table, canonical branch key, row, and content/deletion layer. Pending children could also lose that expected coordinate while waiting for an incomplete parent.

Deletion had a separate semantic leak: deleting a live row used its current content version as the first deletion-register parent. That made `parents` serve both as history ancestry and as a cross-layer dependency.

Example: if content version `C1` is current and a row is deleted, the first deletion version was encoded as `D1.parents = [C1]`, even though `C1` and `D1` belong to different histories.

## After

- A version parent is an exact prior version of the same physical table, canonical branch key, row UUID, and content/deletion layer.
- Mergeable transactions have no general dependency graph. Cross-row or cross-layer transaction causality is not encoded in version parents.
- The first deletion starts deletion-register history with no parent; later delete/restore operations parent only the prior deletion-register head.
- History keys use a frozen canonical branch-key representation and cannot alias default/non-default branches or different physical tables.
- Unknown-parent constraints retain their complete expected coordinate through persistence and reopen.
- Complete parent batches are preflighted atomically, preventing a partially received parent from being accepted and later contradicted.
- Receiver batches and reset ingestion settle pending constraints only after complete parent persistence.

Example: content `C1`, first deletion `D1`, and restoration `D2` are represented as `D1.parents = []` and `D2.parents = [D1]`. `C1` remains content ancestry and never becomes a deletion parent.

## Correctness receipts

Tests cover wrong table, branch, row, and layer; incomplete parent batches; persistence/reopen; receiver/reset settlement; duplicate version rejection; deletion history; and incremental currentness. An adversarial planted mutation removing the layer check made the exact-coordinate regression fail and was restored before publication.

The governing contract is recorded as `INV-HIST-18`; the merge-head storage invariant introduced by #2240 is retained as `INV-HIST-19`.


aeplay:

Issue #2044, opened 2026-08-25
Extend generated-artifact wire and binding compatibility coverage


aeplay:
🤖 Extend direct generated-artifact wire/binding compatibility coverage in alpha.55. Storage freeze is tracked separately by #2160 and #2308.

## Already established

- #2259 pins exact wire/binding decoding and rejection boundaries.
- #2292 triangulates a Rust-owned binding corpus through generated NAPI/WASM artifacts and the production TypeScript decoder.
- Storage epoch, wire version and binding version are independent. Peers negotiate compatible versions or reject before semantic decoding; a shared storage epoch does not imply wire compatibility.

## Remaining work

Broaden direct generated-artifact execution to representative full wire frames, including negotiated features, compression, fragmentation, large values and catalogue/branch payloads. Cover malformed, unsupported, trailing and corrupt-compressed inputs and cross-language integer-width boundaries. Preserve independent golden source-side bytes; two bindings using the same Rust encoder are not independent encoder proofs.

This is additional compatibility-test coverage, not a request to redesign settled storage. If review discovers a persisted replay/codec family whose meaning depends on an unspecified wire representation, raise that concrete defect under #2160 before freezing. The original mixed storage/wire umbrella is intentionally narrowed here.

aeplay:
🤖 Final current-stack codec audit found a concrete cross-language width boundary: Rust binding payloads encode `root_count` and delta positions as `u64` in `binding_codec.rs`, while TypeScript `PostcardReader.u64()` accumulates them into an unchecked JavaScript `number`; native row/runtime code then uses those values as subscription indices and ordering arithmetic. Above `Number.MAX_SAFE_INTEGER`, identity or ordering can silently lose precision. Legitimate counts would be enormous, but malformed or cross-host payloads can still reach this boundary. Please freeze either safe-integer rejection or a BigInt/internal bounded conversion here, with exact `MAX_SAFE_INTEGER ± 1` fixtures.

aeplay:
🤖 Settled compatibility boundary (Anselm, 2026-08-27): storage epoch 1 begins at the storage-settlement commit; all earlier alpha storage is unsupported. Wire and binding compatibility are versioned independently from storage. Peers must negotiate a mutually supported wire/binding version or reject before semantic decoding; sharing a storage epoch never implies wire compatibility, and a wire version must not silently reinterpret persisted replay bytes.

aeplay:
[...]

PR #2259, opened 2026-08-28
Freeze exact wire and binding decoding

aeplay:
🤖 Advances #2044 by freezing exact canonical v14 wire and v1 binding decoding across Rust and TypeScript.

## Before

Rust postcard decoding accepted semantic payload prefixes unless callers inspected the remainder. TypeScript likewise parsed batches/frames without consistently requiring EOF, accepted nonminimal or overflowing varints, and converted some 64-bit values through unsafe JavaScript numbers.

For example, canonical `WireFrame::Error` tag `02` could be replaced by the overlong spelling `82 00`; postcard decoded the same enum and consumed the input even though re-encoding produced different bytes. `PostcardWriter.i64(1n << 63n)` could emit bytes its own reader and Rust reject.

## After

- Rust requires EOF and byte-identical canonical re-encoding for decoded frames/messages.
- TypeScript centralizes canonical 10-byte varint validation: nonminimal, overlong, and overflow spellings reject.
- `u64()` rejects values beyond `Number.MAX_SAFE_INTEGER`; `u64BigInt()` preserves the full canonical unsigned-64 range.
- Signed writers and native-row/query coercion reject unsafe numeric inputs before `BigInt(number)` and enforce exact signed-i64 bounds.
- Fixed-width u32 writers reject truncating negative/overflow inputs.
- WebSocket batches, Hello/Message/Error frames, relation envelopes, malformed UTF-8, feature bits, and compressed payloads require their exact boundaries.
- v14 discriminants, negotiation/intersection, compression, fragmentation, and append/no-migration rules are normative.

For example, `u64::MAX` crosses Rust→TypeScript only through BigInt, while `MAX_SAFE_INTEGER + 1` as a JS number rejects rather than silently rounding. Valid `i64::MIN` and `i64::MAX` BigInts still round-trip.

## Scope

This does not complete #2044. The remaining direct NAPI/WASM physical-fragment matrix and broader branch/catalogue/large-value compatibility corpus remain tracked there and in #2160.

## Verification

- hard-coded v14 semantic↔payload/frame and binding/native-row goldens
- Rust↔TS full-u64 fixture, signed min/max, safe/unsafe-number, overflow, suffix, malformed UTF-8, and nonminimal enum-tag receipts
- real dual-codec and Zstd-only negotiation/intersection receipt
- independent planted removal of Rust canonical re-encoding and TS 10th-byte/safe-integer guards made their exact tests fail
- focused Rust/TS suites, TypeScript checks, invariant registry, Clippy, formatting, lint, and sensitive-data gates


aeplay:

PR #2264, opened 2026-08-28
Freeze transaction and version storage layouts

aeplay:
🤖 Freezes the transaction/version storage layouts and the winner semantics that interpret them.

## Before

Transaction receipts, version records, and merge-head evidence were not all protected by one exact storage contract. Some semantic ingress paths could reach infallible version accessors before validating untrusted deferred records, and the relationship between content winners, deletion winners, direct parents, and exclusive-write evidence was spread across implementation details.

## After

- transaction and version receipts have one frozen canonical layout and fail closed on malformed, trailing, or noncanonical bytes;
- every semantic ingress validates deferred version records before infallible access or mutation, including direct and batched view updates, authorization-scope views, and repair;
- content and deletion winners are tracked independently;
- direct parents are ordered deterministically by transaction time and node UUID;
- accepted/global evidence gates the states that require it;
- mergeable transactions no longer acquire a read set merely to encode causal dependencies.

## Worked examples

1. Two concurrent updates can choose one content winner while independently retaining the correct deletion winner; replay and reopen produce the same pair.
2. A valid first view update followed by a malformed deferred version is rejected atomically: neither update is applied and the node does not panic.
3. An exclusive write through an old schema name after a table rename still compares against the same permanent physical table register, so a matching parent succeeds and a stale parent fails.

## Verification

- independent adversarial review, including planted ingress-guard removal;
- focused malformed-ingress, batch-atomicity, winner, exclusive-CAS, rename, reopen, and ordering receipts;
- invariant registry;
- library clippy and diff checks.

This is a zero-backcompat storage settlement change; the normative contract is recorded in SPEC §2.8 and its cited invariants.


aeplay:

PR #2265, opened 2026-08-28
Freeze V1 large-value lifecycle and locator identities

aeplay:
🤖 Freezes the V1 large-value staging, completed-upload, receipt, and locator lifecycle.

## Before

The lifecycle records were durable, but some keys were treated as existence markers without proving that their embedded upload or receipt ID matched the physical key. A copied completed journal could therefore turn staging into a silent no-op. The descriptor-upload derivation domain also retained stale V2 naming after the first frozen format became V1.

## After

- every staged, pending, completed-upload, and completed-receipt record is decoded through its canonical fixed-width key and must contain the same ID;
- forward and reverse completed journals are checked before retry, staging, acceptance, eviction, and cleanup;
- finalization atomically publishes the descriptor-bound receipt and its retry binding;
- retry returns exactly the durable receipt, while acceptance or explicit TTL eviction removes the binding atomically;
- TTL protects abandoned staging only and is not an admission-correctness mechanism;
- immutable auth locators and shared-DAG reclamation retain their existing semantics;
- the descriptor-upload derivation domain and exact fixture are frozen as V1.

## Worked examples

1. Copying a valid completed-upload record under a different upload key fails closed before a new chunk is staged; it cannot masquerade as an already completed upload.
2. Copying the reverse receipt journal under a different receipt key prevents acceptance or cleanup rather than deleting another upload binding.
3. Retrying after durable promotion returns the same descriptor-bound receipt. If the client abandons staging instead, explicit TTL eviction can reclaim it without affecting an admitted row.
4. A hash collision in an authorization locator never changes the locator already bound to a content root.

## Verification

- independent adversarial review with planted key/ID guard removal;
- full Groove suite;
- corruption/reopen, retry/promotion, TTL, locator, shared-DAG, and reclamation receipts;
- V1 format/manifests, Jazz wire fixtures, and invariant registry.

No compatibility decoder or migration path is retained.


aeplay:

PR #2266, opened 2026-08-28
Harden React Native relay artifact detection

aeplay:
🤖 Hardens the React Native package boundary so staged native relay artifacts are detected consistently on iOS and source-only installs fail explicitly.

## Before

The legacy React Native CocoaPods branch replaced the package header search path while adding Boost. A package containing a valid vendored XCFramework could therefore compile against the ABI-0 fallback header and report that no native relay was installed. The JavaScript wrapper covered missing and incompatible modules, but not the ABI-0 source-only fallback as an executable receipt.

## After

- the legacy CocoaPods branch retains the exact staged relay header directory and extends it with Boost;
- framework-qualified and staged-header imports both resolve;
- ABI 0 is an explicit source-only/native-artifact-unavailable state and never executes a relay command;
- null module, ABI 0, incompatible nonzero ABI, and matching ABI each have executable JavaScript receipts;
- package tests understand pnpm layout without broadening runtime behavior;
- RN source keeps its package-owned Prettier policy while the root formatter avoids contradictory rewrites.

## Worked examples

1. A source-only or Expo Go-style install exposes ABI 0. The first persistent relay call fails with an instruction to install a native build, and native execute is never invoked.
2. An old native build exposing an unsupported nonzero ABI fails before storage is opened.
3. A correctly packaged legacy-RN iOS build retains native/include while adding the Boost search path, resolves the vendored relay header, reports ABI 3, and forwards commands.
4. Bare RN and Expo prebuild still autolink through the same package contract; Expo Go remains unsupported.

## Verification

- independent adversarial review;
- mutation-sensitive exact Podspec path receipt;
- mutation-sensitive ABI-0 no-execute receipt;
- RN Jest, TypeScript, package formatting, and Android+iOS Expo prebuild packaging gate.

This is a focused slice of #1972; producing and signing all platform binary slices remains follow-up work.

### DIFF
--- a/crates/jazz-rn/JazzRn.podspec
+++ b/crates/jazz-rn/JazzRn.podspec
@@ -26,10 +26,16 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/garden-co/jazz.git", :tag => "#{s.version}" }
 
   s.source_files = "ios/**/*.{h,m,mm,swift}", "ios/generated/**/*.{h,m,mm}"
+  # The staged header is the source-of-truth ABI declaration shared with the
+  # Android package. Keep this path when the legacy RN dependency branch below
+  # adds its own headers: assigning pod_target_xcconfig twice would otherwise
+  # compile this source with the unavailable-artifact fallback even though the
+  # package contains a valid XCFramework.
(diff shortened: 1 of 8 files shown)


aeplay:

PR #2267, opened 2026-08-28
Freeze permanent catalogue identities

aeplay:
🤖 Freezes permanent catalogue identities and their durable authority mapping.

## Before

Physical table and column identities were still partly derived from schema-local order. Renames could be represented, but incompatible column epochs, recursive enum identities, out-of-order lineage delivery, and reopen did not all share one permanent UUID contract. A receiver could also accept durable or pending manifests without proving that they matched the authority publication, and UUIDs retired several schema generations earlier could be reused.

## After

- every table, column, and recursive enum variant receives a permanent authority UUID;
- compatible rename/reorder evolution preserves the exact UUID;
- incompatible type, merge-strategy, or structural epochs receive an entirely fresh column and enum subtree;
- UUIDs are globally retired across the retained lineage history and cannot be reused by later tables, columns, or nested variants;
- an out-of-order descendant may inherit IDs first issued by its absent parent, but cannot claim unrelated historical reservations;
- active, staged, pending, snapshot, live-admission, and reopen paths validate the same authority manifest and evolution rules;
- unknown-source pending publications validate their self-contained shape immediately, then validate source evolution before activation;
- genesis construction and descendant authority authorship are distinct APIs; clients may prepare a publication but only trusted serving admission may publish it;
- canonical codecs and deterministic wire fixtures freeze the manifest representation.

## Worked examples

1. Rename todos.body to tasks.text: the table and compatible column keep their UUIDs through sync and reopen.
2. Change a counter column to last-write-wins: the new physical epoch receives a fresh column UUID and fresh nested enum UUIDs.
3. Drop a table in v2, then add an unrelated table in v3: reusing the v1 table UUID is rejected even though v2 no longer contains it.
4. Receive v3 before v2: v3 may remain pending and later inherit UUIDs introduced by v2. A forged inherited UUID is quarantined when v2 arrives, with no partial activation.
5. Tamper a persisted mapping to a nil, colliding, or publication-mismatched UUID: reopen fails closed before exposing catalogue state.

## Verification

- repeated independent adversarial reviews with planted durable tampering, multi-hop UUID reuse, incompatible enum epochs, and history omission;
- catalogue-focused suite, query/node receipts, activation/reopen/out-of-order tests;
- deterministic wire current/decode/current runs;
- cross-crate compilation, invariant registry, library clippy, and diff checks.

This is a zero-backcompat storage settlement change. The permanent identity and retirement rules are normative in the catalogue/migration specs and cited invariants.


aeplay:

Issue #1973, opened 2026-08-25
Wire trusted native relay admission on Android and iOS


aeplay:
🤖 Complete the trusted Android/iOS integration for the native relay admission boundary established by #1832.

## Boundary

Trusted platform code—not arbitrary JavaScript—must derive and admit the complete `{app namespace, storage namespace, auth scope}` configuration, including SQLite path, schema, persistent database identity, and validated session claims. JavaScript receives only the random opaque admission capability used to attach foreground in-memory clients.

Logout must revoke that capability and atomically close all relay/client aliases opened through it. Reuse with changed path, schema, identity, or claims must fail closed. The generic execute/frame channel must never accept or reveal token material or let JavaScript amend trusted scope fields.

## Acceptance example

Two foreground RN runtimes admitted to the same scope share one durable relay. A different scope is isolated. After logout, old and guessed capabilities cannot attach clients; a newly authenticated scope cannot inherit the prior relay. Platform tests must include scope reuse, revocation, auth switching, and process reopen.

Parent roadmap: #1756. The Rust ABI 3 capability model and host contracts are introduced by #1832.

aeplay:
🤖 Closing as superseded by the agreed React Native architecture tracked by #2291 and implemented through the current RN stack. The native component is a scope-isolated client relay, not a trusted backend: it structurally derives only a local cache scope, carries the bearer to Edge for authentication, never evaluates permissions or acts as SYSTEM, and keeps credentials out of ordinary foreground command frames. Multi-root multiplexing remains separately tracked by #2532.

PR #2269, opened 2026-08-28
Require trusted React Native relay admission

aeplay:
🤖 Hardens the trusted React Native relay admission boundary and seals the JavaScript-facing relay ABI.

## Before

The native relay already used opaque capabilities, but trusted JSON admission still permitted a missing authentication scope. The package also relied on a blacklist-style source receipt to prove that JavaScript could not grow configuration or admission methods; alternate TypeScript exports, Android annotations, Objective-C export macros, and preprocessor spelling could evade that receipt.

## After

- every persistent relay scope requires a nonempty authenticated scope;
- omitted, null, empty, and whitespace scopes fail before capability minting, scope admission, or storage open;
- one admitted scope binds the complete path, schema, persistent database identity, claims, and auth scope; any changed field conflicts;
- guessed and revoked 256-bit capabilities fail, and revocation closes aliases opened through that capability;
- trusted Android/iOS platform code owns admission and revocation; JavaScript receives only the opaque capability;
- the public JS ABI is a fixed allowlist:
  - TurboModule Spec: getAbiVersion and execute;
  - package exports: ABI range/constants and opaque execute helper;
  - Android generated module: getAbiVersion and execute;
  - iOS New-Architecture generated selectors only;
- legacy iOS RCT export/remap macros are forbidden raw-source-wide, including comments, strings, categories, and C-preprocessor line-spliced spellings;
- TypeScript AST receipts reject every extra/default/aliased/star/class/property/call/index export shape.

## Worked examples

1. Trusted platform admission with auth_scope null cannot mint a capability or create a relay.
2. Reusing a capability request with a different SQLite path, schema, database identity, claims, or auth scope fails closed.
3. Logout revokes the opaque capability and closes its relay/client aliases; a guessed or old capability cannot attach.
4. Adding getAuthScope, configure, admitScope, or any similarly innocuous extra method to TS, Android, or iOS makes the packaging gate fail.
5. A legacy Objective-C export hidden in a later category, URL-containing line, comment, or backslash-newline macro spelling is rejected. The generic opaque execute channel remains unchanged.

## Verification

- repeated independent adversarial reviews with planted null-scope, config-conflict, TS re-export/default/property, qualified Android annotation, Objective-C category/synchronous/spliced macro, comment, and string probes;
- jazz-native-relay suite;
- C and C++ ABI probes;
- RN package, TypeScript, and Android+iOS codegen/prebuild receipts.

This advances #1973. Actual auth-provider integration and real device scenarios remain tracked separately.

### DIFF
--- a/crates/jazz-native-relay/src/lib.rs
+++ b/crates/jazz-native-relay/src/lib.rs
@@ -958,6 +958,15 @@ impl RelayScope {
                 )));
             }
         }
+        if self
+            .auth_scope
+            .as_deref()
+            .is_some_and(|auth_scope| auth_scope.trim().is_empty())
+        {
+            return Err(RelayError::InvalidScope(
+                "auth scope must not be empty when supplied".to_owned(),
+            ));
+        }
         Ok(())
     }
 }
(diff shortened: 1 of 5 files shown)


aeplay:

PR #2270, opened 2026-08-28
Preserve React provider runtime identity

joeinnes:
## Summary

- Include client-factory identity in the React provider registry key.
- Preserve reference identity for opaque runtime-source values while retaining structural identity for ordinary configuration.
- Keep the React Native client factory stable across rerenders so unchanged providers continue using their current client.

## Before and after

Before, providers with structurally equal configuration shared a registry key even when they used different factories or runtime-source objects. A provider could therefore receive a client created for another runtime.

Making factory identity part of the key also exposed a React Native boundary: its inline factory wrapper had a new identity on every render, so an unchanged rerender would replace the current client.

After, runtime-producing inputs participate in client identity:

```tsx
const source = new Uint8Array([0, 97, 115, 109]);

rerender(<JazzProvider config={{ ...config, runtimeSources: { wasmSource: source } }} />);
// Same source reference: reuse the current client.

rerender(
  <JazzProvider
    config={{ ...config, runtimeSources: { wasmSource: new Uint8Array(source) } }}
  />,
);
// Different opaque source: shut down the old client and create the replacement.
```

Changing `createJazzClient` likewise creates a distinct registry identity. React Native now passes one module-stable wrapper, so an otherwise unchanged rerender does not cause a false replacement.

## Invariants

- Plain JSON-compatible configuration retains stable structural identity.
- Client factories and opaque runtime sources use reference identity.
- A changed factory or opaque source cannot reuse a client created for the previous runtime.
- An unchanged React Native provider keeps its existing client.

## Testing

- `cd packages/jazz-tools && pnpm test:react src/react-core/provider.ssr.test.tsx src/react-core/provider.stable-deps.test.tsx src/react-native/provider.test.tsx` — passed; 8 tests
- `pnpm --filter jazz-tools typecheck:react-native` — passed

### DIFF
--- a/.changeset/bright-foxes-replace.md
+++ b/.changeset/bright-foxes-replace.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Replace React clients when their runtime factory identity changes.
--- a/.changeset/bright-foxes-replace.md
+++ b/.changeset/bright-foxes-replace.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
-Replace React clients when their runtime factory identity changes.
+Preserve runtime factory and source identity when React providers choose a client, while keeping unchanged React Native provider rerenders on the current client.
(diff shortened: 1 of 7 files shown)


aeplay (approved):

PR #2271, opened 2026-08-28
Seal React Native relay artifact inputs

aeplay:
🤖 Seals the native React Native relay artifact cache and manifest boundary after #2266.

Before: a hash-valid Android or iOS relay artifact could be reused after a transitive native input changed, because the manifest and cache identity did not cover the complete Cargo dependency closure.

After:

- artifact manifest format V2 binds the native-source fingerprint;
- Android and iOS cache keys use the same complete input set;
- main-branch workflow triggers cover every fingerprinted native input;
- verification rejects a stale artifact even when its file hashes still match;
- receipts plant changes in idb-tree and jazz-compression and prove rejection.

Example: changing only the IndexedDB tree implementation now changes the native fingerprint, misses the old artifact cache, and prevents a package from claiming that an older relay binary matches current sources.

This does not expose native admission to JavaScript or introduce another runtime path. Actual Android NDK and Xcode link receipts remain in the opt-in React Native artifact workflow.


aeplay:

PR #2273, opened 2026-08-28
Add the React Native relay frame adapter

aeplay:
🤖 Adds the narrow JavaScript frame adapter that lets a foreground in-memory Jazz runtime exchange canonical peer frames with the admitted native React Native relay.

Before: React Native had the native relay ABI, packaging, and trusted admission boundary, but no JavaScript transport primitive for attaching an ordinary in-memory runtime without exposing SQLite or inventing a second query path.

After:

- JavaScript supplies only the opaque native admission capability;
- the adapter exchanges the canonical postcard/base64 peer-frame commands;
- startup, reconnect, backpressure, and shutdown are serialized;
- native-drained frames remain FIFO and are retried without loss or duplication after local delivery failure;
- malformed ABI responses and missing admission fail clearly;
- startup races and invalid attach responses close every opened native handle exactly once.

Example: if an app closes while native attach is still pending, the eventual client and relay handles are both closed and no persistent relay alias leaks. If local delivery temporarily rejects a received frame, the same frame is delivered first on the next pump before later frames.

This remains a transport slice. Wiring it into ReactNativeRuntimeSource persistent startup is the next step and must keep the foreground database memory-only.


aeplay:
🤖 Superseded by the native foreground runtime stack beginning at #2310. The postcard frame adapter would preserve a second JavaScript relay/runtime path; the current design intentionally uses one native `NativeDb`/JSI path so storage, lifecycle, and device behavior share the same implementation.

PR #2274, opened 2026-08-28
Route React Native runtime sources through the native relay

aeplay:
🤖 Wires persistent React Native startup through the reviewed native relay frame adapter while retaining the ordinary foreground in-memory Jazz runtime.

Before: the frame adapter existed, but ReactNativeRuntimeSource still rejected persistent configuration and could not attach its normal direct connection manager to the admitted native relay.

After:

- persistent React Native config accepts only a native executor plus opaque admission capability;
- DirectConnectionManager owns relay readiness, reconnect, disconnect, and shutdown;
- SQLite path, schema, database identity, session claims, and admission derivation remain native-only;
- memory mode is unchanged;
- missing admission, incompatible ABI, and relay startup failures surface explicitly;
- reconnect replaces stale failed readiness without clearing a newer concurrent failure.

Example: a transient native readiness failure can recover after reconnect and queries continue through the same foreground in-memory Db. If a newer relay error arrives during that reconnect, the older success cannot erase it.

This completes the library runtime-source seam. First-party Android emulator and iOS simulator acceptance still belongs in the RN test app/device work.

### DIFF
--- a/packages/jazz-tools/src/react-native/README.md
+++ b/packages/jazz-tools/src/react-native/README.md
@@ -1,14 +1,19 @@
 # React Native bindings
 
 This directory restores the React Native binding shape against the v2 runtime.
-It is compile-level scaffolding only: the React hooks, provider, client factory,
-typed schema exports, and auth helper all typecheck, but persistent storage does
-not run yet.
+The foreground JavaScript runtime is deliberately in-memory; persistent storage
+and upstream relay ownership remain in the installed Android/iOS `JazzRelay`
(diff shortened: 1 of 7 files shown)


aeplay:
🤖 Superseded by the native foreground runtime stack beginning at #2310. Runtime sources now attach through the native `NativeDb`/JSI foreground path; retaining this JavaScript relay source would reintroduce a second persistent-runtime implementation with different lifecycle and storage semantics.

PR #2276, opened 2026-08-28
Harden server authentication admission

joeinnes:
🤖 Hardens external server authentication admission and ports every first-party JWT fixture to the exact configured issuer/audience contract.

## Before

A token signed by a configured key could be admitted for the wrong issuer or audience. Privileged infrastructure-secret aliases could blur ordinary bearer admission, and a forged backend-shaped session could suppress the cookie-origin guard.

First-party browser, NAPI, and Rust test servers also provided JWKS without explicitly configuring the issuer/audience their fixture tokens carried, so enforcing the correct server boundary initially made those consumers fail closed.

## After

- Externally signed JWTs must pass signature verification and exact configured issuer, at least one configured audience member, `exp`, and `nbf` checks.
- Missing issuer/audience configuration fails closed; there is no legacy fallback.
- Local-first identity remains a separate explicitly allowed path.
- Admin/backend infrastructure secrets cannot become ordinary bearer credentials through old aliases.
- Cookie WebSocket origin enforcement follows the credential actually authenticated.
- Forwarded host headers are trusted only when explicitly configured.
- Browser helper, Rust testkit, and NAPI fixtures now pass their matching test issuer/audience explicitly.

For example, `aud: ["another-service", "configured-service"]` is accepted when the configured audience is present, while the same signed token with missing/wrong `iss` or no matching `aud` is rejected before a session exists.

## Verification

- BUG-119 issuer/audience/time boundary; BUG-302 origin/credential cases; BUG-306 privileged-alias rejection
- local-first same-principal refresh and external-subject receipts
- NAPI policy-graph local server and browser test-server consumer receipts with freshly rebuilt artifacts
- independent planted issuer-comparison bypass and reintroduced admin-secret alias each made their exact negative receipt fail
- Jazz Tools/RN typechecks, Clippy, formatting, and sensitive-data gates


aeplay (approved):

PR #2277, opened 2026-08-28
Reject unsafe asymmetric update policy merges

joeinnes:
🤖 Reject unsupported combinations of asymmetric update policies instead of compiling them into authority that no individual rule granted.

## Before

Each asymmetric rule contributes an old-row `USING` predicate and a new-row `WITH CHECK` predicate. If two rules are independently OR-merged, an update can satisfy the old side of rule A and the new side of rule B even though neither rule permits that complete transition.

Example:

- rule A allows the playlist owner to update rows they already own;
- rule B allows an invitation recipient to turn their own pending invitation into accepted;
- naïve independent ORs could combine “owner-authorized old row” with “recipient-authorized new row” across rule boundaries.

## After

The permission builder records whether a table already has an asymmetric update rule and rejects a second one. Multiple symmetric `.where(...)` rules continue to OR-merge normally. The marker is builder metadata rather than a serialization of compiled expressions, so valid literals such as `1n` cannot crash policy compilation.

RecordPlayer now expresses invitation updates as one explicit asymmetric rule whose new-row conditions bind immutable fields back to the persisted row:

- owner path: owner authorization must hold for both old and new states;
- recipient path: the persisted row must already be that recipient’s pending invitation, the new row must be accepted, and invitation identity fields remain unchanged.

Tests reject recipient role changes, playlist changes, and cross-rule old/new mixing.

## Boundary

This is a fail-closed guard, not pair-aware policy representation. A hand-written single rule with independently ORed old alternatives and new alternatives can still express the same mathematical cross-product. Applications that need multiple asymmetric transitions must use a symmetric condition, model each transition as a separate record/action, or mediate it on a trusted backend. The broader representation question remains #2288.

## Stack

This is stacked on #2223 because its repaired RecordPlayer browser receipt consumes the renamed `WriteResult.txId` API. Existing commits are preserved; the review repair is appended.

## Verification

- permissions/asymmetry suite: 57/57, including a BigInt literal regression;
- planted removal of the asymmetry guard makes the cross-product receipt fail;
- RecordPlayer forged-acceptance topology passes with all rejection phases;
- public txId/binding/artifact gates inherited from #2223;
- independent adversarial review and exact ancestry audit.

### DIFF
--- a/.changeset/reject-asymmetric-update-merges.md
+++ b/.changeset/reject-asymmetric-update-merges.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reject multiple asymmetric update permission rules instead of merging their old-row and new-row conditions into unintended combinations.
--- a/examples/record-player/apps/next-betterauth/permissions.ts
+++ b/examples/record-player/apps/next-betterauth/permissions.ts
@@ -5,7 +5,7 @@ import { app } from "./schema";
 
 const recordPlayerPermissions = s.definePermissions(
   app,
-  ({ policy, anyOf, allowedTo, session }) => {
+  ({ policy, allOf, anyOf, allowedTo, session }) => {
     // RecordPlayer deliberately models a shared public catalogue: every
     // authenticated listener can publish album metadata and audio tracks, and
     // everyone can browse them. A production catalog would ordinarily replace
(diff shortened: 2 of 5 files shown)


joeinnes:
The CI failure exposed an existing unsafe RecordPlayer invitation policy: its chained `.whereNew()` discards the intended immutable-field condition, and its owner/recipient update alternatives form an old/new cross-product. Tracked in #2288. This PR remains draft pending that example-policy redesign; the fail-closed asymmetric-merge rejection itself is intentional.

aeplay (approved):

aeplay:
🤖 This PR intentionally fails closed; pair-preserving support for multiple asymmetric rules is tracked in https://github.com/garden-co/jazz/issues/2329.

PR #2278, opened 2026-08-28
Scaffold React Native device acceptance

aeplay:
🤖 Establishes the first-party React Native device acceptance scaffold without pretending source checks are device receipts.

Before: the repository had packaging and runtime seams but no self-contained app, result protocol, or Android/iOS driver boundary for proving the native relay on real devices.

After:

- a self-contained Expo development-build and bare-host scaffold exercises trusted native admission through opaque capabilities;
- a strict JAZZ_DEVICE_RESULT protocol binds every result to the expected platform, device, build fingerprint, run nonce, fresh timestamp, monotonic sequence, and exact scenario set;
- Android adb and iOS simctl drivers reject TODO, partial, duplicate, stale, foreign, or forged receipts;
- the Android fixture remains compile-shaped after Expo prebuild;
- CI is explicitly named scaffold verification, not device acceptance.

Example: copying a passing receipt from an older APK or another simulator cannot make a run green because the build fingerprint, device identity, nonce, and freshness checks all fail.

The full scenario matrix is intentionally TODO until CI actually builds, installs, launches, and collects linked Android/iOS receipts. The scaffold never reports those TODOs as acceptance.


aeplay:

PR #2279, opened 2026-08-28
Reject unsupported compound ExistsRel joins

joeinnes:
🤖 Reject compound `ExistsRel` joins during schema admission instead of silently lowering only their first equality.

## Before

The current core representation supports exactly one equality per relation join. A public policy could nevertheless provide two:

```text
membership.workspace = grant.workspace
AND membership.subject = grant.subject
```

Lowering could retain only the workspace equality. The resulting policy would authorize rows for the wrong subject.

The compound join could also be nested below another join or hidden inside `And`, `Or`, or `Not`, reaching different read/write conversion paths.

## After

- One recursive pre-lowering validator walks the complete relation tree and every policy boolean wrapper.
- Any join whose condition count is not exactly one fails with a contextual schema path before query/runtime evaluation.
- The same rule applies to read/select, insert check, update using/check, and delete using.
- Existing supported one-hop correlated `ExistsRel` and gather-based lowering remain unchanged.

## Verification

- A nested aliased compound join is rejected in all five public policy contexts.
- Existing valid insert/update and gather-based `ExistsRel` receipts pass.
- A planted mutation allowing two equalities makes the schema-admission receipt fail by accepting the malformed policy.
- Focused Rust tests, formatting, and clippy pass.

### DIFF
--- a/crates/jazz/src/node/tests/policies_rls/write_authorization.rs
+++ b/crates/jazz/src/node/tests/policies_rls/write_authorization.rs
@@ -1431,11 +1431,15 @@ fn exists_rel_rejects_nested_outer_correlation_off_the_join_key() {
 }
 
 #[test]
-fn exists_rel_rejects_multiple_join_conditions_in_write_policy() {
+fn exists_rel_rejects_compound_joins_in_every_policy_context_before_lowering() {
     let column = |scope: &str, name: &str| PublicRelColumnRef {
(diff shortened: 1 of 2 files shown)


aeplay (approved):

aeplay:
🤖 Compound joins deliberately remain rejected here; full support is tracked in https://github.com/garden-co/jazz/issues/2328.

PR #2280, opened 2026-08-28
Run React Native iOS device acceptance on Blacksmith

aeplay:
🤖 Continues #2287 with an opt-in Blacksmith iOS simulator build and run-bound native-relay receipt.

## Before

The Linux source workflow could not prove that the XCFramework was consumable by Expo/CocoaPods, that a simulator app linked and launched, or that the native relay loaded in that exact binary. The first Blacksmith trial reached a booted simulator but attempted a runner-forbidden log reset before building the app.

## After

- `react-native`-labeled same-repository PRs build and verify the relay XCFramework on Blacksmith macOS, Expo-prebuild the host, install pods, boot a concrete simulator, build/install/launch, and require a run-bound receipt.
- The iOS fixture uses the same non-secret canonical schema/identity/claims contract as Android and exposes only an opaque capability to JavaScript.
- The automatic `linked-abi-admission` receipt is bound to simulator UDID, executable SHA-256, random run nonce, ordered sequence, and launch timestamp.
- The normal source workflow remains quick, and fork PRs cannot invoke the trusted macOS/device job.

For example, stale simulator logs cannot satisfy the verifier because they do not carry this launch's nonce and executable fingerprint. Source contracts reject the earlier invalid `{}` schema and noncanonical author shapes.

## Honest current boundary

Only the ABI/admission bootstrap scenario is wired. Independent review of #2287 is tightening that scenario so its probe demonstrably executes through the admitted capability rather than merely checking the generic bridge ABI; this iOS layer will inherit that repair before becoming ready. Local write/subscription, reconnect/reopen, scope isolation, and logout/auth-switch remain TODO.

## Verification so far

- `pnpm --filter rn-device-acceptance verify`: 14 tests plus TypeScript.
- The first Blacksmith run passed XCFramework build, Expo prebuild, fixture registration, CocoaPods, and simulator boot before exposing the forbidden-log-reset issue now removed.
- This restacked revision is running the build/install/launch path; it is a draft until the capability-bound probe and independent re-review land.

### DIFF
--- a/.github/workflows/rn-device-acceptance.yml
+++ b/.github/workflows/rn-device-acceptance.yml
@@ -2,7 +2,7 @@ name: React Native device acceptance
 
 on:
   pull_request:
-    types: [opened, reopened, synchronize]
+    types: [opened, reopened, synchronize, labeled]
     paths:
       - "dev/rn-device-acceptance/**"
       - "crates/jazz-rn/**"
--- a/.github/workflows/rn-device-acceptance.yml
+++ b/.github/workflows/rn-device-acceptance.yml
@@ -142,7 +142,9 @@ jobs:
     # artifact -> assemble -> installed-device receipt workflow.
     if: github.event_name == 'workflow_dispatch'
     name: Android device acceptance
-    runs-on: blacksmith-8vcpu-ubuntu-2404
+    # GitHub-hosted Linux runners expose Android hardware acceleration; this
+    # Blacksmith image did not expose a usable /dev/kvm.
+    runs-on: ubuntu-24.04
     timeout-minutes: 45
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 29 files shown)


aeplay:
🤖 Hosted iOS acceptance is now proven on the Blacksmith release simulator.

The successful run built the Release app with its embedded JavaScript bundle, installed and launched it, resolved the codegen `JazzRelay` TurboModule through the native registration boundary, completed the admitted-handle proof, and observed exactly one run-bound receipt persisted atomically in the app sandbox.

The final repair added the missing `RCT_EXPORT_MODULE()` registration. Independent review verified that the empty macro name resolves to the Objective-C class name `JazzRelay`, matching `TurboModuleRegistry.get("JazzRelay")`, with exactly one registration and no legacy remap/export path.

This proves the PR's scoped ABI/admission scenario. It intentionally does not claim two independent foreground Jazz runtimes exchanging application rows; that still requires the foreground engine decision tracked in #2291.

Receipt: https://github.com/garden-co/jazz/actions/runs/33179885690/job/98878114439

PR #2281, opened 2026-08-28
Add epoch-one SQLite and RocksDB compatibility fixtures

aeplay:
🤖 Advances #2160 with committed epoch-one physical SQLite and RocksDB compatibility fixtures, and carries the native transaction-lifetime integration repair required by that physical-open test layer.

## Before

Most reopen tests wrote and read with the same current build. That can stay green even if a future edit changes authoritative bytes, because both halves drift together.

Separately, completed high-level NAPI transactions could retain attached native transaction views until JavaScript garbage collection. `shutdown()` could therefore return while a retained view still owned the database, racing physical RocksDB release and reopen.

## After

- Committed SQLite and RocksDB fixtures record the immutable settlement revision, deterministic generator/provenance, SHA-256, and a human-reviewable backend-neutral ordered-KV pack.
- Tests inspect the frozen store read-only, compare exact logical contents, add current-format writes, reopen, and prove both historical and new data survive.
- Corrupt fixture input and an incorrect pack header fail before a destination database or extraction root is created.
- The SQLite generator reproduces its committed bytes exactly; RocksDB uses a committed release-baseline archive plus hand-spelled epoch-one manifest bytes rather than the current encoder.
- High-level transactions explicitly release their attached native view on successful commit, rollback, and shutdown. A failed commit remains rollbackable; closing a derived schema view releases only that child handle.
- Native transaction close is idempotent, and declaration/runtime surfaces agree.

## Worked examples

Replacing `JAZZ-ORDERED-KV-PACK-1` while retaining an otherwise valid payload is rejected rather than having the parser skip and reinterpret the first line.

If two schema views share one transaction, closing one view cannot invalidate the other or commit the transaction. Conversely, once commit succeeds, context shutdown does not wait for JavaScript GC before RocksDB can be reopened.

## Scope and residual

The physical corpus covers SQLite and RocksDB. Real browser IndexedDB still needs its shared epoch-manifest metadata seam and browser physical-open harness; existing IDB-tree receipts cover page framing but not that missing boundary.

The transaction repair changes resource lifetime only. It does not change transaction attribution, commit semantics, or add retry behavior.

## Verification

- SQLite historical fixture suite: 3/3
- RocksDB historical fixture suite: 3/3
- byte-identical SQLite provenance replay
- independent planted checksum-gate and pack-header bypasses
- repeated native write-path commit/rollback/shutdown receipts
- failed-commit rollback and child-schema-view lifetime receipts
- NAPI declaration/runtime parity and artifact provenance
- Clippy, formatting, and sensitive-data gates


aeplay:

Issue #1774, opened 2026-08-23
Define the portable ordered key-value storage contract


aeplay:
🤖 Specify the minimal ordered-storage interface required by Groove so RocksDB, SQLite, memory, and future constrained environments implement one semantic contract.

## Contract areas

Define byte ordering, inclusive/exclusive bounds, prefix scans, snapshot/read consistency, atomic batches, async suspension, error classes, crash/reopen behavior, and cancellation. A bounded prefix primitive should express every current scan without backend-specific paths.

## Example

Scanning all keys under a logical prefix must stop at the computed upper bound and never read the following keyspace. The same call must return the same ordered bytes on memory, RocksDB, and SQLite, including prefixes ending in `0xff`.

## Pointers

Storage traits and implementations are under `crates/groove/src/storage/` and Jazz storage adapters under `crates/`. Existing scan/reopen receipts and the SQLite work in PR #1720 should inform the contract. The obsolete OPFS B-tree backend is not a compatibility target.

aeplay:
🤖 Storage-settlement decision (portable ordered KV): keys are opaque bytes ordered by unsigned lexicographic order. Bounds are explicit inclusive/exclusive; prefix scan is exactly `[prefix, successor(prefix))`, unbounded above when no successor exists. Each evaluation uses an explicit repeatable-read snapshot that remains the same across async suspension/resume. One atomic write batch contains puts/deletes plus generic `put_if_absent` and `compare_and_delete`; all conditions and mutations commit together or none do. Logical merge/delta behavior stays above storage: RocksDB merge/compaction, SQLite conflict modes, IndexedDB cursor/page quirks never cross the Groove boundary. Cancellation before commit publishes nothing; after success reports committed rather than ambiguously retrying. Reopen observes the complete batch or none. Scans stream ordered pages without duplicate/skip/view changes. Errors distinguish retryable resource/backpressure, corruption/noncanonical bytes, unsupported epoch/format, and permanent backend failure; corruption/incompatibility fail closed. Backend limits are preflight capabilities, not semantic differences. OPFS B-tree is excluded. Acceptance requires normative spec text and one cross-backend conformance corpus for ordering, bounds, snapshots, conditionals, cancellation, crash/reopen, and errors.

aeplay:
[...]

PR #2282, opened 2026-08-28
Share the portable ordered-KV conformance contract

aeplay:
🤖 Advances #1774 by making every current ordered-KV adapter exercise one shared semantic contract instead of maintaining subtly different scan/batch behavior.

## Before

SQLite and RocksDB duplicated prefix-bound calculations, backend receipts covered different cases, and a pre-commit invalid batch could be reported less precisely than its proven `Uncommitted` outcome.

## After

- One unsigned-lexicographic `prefix_successor` contract drives bounded prefix scans.
- Memory, SQLite, RocksDB, and the IDB adapter over `MemoryPageStore` share ordering, bounds, reverse, limit, atomic-batch, and reopen conformance.
- Empty prefixes, all-`0xff`, and interior-`0xff` prefixes such as `[0x12, 0xff]` are explicit receipts.
- Invalid batches detected before adapter submission report `Uncommitted`; errors after submission remain conservatively `PossiblyCommitted`.
- Raw cursors are intentionally non-snapshot. Stable evaluation/window semantics remain a higher-layer responsibility.

For example, reverse-scanning `[0x12, 0xff]` with a limit of two returns only the last two keys in that exact prefix and never spills into the adjacent `[0x13]` namespace.

## Scope and residual

The IDB receipt here exercises `IdbStorage` over the in-memory page store; it does not claim browser IndexedDB/OPFS physical persistence. That missing physical epoch/reopen boundary remains #2160.

## Verification

- exact shared conformance selections for Memory, IDB adapter, SQLite, and RocksDB
- started-write cancellation lifecycle receipts with the required feature enabled
- independent planted `prefix_successor` regressions were caught by both exact-bound and interior-`0xff` adapter receipts
- Clippy, formatting, and sensitive-data gates

### DIFF
--- a/crates/groove/SPEC/2_storage_model.md
+++ b/crates/groove/SPEC/2_storage_model.md
@@ -161,8 +161,9 @@ shared `JSM1` manifest. IndexedDB currently validates its adapter-private
 manifest; therefore it is not covered by the epoch-1 physical-open receipt.
 That remaining acceptance item stays explicitly tracked by #2160.
 
-The only ordering property groove requires from the backing store is
-lexicographic byte order. A range `ScanRequest` returns keys in that order and
+The only ordering property groove requires from the backing store is unsigned
+lexicographic byte order: bytes compare as `0x00 < ... < 0xff`, never as signed
+integers, locale text, or a backend-native collation. A range `ScanRequest` returns keys in that order and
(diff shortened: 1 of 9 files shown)


aeplay:

Issue #2158, opened 2026-08-26
Avoid default serde serialization for persistent data


nrainhart:
Authoritative or high-volume persisted data must use either the normative Groove encoding or an explicitly specified and versioned codec. Serde-derived postcard/JSON should be avoided, as changes such as reordering or adding enum variants can change persistent identities and key ordering.

### `BranchKey` is postcard-encoded

`BranchColumnValue` serializes the general-purpose Groove `Value` enum with postcard, and then `BranchKey` serializes that structure with postcard again (/crates/jazz/src/protocol.rs:1793). These bytes are embedded in rows and used as physical key prefixes.

This should be replaced before v1 with a manual, explicitly versioned encoding whose tags are permanently assigned. It should use the declared column types, not Rust/serde enum layout.

### `authored_columns` is JSON inside every applicable row

`authored_columns` is stored as a nullable Groove `Bytes`, containing JSON for a set of column names (/crates/jazz/src/node/codec.rs:1313).

That is high-volume row data. It should preferably become a native Groove array—ideally stable physical column IDs rather than names. If it remains opaque bytes, it needs an explicit frozen codec version and canonicalization rules.

### Contribution provenance and operation identities use postcard

`contribution_merge` is persisted as raw postcard (/crates/jazz/src/node/codec.rs:1046). Some contribution/GSet operation identities are also derived from postcard serialization of structures containing general-purpose values.

Because these bytes are persistent provenance or semantic identities, ordinary Rust enum/struct evolution could silently alter them. They need either:

- native typed Groove records, or
- a manually specified, versioned binary codec with permanent numeric tags.

### Opaque Groove metadata also uses postcard

Large-value reference, staging and reclaim metadata contains postcard-encoded keys/values. This is less concerning than user history because it is much smaller and may be reconstructible, but it still needs a declared version and migration/rebuild policy.

### Others

Many durable values remain direct serialization of current Rust structs and are worth reviewing:

- Catalogue state is raw serde JSON in (/crates/jazz/src/node/state/lifecycle.rs:1497).
- RocksDB merge operands use postcard-serialized `StorageDeltaKind` and `CurrentWinnerDelta` in (/crates/groove/src/storage/mod.rs:204). These operands may remain physically pending inside RocksDB across an upgrade.

aeplay:
[...]

Issue #2037, opened 2026-08-25
Freeze physical catalogue and enum-registry persistence


aeplay:
🤖 Audit `node::physical::{catalogue,descriptors,enum_lookup,storage}`, `node::state::catalogue`, and catalogue row records before freezing their durable format.

Define the stable meaning and bytes of physical table/column IDs, aliases, schema variants, `TablePhysicalMapping`, scalar/payload/nested enum registries, and catalogue payloads. Test concurrent schema introduction, differing arrival order, retries, torn activation, restart, retained old variants, and conflicts. Semantic enum identity must never depend on a node-local ordinal.

Coordinate with #1779 rather than duplicating its broader activation/recovery work.

aeplay:
🤖 Audit refinement from the full format sweep: enum occurrence identity is currently derived from a truncated 63-bit FNV path hash (`groove/src/records/values.rs`, `groove/src/schema.rs`). The freeze receipts should explicitly cover collision behavior, cross-node replay, and persistence rather than treating the derived value as collision-free semantic identity.

aeplay:
🤖 Cross-audit refinement: include the server-side `cat:` catalogue store and `catalogue_payload_codec`, not only node-owned catalogue rows. The store persists raw map JSON and the SCHEMA/LENS/PERMISSIONS payload decoders are current-version-only despite being restart-authoritative. Freeze their key namespace, canonical payload encoding, unknown-version behavior, and replay/migration fixtures alongside the node catalogue.

aeplay:
🤖 Storage-settlement decision: authoritative persisted structured values use the normative Groove record/scalar encoding, with internal records described by explicit versioned descriptors. Field IDs and enum tags are permanent numeric identities; declaration or Rust source order is not identity. Ordered physical keys use a separately normative order-preserving key codec. Default serde/postcard/JSON serialization is not authoritative storage. Opaque bytes are allowed only at a narrow, registered boundary with an explicit codec version, canonical rules, and golden fixtures. Convenience serialization is allowed only for explicitly discardable/rebuildable caches. Jazz system records use the same typed machinery as app records, under reserved and versioned system descriptors.

aeplay:
[...]

PR #2283, opened 2026-08-28
Freeze server catalogue entry storage codec

aeplay:
🤖 Advances #2037 and #2158 by freezing the restart-authoritative server catalogue entry boundary instead of persisting an unordered, unversioned JSON map.

## Before

An entry lived under `cat:<ASCII UUID>` and its metadata was serialized from a `HashMap` as arbitrary JSON. Equivalent maps could therefore produce different bytes, alternate UUID spellings entered the same logical namespace, and a copied value was not cryptographically or structurally bound to its key.

For example, `{ "a": "1", "b": "2" }` and the same members in reverse order were both accepted persisted spellings.

## After

- Keys are exactly `cat:v1:` followed by the 16 raw UUID bytes.
- Values are exact `JCAT` v1 records containing the same UUID, byte-sorted unique UTF-8 metadata pairs, length-delimited content, and EOF.
- Startup scans the whole reserved `cat:` parent namespace and fails closed on old, alternate, unknown-version, malformed, trailing, reordered, duplicate, or key/value-mismatched entries before exposing a resident catalogue.
- The normative lens/catalogue spec and `INV-LENS-24` record this boundary.

For example, moving a valid value under another object's key, or changing only the order of metadata pairs, now rejects during recovery rather than silently normalizing it.

## Scope and residual

This freezes the server entry envelope. Nested schema and policy payload JSON in `catalogue_payload_codec.rs` remains explicitly tracked by #2037 and is not claimed as settled here.

## Verification

- 166 `jazz-server` tests
- exact golden, unknown-version, malformed, key-binding, reopen, and atomic startup-failure receipts
- independent adversarial review narrowed the scan to `cat:v1:`; the planted unknown sibling then escaped and the startup receipt failed, proving the parent-namespace check is sensitive


aeplay:

Issue #1779, opened 2026-08-23
Make schema lenses and catalogue activation complete and recoverable


aeplay:
🤖 Complete the lifecycle from schema/lens publication through catalogue sync, activation, projection, restart, and retirement.

## Required behavior

Validate a continuous lineage, publish schema+lens atomically, park authored data whose schema/lineage is unavailable, project winners through the ordered lens chain, and activate only when required catalogue state is complete. Restart/reconnect must reproduce the same active lineage; invalid or conflicting transitions fail closed.

## Example

A row authored under schema A arrives before the A→B lens. The receiver parks it, installs the validated lens/catalogue transition, then projects it exactly once into B. It must never decode A bytes as B or expose an unprojected intermediate row.

## Pointers

Catalogue/lens code is in `crates/jazz/src/db/catalogue.rs`, `crates/jazz/src/tools/schema_lens/`, and `crates/jazz/src/node/tests/catalogue_lenses/`. Deployed sync coverage is `crates/jazz-testkit/tests/catalogue_sync_integration.rs` and `schema_migration_policies.rs`. Public atomic administration is tracked by #1771.

aeplay:
🤖 Storage-settlement decision: use one tiny storage-epoch-pinned catalogue kernel as the only hard-coded schema. It defines only enough primitives to decode immutable record/enum descriptors, physical identities, schema lineages, and lenses. All higher Jazz system tables use the same catalogue/record machinery as app tables under a reserved system namespace. Physical tables, columns, and enum variants have permanent global IDs; names/docs are non-identifying metadata. A schema version is the hash of canonical semantic descriptor bytes: documentation and source order are excluded, every decoding-relevant change is included. Descriptors are immutable; evolution publishes a new version plus explicit lineage/lenses. Publication and activation are atomic, and data for unknown/incomplete versions is parked rather than decoded under another schema. The kernel changes only with a new storage epoch. Acceptance requires updating the normative specs with the kernel, identity, canonicalization, activation, failure, and evolution rules—not only implementation and fixtures.

aeplay:
🤖 A local catalogue-integrity repair now rejects standalone schema/lens rows whose stored key does not match content-derived identity and validates lens semantics against durable source/target schemas before residency. This is one prerequisite only; atomic lineage publication/activation, parking/projection, restart/reconnect, and retirement remain open here.

aeplay:
[...]

PR #2284, opened 2026-08-28
Complete catalogue activation and reconnect lifecycle

aeplay:
🤖 Completes the activation/reconnect slice of #1779 on top of #2267's permanent physical catalogue identities.

## Before

Activating a staged schema lineage persisted its active witness but left the corresponding durable pending obligation behind. After reopen, the catalogue could therefore carry both “active” and stale “still needs activation” state for the same publication.

Unknown-schema commit units are intentionally parked only in live link-delivery state. The spec did not clearly distinguish that transient parking from durable catalogue obligations.

## After

- The activation batch deletes the pending lineage in the same atomic write that exposes the schema, lens, physical mapping, and active witness.
- Reopen proves an active lineage has no stale pending obligation.
- An unavailable lineage retains its sequence and target reservations durably across reopen.
- A peer may retransmit the canonical unknown-schema unit after reconnect; it parks again before activation, drains/projects once after activation, and a later duplicate remains idempotent.
- The spec explicitly distinguishes transient commit-unit parking from durable pending catalogue lineage state.

For example, if a node restarts after receiving data for a schema it does not yet know, it does not guess-decode or persist that data under another schema. The peer retransmits it on reconnect; once the exact lineage activates, the row appears exactly once.

## Deferred boundary

Durably storing the parked commit unit itself would be a broader protocol/storage decision. The chosen current contract relies on canonical reconnect retransmission; this PR records that explicitly rather than implying otherwise.

## Verification

- focused activation, crash-boundary, reservation, reconnect, and projection receipts
- independent planted removal of the pending-delete operation made the reopen assertion fail


aeplay:

PR #2286, opened 2026-08-28
Freeze the Groove memory snapshot codec

aeplay:
🤖 Advances #2158 and #2160 by removing the remaining default-serde dependency from Groove's restart-authoritative memory snapshot boundary.

## Before

`MemoryStorage` snapshots used postcard directly over Rust container types. That made the durable test/restart format depend on serde traversal details rather than an explicit, reviewable storage contract.

## After

- Snapshots use an exact `GMS1` v1 byte format with explicit family/entry counts and length-delimited keys and values.
- Families and entries have one canonical byte order; duplicates, reordered aliases, truncation, trailing bytes, malformed names, and unknown versions reject.
- Decoding validates the complete replacement before assigning it, so any failure leaves the resident store unchanged.
- `INV-STORAGE-34` records the boundary and its intended scope.

For example, reversing two otherwise valid column-family records no longer decodes to the same map: it is rejected as a noncanonical snapshot spelling.

## Scope

This is a complete ordered-KV snapshot for the restart/test harness. It is not the logical cross-backend interchange format and does not claim migration of derived state.

## Verification

- exact golden semantic↔bytes fixture
- malformed, trailing, noncanonical, and atomic no-replacement receipts
- full Groove suite: 589 passed, 2 ignored
- independent review disabled canonical re-encoding; the planted reverse-order snapshot was then incorrectly accepted and the golden receipt failed


aeplay:

PR #2287, opened 2026-08-28
Run React Native Android device admission acceptance

aeplay:
🤖 Continues #2278 with the first authoritative Android-emulator receipt from the staged native relay.

## Before

The acceptance app was compile-shaped only. It could verify source contracts, but did not build all relay ABIs, assemble/install an APK, cross the trusted native admission boundary, or bind a claimed pass to the exact installed binary and emulator run.

## After

- A task-local pinned JDK/Android SDK/NDK/cargo-ndk bootstrap builds and verifies the four-ABI relay artifact.
- Expo/Gradle assembles the release APK and registers the native fixture module.
- The strict driver installs and launches on an emulator, then accepts only a fresh receipt bound to a random run nonce, APK SHA-256, platform, ordered sequence/timestamp, and host `Build.FINGERPRINT`.
- The implemented `linked-abi-admission` scenario proves the native library loaded at ABI 3 and trusted admission returned an opaque capability without exposing fixture identity or claims to JavaScript.

For example, stale logcat output from another APK or emulator run is rejected because its nonce, binary hash, or host fingerprint cannot match the current driver expectation.

## Honest current boundary

`local-write-subscription`, reconnect/reopen, scope isolation, and logout/auth-switch remain TODO and are not included in the passed scenario set. This PR proves packaging, loading, and trusted admission on Android; it does not claim full RN end-to-end usability yet.

## Device receipt

On the local KVM emulator:

```text
pnpm --filter rn-device-acceptance device:android
exit 0; linked-abi-admission passed
```

The same run used a freshly assembled 215 MB release APK and verified relay artifacts. Source protocol tests and TypeScript checks pass.

## Friction captured

- The host account lacked its existing `kvm` group membership; a fresh group context was required.
- Expo production assembly requires explicit `NODE_ENV=production`.
- The config plugin must be a direct dependency.
- Android's app-scoped `ANDROID_ID` cannot serve as a host/device receipt identity; the immutable build fingerprint is used instead.


aeplay:
🤖 Hosted Android acceptance is now green on exact tip `98cd6d8e0`.

The receipt proves:
- all four Android relay ABI libraries were built and verified against the exact source revision;
- the pinned, lane-local AVD was created at the configured physical path;
- the emulator reached `sys.boot_completed=1` under a single bounded deadline;
- the release APK assembled, installed, and launched on a KVM-backed x86_64 emulator;
- the run-bound native ABI/admission receipt was observed by the strict driver.

The same run also passed both source-verification jobs and the Blacksmith iOS release-simulator acceptance job.

Run: https://github.com/garden-co/jazz/actions/runs/33189173057

aeplay:

PR #2290, opened 2026-08-28
Freeze nested catalogue schema and policy codecs

aeplay:
🤖 Continues #2037 on top of #2283.

## Before

#2283 freezes the outer catalogue key and row envelope, but two nested semantic fields still crossed the persistence boundary as generic JSON blobs:

- `ColumnType::Json.schema`
- `PolicyExpr::ExistsRel.rel`

That left their byte identity dependent on a general serializer and allowed semantically equivalent map orderings to have different persistent bytes. Several older policy collection decoders also trusted a declared `u32` count when reserving a vector, so a tiny corrupt payload could request a very large allocation before truncation was detected.

## After

Schema v12, lens v5, and permissions v2 use one explicit nested-v1 grammar for every persisted JSON and relation value. It has fixed tags, flags, field order, and bounded collection counts. JSON object keys are byte-sorted and duplicates are rejected. Every outer payload is decoded and re-encoded exactly, so alternate ordering, duplicate entries, suffixes, and non-canonical encodings fail closed.

Example: `{ "z": 1, "a": 2 }` is persisted with keys in canonical byte order (`a`, then `z`) regardless of construction order. An encoded object that presents `z` before `a`, repeats `a`, or appends unused bytes is rejected rather than normalized during recovery.

For permissions, the grammar covers relation inputs, predicates, value references, projections, joins, recursion bounds, and every list/boolean branch. A payload that declares billions of `And` children while containing no child bytes now fails the bounded-count check before allocation.

Schema hashes/metadata/object IDs and lens metadata/derived object IDs are bound before an entry becomes resident during recovery.

## Compatibility

There is intentionally no old nested-blob decoder. This is the zero-user storage-format settlement; unsupported bytes fail closed.

## Verification

- `cargo test -p jazz-server --lib`: 170/170
- canonical codec suite: 30/30
- exact schema and permissions goldens
- corrupt persistent nested payload rejects startup before recovery residency
- malformed ordering/version/suffix/duplicate/huge-count receipts
- planted canonical-check and count-bound regressions both make their receipts fail
- invariant registry, formatting, Clippy, and sensitive-data gate
- independent adversarial review passed after the count-allocation repair

## Remaining boundary

This freezes nested catalogue schema/policy bytes. Catalogue activation/reconnect lifecycle remains the separate stacked #2284; other storage families retain their own milestone issues and PRs.


aeplay:

PR #2292, opened 2026-08-28
Execute the binding corpus through NAPI and WASM

aeplay:
🤖 Continues #2044 on top of #2259.

## Before

#2259 freezes the wire/binding codec and checks Rust and TypeScript fixture sources, but it did not execute one Rust-owned frozen binding corpus through both generated native artifacts. A drift where NAPI and WASM exported different fixture bytes—or where the TypeScript reader stopped understanding one artifact—could therefore escape source-only comparison.

## After

Rust owns one `BINDING_CODEC_GOLDEN_FIXTURE`, generated from the production postcard encoders. Both NAPI and WASM expose that exact frozen corpus through their generated binding surfaces. The TypeScript compatibility matrix:

1. calls both real generated artifacts;
2. compares their parsed ordered records and exact `payload_hex` bytes with the Rust source fixture;
3. decodes every returned payload through the production TypeScript relation/delta reader;
4. asserts the semantic snapshot, delta, occurrence-key/index, terminal-event, and rejection results.

Example coverage includes empty, batched, non-adjacent, and deleted snapshots; added/updated/removed deltas; v1/v2 occurrence keys; indices; and terminal JSON event/rejection families. Literal JSON whitespace is deliberately not storage identity—the parsed record plus exact payload hex is.

## Boundary

This triangulates the binding ABI and production TS decoder. It does not claim that two shared implementations prove the Rust encoder correct: the Rust golden-producer test separately regenerates the fixture through the production postcard encoder. It also does not complete the broader direct wire-frame or physical-backend corpus tracked by #2044/#2160.

## Verification

- exact Rust golden-producer test passes
- artifact pipeline contract tests: 34/34
- source TypeScript fixture tests pass
- planted TypeScript semantic-decoder regression fails the golden expectations
- independent adversarial review found no source-level blocker

The local review lane intentionally did not rebuild the full NAPI+WASM artifact pair under disk pressure. CI's required `build:test-artifacts` path is therefore the authoritative final proof that both freshly generated artifacts execute this matrix; this PR is not ready until that check is green.

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -111,6 +111,14 @@ pub fn native_artifact_fingerprint() -> String {
         .to_owned()
 }
 
+/// Test-only bridge for executing the Rust-owned v1 binding corpus through the
+/// generated NAPI artifact. It intentionally returns the frozen corpus rather
+/// than a TypeScript reimplementation of its byte layouts.
+#[napi(js_name = "__testBindingCodecGoldenFixture", skip_typescript)]
+pub fn test_binding_codec_golden_fixture() -> String {
+    jazz::binding_codec::BINDING_CODEC_GOLDEN_FIXTURE.to_owned()
+}
+
 #[derive(Clone, Debug, Deserialize)]
 struct CoreOpenDbConfig {
     identity: CoreOpenDbIdentity,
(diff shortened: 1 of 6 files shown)


aeplay:

PR #2295, opened 2026-08-28
Pin the IndexedDB physical storage epoch

aeplay:
🤖 Continues #2160.

## Before

The production IndexedDB page store had a page-v2 codec, but no fixed top-level storage-epoch manifest. A store could therefore be opened without first proving that its adapter version, page version, codec, page size, checksum, and manifest shape matched the code interpreting its pages.

## After

IndexedDB stores one fixed `storage-manifest` / `epoch` record containing:

- storage epoch
- adapter/page format versions
- the only supported codec ID
- fixed 16 KiB page size
- page magic and xxh3 checksum identity

Open validates the exact manifest—including its exact key set—before returning a handle or mutating pages. A fresh epoch is installed atomically in the IndexedDB upgrade transaction. Unsupported predecessor layouts abort the upgrade transaction, preserving their version and raw pages; this zero-user settlement does not guess, fall back, or reinterpret them.

Example: an otherwise valid epoch-1 manifest with one additional `futureDecodeParameter` field is rejected. Both the Node raw-IDB receipt and a real Chromium receipt then read the underlying page directly and prove its bytes are unchanged.

## Verification

- physical-store Node suite: 11/11
- dedicated real-Chromium IndexedDB suite: 3/3
- actual committed page-v2 fixture opens and reads
- current write/reopen and stale-generation atomicity
- corrupt/unknown/extra manifest fields reject before mutation
- raw pages remain byte-identical after rejection
- Rust page magic/version/xxh3 malformed/canonical receipts
- planted epoch-validation and exact-key-set bypasses fail the corresponding receipts
- TypeScript typecheck, invariant registry, and independent adversarial review

## Boundary

The Chromium test uses the production TypeScript `IndexedDbPageStore` against real browser IndexedDB. Rust/WASM page decoding is covered separately by the Rust codec receipt; this PR does not claim an additional full WasmDb/catalogue browser corpus. SQLite and RocksDB compatibility fixtures are in #2281, and broader authoritative-codec coverage remains tracked by #2160.


aeplay:
🤖 Approved by Anselm after the IndexedDB adapter/page codec was corrected from the development-history V2 label to the sole supported V1 format.

Issue #2035, opened 2026-08-25
Freeze Groove record, scalar, and ordered-key bytes


aeplay:
🤖 Deep-review and freeze the durable bytes produced by `groove::records::{RecordDescriptor, ValueType}` and `groove::db::encoding::{encode_record, encode_primary_key_part, encode_ordered_bytes, encode_index_prefix_part}`.

Build a byte fixture corpus covering logical versus physical field order, nullable and array offsets, payload enums, signed/floating values, UUIDs, NUL escaping, primary keys, index suffixes, and malformed, trailing, or non-canonical inputs. Verify the same corpus through Memory, RocksDB, SQLite, and IDB, including reopen and ordered scans.

The review must preserve `INV-STORAGE-7..15`, `INV-STORAGE-22`, `INV-STORAGE-25`, and `INV-STORAGE-27`: representation-only changes must not reorder scans or reinterpret existing rows.

aeplay:
🤖 Storage-settlement decision: authoritative persisted structured values use the normative Groove record/scalar encoding, with internal records described by explicit versioned descriptors. Field IDs and enum tags are permanent numeric identities; declaration or Rust source order is not identity. Ordered physical keys use a separately normative order-preserving key codec. Default serde/postcard/JSON serialization is not authoritative storage. Opaque bytes are allowed only at a narrow, registered boundary with an explicit codec version, canonical rules, and golden fixtures. Convenience serialization is allowed only for explicitly discardable/rebuildable caches. Jazz system records use the same typed machinery as app records, under reserved and versioned system descriptors.

aeplay:
🤖 Storage-settlement decision: use one tiny storage-epoch-pinned catalogue kernel as the only hard-coded schema. It defines only enough primitives to decode immutable record/enum descriptors, physical identities, schema lineages, and lenses. All higher Jazz system tables use the same catalogue/record machinery as app tables under a reserved system namespace. Physical tables, columns, and enum variants have permanent global IDs; names/docs are non-identifying metadata. A schema version is the hash of canonical semantic descriptor bytes: documentation and source order are excluded, every decoding-relevant change is included. Descriptors are immutable; evolution publishes a new version plus explicit lineage/lenses. Publication and activation are atomic, and data for unknown/incomplete versions is parked rather than decoded under another schema. The kernel changes only with a new storage epoch. Acceptance requires updating the normative specs with the kernel, identity, canonicalization, activation, failure, and evolution rules—not only implementation and fixtures.

[...]

PR #2296, opened 2026-08-28
Pin native record byte receipts

aeplay:
🤖 Completes the remaining native-backend receipt scope in #2035.

## Existing frozen boundary

The merged epoch-one Groove corpus already pins `INV-STORAGE-7..15`, `-22`, `-25`, and `-27`: scalar/array/payload-enum bytes, nullable offsets, ordered primary/index keys, malformed/trailing/non-canonical rejection, and Memory reopen behavior.

The remaining gap was showing the same literal semantic-table row bytes and scan order through the native persistent adapters, not merely through the in-memory codec.

## Added receipts

RocksDB and SQLite now persist two deliberately awkward records through their production ordered-KV adapters, inspect the raw prefix scan, close/drop every handle, reopen physically, and decode through the public record declaration.

Frozen raw keys:

- `01 00 02`
- `01 12 34`

Frozen raw values:

- `00 02 00 00 02 61`
- `00 34 12 01 02 68 69`

The leading `00` is the canonical record variant tag. The schema deliberately declares variable-width `label` before fixed-width `id`/`enabled`; the bytes prove fixed fields are stored first for canonical layout while reopened public rows preserve logical declaration order. The keys prove lexicographic U16 primary-key ordering.

## Verification

- RocksDB plain-row receipt suite: 5 passed (plus one pre-existing ignored test)
- SQLite ordered-KV suite: 13 passed
- Groove `epoch_1_*` corpus: 10 passed
- fresh physical close/reopen on both native backends
- planted record-variant-byte regression fails exact raw-row comparison
- Clippy, rustfmt, sensitive-data gate, and independent adversarial review pass

## Non-goals

This does not define backend-file interchange; the invariant is the Groove semantic/key bytes observed through each adapter. IndexedDB’s physical manifest and real-browser receipt are covered separately by #2295/#2160.

### DIFF
--- a/crates/jazz-storage-rocksdb/tests/plain_row_receipt.rs
+++ b/crates/jazz-storage-rocksdb/tests/plain_row_receipt.rs
@@ -29,6 +29,21 @@ fn schema() -> DatabaseSchema {
     .with_primary_key(PrimaryKey::new("id", IntegerKeyType::U64))])
 }
 
+fn epoch_1_codec_fixture_schema() -> DatabaseSchema {
+    // Declaration order is deliberately variable-before-fixed. The frozen
+    // stored bytes below prove that every durable backend shares Groove's
+    // physical fixed-first record layout without changing public row order.
+    DatabaseSchema::new([TableSchema::new(
+        "records",
+        [
+            ColumnSchema::new("label", ColumnType::String),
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2298, opened 2026-08-28
Serialize auth secret and database transitions

joeinnes:
🤖 Serialize authentication-secret and database ownership transitions so the newest user/session action wins, independent of asynchronous completion order.

## Before

- Expo serialization/cache lived on each `ExpoAuthSecretStore` object. Two instances targeting the same SecureStore key could race: an older `getOrCreateSecret()` read could finish after a newer imported secret or logout and restore stale state.
- Hybrid session transitions could finish out of order and install an older database.
- JWT-expiry listeners belonged to the original database while closing over the mutable current `db`. A delayed token refresh started by session A could therefore install A’s token on session B’s replacement database, which itself had no listener.
- A session change during the initial asynchronous database open could occur before subscription and never be reconciled.

## After

- Operation state and cache are shared per SecureStore object/key, including direct store instances; reads, creation, import, and clearing run through the same serialized tail.
- Authentication generations prevent stale database opens from becoming current and close any stale result.
- Every owned database gets its own expiry listener. Replacement attaches the new listener, swaps the app database, unsubscribes the old listener, and closes the old database.
- Token refresh checks database ownership both before and after fetching, so a delayed old-session response cannot reach its replacement.
- The current session is read again after subscription to reconcile changes during startup.

## Verification

- Direct same-key Expo instances, create/import/clear ordering, and cache failure recovery are covered.
- Hybrid receipts cover stale database opens, delayed stale JWT refresh, refresh on the new database, and a startup session race.
- A planted removal of the ownership check installs the stale token and fails its receipt.
- React/TypeScript hybrid starter test globs now run the matching source tests.
- Focused Expo, TS-hybrid, React-hybrid, parity, formatting, and TypeScript gates pass.

### DIFF
--- a/.changeset/serialize-auth-secret-transitions.md
+++ b/.changeset/serialize-auth-secret-transitions.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Serialize authentication secret and hybrid database transitions so stale asynchronous work cannot restore superseded credentials or database instances.
--- a/packages/jazz-tools/src/expo/auth-secret-store.test.ts
+++ b/packages/jazz-tools/src/expo/auth-secret-store.test.ts
@@ -23,6 +23,34 @@ function recordingStore() {
   return { secureStore, getItemAsync, setItemAsync, deleteItemAsync };
 }
 
+function deferred<T>() {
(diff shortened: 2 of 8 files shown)


aeplay (approved):

PR #2299, opened 2026-08-28
Reject unorderable TopBy fields

joeinnes:
🤖 Reject `TopBy` plans whose order or tie-breaking keys cannot be compared, before hydration reaches the runtime comparator.

## Before

A query could validate and start with a structured field such as a record or array in either the primary `orderBy` position or a later tie field. The runtime comparator has no total order for those values, so ticking the graph could panic.

For example, both of these are invalid and now fail during graph validation:

```text
TopBy(order = profile_record, tie = id)
TopBy(order = score, tie = metadata_array)
```

## After

- Graph validation and GraphBuilder compilation inspect every flattened order and tie-field index.
- Structured/unorderable fields return `TopBySortFieldMustBeOrderable` before subscription hydration or ticking.
- Supported nullable and scalar values keep their existing ordering semantics.
- Fixed-width sort keys remain inline; strings and bytes own their variable-width payload so retained keys cannot borrow transient row memory.

## Verification

- Direct graph-level receipts cover invalid primary and tie fields.
- A planted regression that validated only the first key was caught by the new tie-field receipt.
- `cargo test -p groove top_by -- --nocapture` — 20 passed.
- `cargo fmt --check` and focused clippy passed.

### DIFF
--- a/crates/groove/src/db/tests/queries/windows.rs
+++ b/crates/groove/src/db/tests/queries/windows.rs
@@ -563,9 +563,10 @@ async fn top_by_rejects_record_array_sort_fields_before_evaluation() {
             "ranked",
             vec![
                 Value::U64(id),
-                Value::Array(vec![Value::Record(
+                Value::Array(vec![Value::Record(crate::records::OwnedRecord::new(
                     child.create(&[Value::U64(rank)]).unwrap(),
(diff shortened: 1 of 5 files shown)


aeplay (approved):

Issue #2275, opened 2026-08-28
RecordPlayer edge query can stall after persistent reopen


aeplay:
🤖 A deterministic RecordPlayer topology receipt intermittently stalls after reopening the editor persistent store following authorization revocation.

Observed in CI run 33161730693, job 98817548467:

- scenario: record-player.playlist-auth-reconnect
- seed: 41
- query tier: edge
- expected result: empty, because the reopened editor remains revoked
- observed local result state: lastRowsCount=0 and lastRows=[]
- failure: db.all did not resolve within 4.4 seconds; the scenario retried until its 15 second deadline

The same seed passed on main, the parent branch twice, the prior RN branch tip, and the current RN tip, with the reopen phase completing in roughly 300ms. The RN diff does not touch browser or topology code. This therefore looks like a pre-existing intermittent liveness failure rather than a result-correctness disagreement.

Investigate why an edge-tier one-shot query can retain the correct empty result yet fail to settle after persistent reopen. Add a deterministic lower-level receipt around the suspected connection/subscription transition rather than increasing the timeout.

Replay:

JAZZ_EXAMPLE_TOPOLOGY_SEED=41 pnpm --dir examples/record-player/apps/next-betterauth test:browser -- topology.e2e.test.ts

aeplay:
🤖 Verified resolved on current `main` by #2193 (`b16dabaee`). The relay worker now advances the subscriber-dirty generation and schedules an immediate tick after applying an authority-view update, so a reopened tab no longer waits for unrelated traffic before its one-shot Edge settlement.

The exact public regression replay now passes unchanged:

```text
JAZZ_EXAMPLE_TOPOLOGY_SEED=41 pnpm --dir examples/record-player/apps/next-betterauth test:browser -- topology.e2e.test.ts
9 passed; persistent reopen phase: 316ms
```

The deterministic lower-level receipts also pass:

- `browser_relay_hydrates_fresh_included_edge_subscription_from_authority`
- `reopened_browser_tab_hydrates_from_worker_authority_state`
- `browser_relay_publishes_an_explicit_settled_empty_handoff`

Closing because both the inferred mechanism and the exact adopter-style replay are now covered.

aeplay:
🤖 Reopening: a later current-main CI receipt contradicts the single successful replay above and proves this is intermittent.

In #2279, CI run `33164784071`, job `98827489942`, failed the same seed and exact persistent-reopen phase:

```text
[...]

Issue #2288, opened 2026-08-28
Fix unsafe RecordPlayer invitation update permissions


joeinnes:
## Problem

RecordPlayer authorises two different updates to `invitations`:

1. A playlist owner may manage invitations for their playlist.
2. A recipient may accept their own pending invitation.

The current policy does not preserve those transitions as separate old/new pairs.

First, it chains `.whereNew(...)` twice. `UpdateRuleBuilder.whereNew()` replaces the stored new-row condition, so the second call discards the first call's checks intended to preserve the invitation ID, playlist, subject and role.

Second, the owner and recipient rules are merged by independently ORing their `USING` and `WITH CHECK` predicates. The resulting policy is effectively:

```text
old: owner_of(old.playlist_id)
     OR (old.subject = session.user AND old.status = pending)

new: owner_of(new.playlist_id)
     OR (new.subject = session.user AND new.status = accepted)
```

This admits cross-rule combinations that neither source rule authorises.

## Exploit

A user who owns playlist A can take an invitation attached to A and update it to:

- reference another user's playlist B;
- name themselves as the subject;
- grant the editor role;
- set the status to accepted.

The old row passes the owner branch. The new row passes the recipient-accepted branch. RecordPlayer can then treat the forged row as an accepted editor invitation for playlist B.

## Required behaviour

- A recipient may only perform the intended `pending` to `accepted` transition on their own invitation.
- Recipient acceptance must preserve the invitation ID, playlist, subject and role.
- A playlist owner must not transfer an invitation or its authority to a playlist they do not own.
- Alternative update rules must not form an old/new cross-product.
- Add an end-to-end regression proving the forged transfer is rejected and grants no playlist-entry authority.

## Design constraint

The current policy representation carries one independently evaluated old-row predicate and one new-row predicate. Correctly supporting alternative asymmetric transitions requires preserving rule pairs through evaluation:

```text
(old_A AND new_A) OR (old_B AND new_B)
```

Until that representation exists, use a fail-closed design such as a separate invitation-acceptance record or a server-mediated acceptance action. PR #2277 adds rejection of unsupported asymmetric merges.

aeplay:
🤖 Verification receipt

**PR:** #3283, which adds the missing end-to-end regression. It is test-only: the policy fix on `main` (#2277 plus the current RecordPlayer `permissions.ts`) is unchanged.

[...]

PR #2301, opened 2026-08-28
Deliver NAPI mutation errors as one callback argument

joeinnes:
🤖 Deliver NAPI mutation events with the same one-argument callback shape declared by the public runtime-source contract.

## Before

The napi-rs threadsafe function used its default error-first convention. A native mutation event therefore reached JavaScript as:

```text
callback(null, event)
```

The Jazz runtime registered `callback(event)`, so it observed `null` instead of the rejection. A rejected mutation could consequently appear as a timeout or fatal native callback error rather than a structured transaction rejection.

## After

This one threadsafe function disables napi-rs error-first delivery and emits exactly:

```text
callback(event)
```

The event retains its transaction ID and rejection details. Other NAPI callback APIs keep their existing conventions; this is not a repository-wide callback reinterpretation.

## Verification

- A real rejected native mutation reaches JavaScript as exactly one structured event argument.
- A planted restoration of the default convention fails with the observed `[null, event]` shape.
- Fresh native binding E2E and public declarations pass.
- The package changeset covers the binding/runtime consumers.
- Full CI rerun is green; the earlier RecordPlayer failure was independently tracked as #2275.

### DIFF
--- a/.changeset/fix-napi-mutation-error-callback.md
+++ b/.changeset/fix-napi-mutation-error-callback.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Deliver native mutation errors as the single callback argument declared by the JavaScript API.
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -2377,13 +2377,13 @@ impl NapiDb {
     )]
     pub fn on_mutation_error(
         &self,
-        callback: ThreadsafeFunction<JsonValue, ()>,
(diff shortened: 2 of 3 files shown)


joeinnes:
CI failed in the existing RecordPlayer authenticated playlist topology test while reopening persistent editor storage after revocation. The branch-specific NAPI mutation callback test passed in the same run, as did lint and both Rust suites. This appears unrelated to this PR and is tracked in #2288, so I am leaving the PR in draft.

aeplay (approved):

PR #2302, opened 2026-08-28
Enforce migration integrity before publication

joeinnes:
🤖 Enforce migration integrity and keep executable migration inputs inside the project storage boundary.

## Before

- Lens row operations could be applied to a table other than the table declared by the operation.
- A migration filename could claim one transition while its embedded metadata or schema witnesses described another.
- Concurrent pushes shared predictable staging paths.
- A symlinked migrations directory or executable `.ts` migration could be followed before publication checks. For example, `migrations/2026…ts -> ../outside.ts` could make the CLI bundle source outside the migration store.

## After

- Each lens operation is applied only to its declared table.
- Filename hashes, embedded transition metadata, and both schema witnesses must agree with the schemas being published.
- Every push uses a private process-unique bundle and removes it on success or failure.
- `pushMigration` initializes and validates the migration-storage boundary before network access. Symlinked directories and executable migration inputs fail closed before bundling or publication; project deploy applies the same rule before enumerating migrations.

## Verification

- Cross-table lens operations are ignored by the wrong table and exercised against the intended table.
- Tampered hash, operation metadata, and schema-witness cases are rejected before publication.
- Concurrent/private staging and cleanup are covered.
- Planted symlinked-file and symlinked-directory cases prove that neither the server nor publication endpoint is reached.


aeplay (approved):

PR #2303, opened 2026-08-28
Freeze the Jazz class storage layout

aeplay:
🤖 Closes the remaining physical-layout gap tracked by #2160.

## Before

`StorageLayout::JazzClassV1` already wrote a durable marker and framed logical keys, but the exact marker, classifier precedence, and key grammar were not fully normative or covered by raw backend receipts. A public partial-selection path could also interpret the same marked store differently on later opens.

For example, a logical history key was physically stored as:

```text
u32be(len("jazz_albums_history")) | "jazz_albums_history" | logical_key
```

but that byte contract was not frozen end to end.

## After

- V1 mapping is total: every classifier match is mapped identically on every open.
- The exact marker, ordered classifier, class families, and framed-key grammar are normative in SPEC 2 / `INV-STORAGE-34`.
- Logical family names are validated before framing or atomic writes.
- Marker creation fails closed when family enumeration is unavailable or any classifier-matching legacy family exists, even if that legacy family is empty.
- Memory, SQLite, and RocksDB receipts pin the raw marker and keys, including the special `indices` family.
- Malformed, old, and future markers fail without normalizing or mutating stored data.

## Stack

This is stacked on #2281 because it extends the same epoch-one physical compatibility corpus. It introduces no legacy migration or fallback behavior.

## Verification

- Groove class-layout receipts: 9/9
- SQLite raw/reopen receipts: 2/2
- RocksDB raw receipt: 1/1
- invariant registry: 0 missing citations
- independent adversarial review with planted partial-mapping and non-enumerating-adapter regressions


aeplay:

PR #2304, opened 2026-08-28
Compose complete codec profiles into storage manifests

aeplay:
🤖 Advances the final codec-inventory work in #2160.

## Before

The storage epoch manifest claimed to list every authoritative persistent codec, but it named only Groove ordered-KV. A Jazz database could therefore change a durable Jazz payload—or Groove's large-value/chunk envelopes—without changing the manifest that guards its physical epoch. Some shipping opens also constructed a generic Groove-only manifest even though they persisted Jazz data.

For example, NAPI persistent storage previously admitted this incomplete profile:

```text
groove.ordered-kv.v1
```

even though the same root could contain transaction, catalogue, branch-index, and large-value payloads.

## After

- Groove owns and requires a complete generic base:
  - `groove.ordered-kv.v1`
  - `groove.large-value.v1`
  - `groove.ordered-chunk-storage.v1`
- Jazz supplies a deterministic, sorted closed profile for every Jazz-owned durable codec family.
- The server catalogue adds its root-local catalogue-entry codec without teaching Groove Jazz semantics.
- NAPI's three persistent opens, the native relay SQLite root, server catalogue, RocksDB/SQLite factories, and browser IndexedDB all enforce the appropriate composed profile before mutation.
- Missing, additional, duplicate, unsorted, or future codec profiles fail closed.
- Exact Rust and TypeScript manifest receipts pin the same 14-codec Jazz profile.

## Stack

This is stacked on #2295 because IndexedDB's newly pinned physical epoch must carry the same final profile as native adapters. It should be linearized with the catalogue stack before merge; it does not introduce compatibility fallback for earlier alpha manifests.

## Verification

- Groove manifest suite: 9 tests
- RocksDB: 18 tests
- SQLite: 4 tests
- IndexedDB page-store: 11 tests
- jazz-server: 163 tests
- jazz-napi: 24 tests
- native relay: 20 tests
- TypeScript, invariant registry, formatting, Clippy
- independent planted omissions at NAPI, relay, server catalogue, Groove base, and IDB boundaries


aeplay:

PR #2306, opened 2026-08-28
Freeze Node catalogue lineage and lens codecs

aeplay:
🤖 Completes the Node-local durable catalogue boundary tracked by #2037 and #1779.

## Before

Server catalogue entries had canonical versioned codecs, but restart-authoritative Node state still persisted `MigrationLens`, staged lineage, and pending lineage as raw `serde_json`. The wire representation could also deserialize arbitrary lens fields before validation, allowing equivalent mappings to hash differently, oversized nested input to allocate before bounds, or invalid default values to reach a panic in content-ID calculation.

For example, these equivalent table-wise transforms could previously produce different identities—or ambiguous reconciliation:

```text
[a → b, c → d]
[c → d, a → b]
[a → b, a → c]
```

## After

- Node-local standalone lenses and staged/pending lineage use versioned, exact-consumed typed envelopes; no runtime persistence fallback to JSON remains.
- Protocol `MigrationLens` is explicitly distinct from the server's schema-editing `LensTransform`; no invented conversion blurs their semantics.
- `MigrationLens::new` is fallible, canonicalizes table order, and requires unique source and target tables.
- Public fields are read-only; invalid storage-only or descriptor-less defaults cannot be manually constructed or hashed.
- Add/drop defaults are checked against endpoint schema types before admission.
- Storage and postcard wire both use the same canonical lens-byte grammar.
- Postcard deserializes borrowed bytes, checks the 32 MiB outer budget before allocation/parsing, then applies aggregate item, byte, and nesting limits.
- Wrong IDs, duplicate mappings, malformed/future tags, truncation, trailing bytes, oversized declarations, and corrupt reopen fail closed without partial residency.

## Worked failure example

A remote `PublishLens` carrying `a → b, a → c` now fails during bounded deserialization/validated construction. It never becomes a `MigrationLens`, never reaches `content_id()`, and cannot alter catalogue state.

## Stack

This is the top of catalogue stack #2300:

1. #2267 permanent physical identities
2. #2283 outer server catalogue codec
3. #2290 nested server schema/policy codecs
4. #2284 activation/reconnect lifecycle
5. this PR: Node-local lineage/lens storage and wire codec

There is no compatibility path for earlier experimental JSON payloads.

## Verification

- full Jazz all-target compile and no-run test inventory
- exact lens/staged/pending byte receipts and trusted snapshot reopen coverage
- malicious postcard Record/Enum/Large, duplicate-source/target, wrong-ID, count/byte/depth bomb receipts
- direct borrowed-byte outer-budget error receipt plus separate postcard no-panic receipt
- independent planted removal of bijection and outer-bound guards


aeplay:

Issue #2291, opened 2026-08-28
Complete native React Native foreground acceptance


aeplay:
🤖 Alpha.54 single-root native React Native foreground acceptance is complete on release runtime `a848ea51ef2fbe5e697e17e17fe13fd8dee4a2c7`, packaged as `193f5894d1d1f305fd2045b4a149c0305a93ab13`.

The native foreground → relay → Edge → Core path is implemented. The final installed Android emulator and iOS simulator journeys passed persistent write, fresh foreground readback, offline process restart/readback, scope isolation/revocation and the independently observed run-bound Core marker: https://github.com/garden-co/jazz/actions/runs/33998136838. Source-scaffold jobs are not the basis for this result, and no physical-iPhone RN receipt is claimed.

All final candidate CI partitions passed: https://github.com/garden-co/jazz/actions/runs/33998110029. Packaged RN consumers and pkg.pr.new publication also passed: https://github.com/garden-co/jazz/actions/runs/33998124267. These supersede the older intermediate integration receipts.

The currently remaining release fixes are test-harness and type-import changes rather than native runtime/storage changes. #1974 retains the broader installed scenario work, including the stronger reconnect test in #2604. Simultaneous physical roots remain explicitly deferred to alpha.55/#2532. The forthcoming realistic app-building session will exercise the refreshed preview as an external consumer; it is not claimed complete here.

aeplay:
🤖 Overnight feasibility work confirms the remaining acceptance gap cannot be filled by a small binding repair on the current Expo SDK 54 / React Native 0.81.5 host. The existing acceptance app has one physical installed JSI runtime; its two Jazz instances are aliases within that runtime.

On Android, Expo’s `createFromReactNativeHost` path owns a process-global React host singleton. Constructing a genuinely second physical runtime would require an app-owned host lifecycle around unstable React Native host APIs while preserving Expo delegate/setup behavior. On iOS, the equivalent requires a second app-owned factory lifecycle; current Expo integration also owns one factory and its own source notes incomplete multi-host behavior. A raw second Hermes runtime would not prove a React Native app lifecycle, and a second OS process would not exercise the required process-local relay/subscription sharing.

Recommended acceptance design:

1. Pin exact RN/Expo versions and define an app-owned two-host lifecycle abstraction for Android and iOS.
2. Prove distinct native runtime tokens and distinct JSI installations without using a database side channel.
3. Have runtime B subscribe, then report only after a Rust-produced delta from runtime A.
[...]

PR #2310, opened 2026-08-28
Specify the React Native foreground runtime boundary

aeplay:
🤖 Specify and scaffold the native JSI boundary required for real foreground Jazz runtimes on React Native. This advances #2291 without claiming the two-runtime device scenario is implemented yet.

## Decision

React Native will not ship the browser WASM runtime or depend on Hermes gaining `WebAssembly`. Each foreground app instance owns a memory-only Groove/Jazz engine exposed through a thin native JSI byte API. Those engines attach, using an admitted 32-byte capability, to the separately owned persistent native relay already established by the lower stack.

This keeps the semantic split explicit:

- **foreground runtime:** app/query/IVM state, one per JS runtime, memory-only;
- **native relay:** trusted admission, persistent SQLite-backed sync/authority transport, shared per auth scope;
- **JS adapter:** existing encoded `NativeDb` contract, lifecycle ownership, and callbacks;
- **future Swift/Kotlin:** consume the same native owner/command ABI without inheriting React Native objects.

## Call flow

```text
JS runtime
  → install native JSI foreground factory
  → openAttached(admittedCapability)
  → native owner thread creates memory-only Jazz engine
  → engine exchanges frames through the admitted persistent relay
  → bounded callbacks return encoded query/write/subscription events
```

Opening, callback delivery, close, capability revocation, bridge reload, and native-host shutdown have distinct documented ownership and ordering rules. Expo prebuild and bare RN include the native module; Expo Go remains unsupported because it cannot contain this code.

## This PR

- Adds the normative architecture, threading, teardown, packaging, and staged two-runtime acceptance plan.
- Adds a capability-only V1 factory installer scaffold.
- Clears the private JSI slot before every install and requires native code to synchronously create a fresh own factory; a same-ABI HostObject from an old bridge cannot be reused.
- Rejects non-`Uint8Array` and non-32-byte capabilities before crossing JSI. Native admission remains authoritative.
- Keeps the current relay-only public surface working while the actual native `NativeDb` engine is implemented.

## Not yet claimed

The C++/Rust JSI `NativeDb` byte adapter, real foreground engine owner, complete callback bridge, and two-foreground device receipt remain follow-up work under #2291. This PR deliberately throws a clear installation error when the packaged native build does not yet provide that engine.

## Verification

- RN Jest suite: 8/8.
- RN TypeScript check passes.
- Tests reject missing, malformed, ABI-incompatible, stale/no-op installers and malformed capabilities.
- Independent review planted a same-ABI stale factory; the repaired installer rejects it before any call.

### DIFF
--- a/crates/jazz-native-relay/include/jazz_native_relay.h
+++ b/crates/jazz-native-relay/include/jazz_native_relay.h
@@ -3,6 +3,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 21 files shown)


aeplay:

PR #2311, opened 2026-08-28
Prove a fresh packed React Native install

aeplay:
🤖 Prove that an adopter can install the published `jazz-rn` package into a fresh Expo or bare React Native app without workspace-only wiring.

## Before

The repository acceptance app proved native code could build, but it lived inside the monorepo and reused its dependency graph. A missing peer dependency, omitted podspec/config file, or workspace-only source import could therefore pass while an installed npm tarball failed to prebuild or autolink.

## After

The packaging gate now:

1. packs the actual `jazz-rn` npm tarball;
2. installs that tarball into an isolated directory using offline npm resolution;
3. copies only the installed package into a newly generated minimal Expo SDK 54 app;
4. provides only Expo/React/React Native/TypeScript peers;
5. typechecks the package's published declarations;
6. runs clean Android and iOS Expo prebuilds with New Architecture enabled;
7. verifies Expo autolinking on both platforms;
8. verifies bare RN discovers the Android package and iOS podspec;
9. proves resolution comes from the npm-installed tarball, never workspace source.

The install docs describe the direct dependency/config-plugin path, bare RN New Architecture requirement, and why Expo Go cannot work with a custom native module.

## Boundary

This receipt proves installation, code generation, and native discovery. It does not claim that an arbitrary fresh app already contains sealed relay artifacts, trusted admission configuration, or the foreground engine from #2291; those remain explicit runtime requirements.

## Verification

- Packaging gate: 11/11.
- Planted Android package-name corruption fails bare RN discovery.
- Planted missing `react-native` peer metadata fails the packed-manifest assertion.
- A manual current `create-expo-app` scaffold also accepted the packed tarball and completed Android/iOS prebuild plus TypeScript checking.

### DIFF
--- a/crates/jazz-rn/README.md
+++ b/crates/jazz-rn/README.md
@@ -57,6 +57,43 @@ artifact.
 The wrapper accepts ABI 3, which uses opaque host-generated admission
 capabilities and trusted revocation.
 
+## Expo development-build install path
+
+`jazz-rn` is a direct application dependency: React Native codegen and Expo
+autolinking must discover it from the application, not through `jazz-tools`.
+For an Expo development build, the minimal configuration is:
+
+```bash
+pnpm add jazz-rn@alpha
+```
+
+```json
+{
+  "expo": {
+    "plugins": ["jazz-rn"]
+  }
+}
+```
+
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2312, opened 2026-08-28
Canonicalize server catalogue metadata bytes

aeplay:
🤖 Fixes a persistent-byte nondeterminism discovered while auditing the complete storage-codec registry.

## Before

`CatalogueEntry::encode_storage_row` serialized its durable `metadata: HashMap<String, String>` directly with `serde_json`. Two processes could therefore encode the same logical metadata in different key orders, depending on each process's hash seed.

For example, `{ "region": "iad", "role": "core" }` and `{ "role": "core", "region": "iad" }` were semantically equal but could produce different persistent bytes. The existing receipt used one metadata key, so it could not detect this.

## After

- Encoding first projects metadata into lexicographically sorted keys and emits one compact JSON spelling.
- Decoding consumes the value, validates the catalogue key, and byte-for-byte re-encodes it. Reordered keys, whitespace, duplicate keys, trailing bytes, and alternate spellings fail closed before catalogue admission.
- The exact V1 bytes are pinned by a hard-coded multi-key fixture and remain identical across insertion orders.
- The Jazz storage spec now states this canonical root-local codec rule explicitly.

This changes only the opaque server-catalogue entry codec. It does not change application-row record encoding, the Groove scalar codec, catalogue ordering semantics, or wire messages.

## Verification

- Exact multi-key storage-row fixture
- Equivalent insertion-order receipt
- Reordered persistent-byte rejection
- Malformed `cat:` key rejection before publication
- Independent adversarial review with a planted canonical-equality bypass

Stacked on #2304 because the server root composes this root-local family with the complete Jazz/Groove codec manifest.


aeplay:
🤖 Superseded by #2283 in the unified non-RN stack.

Once the canonical server-catalogue JCAT representation from #2283 is below the codec registry, this PR's sorted-JSON repair is no longer needed: #2283 removes that representation entirely. Keeping both would reintroduce a second persistent encoding path. The registry coverage from the upper stack now points at #2283's canonical fixtures instead.

PR #2313, opened 2026-08-28
Register every persistent codec family

aeplay:
🤖 Adds one machine-checked inventory for every persistent codec family that participates in the settled Jazz/Groove storage profiles.

## Why

The storage freeze had exact fixtures and malformed-input receipts distributed across many crates, but no executable answer to: "which durable family is this profile promising, and where is its compatibility proof?" That allowed a one-key server-catalogue fixture to hide nondeterministic multi-key metadata encoding.

## What is registered

Each family names:

- its owning specification section;
- an exact-byte or historical fixture;
- a malformed/noncanonical rejection receipt;
- reopen or backend evidence where applicable;
- membership in the Groove base, Jazz base, browser, native, and server-root profiles.

For example, `jazz.server-catalogue-entry.v1` points to the canonical multi-key fixture introduced by #2312 and its alternate-spelling rejection test. Large-value descriptors/nodes, branch keys, transaction/version rows, catalogue payloads, result members, program facts, binding/wire boundaries, and adapter manifests are similarly explicit.

## Fail-closed maintenance

The registry test rejects:

- a profile family added without a registry entry;
- a registered family omitted from its declared profile;
- stale or non-unique source anchors;
- missing, absolute, escaping, or symlink-escaping paths;
- unknown JSON fields;
- duplicate IDs and misplaced wire/auth-only boundaries.

This is a test/maintenance contract, not a new runtime registry or a decoder-dispatch path. Runtime storage admission remains the manifest/profile mechanism from #2304.

Independent adversarial review planted profile omissions, stale anchors, path escapes, unknown fields, and a canonical-decoder bypass; the corresponding receipts failed as intended.

### DIFF
--- a/crates/jazz/fixtures/persistent_codec_family_registry.json
+++ b/crates/jazz/fixtures/persistent_codec_family_registry.json
@@ -400,16 +400,16 @@
       },
       "semantic_fixture": {
         "path": "crates/jazz-server/src/server/catalogue_entry.rs",
-        "anchor": "catalogue_entry_metadata_json_is_sorted_and_noncanonical_spelling_is_rejected"
+        "anchor": "storage_row_v1_golden_is_exact_and_rejects_alternates"
       },
       "rejection_receipt": {
         "path": "crates/jazz-server/src/server/catalogue_storage.rs",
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2315, opened 2026-08-28
fix(inspector): attach with the host's verified session

nrainhart:
## Stack

1. **This PR** — fix inspector reads and CI coverage
2. `refactor/remove-version-bundles` — remove the duplicate `ViewUpdate.version_bundles` wire field
3. `fix/revert-local-rejected-writes` — propagate rejected optimistic writes to every attached in-memory DB

Review and land this stack from the bottom up.

## Problem

The embedded inspector opened its own persistent-worker connection, but the host bridge did not carry the host's verified runtime session into that connection. The default runtime therefore had neither a verified user session nor an admin credential and correctly rejected row queries with:

> Default runtime requires a verified session or admin credential

The inspector UI initially hid that query error behind an empty grid, making the failure look like a replication or table-selection problem. Inspector browser coverage also existed outside the canonical TypeScript CI browser lane, which allowed the broken path to escape CI.

## Solution

- Forward the host DB's internal session only on the same-origin, pre-attached inspector worker path.
- Reconstruct trust from the reserved-issuer identity token and require its issuer, user id, and auth mode to match the attached session before opening the peer.
- Preserve the host's existing runtime-source configuration when installing the inspector worker port.
- Strip the live worker port and forwarded session from public `Db.getConfig()` copies.
- Surface row-query failures directly in the data grid.
- Harden the overlay browser fixture and run the inspector browser suite alongside the jazz-tools browser suite in CI.

## Governing invariants

- A session object received across the overlay boundary is never trusted on its own.
- The forwarding path is admitted only with a same-origin attached `MessagePort` and a matching reserved-issuer identity token; native token verification remains fail-closed.
- Claims come from the verified token, not from the cross-realm session object.
- This does not add an admin bypass, a runtime-context selector, or broader inspector permissions.

### Normal path

A logged-in todo app opens the embedded inspector. The inspector joins the host's IndexedDB-backed worker using the same verified principal, subscribes to `todos`, and displays the rows already visible to the app.

### Failure path

If the forwarded session does not match the identity token, connection setup fails explicitly. If the connected principal cannot read a table, the grid displays `Could not load rows: ...` instead of silently rendering an empty result.

## Testing

- Adds/updates unit coverage for attached-session admission and mismatch rejection.
- Repairs the inspector overlay browser fixture and its synchronization canary.
- Adds the inspector browser suite to the canonical parallel TypeScript browser lane.

### DIFF
--- a/dev/gates/run-ts-tests.sh
+++ b/dev/gates/run-ts-tests.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
 # The test-ts runner has 16 CPUs. Turbo gets at most two test tasks, while the
-# jazz-tools browser suite caps Vitest at four file workers so Chromium retains
-# scheduling headroom; both reuse the artifact build completed before this
-# script starts.
+# browser lane runs the Jazz Tools and inspector suites in parallel. Jazz Tools
+# caps Vitest at four file workers so Chromium retains scheduling headroom; all
+# suites reuse the artifact build completed before this script starts.
 set -u
 
 # The runner accepts small command overrides solely so its process-management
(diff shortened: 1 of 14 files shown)


dev/gates/run-ts-tests.sh
```diff
@@ -21,7 +21,7 @@ if [[ "${JAZZ_REQUIRE_CI_TEST_COMMANDS:-0}" == "1" ]]; then
 fi
 
 node_tests_command=${JAZZ_NODE_TEST_COMMAND:-"pnpm test --filter=!moon-lander-react --filter=!@jazz/rust --filter=!auth-simple-chat --filter=!auth-workos-chat --filter=!auth-betterauth-chat --filter=!chat-react --filter=!world-tour --filter=!jazz-rn --concurrency=2"}
-browser_tests_command=${JAZZ_BROWSER_TEST_COMMAND:-"pnpm --filter jazz-tools test:browser"}
+browser_tests_command=${JAZZ_BROWSER_TEST_COMMAND:-"pnpm --parallel --filter jazz-tools --filter inspector test:browser"}
```
nrainhart, line 24: @aeplay looks like `jazz-tools` browser tests are already run as part of CI on every PR. Added the inspector browser tests here as well, if you prefer to move them elsewhere feel free

aeplay (approved):

PR #2316, opened 2026-08-28
refactor(sync): remove duplicate ViewUpdate version bundles

nrainhart:
## Stack

1. #2315 — inspector reads and CI coverage
2. **This PR** — remove the duplicate `ViewUpdate.version_bundles` wire field
3. `fix/revert-local-rejected-writes` — propagate rejected optimistic writes to every attached in-memory DB

Depends on #2315. Review and land this stack from the bottom up.

## Problem

Wire v14's `ViewUpdate` contained two representations of the same version payloads:

- `version_carriers`, the current compact representation; and
- `version_bundles`, a transitional duplicate retained for older receive paths.

Current senders populated only `version_carriers`, while receivers expanded carriers and merged both fields into one semantic bundle stream. Keeping both fields made every producer, consumer, fixture, benchmark, and test account for an impossible mixed representation and obscured which field was canonical.

Because postcard encodes struct fields positionally, simply deleting the field without a protocol cut could make a v14 frame decode under the wrong layout.

## Solution

- Remove `ViewUpdate.version_bundles` and the receive-time normalization shim.
- Make every view consumer read `version_carriers` directly and use borrowed bundle references where expansion is unnecessary.
- Keep `VersionBundle` as the semantic singleton produced by carrier expansion and as the payload used by `RowVersionPayloads` repair responses.
- Bump the exact wire protocol from v14 to v15 and regenerate the golden wire fixtures.
- Reject v14 during handshake, before payload decoding.
- Update the sync specification, invariants, tests, and benchmarks to describe the single canonical representation.

## Governing invariants

- A `ViewUpdate` carries version payloads through exactly one field: `version_carriers`.
- `VersionCarrier::Bundle` and packed runs remain semantically equivalent after expansion.
- Repair traffic continues using `RowVersionPayloads.version_bundles`; this PR removes only the duplicate field on `ViewUpdate`.
- Wire v14 and v15 never attempt cross-layout decoding.

## Compatibility and non-goals

This is a **sync-protocol-breaking** change: v15 peers reject v14 peers during negotiation. It is **not a storage-format migration**; persisted database rows do not contain serialized `ViewUpdate`/`SyncMessage` frames and need no rewrite.

This PR does not remove the `VersionBundle` type, change transaction semantics, or alter row storage.

### Normal path

A v15 sender ships a packed carrier run. The receiver borrows or expands its bundle bodies and applies the same transaction/version/fate information without consulting a second field.

### Version mismatch

A v14 peer advertises protocol 14 to a v15 peer. Negotiation returns `UnsupportedProtocolVersion` before either side tries to decode the positionally incompatible payload.

## Testing

- Regenerates and validates the v15 golden wire fixtures.
- Updates singleton/packed-carrier equivalence and wire negotiation coverage.


aeplay (approved):

PR #2317, opened 2026-08-28
fix(sync): roll back rejected writes in every attached DB

nrainhart:
## Stack

1. #2315 — inspector reads and CI coverage
2. #2316 — remove the duplicate `ViewUpdate.version_bundles` wire field
3. **This PR** — propagate rejected optimistic writes to every attached in-memory DB

Depends on #2316. Review and land this stack from the bottom up.

## Problem

When two in-memory `Db` clients were attached to the same persistent browser worker, an optimistic write from one client was delivered to both clients, but the shared runtime remembered a terminal-fate route only for the client that originated the `CommitUnit`.

If server authorization later rejected the write:

- the persistent worker rolled it back;
- the writer DB received `FateUpdate::Rejected` and rolled it back; but
- a sibling DB that had observed the optimistic transaction through a `ViewUpdate` never received the rejection and retained the row in local queries until restart.

A result-membership removal was insufficient: the sibling's in-memory node still retained the pending transaction/version state and needed the terminal fate to remove it correctly.

## Solution

Whenever a subscriber successfully receives a `ViewUpdate`, inspect its `version_carriers` and register that subscriber as an observer of each `Fate::Pending` transaction it was shown. The existing local fate router then fans a later rejection out to the originating peer and every observing peer.

Observer routes start with local durability already acknowledged because the `ViewUpdate` represented that local state. Registration happens only after the transport accepts the update, so fate routing cannot reveal a transaction to a peer that never observed it.

## Governing invariants

- Every live peer that observes a pending transaction receives its terminal rejection.
- Only peers that were actually sent the transaction are registered as observers.
- The originating writer retains the existing local-acknowledgement behavior; observer registration does not emit a duplicate local acknowledgement.
- Only `Fate::Pending` transactions need observer routes: an accepted transaction cannot later transition to rejected under the fate state machine.
- Terminal rejection still removes the optimistic edit; this PR does not restore rejected inserts, updates, or deletes to the inspector/editor.

### Rejected insert

App DB A and inspector DB B share one persistent worker. B inserts a row while offline from the server, and A receives it optimistically. After reconnect, the server rejects the insert. The worker routes the rejection to both A and B, and both local queries remove the row immediately.

### Unobserved peer

DB C attaches only after the transaction has already been rejected and was never sent its pending payload. C is not registered as an observer and receives no stale live rejection; it simply hydrates the authoritative post-rejection state.

## Testing

The retained browser regression `rejected write from one live peer reverts every attached peer` now passes unchanged.

Focused receipts from development:

- Target cross-peer rejection regression: passed.
- Related rejection, notification, and restart group: 11/11 passed.
- Rust focused configuration check: passed.
- Correctness NAPI/WASM artifact rebuild: passed.
- Formatting and diff checks: passed.

A full `worker-bridge.test.ts` run also exposed separate pre-existing A→B/B→A ordinary server-propagation timeouts; those are not changed or masked by this PR.

### DIFF
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -1628,6 +1628,23 @@ fn register_local_fate_route(
     tx_id: TxId,
     queue: &PendingDownstreamFates,
 ) {
+    register_local_fate_route_with_acknowledgement(routes, tx_id, queue, false);
+}
+
+fn register_local_fate_observer(
+    routes: &LocalFateRoutes,
+    tx_id: TxId,
+    queue: &PendingDownstreamFates,
+) {
+    register_local_fate_route_with_acknowledgement(routes, tx_id, queue, true);
+}
+
+fn register_local_fate_route_with_acknowledgement(
+    routes: &LocalFateRoutes,
+    tx_id: TxId,
+    queue: &PendingDownstreamFates,
+    local_acknowledged: bool,
+) {
     let mut routes = routes.borrow_mut();
     routes.retain(|_, pending| {
         pending.retain(|candidate| candidate.queue.upgrade().is_some());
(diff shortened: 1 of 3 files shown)


aeplay (approved):

PR #2318, opened 2026-08-28
Canonicalize nested settled result values

aeplay:
🤖 Fixes the remaining nested-postcard persistence gap tracked by #2158.

## Before

The outer `JPFK` program-fact and `JRME` result-member records used settled Groove encoding, but some durable fields inside those records were still opaque postcard bytes:

- a `ResultPayload` fact's `RecordDescriptor`;
- a synthetic result member's row value;
- a synthetic result member's replacement token.

That made persistent identities depend on Rust/serde layout, accepted noncanonical nested bytes, and left those fields outside the codec registry's exact-byte guarantees.

## After

Those nested values use explicit canonical Groove record codecs with permanent numeric identities. Production writers, rehydration readers, and descriptor-derived terminal hashing all use the same representation. Decode rejects malformed, trailing, and noncanonical nested bytes before they can become maintained state.

## Worked examples

- An aggregate result persisted as a synthetic member reopens from RocksDB with the same canonical row and replacement-token bytes.
- A program fact containing a result descriptor round-trips through one fixed Groove record representation; appending bytes or changing its nested tag is rejected.
- Corrupting only a nested synthetic value fails rehydration rather than being accepted because the outer `JRME` envelope itself was valid.

## Non-goals

- This does not change query semantics, result membership, or public APIs.
- This does not add compatibility decoding for the former postcard representation; the current core has no storage backcompat requirement.

## Verification

- exact Groove descriptor byte fixtures;
- synthetic/JPFK corruption receipts;
- RocksDB reopen receipt;
- maintained-vs-one-shot aggregate differential;
- persistent codec-family registry receipt.


aeplay:

PR #2320, opened 2026-08-28
Bound deep query graph preparation stack use

aeplay:
🤖 Fixes the default-thread stack overflow discovered while exercising the fully integrated structured-subscription path.

## Before

Deeply lowered policy/query graphs used recursively owned `Box<GraphBuilder>` children. Cloning, inspecting, rewriting, hashing, or comparing a sufficiently deep graph recursively consumed the owner thread's stack. A realistic propagated structured subscription could therefore abort the process on Rust's normal 2 MiB test-thread stack. Merely splitting the first large async frame moved the failure to the next recursive graph operation.

## After

- Recursive graph edges are immutable `Arc<GraphBuilder>` children, so ordinary graph clones are shallow.
- Runtime rewrite and inspection paths use bounded postorder traversals.
- Shape rewriting memoizes original child pointers, retaining shared DAG structure instead of expanding shared subgraphs.
- Auto-direct family identity uses a bounded structural fingerprint plus exact iterative equality, so hash collisions remain harmless without recursively hashing/comparing the graph.
- The query source dispatcher is split into smaller behavior-preserving helpers, keeping its synchronous frame bounded.

## Worked examples

- A 4,096-node unary graph can be cloned, inspected, shape-rewritten, fingerprinted, and compared on an explicit 2 MiB thread.
- Two parents that share one child before a shape rewrite still share one rewritten child afterward.
- A propagated structured subscription whose policy graph previously aborted now installs and rehydrates normally.

## Non-goals

This does not change query semantics, lowering choices, or subscription results. It only changes graph ownership and traversal mechanics so depth controlled by application schemas and policies cannot exhaust the runtime stack.

## Verification

- Full Groove library suite: 605 passed, 2 ignored.
- Full Jazz library suite: 1,563 passed, 2 ignored in the implementation receipt.
- Independent adversarial review reproduced the old 2 MiB overflow, verified shared-DAG retention, and planted a wrong-shape mutation that the new bounded-stack receipt caught.
- Final stack CI-equivalent local gate passed.

### DIFF
--- a/crates/groove/src/ivm/graph.rs
+++ b/crates/groove/src/ivm/graph.rs
@@ -7,7 +7,10 @@
 //! [`super::op_types`]; lowering from queries lives in [`super::planner`]; the
 //! tick loop that evaluates the graph lives in [`super::runtime`].
 
-use std::hash::{BuildHasher, Hash, Hasher};
+use std::{
+    hash::{BuildHasher, Hash, Hasher},
+    sync::Arc,
+};
 
 use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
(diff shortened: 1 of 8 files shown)


aeplay:

PR #2321, opened 2026-08-28
Share embedded server shutdown completion

aeplay:
🤖 Replaces merged-in-stack #2268 at the earliest layer whose native-artifact corpus exercises embedded-server shutdown.

## Before

The first `JazzServer.stop()` caller took ownership of the native server. Concurrent or later callers could resolve without observing whether native finalization succeeded, and test cleanup could race a still-open RocksDB directory.

## After

Shutdown is one shared `Running → Stopping → Stopped` transition. Every concurrent and repeated caller observes the same immutable success or failure, and success requires native storage to be closed.

```ts
const first = server.stop();
const second = server.stop();
await Promise.all([first, second]);
await server.stop(); // replays the same terminal result
```

If finalization fails, times out, or panics, every waiter rejects with that terminal result.

## Non-goals

This does not change client database shutdown, durability semantics, or add shutdown retries.

## Verification

- Concurrent/repeated NAPI stop receipts pass.
- Embedded finalization-failure receipt passes.
- Jazz Tools dev-server tests pass.
- The final CI-equivalent stack gate passes.

### DIFF
--- a/.changeset/clear-moths-stop.md
+++ b/.changeset/clear-moths-stop.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reject embedded server stop calls when shutdown finalization fails.
--- a/.changeset/clear-moths-stop.md
+++ b/.changeset/clear-moths-stop.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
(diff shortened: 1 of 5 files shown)


aeplay:

PR #2322, opened 2026-08-28
Refresh ownerless relay authority membership on reopen

aeplay:
🤖 Replaces #2289 at the earliest stack layer whose browser topology gate exercises the fix.

## Before

A durable browser worker could reopen with a settled relay-authority result set but without the process-local wire registration that originally owned it. A fresh foreground Edge query could then be covered by stale recovered membership before Core sent its current reset.

Example: an editor reads row B, every tab closes, Core revokes B while the worker is offline, and a fresh tab reopens the persistent worker. The new query could settle from the recovered result containing B instead of waiting for Core's authoritative empty reset.

## After

When a fresh local subscriber establishes a `RelayAuthoritySession` view, the worker discards that exact settled view only when it has no live registration owner. The replacement upstream registration must receive a current reset before the Edge usage site settles. Live sibling registrations remain intact.

## Boundary

This is limited to authority handoff for read-policy-scoped/windowed queries. Ordinary relay reads retain their existing semantics.

## Verification

- Deterministic reopen regression passes.
- Full browser relay durability suite passes.
- Removing the invalidation reproduces the stale-row failure.
- Independent adversarial review of #2289 found no blocker.

### DIFF
--- a/crates/jazz/src/db/peer_connection.rs
+++ b/crates/jazz/src/db/peer_connection.rs
@@ -3130,14 +3130,6 @@ where
                                 self.node.borrow().permissions_ready(),
                                 ingest_context.trust,
                             );
-                            let local_waiting_for_upstream_settlement = local_subscriber
-                                && opts.propagate_upstream
-                                && opts.tier > DurabilityTier::Local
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2324, opened 2026-08-28
fix: update subscription with optimistic writes when server is down

nrainhart:
🤖 Keep local subscriptions alive after a terminal upstream transport failure so locally durable and optimistic writes remain reactive while the foreground runtime continues.

## Before

When the native runtime received a terminal server transport error, it closed every active subscription. That included `local` subscriptions whose result depends only on the still-live local database.

For example:

1. An app and its persistent worker are connected to an edge.
2. The edge process terminates unexpectedly.
3. The app writes a todo and waits only for local durability.
4. The write is successfully applied and persisted locally, but the existing local subscription has already been closed and never publishes it.

The app therefore appeared frozen offline even though local storage and mutation processing were healthy.

## After

- A terminal upstream transport error fails only subscriptions whose internal tier is `edge` or `global`.
- `local` subscriptions remain registered and continue publishing local optimistic and durable changes.
- Edge/global transaction waits still reject with the exact transport error.
- Transport failure remains distinct from authoritative mutation rejection and does not synthesize a transaction fate or invoke `onMutationError`.

The public `local-first` tier lowers to this local behavior. This PR does not make remote reads silently fall back to local state and does not change the strict remote-tier contract.

### Local subscription

A local subscription opens, the server stops, and an offline insert reaches local durability. The same subscription publishes the inserted row without being recreated.

### Remote subscription

An edge/global subscription observing the same terminal protocol failure closes with that error rather than publishing a result that claims remote settlement.

### Temporary connectivity loss

Normal reconnectable offline/connecting transitions remain governed by the existing tier state machine. This change is specifically the terminal transport-error cleanup path.

## Verification

- Browser SharedWorker/IndexedDB receipt stops the real test server, performs a locally durable insert, and observes it through the pre-existing local subscription.
- Native-runtime receipt opens local, edge, and global subscriptions together and proves only the remote subscriptions fail.
- The same receipt proves edge/global waits reject and no mutation-rejection callback is invented.
- Rebased stack retains wire-malformation, transaction-ID, and native-artifact receipts.


aeplay (approved):

PR #2330, opened 2026-08-28
Reset the sync wire protocol to v1

aeplay:
🤖 Reset the sole supported Jazz sync-wire protocol to V1 before settlement.

## Before

After #2316 removed the transitional `ViewUpdate.version_bundles` field, the exact wire protocol was labeled v15. Versions 1–14 were development history, not supported compatibility formats, but the label made the first settled protocol look like the fifteenth public epoch.

Negotiation was also expressed as generic range intersection. Even after changing the local constant to 1, a peer advertising `1..15` could be admitted as V1, leaving an unnecessary compatibility-shaped interpretation.

## After

- Rust and TypeScript encode and accept only wire protocol V1.
- A Hello is admissible only when both advertised endpoints are exactly `1`.
- `0..1`, `1..2`, `1..15`, v14, v15, and every other singleton/range reject before incompatible payload decoding.
- Exact Hello and message fixtures are regenerated with V1 bytes.
- The sync spec and invariants explicitly distinguish wire V1 from independently versioned storage, catalogue, and binding V1 formats.
- There is no compatibility alias, range negotiation, migration, or old decoder.

## Worked examples

Accepted:

```text
min_version = 1
max_version = 1
```

Rejected despite containing V1:

```text
min_version = 1
max_version = 15
```

A v14 or v15 peer is rejected during Hello negotiation; its positionally incompatible `ViewUpdate` payload is never decoded.

## Storage boundary

This changes only live sync-wire identity and exact wire fixtures. Persisted rows, storage manifests, catalogue records, and binding payload formats are unchanged.

## Verification

- exact Rust Hello/message fixture decode and re-encode;
- exact V1 negotiation and explicit v14/v15 rejection;
- Rust and TypeScript rejection matrices for ranges containing V1;
- TypeScript wire/binding gate: 182 passed, 1 skipped;
- independent adversarial review with a planted exactness bypass caught by the negative receipt;
- formatting, Clippy, and commit hooks.


aeplay:

PR #2331, opened 2026-08-29
Restore permanent catalogue identity retirement

aeplay:
🤖 Restore two fail-closed guarantees exposed by the final integrated Rust workspace gate.

## Before

A later catalogue refactor stopped invoking permanent physical-identity retirement validation at two publication boundaries:

- immediate trusted catalogue admission; and
- deferred activation of a previously pending lineage.

A forged later schema could therefore reuse a table, column, or nested enum-occurrence UUID retired earlier in the same durable lineage. Two existing adversarial receipts caught this on the final stack.

The persistent-codec registry also still named the pre-settlement IndexedDB page-v2 fixture after #2295 correctly froze the sole page codec as V1.

## After

- Live admission and deferred pending activation both call `validate_evolution_to_with_history` against durable lineage history.
- Structural publication validation applies the same manifest constraints before mutation.
- Reopen continues validating active, staged, and source-resolved pending mappings.
- The codec registry names `jazz.idb-page.v1` and points to the exact V1 fixture, decoder rejection receipt, and browser raw-IDB evidence.

## Worked examples

A table UUID used in schema A, retired in schema B, and assigned to an unrelated table in schema C now rejects both when C arrives live and when C was parked pending a missing parent. The catalogue remains unchanged.

A scalar enum UUID likewise cannot later be reused as an unrelated nested enum occurrence.

## Verification

- exact five failing CI receipts now pass;
- full Jazz crate suite: 1,705 passed, 4 skipped;
- idb-tree: 19/19;
- multihop, scalar/nested-enum, parked/reopen, malformed-deferred, and registry planted tests;
- independent adversarial review with a planted retirement-validation bypass.

### DIFF
--- a/crates/jazz/fixtures/persistent_codec_family_registry.json
+++ b/crates/jazz/fixtures/persistent_codec_family_registry.json
@@ -481,24 +481,24 @@
       ]
     },
     {
-      "id": "jazz.idb-page.v2",
+      "id": "jazz.idb-page.v1",
       "boundary": "durable-storage",
       "spec": {
         "path": "crates/groove/SPEC/2_storage_model.md",
         "anchor": "### IndexedDB page-store physical format"
       },
       "semantic_fixture": {
-        "path": "crates/idb-tree/fixtures/page-v2-leaf.hex",
+        "path": "crates/idb-tree/fixtures/page-v1-leaf.hex",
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2332, opened 2026-08-29
Run browser topologies in their complete CI harnesses

aeplay:
🤖 Make the canonical TypeScript CI partition execute every browser topology in its owning browser harness and make Inspector attachment deterministic and credential-complete.

## Before

- Turbo’s broad Node partition invoked RecordPlayer’s `test`, which chained into a browser topology suite without the Jazz server command fixture.
- BandChat had the same latent partitioning mistake.
- Inspector browser tests conditionally reused ignored `dist-embedded` output when it existed, so a reused checkout could exercise stale code.
- The inspector host forwarded its attached worker session only from private WeakMap state. When that state was unavailable even though the resolved reserved-issuer JWT was present, the inspector opened without a verified session and failed row reads.

## After

- BandChat and RecordPlayer `test` commands are Node-only.
- Their topology suites run explicitly alongside Jazz Tools and Inspector in the canonical browser CI command, with all required server lifecycle/JWT commands.
- `inspector test:browser` always rebuilds the embedded bundle before Playwright.
- If private host-session state is absent, the bridge deterministically derives the claimed session from the resolved reserved-issuer JWT.
- The receiver still verifies the token and requires exact issuer, user, and auth-mode equality before trusting the forwarded session; claims come from the token, not the cross-realm object.

## Worked examples

A reused worktree with an old ignored Inspector bundle now rebuilds before testing. An attached local-first host with a resolved JWT but no private WeakMap entry forwards matching session claims and the inspector reads the seeded row. A malformed, external-issuer, or mismatched token/session remains untrusted.

## Verification

- exact TypeScript CI-equivalent partition: Node exit 0, browser exit 0;
- Inspector fresh browser 12/12 and repeated overlay 12/12;
- Jazz Tools browser: 270 passed, 1 expected failure, 3 skipped;
- BandChat browser 5/5 and RecordPlayer browser 9/9;
- build:test-artifacts and 122 CI contracts;
- planted missing server command, missing embedded build, missing session derivation, and receiver equality bypasses;
- independent security/coverage review.

### DIFF
--- a/dev/gates/run-ts-tests.sh
+++ b/dev/gates/run-ts-tests.sh
@@ -20,8 +20,11 @@ if [[ "${JAZZ_REQUIRE_CI_TEST_COMMANDS:-0}" == "1" ]]; then
   done
 fi
 
+# Every workspace package's `test` target belongs to this Node/Turbo partition.
+# Browser-only receipts keep their topology out of that target and run through
+# `test:browser` below, where their Vitest projects own the Jazz server commands.
 node_tests_command=${JAZZ_NODE_TEST_COMMAND:-"pnpm test --filter=!moon-lander-react --filter=!@jazz/rust --filter=!auth-simple-chat --filter=!auth-workos-chat --filter=!auth-betterauth-chat --filter=!chat-react --filter=!world-tour --filter=!jazz-rn --concurrency=2"}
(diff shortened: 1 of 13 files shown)


aeplay:

PR #2334, opened 2026-08-29
Build a full native Jazz storage compatibility corpus

### LINKED ISSUE #2307: Build a full native Jazz storage compatibility corpus
🤖 Build one committed epoch-one Jazz database corpus that exercises the complete native persistence profile together, rather than proving each codec only in isolation.

## Why

#2281 freezes real SQLite/RocksDB files containing generic Groove records and indices. Node and codec suites separately cover Jazz branch/history/version/provenance/catalogue/result/program-fact/large-value state. We still lack one historical physical Jazz root proving that all of those families compose across an actual version boundary.

For example, the fixture should contain a branched row with history and contribution provenance, catalogue lineage/lens state, settled result/program facts, and a referenced large value. Current code must read those exact old bytes, append new writes, rebuild only declared derived state, and reopen without changing the historical bytes.

## Acceptance

- Generate the fixture from the pinned storage-settlement producer commit with recorded command, schema/profile IDs, checksums, and logical snapshot.
- Commit backend-neutral authoritative key/value packs plus small SQLite and RocksDB stores.
- Current code opens each store read-only without mutation and compares the full logical snapshot.
- Current code performs mixed writes, reopens, and verifies both historical and new state.
- Unknown/corrupt manifest and one corruption per authoritative family fail before publication or mutation.
- The canonical storage-compat CI partition runs the corpus.
- Specs and the #2160 family registry link the fixture to every covered codec family.

aeplay:
🤖 Completes #2307 with a full epoch-one native Jazz compatibility corpus on both production native backends.

## Before

The compatibility suite froze individual codecs and generic SQLite/RocksDB stores, but it did not prove that a complete Jazz root containing its authoritative families could cross a producer/current-reader boundary, accept a current write, and reopen without losing historical meaning.

## After

- builds a deterministic Jazz root with catalogue lineage, physical application history/current projections, transactions/fates/merge heads, settled query state, deletion history, and an authenticated indirect large value;
- pins a backend-neutral canonical logical pack plus physical SQLite and RocksDB fixtures with reviewed checksums and producer provenance;
- includes the canonical large-value metadata records while leaving the backend-owned raw chunk plane in the physical fixtures;
- stores settled program facts and result members with their new fixed digest keys and canonical values;
- materializes each candidate into a fresh root, reopens it through the current reader, writes a current-format row, and reopens again;
- rejects corrupt archives before target creation and rejects normalized, symlink, hardlink, and existing-output publication aliases;
- proves the logical digest changes when either ordinary application rows or large-value content changes;
- runs from the dedicated storage-compat partition, which also compiles the non-test Jazz surface before enabling corpus-only helpers.

For example, the historical root contains a non-default branched row, globally settled deletion history, populated settled program/result state, and a large byte attachment. The current reader must reproduce those identities exactly, materialize the attachment, add a new row, and preserve both generations after another reopen.

## Format boundary

The logical pack is deterministic and reviewable. SQLite pages and RocksDB files remain backend-owned physical receipts; they are independently reopened but are not claimed as interchange encodings or byte-deterministic outputs.

There is intentionally no reader or migration for pre-settlement alpha stores.

## Verification

- all eight required storage-compat receipts
- fresh staged producer run repeated twice
- fresh-root SQLite and RocksDB reopen/current-write receipts
- corruption and publication-alias receipts
- non-test Jazz feature build

Closes #2307 and contributes to #2160.


aeplay:
🤖 Approved by Anselm in the orchestrator review. Confirmed this is compatibility-corpus and test infrastructure; production runtime semantics are unchanged.

Issue #2308, opened 2026-08-28
Build a full browser Jazz storage compatibility corpus


aeplay:
🤖 Complete the remaining full-browser historical Jazz storage receipts. The harness and pinned fixture are already implemented by merged #2335 and #2546; this is no longer a request to build them from scratch.

## Current evidence

The committed raw IndexedDB corpus is installed at the authenticated principal root production opens, then read through public createDb/SharedWorker/WasmDb. Historical main/draft large values and reopen are exercised, raw fixture bytes are compared, and manifest corruption is rejected. A planted HLC change proves the historical-value comparison is sensitive.

Exact a0708063b1 TS job reports two browser corpus tests passing:
https://github.com/garden-co/jazz/actions/runs/33912712674/job/101152833665

## Remaining against the original acceptance

- #2561 implements current writes appended to the historical root, close/reopen with server network blocked, and verification of historical plus newly written state. Coordinator Chromium2/2 and planted old-root replacement sensitivity pass; CI green, awaiting user review/merge.
- Make the intended history, catalogue, schema/profile and producer-provenance receipts explicit instead of inferring them from successful current reads alone.
- Plant page and authoritative-payload corruption and prove rejection before a usable handle or target mutation. Current corrupt-open coverage is for the manifest; the HLC comparison mutation is not a corrupt-open receipt.
- Register the browser historical corpus in the canonical storage-compat partition and link its family coverage from #2160. The current partition executes native receipts; the browser corpus runs in broader TS CI.
- Revalidate against the final descriptor/V1 settlement from #2548 or #2558 and #2461.

Preserve the pinned historical producer/checksums. Do not regenerate fixtures simply to hide an incompatible encoding. #2309 separately pins the actual published settlement release after publication.

PR #2335, opened 2026-08-29
Build a full browser Jazz storage compatibility corpus

aeplay:
🤖 Draft implementation of #2308: a complete browser Jazz storage compatibility corpus.

## Before

The browser compatibility receipt froze raw IndexedDB page framing, but did not open and extend a complete historical Jazz root through the public browser runtime.

## Current slice

The canonical real-Chromium project now exercises the production path from createDb through SharedWorker and WasmDb into IndexedDB.

It persists deployed catalogue permissions, application rows, main and draft branch state, an updated history/current row, and a large text value, then shuts down and reopens the database through the public API.

## Still in progress

- pin the producer fixture and checksums;
- prove read-only inspection preserves raw IndexedDB records;
- add manifest, page, and authoritative-payload corruption receipts;
- register the complete historical corpus in canonical storage-compat CI and specs.

### Example

A browser store with a historical row, its draft-branch edit, and a referenced large text value is reopened by the current runtime. Both branch views and the large value must remain readable; a corrupt authoritative record must fail before a usable handle or mutation is published.

Tracks #2308 and contributes to #2160.


aeplay:
🤖 Approved by Anselm.

PR #2338, opened 2026-08-29
Exercise frozen wire frames through generated artifacts

aeplay:
🤖 Execute the complete frozen V1 wire-frame corpus through every generated runtime artifact.

## Before

Rust froze exact Hello/message/error frame fixtures, and generated NAPI/WASM bindings had lower-level codec receipts. We did not prove that every frozen complete frame reached the same production semantic decoder through Rust, NAPI, WASM, and TypeScript. A generated artifact could therefore drift or accept a malformed inner payload while source-level fixtures remained green.

## After

- all 9 frozen Hello fixtures and all 22 frozen message fixtures execute through Rust, generated NAPI, generated WASM, and TypeScript;
- exact executed-name equality fails if a fixture is omitted or a new family is not added;
- the codec registry retains the complete source corpus as its semantic anchor and records generated-artifact execution as additional evidence;
- malformed, trailing, unnegotiated-auxiliary, and corrupt-compressed frames are rejected through the production semantic decoder;
- generated artifact fingerprints are rebuilt and pinned from the final source closure.

### Examples

A valid V1 operation frame produces the same semantic result through every runtime path. A valid outer envelope containing a malformed or trailing inner message is rejected everywhere; merely accepting the postcard prefix is insufficient.

## Sensitivity

- removing one executed fixture or adding an unexecuted family fails exact corpus equality;
- bypassing `decode_sync_message_for_features` makes the malformed-inner receipt fail;
- stale generated NAPI/WASM artifacts fail their fingerprint gate.

This completes the remaining full-frame/generated-artifact scope of #2044.


aeplay:
🤖 Approved by Anselm during coordinated review.

PR #2340, opened 2026-08-29
Prevent browser foreground transaction identity aliasing

aeplay:
🤖 Prevent transaction-identity aliasing between fresh browser foreground runtimes.

## Before

Every persistent browser foreground runtime deterministically reused `main-node`, while each disposable in-memory runtime began with an empty HLC register. Two fresh runtimes attached to the same persistent worker could therefore mint the same first transaction ID when their clocks matched. Public behavior could lose liveness rather than reliably surface the internal conflict.

## After

Each foreground runtime receives a fresh random node identity. The persistent worker remains deterministic and owns durable recovery; disposable foreground identities do not need to survive tab lifecycle.

### Example

Two tabs attach to one worker. Their first writes observe the same fixed millisecond. Previously both could produce the same public `txId`; now their IDs differ and both project/document writes reach global settlement.

## Receipt

A real Chromium + SharedWorker + server test creates two fresh foreground runtimes, pins `Date.now()` only around their first writes, asserts distinct transaction IDs, and awaits both global settlements with no sleep or retry. Restoring deterministic `main-node` produces equal IDs and a liveness failure.

This fixes the proven cross-foreground aliasing. The distinct full-server single-runtime sequence still under investigation in #2335 is intentionally not claimed fixed here.

### DIFF
--- a/packages/jazz-tools/src/runtime/default-runtime-source.ts
+++ b/packages/jazz-tools/src/runtime/default-runtime-source.ts
@@ -208,14 +208,12 @@ export class DefaultRuntimeSource extends RuntimeSource<DbConfig> {
     const session = sessionFromConfig(config);
     const selfSignedClientProof = selfSignedClientProofFromConfig(config, session);
     const backendMode = isBackendRuntime(config);
-    const identitySeed = persistentIdentitySeed(config, session);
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Approved by Anselm during coordinated review.

PR #2341, opened 2026-08-29
Gate browser local subscriptions on storage readiness

### LINKED ISSUE #2339: Propagate browser storage-open failures to the initiating operation
🤖 The public browser facade does not reliably propagate an IndexedDB storage-open failure to the operation that triggers lazy schema/storage initialization.

## Reproduction

1. Create and close a browser Jazz database.
2. Release the worker context.
3. Corrupt the persisted manifest to an unsupported `storageEpoch`.
4. `createDb` resolves because browser storage is schema-lazy.
5. The first schema-backed operation, such as `db.all(...)`, causes the SharedWorker to emit `runtime-error: Missing or invalid IndexedDB storage epoch manifest`.
6. The caller may observe an asynchronous worker error rather than a clean rejection of the returned operation, allowing corruption to present as a timeout.

This was found while building the full browser compatibility corpus in #2335.

## Required behavior

The operation that first requires storage readiness must reject deterministically with the structured storage-open error. No usable result, subscription settlement, or mutation may be published. The failure must remain observable after worker handoff/reconnect rather than only through an ambient error event.

Add a real-Chromium public-path receipt and a planted case proving that swallowing the readiness failure makes the test fail. Keep `createDb` schema-lazy unless a separate API decision changes that contract.

aeplay:
🤖 Fixes #2339 and supplies complementary coverage for #2335.

## Before

createDb intentionally resolves before schema-backed IndexedDB opens. Point reads awaited worker readiness, but a newly attached local subscription could publish an empty seed first. Corrupt browser storage could therefore briefly look like a valid empty result before its structured open failure arrived.

The first readiness fix also exposed a deeper scheduler bug: bounding resident IVM work could leave direct/no-owner subscriptions incomplete, while an older storage-pending evaluation at the queue head could hide an unrelated later resident continuation. In real browser topologies this surfaced as “authorization hydration suspended outside an owner-loop subscription.”

## After

- Browser followers defer local seed publication and native subscription setup until worker readiness succeeds.
- Resident-only direct progress scans every queued evaluation for explicitly runnable CPU continuation work, so an older cold hydration cannot starve an independent resident subscription or write.
- Cold storage futures are never repolled merely because another direct operation starts.
- Only storage implementations that explicitly attest an executor-local eager retry may be retried without an owner; MemoryStorage does, while opaque/cold backends remain suspended for their durable runtime owner.
- Owner-attached polling remains bounded and preserves wake ownership and fairness.

## Examples

- A closed IndexedDB store with an unsupported epoch produces no subscription callback before the structured manifest error; cancellation before readiness performs no setup.
- A cold docs hydration may remain parked while a later recursive edges write drains and publishes its resident result.
- A no-owner maintained point subscription still materializes its seed before its caller immediately receives it.
- A self-woken cold storage request remains parked until the owner that installed its durable waker resumes it.

## Verification

- Real Chromium + SharedWorker + IndexedDB corrupt-manifest lifecycle, raw-record preservation, cancellation, and no ambient error.
- Full Groove library suite: 615/615.
- Mixed multiple-cold-head/resident-tail, recursive/cycle, direct no-owner, and owner-wake receipts.
- Exact Jazz maintained physical-point materialization and authorization regressions.
- Independent adversarial review with planted head-only/resident-scheduler regressions.

### DIFF
--- a/crates/groove/src/db/commit.rs
+++ b/crates/groove/src/db/commit.rs
@@ -268,6 +268,15 @@ impl Database {
                 return Err(Error::IvmRuntime(error));
             }
         };
+        // The public direct write APIs are themselves the runtime owner for
+        // CPU-only continuations which this tick scheduled.  In particular,
+        // a bounded recursive evaluation may yield after making resident
+        // progress and self-wake for another slice.  Finish those slices
+        // before returning the publication so its subscription output is
+        // observable in the same direct write turn.  A genuinely cold input
+        // never self-wakes here, so it remains pending for an external owner
+        // rather than making a write wait on storage.
+        self.drain_self_scheduled_resident_progress()?;
         let ivm_tick_time = tick_start.elapsed();
         // The IVM's resident observer uses the staged overlay. It takes the
         // lifecycle lock itself only when another resident publication does
--- a/crates/groove/src/db/commit.rs
+++ b/crates/groove/src/db/commit.rs
@@ -273,9 +273,9 @@ impl Database {
         // a bounded recursive evaluation may yield after making resident
         // progress and self-wake for another slice.  Finish those slices
         // before returning the publication so its subscription output is
-        // observable in the same direct write turn.  A genuinely cold input
-        // never self-wakes here, so it remains pending for an external owner
-        // rather than making a write wait on storage.
+        // observable in the same direct write turn. A genuinely cold input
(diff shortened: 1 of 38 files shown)


aeplay:
🤖 Anselm approved the revised storage-readiness and resident-scheduler design. Subsequent fixes that preserve this contract may proceed without another product review.

PR #2347, opened 2026-08-29
Complete browser foreground identity and recovery

### LINKED ISSUE #1933: Worker restart drops persisted mutation rejections

aeplay:
🤖 Complete the browser foreground identity and recovery path above #2351.

This PR now also contains the work originally reviewed as #2345 and #2346. GitHub marked those PRs merged when their branches became ancestors of this corrected stack head; keeping the implementation here produces the shorter dependency chain #2340 → #2351 → #2347.

## Before

- A browser replica could reuse a process-default node identity instead of a durable per-database identity.
- Foreground tabs could not safely reuse transaction-node identities across clean handoff, nor reliably retire them after an uncertain crash.
- The foreground lease and runtime could accidentally connect to different SharedWorker generations.
- Wiping storage could leave the worker holding a lease owner for the erased storage epoch.
- After worker restart, an unresolved foreground transaction could be replayed and rejected, correcting durable state without notifying the currently attached successor runtime.

## After

- each browser replica persists its own physical node identity;
- foreground identities are leased with high-water handoff, safely reused only after clean release, and quarantined after uncertain ownership;
- lease and runtime traffic use the same generation-qualified SharedWorker;
- storage reset closes and forgets the lease owner associated with the erased epoch;
- persistent creation fails operation-scoped if its required foreground lease cannot open storage;
- restored unresolved transaction IDs remain process-local markers: an attached runtime receives exactly one header/fate-only rejection, while an app-less interval retains no notification backlog or foreign rejected payload.

### Example: clean handoff

Tab A obtains a foreground node lease and mints transaction IDs locally. It reports its high-water mark and closes cleanly. Tab B may reuse that node identity starting strictly above A's high-water mark. If A crashes without a safe handoff, the worker retires that identity instead of risking an alias.

### Example: restart and rejection

Tab A writes optimistically and closes. The persistent worker restarts and replays A's unresolved transaction. Tab B is attached when the edge rejects it. B observes exactly one rejection and the optimistic row disappears. If no tab is attached at rejection time, a later tab sees only corrected state and receives no historical error toast.

## Receipts

- durable replica identity and owner/replica metadata ordering;
- foreground lease high-water handoff, quarantine, storage reset, and worker-generation alignment;
- exact real-Chromium restart/acceptance/rejection composition;
- successor fallback notification and active `.wait()` suppression;
- accepted/rejected marker retirement;
- app-less interval has no later-launch callback;
- full browser relay durability suite;
- INV-TX-9 remains strict: no foreign rejected payload is retained.

Fixes #1933.


aeplay:
🤖 Approved by Anselm.

PR #2351, opened 2026-08-29
fix(browser): bind explicit IndexedDB roots to one owner

### LINKED ISSUE #2344: Give explicit browser database names one safe owner
🤖 Explicit browser `driver.dbName` can select one physical IndexedDB namespace while authentication scope still selects different SharedWorker realms.

## Risk

`resolveDefaultPersistentDbName` returns an explicit database name before adding auth scope, but the worker name and persistent identity seed include issuer/subject. Alice and Bob using the same explicit `dbName` therefore create separate workers and runtimes that can concurrently open the same physical IndexedDB namespace without a shared ownership lock.

Default derived names are user-scoped and safe. The explicit-name path silently violates the browser runtime’s one-worker-per-persistent-namespace ownership assumption. Concurrent write/cleanup corruption has not yet been reproduced, but the split ownership is concrete.

## Required work

Define whether explicit names are:

- logical names that are still physically scoped by app and authenticated subject; or
- intentionally shared physical namespaces, in which case every auth scope must converge on one safe owner and permissions must prevent data leakage.

Add a real-Chromium receipt with two tabs, distinct authenticated subjects, the same explicit name, concurrent writes, worker restart, and cleanup. The chosen behavior must either isolate the stores or reject the second incompatible owner before either runtime can mutate the namespace. Never permit two independent page-store owners over the same IndexedDB root.

aeplay:
🤖 Closes #2344.

## Before

driver.dbName selected one physical IndexedDB root. Switching accounts under the same configured name either collided at the durable owner marker or required destructive reset, so offline caches for multiple auth scopes could not coexist.

## After

driver.dbName and the deprecated top-level dbName are logical base names. Jazz derives the physical IndexedDB and SharedWorker namespace from that base plus the exact app, environment, and canonical non-secret auth scope. Different accounts therefore coexist in separate caches; the same account deterministically reopens and shares its existing cache.

A durable exact owner marker remains beside each physical page store as defense in depth. Low-level attempts to open a physical root with the wrong owner still fail before WASM/runtime initialization or storage mutation. Reset/destruction affects only the current derived scope.

## Examples

- Alice and Bob both use the logical base offline-cache: each opens a different physical database and both caches remain available.
- Alice reloads or opens another tab: it derives her same physical database and SharedWorker realm.
- The inspector overlay forwards the logical base and independently derives the host’s exact physical namespace; it cannot append the scope twice.
- App, environment, auth class, issuer/subject, Unicode, anonymous, and token-boundary differences are covered explicitly.

## Invariants and non-goals

- The physical namespace contains no token, backend secret, or arbitrary JWT claim.
- The owner marker is non-secret and is not an authorization credential.
- Physical database ownership remains distinct from foreground replica/node identity.
- No migration or backward compatibility for older unowned/naming formats is provided.

## Verification

- Focused scope, worker configuration, owner-marker, and inspector tests.
- TypeScript checks and repository formatting/lint/sensitive-data gates.
- Independent adversarial review with a planted inspector double-derivation regression.
- The Chromium behavioral receipt is presently blocked before its test body by the known stale WASM ABI-fingerprint artifact issue tracked in the surrounding stack.


aeplay:
🤖 Approved by Anselm.

PR #2352, opened 2026-08-29
Bound settled-state storage keys

aeplay:
🤖 Bounds authoritative settled-state storage keys and preserves exact canonical identities.

## Before

Three direct-store paths encoded authored data directly into ordered keys:

- branched row versions could lose their non-default branch coordinate when reconstructed;
- settled program facts lived in B-tree keys;
- settled result-member payloads, including large path values, lived in B-tree keys and could grow without a fixed bound.

This coupled storage correctness and backend key limits to arbitrary authored payload size.

## After

- commit units retain canonical branched versions, so reopening a maintained witness yields the exact non-default branch version;
- settled program facts are stored as fixed, domain-separated digest keys with canonical values;
- settled result members use fixed v1 BLAKE3 digest keys and canonical member values;
- recovery verifies key/value digest, type, length, and canonical decoding and fails closed on corruption;
- add, idempotence, remove, rewrite, and native RocksDB reopen preserve exact large members;
- a direct-store audit records why remaining known-state and marker keys are bounded UUIDs or constants.

For example, a roughly 900 KiB real result member no longer creates a roughly 900 KiB B-tree key. Its key is fixed-size; the canonical value carries the full member and is verified against the key on recovery.

## Compatibility

There is intentionally no legacy reader, migration, or v2 format. This establishes the permanent v1 representation on the pre-release storage line.

## Verification

- focused large settled-result add/remove/rewrite/reopen/corruption receipt
- 17/17 settled-state library tests
- exact large non-default-branch witness reload receipt
- direct-store key audit

Stacked on #2336. #2334 will regenerate its native compatibility corpus on top of this representation.

### DIFF
--- a/crates/jazz/src/node/currency.rs
+++ b/crates/jazz/src/node/currency.rs
@@ -1198,7 +1198,7 @@ where
         .await
     }
 
-    async fn query_version_by_alias_with_storage_in_schema(
+    pub(super) async fn query_version_by_alias_with_storage_in_schema(
         &mut self,
         schema_version: SchemaVersionId,
         table: &str,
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Approved by Anselm in the orchestrator review.

PR #2354, opened 2026-08-29
Seal correctness artifact consumers to one exact build

aeplay:
🤖 Separates correctness-artifact production from consumption and makes every in-scope consumer execute one admitted WASM/NAPI snapshot.

## Before

TypeScript and browser correctness commands could rebuild or resolve generated artifacts independently. A consumer could therefore combine NAPI or WASM outputs from different source states, follow a mutable workspace pointer after checking it, or silently use a fallback package. The resulting failures looked like product regressions even when the code under test was not one coherent build.

## After

- one producer builds and snapshots the exact-head correctness WASM and NAPI artifacts;
- the manifest binds source identity, inventory, permissions, link count, and content hashes;
- consumers fail closed before execution and revalidate the complete snapshot after the child exits, detecting accidental concurrent workspace/build mutation;
- BandChat, RecordPlayer, Jazz Tools, and Inspector correctness/browser topology entrypoints resolve the admitted artifacts rather than mutable workspace outputs;
- hardlinks, mutable files/directories, stale source identities, pointer swaps, and post-preflight mutations fail admission or lifecycle validation;
- parallel snapshot publication preserves the winning snapshot and safely removes the losing private stage;
- every snapshotted artifact has a proven runtime consumer.

For example, if another local build replaces the admitted WASM package after preflight while a browser test is running, the wrapper detects the changed inode/content/permissions after the child exits and fails the run rather than reporting it as a valid correctness result.

## Deliberate boundary

This protects correctness runs from accidental concurrent build/workspace mutation. It is not a security boundary against a hostile same-UID local process; portable path-based consumers cannot enforce that guarantee.

Dedicated performance benchmarks remain outside the correctness snapshot. Their workflows explicitly build release/performance CLI, NAPI, and WASM artifacts, and contracts prevent them from falling back to fast correctness WASM.

## Verification

- 76/76 artifact-store, pipeline, workflow, consumer, Inspector, and RecordPlayer contracts
- exact-head producer and verifier
- real sealed browser consumer: 12/12
- Jazz Tools and Inspector TypeScript checks


aeplay:
🤖 Approved by Anselm. The broader consolidation of artifact and cache identities is tracked in #2384.

PR #2355, opened 2026-08-29
feat(rn): establish native foreground foundation

aeplay:
🤖 Reconstructs the reviewed React Native relay work as one clean native foreground foundation on current `main`.

## Before

The RN work was split across a stale native stack that still contained a now-obsolete JavaScript postcard relay path and a merged PR that GitHub would not let us remove from that stack. The local continuation also carried months of already-integrated non-RN commits, making a normal rebase misleading.

## After

- preserves the reviewed trusted native admission, package/artifact detection, device harness, Android admission, and fresh packed-install commits;
- removes the superseded JavaScript frame adapter/runtime-source path rather than maintaining two runtimes;
- establishes one native `NativeDb`/JSI foreground owner, factory, command, lease, and wake substrate;
- keeps persistent relay ownership native while foreground Jazz state remains memory-only;
- rejects stale factories and runtime leases and isolates foreground handles by runtime;
- retains Android/device and packed-install admission receipts.

For example, two RN runtimes no longer establish separate JavaScript relay implementations. Each attaches a native foreground handle to the same trusted native relay boundary, while stale factories and retired runtime leases fail closed.

## Support boundary

This is still a WIP implementation foundation, not a public high-level RN support claim. Public docs continue to say high-level RN Jazz is unsupported until rebuilt Android/iOS artifacts and a real app device E2E prove the complete path.

This replacement preserves the authored commits and reviewed behavioral boundaries from #2269, #2266, #2271, #2278, #2287, #2311, and #2310, while superseding #2273/#2274's second runtime path.

## Verification

- RN Jest and typecheck
- native relay/source/codegen contracts
- Android device admission/source receipts
- fresh packed-install receipt retained in history


aeplay:

PR #2356, opened 2026-08-29
feat(rn): restore native foreground facade

aeplay:
🤖 Adds the public-to-native foreground query and transaction facade on the single RN `NativeDb`/JSI substrate.

## Before

The native foreground owner could be created and scheduled, but Jazz Tools could not drive a useful app database through that path.

## After

- exposes canonical byte commands for foreground reads, subscriptions, inserts, updates, and transaction lifecycle;
- wires Jazz Tools `Db` to the native foreground facade without reviving the JavaScript relay path;
- validates mutation envelopes and keeps pending/retryable/cancelled subscription batches correctly owned;
- derives the foreground factory ABI from the relay implementation.

For example, a local-first RN app can open its native foreground database, subscribe to a query, commit a transaction, and receive the resulting update through the same native owner.

Unsupported operation families remain explicit; this is a WIP capability rather than a general RN support declaration.


aeplay:

PR #2357, opened 2026-08-29
test(rn): prove native foreground persistence E2E

aeplay:
🤖 Proves the RN native foreground path across persistence, restart, authentication scopes, node leases, writes, and subscriptions.

## Before

Host/source tests exercised individual commands, but did not compose them through native SQLite ownership and foreground lifecycle boundaries.

## After

- reopens native relay SQLite state across process restart;
- isolates authenticated scopes across restart and logout;
- allocates per-runtime foreground node leases, safely reuses only cleanly handed-off leases, and retires uncertain identities;
- proves foreground write/subscription behavior and a packed foreground install;
- keeps foreground Jazz state memory-only while rehydrating from the persistent native relay.

For example, an authenticated runtime writes a row, the foreground process terminates, and a new runtime for the same scope reopens the native relay and observes the durable row; a different scope cannot observe it.


aeplay:

PR #2358, opened 2026-08-29
fix(rn): seal foreground artifact lifecycle

aeplay:
🤖 Caps the RN native foreground stack with artifact provenance and failure-safe lifecycle teardown.

## Before

Scope revocation and runtime invalidation could remove a public foreground handle without proving that queued native wake callbacks could no longer reach the released raw context, especially near queue saturation.

## After

- seals transitive relay artifact inputs and keeps generated relay details private;
- centralizes close, runtime invalidation, and scope revocation teardown;
- atomically inertizes the raw wake context before clearing its owner scheduler and emitting cancellation;
- reserves lifecycle queue capacity so ordinary commands cannot crowd out close/revoke/invalidation;
- retires forced-teardown leases, preserves unrelated scopes, and makes late finalizers idempotent;
- retains honest public docs: high-level RN remains unsupported pending real platform E2E.

For example, revoking scope A with a saturated owner queue cancels A without allowing a later scheduler callback to touch released memory, while scope B remains live.

## Verification

- native relay: 38/38
- RN Jest: 9/9 plus typecheck
- native host/header: 14/14 and ABI mirror
- RN device-source acceptance: 53/53 plus typecheck
- planted removal of wake-state inertization fails the saturation receipt


aeplay:

PR #2360, opened 2026-08-29
feat(rn): restore a coherent native foreground foundation

aeplay:
🤖 Restore a coherent native React Native foreground foundation as one independently buildable increment.

This supersedes the artificial four-layer draft stack #2355–#2358. That split left intermediate source packages depending on later ABI and lifecycle work. This replacement preserves the complete implementation and receipts in one honest boundary that installs, compiles, persists, restarts, and tears down coherently.

## Architecture

- `jazz-native-relay` owns the Rust byte-command ABI, SQLite-backed durable host, wake lifecycle, and process-restart behavior.
- `jazz-rn` packages the authoritative native header plus Android/iOS host bridges and validates the native ABI before exposing it to JavaScript.
- Jazz Tools' React Native adapter translates the public database operations onto that byte ABI; unsupported operation families fail closed.
- platform fixtures prove the source/package boundary on Android and iOS. Public docs still describe high-level React Native support as experimental rather than claiming a finished production integration.

The native host is deliberately thinner than Groove/Jazz semantics: it transports canonical command/result bytes and owns persistence/lifecycle. Query, transaction, permission, and sync meaning stays in Jazz/Groove.

## Before

- the checked-in packaged header could drift from the authoritative Rust C ABI;
- native foreground ownership and teardown were incomplete across queue saturation, logout, and scope changes;
- fresh-install fixtures could hard-code an obsolete ABI and pass only because tests staged a different header;
- macOS source verification depended on GNU `/usr/bin/timeout`;
- no coherent public increment proved persistent write, process restart, and readback through the packaged RN boundary.

## After

- the packaged header is byte-equal to the authoritative header and source receipts fail on drift;
- token-keyed foreground installations prevent one runtime from retaining or revoking another runtime's native host;
- teardown makes queued wakes inert before their context can be released, including saturation and cancellation;
- scope switch and logout explicitly revoke the admitted native capability;
- fresh packed Expo verification derives the ABI from the packed JS entrypoint and plants an incompatible-version rejection;
- Android boot probing is bounded on Linux and macOS without assuming GNU timeout;
- SQLite persistence is exercised through write, host close/process restart, reopen, and readback;
- current public write receipts expose `txId`; private native binding-field differences are normalized inside the adapter and never leak into the TypeScript API.
- unsupported in-place auth refresh rejects atomically while leaving the previously admitted scope usable; explicit logout or scope switch remains the revocation/recreate boundary.

### Example: persistent restart

An admitted foreground opens its native host, writes canonical bytes, closes, starts a fresh host process over the same SQLite file, and reads the bytes back. The receipt crosses the packaged JS/native boundary rather than reading SQLite directly.

### Example: logout during pressure

A foreground fills the command queue and has a pending native wake. Logout revokes that runtime token; later queue progress cannot call the released context, and another admitted runtime remains independently usable.

## Verification

- CI workflow contracts: 125/125
- RN device/source protocol and type receipts: 55/55
- `jazz-native-relay` library tests: 38/38
- Jazz Tools and `jazz-rn` typechecks
- React Native public adapter typecheck
- repository format gate
- authoritative/package header byte equality
- planted ABI mismatch, stale-header, and cancelled-wake positives

## Remaining boundary

This establishes the native foreground substrate and a real persistence receipt. The device acceptance fixture now seeds a row through `createJazzClient`, force-stops the process, launches a fresh process over the same native SQLite host, and requires that exact run-bound row through a new `createJazzClient`. The host-issued per-run marker prevents retained app data from producing a false pass. A polished product application running on physical Android/iOS devices remains follow-up work; the docs do not claim otherwise.


aeplay:
🤖 Approved by Anselm in the orchestrator review.

aeplay:
🤖 Approved by Anselm, including the public restart receipt and RN device-workflow/workspace follow-ups.

aeplay:
🤖 Approved by Anselm. The exact head has now also completed the full installed-app acceptance contract on real iOS and Android devices: initialization, persistent writes, sibling foreground delivery, clean shutdown, auth-scope reopen, full process restart, and SQLite readback.

aeplay:
🤖 Superseded by #2552, now merged as ab976caee4 after Anselm’s explicit approval. Independent preservation review verified the meaningful behavior from this PR at 09a6185b9b6f, including the governing single-foreground relay scope, remains in the integrated implementation. Both platform device suites, native linkage, isolated packaged consumers and full CI passed at a0708063b1. This PR is closed as superseded; its history and recorded decisions remain available here.

PR #2366, opened 2026-08-29
Make maintained equality sources index-selective

aeplay:
🤖 ## Before

A maintained filtered subscription always hydrated from the complete physical-current table. Even when `WHERE team = $team` had an equality index, first attachment grew with every unrelated row. Intersected equality probes could not use the one-shot fused index source because that source had no live-update semantics.

Example: with 500 owned rows, `WHERE ownerId = session.user` took roughly the same one-shot work at 10k and 100k total rows, while the maintained subscription grew from about 160 ms to 1.2 s.

## After

Maintained equality access paths use the same IVM program for hydration and later deltas:

- one equality becomes a live `variant_index_scan` source;
- multiple equalities become live index sources joined by ordinary IVM semi-joins;
- Local and Edge reads still include the complete ahead overlay before winner selection;
- no separate snapshot RPC or continuation path is introduced.

An initially empty indexed prefix remains live and delivers its first matching insert exactly once. A row entering or leaving either side of an intersected equality query updates membership through the same graph.

## Current verification

- focused maintained-view and public database tests pass;
- new receipts cover empty-prefix first insertion and enter/leave through either equality index;
- the lane is adding Global/async-storage coverage and fixed-cardinality 10k/100k/1M CodSpeed receipts before this leaves draft.

This is the first coherent checkpoint for #2077 and is intentionally draft while hosted performance evidence and the remaining acceptance matrix run.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -142,11 +142,11 @@ jobs:
       - name: Install cargo-codspeed
         run: cargo install cargo-codspeed --version 5.0.1 --locked
 
-      - name: Build maintained selective-hydration receipt
-        run: cargo codspeed build --measurement-mode walltime --package jazz --features testing --bench selective_global_hydration
+      - name: Build maintained selective-hydration benchmark
+        run: cargo codspeed build -m walltime --package jazz --features testing --bench selective_global_hydration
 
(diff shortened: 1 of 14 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2367, opened 2026-08-29
Materialize simple maintained results from storage

### LINKED ISSUE #2129: Perf thesis: same-shape subscription bindings must share source witness state

aeplay:
🤖 ## Before

Every binding of a simple maintained query retained source-wide version and replacement witnesses. With one query shape bound to 100 tenants over 2,000 source rows, the same immutable evidence was multiplied into 200,000 version identities and 200,000 replacement entries, consuming about 683 MB of maintained/control heap.

This happened even when each binding delivered only a small bounded result.

## After

For the proved simple subset—default current read, one root table, no relations, recursion, includes, array subqueries, or aggregate, with predicate-only policy branches—the maintained view retains result membership only.

When a row enters the delivered result, Jazz loads its exact immutable `(table, row, tx)` content from authoritative node storage. A deletion removes the old occurrence without reading a newer content winner; a later restore re-enters using the exact original content transaction. Unsupported query shapes keep the existing self-contained witness path.

The 100-binding route receipt now reports:

- `0` retained version identities;
- `0` replacement entries;
- about `606 KB` maintained/control memory versus the issue baseline of `683.39 MB`;
- matching, unrelated-write, and below-window assertions still exact.

## Current verification

The existing `route_subscription_curve` receipt passes with its semantic canaries. Focused differential and authorization coverage is still running, so this PR remains draft. The `performance` label intentionally starts hosted CodSpeed measurement and flamegraph collection at this coherent checkpoint.

Closes #2129 once the remaining acceptance matrix and independent adversarial review are green.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -150,7 +150,11 @@ jobs:
         uses: CodSpeedHQ/action@4296e51e7041e24dadb86d1d6e8b9320d223dbe8 # v5.0.3
         with:
           mode: walltime
-          run: JAZZ_ROUTE_CURVE_ROUTES=100 cargo codspeed run --package jazz --features testing --bench route_subscription_curve
+          # Features are selected at `cargo codspeed build` time. `run` only
+          # executes that copied target; passing Cargo feature flags here is
+          # rejected by cargo-codspeed and used to leave this job reporting no
+          # walltime benchmarks.
+          run: cargo codspeed run --package jazz --bench route_subscription_curve
 
   maintained-selective-hydration-walltime:
     # This is the direct receipt for #2077: fixed owner/result cardinality
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -118,6 +118,40 @@ jobs:
           mode: walltime
           run: cargo codspeed run --package jazz-example-benchmark-w1 --bench reads_memory_walltime --bench reads_rocksdb_walltime
 
+  route-subscription-walltime:
+    # #2129: this is the production-shaped same-query/many-binding receipt.
+    # Keep the issue's 100 bindings: it is large enough to expose accidental
+    # source-witness multiplication while remaining comfortably inside the
+    # hosted macro-run budget.
+    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'benchmark')
+    name: Route subscription wall time
+    runs-on: codspeed-macro
+    timeout-minutes: 25
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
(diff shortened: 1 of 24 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2369, opened 2026-08-29
Guard stack operations from dirty worktrees

aeplay:
🤖 Prevent stack propagation, rebasing, or submission from silently carrying a dirty index or worktree into another branch.

## Before

Lane 1 was reused during sequential restacking while its index still contained staged content. Later checkouts and merges preserved that index, leaving 22 staged changes that collectively reversed already-published #2341 scheduler work and #2351 browser-storage isolation work. A normal conflict-free restack therefore looked successful while the lane held a latent regression.

## After

- `dev/gates/require-clean-worktree.sh <checkout>` rejects staged changes, unstaged tracked changes, and non-ignored untracked files separately.
- ignored build/cache outputs remain allowed;
- `pnpm test:stack-preflight` exercises every state in a temporary repository;
- `AGENTS.md` requires the preflight immediately before stack linking/submission, rebases, and manual restack propagation.

Example:

```sh
dev/gates/require-clean-worktree.sh /work/lane-2
gh stack submit
```

The helper is intentionally independent of `gh-stack`: it can guard the manual merge-based restacks we also use, without changing stack semantics.

## Verification

- `pnpm test:stack-preflight`
- planted staged, unstaged, and untracked states each fail closed;
- an ignored artifact succeeds.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,16 @@ handoff cases. For nuanced or large changes, make these concrete enough for an
 adversarial reviewer to verify the behavior and for a future reader to recover
 the decision without reconstructing it from the diff or conversation history.
 
+### Stack and restack preflight
+
+Before every `gh stack link`, `gh stack submit`, rebase, merge-based restack,
+or branch propagation, run `dev/gates/require-clean-worktree.sh <checkout>`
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2371, opened 2026-08-30
Retain async mutation preparation and finalization

### LINKED ISSUE #2370: Split synchronous mutation admission from async finalization
🤖 The synchronous Jazz client facade still bridges several potentially suspendable CoreDb mutation paths with `db::block_on`: insert, upsert, update, delete, exclusive transaction operations, and exclusive commit.

This became visible while fixing the maintained-query liveness regressions in #2367. Tick, read, subscribe, provenance, and storage-open paths can be made owner-driven/async without changing mutation admission. The mutation paths are different: Core may publish resident state before persistence or subscription refresh later returns Pending. Polling once and dropping the future can lose finalization; globally deferring local persistence stranded immediate resident subscription updates in focused join receipts.

Design a clean split, likely one of:

- synchronous resident admission that returns the transaction identity plus an owned/retained async finalization operation; or
- a public async mutation API, with the synchronous facade limited to storage/runtime arrangements that prove completion cannot suspend.

Acceptance should include yielding-storage receipts for ordinary and exclusive writes, retained finalization across Pending, immediate resident subscription behavior, persistence completion, rejection delivery, and no synchronous reentrant polling. Do not convert the remaining bridges mechanically.

aeplay:
🤖 Fixes #2370.

## Why

Jazz's synchronous client facade polled Core mutation futures with a no-op waker. That worked only when preparation was immediately ready. A genuinely asynchronous storage backend such as IndexedDB can yield while reading parents, staging a transaction, preparing a partial large-value edit, or finalizing a write. Re-polling synchronously could spin forever; dropping a pending future could lose required work.

The public API should remain synchronous and local-first. The asynchronous work belongs to the runtime owner, not to the caller's stack or the lifetime of a returned handle.

## Before

- WASM and NAPI mutation bindings synchronously polled Core futures.
- Cold preparation could block, spin, or be dropped after returning `Pending`.
- Ordinary writes and transaction stages used different ownership paths.
- A returned transaction identity could not safely be reserved before cold parent discovery.
- A reserved identity could incorrectly look resident before publication.
- Dropping a write handle could also drop the only owner of unfinished preparation.

## After

- There remains one synchronous public mutation API.
- A call reserves its definitive row/transaction identity, enqueues one Node-owned operation, and returns the existing write handle.
- One FIFO owns ordinary, attributed, restore, partial-large-value, mergeable, and exclusive transaction operations.
- Memory-backed foreground calls receive one bounded inline owner turn, preserving immediate local read-your-writes behavior without spinning.
- Yielding persistent/WASM operations remain queued and are resumed by scheduler-owned turns.
- Exact reserved IDs survive cold preparation and same-node reopen because every observed/minted transaction time advances a shared high-water clock.
- A failed transaction stage poisons the whole transaction, prevents partial commit, exposes the original terminal error, and queues cleanup of the abandoned Core transaction.
- Dropping a handle cannot cancel queued work.
- `.wait(...)` only observes the requested milestone; reservation alone does not satisfy resident publication.
- Storage and wire formats do not change.

Example: `db.update(...)` returns its normal handle and definitive identity immediately. Internally, preparation may await an IndexedDB parent read, then publish resident state, persist, upload, and settle. `handle.wait("local")` stays pending until publication; dropping the handle does not stop the operation.

For a transaction, `begin`, stages, commit, rollback, and failure cleanup retain FIFO order. If an early stage fails, a later otherwise-valid stage cannot commit a partial transaction.

## Verification

- queued mutation/persistence receipts: 8/8;
- WASM32 and NAPI host checks;
- filtered WASM/NAPI transaction diagnostics;
- plants for dropped first poll, FIFO ownership, exact reserved-ID commit, poisoned-stage partial commit, reopen high-water recovery, and reservation-versus-publication waits;
- no remaining WASM/NAPI synchronous mutation bridges for insert/update/upsert/delete/restore/begin/commit.

## Review boundaries

The independent review should explicitly examine process-crash recovery of pre-admission in-memory operations, close-time queue semantics, preparation-error delivery, transaction poisoning/cleanup, identity dominance across reopen, and Memory versus yielding-storage parity.

Transaction-local query bindings remain synchronous read bridges and are being evaluated as an orthogonal read API concern; this PR does not introduce a second mutation API.

The branch is temporarily based on the integrated alpha.54 stack, so GitHub may show lower-stack changes until the native stack is propagated.

### DIFF
--- a/crates/jazz-napi/close-pollable.cjs
+++ b/crates/jazz-napi/close-pollable.cjs
@@ -1,0 +1,8 @@
+async function completePollableClose(pending) {
+  for (;;) {
+    if (pending.poll() !== null) return;
+    await new Promise((resolve) => setImmediate(resolve));
+  }
+}
+
+module.exports = { completePollableClose };
--- a/crates/jazz-napi/index.cjs
+++ b/crates/jazz-napi/index.cjs
@@ -2,6 +2,7 @@
 // generated loader. It gives a fresh checkout the same deterministic failure
 // as a rebuilt checkout.
 const { nativeBinding, expectedNativeArtifactFingerprint } = require("./native-binding.cjs");
+const { completePollableClose } = require("./close-pollable.cjs");
 
 if (typeof nativeBinding.nativeArtifactFingerprint !== "function") {
   throw new Error(
@@ -25,4 +26,21 @@ if (
       "Rebuild the monorepo binding or reinstall matching package versions.",
   );
 }
+
+const closePollable = nativeBinding.NapiDb.prototype.__closePollable;
+if (typeof closePollable !== "function") {
+  throw new Error(
+    "Jazz NAPI artifact is incomplete despite a matching fingerprint (missing NapiDb close operation). " +
+      "Rebuild the monorepo binding or reinstall matching package versions.",
+  );
+}
+Object.defineProperty(nativeBinding.NapiDb.prototype, "close", {
+  configurable: true,
+  writable: true,
+  value: async function close() {
+    const pending = closePollable.call(this);
+    if (pending instanceof Uint8Array) return undefined;
+    await completePollableClose(pending);
+  },
+});
 module.exports = nativeBinding;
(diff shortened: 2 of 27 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2374, opened 2026-08-30
Evaluate deep IVM graphs without recursive polling

aeplay:
🤖 Deep Jazz policy/query graphs currently recurse through async IVM polling. On the server shell's bounded stack, a legitimate inherited-policy graph can overflow and abort the process before producing a result.

Before:

- `TickEvaluator::update_node` recursively awaited child nodes;
- recursive seed/step nodes could accidentally participate in the root traversal;
- Jazz's declared-output and binding-parameter walkers were recursive as well;
- the reproduced inherited scalar-path policy aborts under a 2 MiB stack.

After:

- Groove evaluates ordinary dependencies using an explicit postorder worklist;
- one node is evaluated at a time, while recursive seed/step evaluation remains owned by the existing frontier-scoped recursive evaluator;
- shared DAG inputs retain the existing memoization behavior;
- Jazz's corresponding graph walkers are iterative;
- deep compile, retained evaluation, async hydration, recursive-cycle, and policy receipts run on a 2 MiB stack.

Example: an inherited read policy following a deeply nested chain now evaluates to its normal grant/deny result. It no longer consumes one Rust call frame per IVM node. Recursive cycles remain set-convergent and fail closed; this PR does not change policy semantics.

Adversarial review restored the old recursive bridge and reproduced the stack overflow; restoring this implementation made the same receipt pass.

### DIFF
--- a/crates/groove/src/ivm/runtime/evaluator.rs
+++ b/crates/groove/src/ivm/runtime/evaluator.rs
@@ -451,19 +451,24 @@ impl TickEvaluator<'_> {
                 .node(node)
                 .ok_or(IvmRuntimeError::GraphNodeNotFound(node))?;
             pending.push((node, true));
-            pending.extend(
-                graph_node
-                    .descriptor
-                    .inputs
-                    .iter()
-                    .rev()
-                    .map(|input| (*input, false)),
-            );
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2375, opened 2026-08-30
Scope maintained views by session claims

aeplay:
🤖 Maintained query compilation was partly keyed by author identity while session claims lived in an author-keyed ambient map. Two simultaneous sessions with the same issuer/subject but different claims could therefore replace or reuse one another's maintained source and authorization-support scope.

Before:

- session A and session B could share an author while carrying different claims;
- compiling or rehydrating B could make A's maintained stream retract an authorized row;
- RegisterShape preflight, authorization advice/support, and post-rehydrate refresh still consulted ambient author claims;
- cancellation during an awaited scoped operation could leave the temporary claim scope installed;
- one cache identity retained a formatted claim string.

After:

- maintained source/cache identity includes canonical, domain-separated identity-and-claims input;
- transport read paths pass the connection's claims explicitly through preflight, support-scope compilation, rehydrate, and refresh;
- an RAII scope restores prior state even when a pending operation is cancelled;
- legacy author claim state remains only for existing write admission/upstream behavior;
- cache keys retain only a digest, not raw claim values.

Example: an `editor` and a `viewer` session for the same user can subscribe concurrently. The editor keeps receiving rows authorized by its claims; the viewer cannot see them, and registering or reconnecting the viewer cannot retract the editor's stream or reuse its support receipt.

Planted reviews narrowed the source key, restored ambient support-scope lookup, and cancelled a pending scoped operation. Each corresponding isolation/restoration receipt failed as intended.

### DIFF
--- a/crates/jazz-testkit/tests/branch_claims_integration.rs
+++ b/crates/jazz-testkit/tests/branch_claims_integration.rs
@@ -585,6 +585,13 @@ async fn subscription_matches_claims_select_query() {
 }
 
 #[tokio::test(flavor = "current_thread")]
+/// Keeps alice's authorized subscription and receipt-derived view isolated
+/// from mallory's second JWT for the same authenticated author.
+///
+/// ```text
+/// writer ──insert──► server ──authorized view──► alice
+///                              └──denied view────► mallory
+/// ```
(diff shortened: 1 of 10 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2376, opened 2026-08-30
Fingerprint tracked native artifact inputs only

aeplay:
🤖 Native correctness-artifact fingerprints currently walk package directories recursively. Ignored or untracked build output left in a worktree can therefore change the fingerprint and become embedded in otherwise source-identical NAPI/WASM artifacts.

Before:

- a stale nested ignored `index.js` changed the NAPI source fingerprint;
- two clean checkouts of the same commit could build different artifact identities;
- generated expectation receipts could participate in their own identity and cause reseal churn;
- a valid schema-3 producer receipt could still canonize the wrong local input inventory.

After:

- real Git checkouts fingerprint an explicit tracked-file inventory;
- ignored and untracked generated outputs are excluded regardless of nesting depth;
- authoritative artifact tooling remains part of the identity;
- generated expectations and producer/test helpers are explicitly non-authoritative, preventing self-reference;
- non-Git hermetic fixtures retain a documented recursive fallback.

Example: leaving `crates/jazz-napi/crates/jazz-napi/index.js` from an older NAPI-RS build no longer changes the next native fingerprint. Editing the authoritative artifact builder still changes it.

Independent review planted nested ignored output, generated receipt edits, authoritative tooling drift, and a stale schema receipt. The first two preserved identity, tooling drift changed it, and stale schema verification failed closed. A clean source build sealed and verified the schema-3 producer snapshot.


aeplay:
🤖 Approved by Anselm. Follow-up architectural cleanup is tracked in #2384.

PR #2377, opened 2026-08-30
Publish deferred maintained writes from resident state

aeplay:
🤖 Deferred local persistence commits resident state before storage completes, but the maintained-subscription refresh at that boundary was removed. Warm local subscriptions could therefore remain stale forever because no cold query work remained to schedule another refresh.

Before:

- a local mutation updated resident one-shot reads;
- persistence was correctly queued for a later owner tick;
- an already-open maintained subscription had no pending query/chunk flag, so its delta was never drained;
- example symptoms included a Todo delete that never disappeared and a newly inserted BandChat room that never appeared.

After:

- deferred write admission refreshes resident maintained subscriptions before queueing persistence;
- insert/delete deltas are visible immediately from resident state;
- persistence remains asynchronous and host-tick owned;
- no storage polling, nested owner tick, or busy loop is introduced.

Example: with `WriteMany` deliberately paused, an empty maintained subscription observes a local insert and subsequent delete before persistence begins. Once storage resumes, the existing queue settles the write normally.

Independent review planted removal of the refresh; the receipt failed on the first missing insert delta. Groove ordering/window receipts remained green, confirming this is a publication-liveness repair rather than a pagination semantics change.


aeplay:
🤖 Approved by Anselm.

PR #2378, opened 2026-08-30
Preserve flat-join contributors across forwarded resets

aeplay:
🤖 Forwarded maintained-view resets retained complete result payloads but discarded the validated flat-join source vocabulary. A receiver could render the authoritative payload, then immediately reconstruct the same generation from incomplete contributor facts and transiently retract a valid occurrence.

Before:

- normal publication paths supplied flat-tuple source tables;
- one forwarding/coverage constructor supplied an empty vector;
- no `ContributingMembers` facts were emitted for joined sources;
- an authoritative `root × child` result appeared, disappeared at the same generation, then reappeared through payload recovery.

After:

- flat-tuple source metadata is a private type derived only from `ValidatedQuery`;
- every initial, reset, forwarding, and test-capture constructor receives the same validated positional vocabulary;
- repeated physical tables remain distinct positional join sources;
- non-flat queries retain an empty vocabulary;
- forwarded receivers can reconstruct contributor facts consistently with the authoritative payload.

Example: for a two-source flat join where both positions happen to use the same physical table, the receiver records two positional contributor entries. It no longer collapses them or treats the joined side as absent.

Independent review planted both complete metadata omission and narrowing to the first positional source. Each deterministically reproduced the same-generation occurrence retraction; restoration passed the focused forwarding receipt.

This PR fixes forwarded insertion/update reconstruction. A separately tracked follow-up handles a newly exposed joined-source deletion invalidation timeout rather than folding that distinct lifecycle into this repair.

### DIFF
--- a/crates/jazz-testkit/tests/output_occurrence_id.rs
+++ b/crates/jazz-testkit/tests/output_occurrence_id.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 use jazz::query::{Query, col, eq, lit, table};
 use jazz::row_input;
 use jazz::tools::{
-    ColumnType, DurabilityTier, QueryResult, ResultKey, Schema, SchemaBuilder,
+    ColumnType, DurabilityTier, QueryResult, ReadTier, ResultKey, Schema, SchemaBuilder,
     SubscriptionStreamItem, TableSchema, Value,
 };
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2380, opened 2026-08-30
Avoid dropped port reservation in replica restart test

aeplay:
🤖 One replica-restart test reserves a loopback port by binding port 0, reads the assigned number, then drops that listener before starting the server. Under parallel CI another process can acquire the released port, producing an unrelated `AddrInUse` failure.

Before:

- the test guessed a future server port through a dropped reservation;
- ownership was lost between reservation and actual server bind;
- the restart scenario intermittently failed before exercising any Jazz behavior.

After:

- the first server starts directly on port 0 and therefore owns its selected listener;
- the test reads the actual bound port from the running server;
- it awaits shutdown, then restarts the second server on that same port as the scenario requires.

This changes test setup only. The replica settlement/reconnect behavior under test is unchanged.

Independent review ran the exact receipt successfully and repeated it 12 ways concurrently (12/12). The old reserve/drop behavior remains inherently nondeterministic rather than supplying a reliable negative plant; the new ownership sequence removes the TOCTOU structurally.

### DIFF
--- a/crates/jazz-testkit/tests/replica_settlement.rs
+++ b/crates/jazz-testkit/tests/replica_settlement.rs
@@ -629,11 +629,6 @@ mod client_transport {
             .build()
     }
 
-    fn reserve_local_port() -> u16 {
-        let listener = std::net::TcpListener::bind(("127.0.0.1", 0)).expect("reserve local port");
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2381, opened 2026-08-30
Queue transaction-local reads on async storage

### LINKED ISSUE #2379: Make transaction-local WASM reads truly asynchronous
🤖 Transaction-local reads are publicly Promise-based, but the WASM implementation still synchronously polls Core read futures with a no-op waker. A genuinely cold IndexedDB `Get`/`ScanOpen`/`ScanBatch` can therefore block or re-enter the browser owner thread instead of yielding.

There are two related boundaries to settle in one focused follow-up:

- implement an owned asynchronous WASM transaction-read boundary, equivalent in lifetime safety to NAPI's retained pending read;
- await an explicit per-transaction mutation FIFO barrier before evaluating so queued begin/stage work is visible in order.

Relation and array-subquery transaction query branches currently run before `pendingTxFromOptions` and therefore bypass transaction-local evaluation. Either make those paths transaction-aware or reject the shapes explicitly until supported; do not silently read outside the transaction.

Required receipts:

- cold reopened transaction read over yielding IndexedDB storage;
- pause begin/stage on `Get`, `ScanOpen`, and `ScanBatch`, then query without blocking;
- multiple stages preserve FIFO and read-your-writes;
- failed stage makes the transaction read fail consistently;
- relation and array-subquery queries have explicit transaction semantics.

This is an asynchronous read/API implementation issue. It must not introduce a second mutation API or make mutation callers drive queued work.

aeplay:
🤖
## Before

Transaction-local reads in the WASM and NAPI adapters could synchronously poll genuinely async storage. Trusted-serving reads also lacked an enforced immutable opening identity at the raw mergeable-transaction boundary, so a direct caller could open as Alice and query the same overlay as Bob.

## After

Transaction reads are owned async operations in the per-transaction FIFO. Cold storage suspends and resumes without blocking. Trusted-serving reads evaluate snapshot and staged rows under the identity captured when the transaction opened, and the core rejects mismatched identities for both exclusive and mergeable transactions.

For example, an Alice transaction containing an Alice-only staged row still returns that row if a later query call carries Bob session context; a direct raw request to read that transaction as Bob is rejected rather than changing the policy subject.

Relation and array-subquery shapes that cannot be evaluated transaction-locally remain explicitly rejected instead of silently reading a stale owner view.

## Verification

- Real raw NAPI Alice-allow / Bob-deny receipt
- Cold Get, ScanOpen, and ScanBatch suspension
- FIFO read-your-writes and failed-stage handling
- Rust, WASM, NAPI, and TypeScript focused gates
- Independent adversarial review

Stacked directly on #2371. Closes #2379.

### DIFF
--- a/crates/jazz-napi/index.d.ts
+++ b/crates/jazz-napi/index.d.ts
@@ -96,6 +96,14 @@ export declare class NapiDb {
   all(query: PreparedQuery, opts?: { tier?: string; local_updates?: string; propagation?: string; include_deleted?: boolean } | undefined | null): Uint8Array | PendingNativeRead
   /** Read through an open transaction using the identity bound at begin. */
   allInTransaction(query: PreparedQuery, tx: Tx, opts?: { tier?: string; local_updates?: string; propagation?: string; include_deleted?: boolean } | undefined | null): Uint8Array | PendingNativeRead
+  /**
+   * Read through an open transaction as its identity bound at begin.
+   *
+   * The core verifies that `author` matches the capability retained by the
+   * transaction; this ABI exists so trusted-serving hosts never fall back
+   * to their ambient/default read identity.
+   */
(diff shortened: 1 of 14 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2385, opened 2026-08-31
Integrate reviewed alpha.54 non-RN release prefix

aeplay:
🤖 This is the clean, linear, independently reviewed non-RN alpha.54 integration prefix. It exists to make the exact combined tree and CI state visible while the individual approved PR branches are normalized into the same native stack.

## Exact logical order

1. #2336 canonical durable identity encodings
2. #2352 bounded settled-state keys
3. #2334 native compatibility corpus
4. #2341 browser storage-open readiness
5. #2340 foreground transaction identities
6. #2351 auth-scoped IndexedDB ownership
7. #2347 browser foreground identity/recovery
8. #2364 packaged zstd negotiation
9. #2366 maintained equality selectivity
10. #2367 storage-backed maintained results
11. #2354 sealed correctness-artifact consumers
12. #2335 browser compatibility corpus
13. #2371 async mutation preparation/finalization
14. #2381 async transaction-local reads
15. #2374 iterative deep-graph evaluation
16. #2375 session-claims maintained-view isolation
17. #2376 native artifact fingerprinting
18. #2377 deferred maintained publication
19. #2378 forwarded flat-join contributors
20. #2380 replica-restart test repair
21. reviewed NAPI packaging/admission repairs
22. reviewed IndexedDB compatibility + inherited-large-value repairs
23. reviewed Groove hydration admission
24. reviewed authoritative reset/shared dependency-fact repairs
25. reviewed storage-backed replacement-witness repairs
26. planted black-box reset/restoration receipts
27. reviewed structured reset occurrence repair
28. reviewed follower causal-error/reconnect repair
29. integration-only duplicate dependency cleanup
30. writer-disconnect durability receipt
31. reviewed nullable-reference equality repair
32. reviewed queued-mutation tick-yield repair
33. restored #2377 maintained-publication receipt

## Deliberately excluded

- React Native work (#2360 and descendants).
- The structured Global relay handoff candidate: adversarial review still reproduces a duplicate CompleteTransaction on the raw Global path.
- Other unreviewed diagnosis work.
- The preview-only trigger commit; it will move to the absolute final tip once this prefix settles.

## Merge policy

This PR is the truthful exact-tip CI/integration surface, not permission to bypass existing per-PR approvals. The final merge remains the normalized native stack. Current local focused receipts are green; GitHub CI is authoritative for this pushed exact tip.

PR #2387, opened 2026-08-31
fix(jazz): resume structured relay handoffs

aeplay:
🤖 Restores cold structured relay handoff without introducing a second authorization or source-selection path.

Before:
- A maintained structured include could attach to an already-open receiver after relay handoff but never resume it.
- The subscriber remained stuck at the initial empty snapshot, even though the relay had the nested rows.
- A broader attempted fix introduced source-binding and terminal-forwarding machinery and conflicted with complete transactions.

After:
- Only the cold-handoff branch resumes the existing maintained receiver through the ordinary query-update path.
- Its first complete publication is marked as a reset because nothing from that group has previously been published.
- Edge, Global, and Local semantics continue to use the same authorization and maintained-view machinery.

Example:
1. A browser subscribes to a depth-3 include through a relay.
2. The foreground owner detaches; the persistent relay retains the cold maintained view.
3. A new foreground owner attaches.
4. The relay resumes that receiver and publishes exactly one fully materialized reset, followed by ordinary updates.

Verified:
- Cold Edge, Global, and Local structured receipts.
- Abrupt detach/reconnect and reopened-tab ownership receipts.
- Browser depth-3 include 2/2.
- Independent planted review: disabling the resume branch leaves snapshots stuck at `[[]]`; restoring it passes.

Stacked on #2385.

### DIFF
--- a/crates/jazz/src/db/peer_connection.rs
+++ b/crates/jazz/src/db/peer_connection.rs
@@ -4024,7 +4024,36 @@ where
                                 session_claim_binding.as_ref().expect("subscriber claims").0,
                                 session_claim_binding.as_ref().expect("subscriber claims").1.clone(),
                             );
-                            if settled_handoff {
+                            if settled_handoff && peer.has_maintained_subscription(group_subscription) {
+                                // A cold handoff opens the maintained view on
+                                // its first turn. Retrying that full rehydrate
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2388, opened 2026-08-31
fix(jazz): retain maintained replacement support

aeplay:
🤖 Preserves replacement witnesses while maintaining storage-backed views.

Before:
- The maintained fallback stripped replacement witnesses from an otherwise valid change bundle.
- A row replacement could therefore lose the removal side of the result transition.

After:
- Maintained evaluation retains the complete replacement information.
- Public result collection can emit the correct removal and addition without reopening or reconstructing the update.

Example: replacing two joined source rows in one transaction now yields one coherent maintained bundle containing both removals and replacements.

Verified by focused storage-backed and multi-row replacement receipts. Independent planted review restored the stripping behavior and made the receipt fail.

### DIFF
--- a/crates/jazz/src/node/query_eval.rs
+++ b/crates/jazz/src/node/query_eval.rs
@@ -957,9 +957,6 @@ where
             output_request
                 .facts
                 .remove(&ProgramFactKey::VersionWitnesses);
-            output_request
-                .facts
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2389, opened 2026-08-31
fix(jazz): publish complete deletion result deltas

aeplay:
🤖 Prevents deletion-witness bookkeeping from discarding already-proven public result deltas.

Before:
- A transaction could change a deletion witness and also produce a real public delta.
- Peer publication reopened authoritatively because of the witness change and discarded the proven delta.
- Flat joins, bounded pages, and route subscriptions could then miss removals or time out.

After:
- A complete non-empty public delta is published directly.
- Authoritative reopen remains available only when the collector does not already have a complete delta.

Example: deleting a joined source row now emits the joined-result removal in the same publication instead of timing out while waiting for a reopen.

Verified across flat-join netting/replacement, pagination shift, storage-frontier, and route-smoke receipts. Independent planted review restored the old reopen and reproduced the missing-delta timeout.


aeplay:
🤖 Approved by Anselm.

PR #2390, opened 2026-08-31
test(jazz): assert nullable join predicate literal

aeplay:
🤖 Aligns the graph-lowering structural receipt with the schema-driven nullable literal representation.

The nullable reference predicate is lowered as `Nullable(Some(String))`, not an unwrapped `String`. Runtime behavior is unchanged; this corrects the exact expected IR shape after nullable-reference normalization.

Independent planted review restored the stale unwrapped expectation and made the receipt fail.


aeplay:
🤖 Approved by Anselm.

PR #2391, opened 2026-08-31
fix(wasm): close database bindings synchronously

aeplay:
🤖 Makes WASM database closure reentrancy-safe without weakening the existing lifecycle boundary.

Before:
- `WasmDb.close(&mut self)` required an exclusive wasm-bindgen receiver.
- Cleanup could therefore throw a recursive-borrow error while an async binding operation still held a shared receiver.
- A first attempted latch-only fix avoided that borrow but accidentally left ordinary methods usable after close admission.

After:
- The binding stores its runtime in a shared interior lifecycle slot.
- `close(&self)` synchronously takes that slot, so every later operation immediately receives `WasmDb is closed`.
- Exactly one detached runtime performs physical asynchronous close; repeated closes resolve `false`.
- A physical close failure remains terminal, matching the previous eager-Closed behavior.

Example: calling `close()` and immediately calling `setNonDurableClient()` now rejects the latter before awaiting close, while the first close resolves `true` and the second resolves `false`.

Verified by native jazz-wasm tests (13/13), wasm32 checks, clippy, and two independent adversarial review passes.

### DIFF
--- a/crates/jazz-wasm/src/lib.rs
+++ b/crates/jazz-wasm/src/lib.rs
@@ -440,7 +440,10 @@ impl WasmWrite {
 
 #[wasm_bindgen]
 pub struct WasmDb {
-    inner: WasmDbInner,
+    // The runtime leaves this slot synchronously at close admission.  The
+    // close future owns the removed handle, while every later binding call
+    // observes `None` and fails before touching native state.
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2392, opened 2026-08-31
test(record-player): isolate parallel browser topology

aeplay:
🤖 Makes RecordPlayer browser topology tests coexist reliably with the parallel workspace gate.

Before:
- RecordPlayer used Vitest's default browser API ports.
- Other browser suites could claim those ports first, causing RecordPlayer to import from the wrong Vite server.
- The selection contract itself launched Vitest, creating a second hidden RecordPlayer server and making fixed-port attempts collide with themselves.

After:
- The one real RecordPlayer topology process uses reserved strict port `63318`.
- Selection checks consume shared data-only receipt patterns and never launch Vitest or Vite.
- A repository contract rejects any other tracked Vitest config that claims `63318`.

Example: workspace TS tests may run Jazz Tools, provider, and RecordPlayer gates concurrently; RecordPlayer now has one deterministic browser server, while a duplicate port declaration fails immediately with the conflicting config path.

Verified by the full RecordPlayer package gate (13 tests), correctness contract (7/7), and an independent planted provider-config collision.

### DIFF
--- a/dev/gates/test/correctness-consumer-contract.test.mjs
+++ b/dev/gates/test/correctness-consumer-contract.test.mjs
@@ -41,30 +41,54 @@ test("every direct Node/browser correctness entrypoint uses the one sealed consu
   assert.doesNotMatch(aggregate, /env:\s*process\.env/);
 });
 
-test("parallel RecordPlayer browser topology owns a unique strict Vitest API port", () => {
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2393, opened 2026-08-31
test(browser): target auth-scoped physical roots

aeplay:
🤖 Updates low-level browser-worker receipts to operate on the auth-scoped physical IndexedDB root that `createDb` actually opened.

Before:
- The tests advanced worker generation, replaced a storage manifest, or deleted IndexedDB using the caller's logical `driver.dbName`.
- Persistent browser storage now scopes that name by normalized auth identity.
- The operations therefore touched a different empty database, making correct runtime behavior look like missing fencing, failed recovery, or missing reload.

After:
- Receipts derive or retain the resolved physical root before performing low-level generation/storage operations.
- The derivation uses the production helper and preserves isolation between auth scopes.
- Runtime behavior is unchanged.

Example: corrupting the manifest now targets the scoped database opened by the current local-first identity, so the next `createDb` rejects for the intended corruption rather than opening a fresh unrelated store.

Verified in Chromium: generation fencing, corrupt-open rejection, external deletion/reload, and two-principal isolation 4/4. Independent planted regression back to the logical name reproduced the exact `NotFoundError`.

### DIFF
--- a/packages/jazz-tools/tests/browser/worker-bridge.test.ts
+++ b/packages/jazz-tools/tests/browser/worker-bridge.test.ts
@@ -288,28 +288,33 @@ describe("SharedWorker bridge with IndexedDB", () => {
     const appId = uniqueDbName("physical-worker-epoch-app");
     const dbName = uniqueDbName("physical-worker-epoch-root");
     const secret = generateAuthSecret();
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2394, opened 2026-08-31
test(browser): retain scoped worker root across restart

aeplay:
🤖 Corrects worker rejection/recovery receipts to retain the auth-scoped physical database root across restart.

Before:
- Inspector contexts correctly reported the physical scoped database name.
- The receipts later waited for release or terminated a worker using the raw logical `driver.dbName`.
- Those checks targeted a different root, returned too early, and left live runtime contexts behind.

After:
- Each receipt captures the inspector-provided physical name as an opaque same-root handle.
- Shutdown, restart, release, and termination checks consistently use that handle.
- Runtime rejection delivery and rehydration semantics are unchanged.

The intended behavior remains:
- A runtime attached only after a rejection receives no historical notification.
- A runtime attached while rehydration is active receives the live outcome exactly once.
- Rehydrated state excludes rejected writes.

Focused Chromium receipts pass 3/3. Independent planted regression back to the logical name makes all three fail with live runtime contexts, proving the corrected target is meaningful.

### DIFF
--- a/packages/jazz-tools/tests/browser/worker-bridge.test.ts
+++ b/packages/jazz-tools/tests/browser/worker-bridge.test.ts
@@ -1743,14 +1743,17 @@ describe("SharedWorker bridge with IndexedDB", () => {
 
     const inspectorControl = await dbBeforeRestart.openInspectorControlPort();
     inspectorControl.start();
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2395, opened 2026-08-31
test(wasm): await transaction read promises

aeplay:
🤖 Repairs the real wasm32 transaction-read receipts after those APIs became Promise-returning.

Before:
- Four WASM tests still typed transaction reads as synchronous `Result<Vec<u8>, JsValue>`.
- The wasm32 test target could not compile, so no wasm-bindgen integration receipt ran.
- A first rewrite accidentally dropped the Alice-bound/Bob-attributed security assertion.

After:
- Valid transaction-read overloads create and await Promises.
- Alice-bound reads attempted as Bob create a Promise and reject asynchronously with the bound-identity diagnostic.
- Cross-runtime object misuse still rejects synchronously before Promise creation.
- Dropping a non-owning transaction facade remains safe for both mergeable and exclusive owner batches.

Runtime behavior is unchanged.

Verified by wasm32 test compilation and real Node wasm-bindgen tests 6/6. Independent planted review disabled the identity guard and the Bob-attribution receipt failed exactly.

### DIFF
--- a/crates/jazz-wasm/src/lib.rs
+++ b/crates/jazz-wasm/src/lib.rs
@@ -4281,7 +4281,9 @@ fn unknown_transaction_kind_message(kind: &str) -> String {
 #[cfg(test)]
 mod dynamic_schema_view_tests {
     use super::*;
-    use jazz::db::{DbConfig, DbIdentity, ExclusiveTxOps};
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2396, opened 2026-08-31
feat(jazz): read relations inside transactions

aeplay:
🤖 Adds relation/include reads inside mergeable and exclusive transactions through the existing query compiler and transaction overlay.

Before:
- Primitive transaction reads worked.
- Any relation/include query threw `Native runtime does not support relation reads inside a transaction`.
- A first implementation reached the unified planner but transaction sources omitted version-witness metadata, so staged and include-deleted relations still failed before evaluation.

After:
- Transaction relation snapshots compile through the same relation query program as ordinary reads.
- The open-transaction overlay supplies current-row metadata/version witnesses, including staged deletes and restores.
- Staged roots and staged children are visible; relation predicates join the transaction read set and conflict detection.
- Storage-backed evaluation may suspend and resume without materializing a second query path.
- NAPI and WASM expose the same capability.

Example: inside an exclusive transaction, insert a parent and child, then query the parent with a nested child include. Both staged rows are returned. A concurrent child update after that read causes the transaction to reject with `transaction_conflict`.

Verified:
- Full query API: 171 passed, 6 skipped.
- Staged root/child, deleted/restore, and conflict receipts.
- Cold async-storage suspension.
- NAPI 29/29 and WASM 13/13.
- Exact-tip authoritative artifacts and planted ABI-bypass regression.
- Three adversarial review passes.


aeplay:
🤖 Approved by Anselm.

PR #2397, opened 2026-08-31
fix(jazz-tools): observe post-admission write rejections

aeplay:
🤖 Aligns TypeScript mutation receipts with the asynchronous post-admission write lifecycle.

Before:
- `insert`, `update`, `delete`, `restore`, `upsert`, and branch writes reserve/enqueue synchronously.
- Several tests still expected row-state failures to throw from the call itself.
- An over-broad normalization attempt could misclassify transport or `NotObserved` errors as permanent write rejection.

After:
- Post-admission row-state failures are observed through `WriteHandle.wait({ tier: "local" })`.
- Only native errors beginning with the exact Rust display prefix `WriteRejected:` become `PersistedWriteRejectedError` with stable `txId`, `code`, and `reason`.
- Protocol, Schema, `NotObserved`, cancellation, and unknown lifecycle errors pass through unchanged.
- Synchronous pre-admission validation remains synchronous.

Example: updating a deleted row returns a handle immediately; awaiting its local settlement rejects with `{ name: "PersistedWriteRejectedError", code: "write_rejected", reason: "row already deleted: …" }`. A transport error mentioning the words `WriteRejected` elsewhere remains the exact original transport error.

Verified by 74 focused runtime/client receipts, exact API suites, typecheck, and a planted false-positive classifier regression.

### DIFF
--- a/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts
+++ b/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts
@@ -4489,7 +4489,22 @@ function rejectedWaitError(
   message: string;
 } | null {
   const message = errorMessage(error);
-  if (!message.includes("WriteRejected")) return null;
+  if (extractWriteRejectedReason(message) === null) return null;
+  return queuedWriteRejection(transactionId, error);
+}
+
+function queuedWriteRejection(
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2398, opened 2026-08-31
fix(jazz): preserve queued no-op update identity

aeplay:
🤖 Preserves the public identity and settlement semantics of queued empty updates without publishing a fake row version or retaining unbounded alias state.

Before:
- A synchronous queued update reserves request transaction A.
- If its validated patch is empty, the direct mutation correctly returns the row's existing transaction B.
- The queue asserted A == B and panicked.
- Naive fixes either leaked an A→B map forever, lost the alias at NAPI/WASM boundaries, stole B's rejection, or hot-looped after durable acknowledgement failure.

After:
- The queued `WriteHandle` keeps stable public identity A and a bounded completion target that may resolve to B.
- NAPI/WASM write wrappers retain that handle; `writeState` and `wait` observe B while reporting A.
- No row version is published and direct `Db::write_state(A)` remains `NotObserved`, because A is not a durable transaction.
- Alias lifetime ends with the handle/wrapper; no runtime-global map exists.
- Waiting A observes B's rejection without consuming B's callback/error ownership.
- Durable rejection acknowledgements are flushed before clean close; failures are retained and surfaced.
- A poisoned Groove database requires reopen and does not schedule an infinite owner-turn retry loop.

Example: queue `update(id, {})` behind another write. Its handle is initially pending under A, then settles by observing current row transaction B. The row history remains unchanged. If B is rejected, B's observer sees B and A's observer sees A exactly once.

Verified across core FIFO/history/lifetime, both observer orders, durable close/reopen and injected storage failure, NAPI/WASM wrappers, and planted retention/reschedule regressions. Seven adversarial review cycles completed.

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -240,11 +240,11 @@ impl NapiDbInnerStorage {
 enum NapiWrite {
     Memory {
         db: Rc<CoreDb<CoreMemoryStorage>>,
-        tx_id: TxId,
+        write: WriteHandle<CoreMemoryStorage>,
     },
     Persistent {
         db: Rc<CoreDb<CoreRocksDbStorage>>,
-        tx_id: TxId,
+        write: WriteHandle<CoreRocksDbStorage>,
     },
 }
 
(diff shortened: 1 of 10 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2400, opened 2026-08-31
test(jazz-tools): await recursive permission setup

aeplay:
🤖 Repairs the recursive same-table double-reference permission receipt without changing permission lowering.

Before:
- The test enqueued all trusted setup writes and immediately queried from an untrusted session.
- After async mutation persistence, even the trusted setup read could still see an empty database.
- The resulting failure looked like a `team_id` versus `target_id` gather-lowering bug.

After:
- Setup writes await local settlement before the authorization query starts.
- The untrusted query and policy path are unchanged.
- A source-only negative row proves the recursive gather cannot accidentally follow the wrong same-table reference.

Example: the seed projects `team_entry.target_id` for the authenticated user, while recursion follows the `team_id` frontier. The intended target is visible; a row reachable only through the source-side reference remains hidden.

Verified by the full permission repro suite (4 pass, 1 expected failure), exact sealed artifacts, and planted wrong-seed/wrong-frontier regressions.

### DIFF
--- a/packages/jazz-tools/src/runtime/permissions.repro.test.ts
+++ b/packages/jazz-tools/src/runtime/permissions.repro.test.ts
@@ -574,32 +574,53 @@ describe("runtime permission repros for recursive gather and qualified predicate
     });
 
     const db = context.db(doubleRefReproApp);
-    const { value: userTeam } = db.insert(
-      doubleRefReproApp.teams,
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Approved by Anselm.

PR #2401, opened 2026-08-31
fix(browser): harden worker lifecycle and peer progress

aeplay:
🤖 Closes foreground lease admissions that outlive the caller's bounded timeout.

Before this change, a browser runtime could stop waiting for its SharedWorker lease while the worker continued allocating it. A late allocation then remained live without a foreground owner, and repeated attempts could accumulate retained ports.

After this change:

- the caller still fails after the existing bounded timeout;
- it sends a cancellation request tied to the exact admission request;
- if allocation is already in flight, the worker waits for it, retires the resulting lease, and acknowledges cleanup;
- pending cleanup is coalesced by physical worker and durable auth scope;
- ownership tokens prevent one concurrent successful admission from clearing another admission's retained cleanup.

Example: A times out while allocating; B for the same database succeeds concurrently; C cannot create another cleanup port while A remains pending. Once A receives its late lease and retires it, a subsequent attempt proceeds normally.

Verification:

- focused connection receipts: 7/7 passed, including timeout/success/retry interleaving;
- real Chromium delayed-allocation cancellation, retirement, and reacquisition receipt passed;
- TypeScript checks passed;
- independently adversarially reviewed through the final ownership-race repair.

This changes browser foreground lease lifecycle only. Storage and wire formats are unchanged.

It also fixes the Inspector reload symptom that exposed this work in CI. The
overlay previously selected the first registered runtime after reload, but
provider registration order is not stable. With multiple auth scopes it could
therefore select a sibling runtime that correctly contained no matching rows.
The host now derives its exact auth-scoped physical database root from the
verified session and passes that non-secret coordinate as Inspector-only
selection metadata. The iframe selects the exact `appId` + physical root while
preserving any later explicit user selection.

Additional verification:

- reversed same-app/base, different-auth-scope selector receipts: 2/2 passed;
- full real Chromium Inspector suite: 12/12 passed;
- the old registration-order selector reproduces the CI failure.

Finally, the foreground cancellation receipt no longer guesses allocation
timing or decodes a private pool. A test-only worker entry observes the exact
post-commit allocation point, proves the allocated node becomes durably
`retired`, then proves the successor uses a distinct node. The production entry
injects no hooks, exports no injection surface, ignores forged test fields, and
does not ship the test worker entry.

The full exact-tip gate then exposed concurrent first-tab admission races in
the production worker. Two initial ports could both observe no owner before
asynchronous Web Lock/IndexedDB admission published one; two queued lease
allocations could both decide abandoned recovery was necessary; and returning
the first active lease could close the physical owner while a second allocation
was still in flight.

The worker now:

- single-flights physical database admission per database and validates the
  established storage owner for followers;
- rechecks the foreground owner after asynchronous admission;
- serializes lease allocation so each active node is published before the next
  recovery decision;
- counts queued and in-flight allocations as owner liveness until active
  publication or terminal failure.

Production-path receipts concurrently admit two first tabs, return the first
while the second is queued, require both distinct live nodes to retire cleanly,
and prove terminal allocation failure permits a clean successor reopen.
Planted removal of physical coalescing, allocation serialization, pending
liveness, and failure cleanup each reproduces the corresponding Busy, unknown
lease, timeout, or leaked-owner failure.

Two exact-tip CI integration repairs accompany these lifecycle changes:

- external peer-frame admission now requests a distinct pass from every pump
  sharing the owner, so work arriving while a sibling subscription is mid-pass
  cannot remain stranded; routine end-of-pass notifications remain coalescible;
- RecordPlayer's concurrently executed unit, provider, and topology Vitest
  projects use disjoint Vite caches, preventing one project from re-optimizing
  another project's served module graph during Chromium import.

The peer-admission boundary is covered at both adapter and two-peer pump levels,
with repeated Todo, A-to-B, and worker-restart receipts. The cache boundary is
covered by exact concurrent package/browser execution and a planted shared-cache
hang. Browser app fixtures no longer pass the backend admin credential into
ordinary self-signed clients; server startup and schema/permission deployment
retain it.

The exact-tip gate subsequently exposed the analogous peer-progress gap for
frames arriving from the server rather than a foreground `MessagePort`. After
each real WebSocket carrier batch is auxiliary-routed and canonically admitted,
the adapter now requests one distinct peer pass. This happens after admission
(not at the generic scheduler or premature socket callback), includes
auxiliary-only batches, and cannot form a feedback loop because ordinary ticks
and outbound sends do not enter this boundary. Repeated A-to-B and B-to-A
cross-worker receipts pass; removing the notification fails the new real
WebSocket receipt.

The final CI retry found no failing behavioral test, but three nested framework
test runners lost their already-admitted correctness-artifact capability at the
Turbo boundary and performed fresh source admission after concurrent browser
outputs changed the workspace. Turbo now passes the two private capability
coordinates through test tasks only. The capability remains bound to its
0600 regular file, random token, repository root, live owner process identity,
and descendant process tree; build tasks, unrelated processes, replay, and
tampered artifacts remain rejected. React (38/38), Svelte (51/51), and Solid
(32/32) nested suites pass, and planted capability removal fails the contract
receipt.

Exact stack-tip CI at `729ce8389` is green: TypeScript, lint, storage
compatibility, Rust workspace, Rust differential, and both Rust completion jobs
all passed. Every behavior-changing repair above received an independent
adversarial review with planted regression sensitivity.


aeplay:
🤖 Approved by Anselm.

Issue #1995, opened 2026-08-25
Restart receipt: default/Global read after reopen includes writes already acknowledged Global


aeplay:
🤖 Define whether an Edge read after restart must preserve read-your-writes for a transaction that previously completed `wait({ tier: "edge" })` but has not reached Global durability.

Concrete observed sequence:

1. A client inserts two rows and both waits complete at Edge.
2. The runtime shuts down cleanly and reopens the same durable local database.
3. A Local read returns both rows.
4. The first default/Edge read can return zero rows.
5. Immediately repeated Edge reads return both rows after replay/upstream settlement progresses.

This is reproducible in `examples/docs/todo-server-ts`'s `survives a server restart` receipt. It is not storage loss: probes confirm the reopened Local view already contains both rows.

The implementation currently gives ordinary nodes a Global `upstream_durability_floor`. An Edge read therefore consumes Global upstream coverage, while reopened transactions at `Fate::Accepted` / `DurabilityTier::Edge` remain pending below that floor. The first canonical Global authority snapshot may exclude them; replay later makes them visible.

There is an explicit governing contract in `crates/jazz/src/db/tests/subscriptions/materialization.rs` stating that Edge reads consume the canonical Global settled member set and excluding an Edge-accepted local row absent from the server result. Making the restarted read causally preserve the accepted rows would require changing that contract and broadening reconciliation beyond unresolved/non-durable browser transactions.

Decision needed: should successful Edge wait establish durable read-your-writes across runtime restart for later Edge reads, or is the temporary canonical-Global result intentional? Once decided, replace the flaky todo-server receipt with a deterministic topology test for the chosen behavior and align the materialization contract/spec.

aeplay:
🤖 Restated for Core-only after #3235, with @aeplay's approval. The title changed and the stable-blocker label was removed. The original text above is kept for history.

**What changed.** The original question was whether a write acknowledged at the Edge tier survives a restart for later Edge reads. That no longer applies, because Edge acceptance is gone. Legacy tag-2 rows now reopen as Pending and are replayed to Core. That replay is covered by #3265 (a corpus fixture for tag 2) and #3266 (the app observes Core rejecting a replayed write).

**What remains.**
- The "survives a server restart" test in `examples/docs/todo-server-ts` still reads at `tier: "local"` to sidestep this issue (`tests/integration.test.ts:320-326`). The server now waits at `global` before returning 201 (`src/main.ts:81,165`).
[...]

PR #2402, opened 2026-08-31
fix(todo-server): harden restart and startup receipts

aeplay:
🤖 Hardens the public Todo server example along the same paths a fresh backend adopter uses.

Before this change, two useful receipts stopped short of the public boundary:
restart persistence was checked through the server's private admin handle, and
`startServer()` did not handle the Node listener's `error` event. Starting a
second server on an occupied port therefore left programmatic callers waiting
forever while Node emitted an unhandled `EADDRINUSE` error.

After this change:

- `startServer()` rejects with the original listener error and closes both the
  listener and Jazz context;
- restart persistence is read through the authenticated HTTP endpoint rather
  than a private local handle;
- a dense-history receipt creates one row, performs 256 authenticated updates,
  flushes and restarts Fjall, then reads revision 256 through HTTP.

Example: if port 3000 is already occupied, `await startServer({ port: 3000 })`
now rejects with an error whose `code` is `EADDRINUSE`; it no longer hangs.

This does not change `edge`/`remote` read semantics. The separate question of
globally settled membership after restart remains tracked in #1995.

Verification:

- Todo server focused suite: 27/27 passed;
- TypeScript build passed;
- trusted-backend policy-graph restart receipt passed;
- focused NAPI authorization/permission receipts passed;
- independent adversarial review planted stale restart responses and removed
  listener error handling; each corresponding receipt failed, then the lane was
  restored clean.

### DIFF
--- a/examples/todo-server-ts/src/main.ts
+++ b/examples/todo-server-ts/src/main.ts
@@ -285,8 +285,16 @@ export async function createServer(
  * @returns RunningServer with server instance and actual port
  */
 export function startServer(todoServer: TodoServer, port: number = 0): Promise<RunningServer> {
-  return new Promise((resolve) => {
-    const server = todoServer.app.listen(port, () => {
+  return new Promise((resolve, reject) => {
+    const server = todoServer.app.listen(port);
+
+    const onError = (error: Error) => {
+      server.off("listening", onListening);
+      reject(error);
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2403, opened 2026-08-31
fix(ci): harden release artifact provenance

aeplay:
🤖 Makes release and correctness artifacts fail closed at the exact byte and
process boundaries that consume them.

This addresses three independent gaps exposed while validating the freshly
merged release candidate:

1. Rust-shadow runners could begin from a contaminated reused workspace, but
   the old workflow emitted no status before checkout potentially erased the
   evidence.
2. NAPI staging verified a source binding and copied it later, leaving a
   validation-to-copy race.
3. A correctness consumer could restore the committed release fingerprint
   modules after producing lane-local sealed WASM/NAPI artifacts. Admission
   still succeeded, then browser startup failed with a confusing ABI mismatch.

After this change:

- shadow jobs inspect the workspace before `actions/checkout`, log tracked and
  untracked residue, and abort without deleting it; checkout cleaning is
  disabled, and the normal post-checkout source baseline remains mandatory;
- NAPI staging reads each source once, validates those captured bytes, then
  writes those same bytes to the package-visible destination;
- correctness-consumer preflight and postflight require the generated runtime
  expectation modules to match the sealed artifact snapshot.

Example: restoring the committed WASM expectation after a local correctness
build now fails during consumer admission with both expected fingerprints. It
no longer reaches Chromium and times out after a runtime ABI error.

Verification:

- focused artifact, provenance, release, CI, and source suites: 87/87 passed;
- exact workflow payload accepts an empty workspace, while planted tracked and
  untracked residue is logged, rejected, and left intact;
- an interposed NAPI source replacement cannot change staged bytes after their
  validation;
- planted fingerprint restoration is rejected at both consumer preflight and
  postflight;
- independently adversarially reviewed twice, including repairs for the
  original checkout-clean and staging-TOCTOU findings.

Storage and wire corpus audits found no additional format defect. This PR does
not change any persisted or wire encoding.


aeplay:

aeplay:

PR #2405, opened 2026-08-31
fix(jazz): bind trusted backend policy to link identity

aeplay:
🤖 Binds trusted-backend policy evaluation to the authenticated transport
identity instead of a principal supplied by the uploaded transaction.

Before this change, a trusted backend B could upload a transaction attributed
to user U and set `permission_subject=U`. Jazz then evaluated U's claims and
write policy even though B was the authenticated link. If U had an editor role
and B did not, B could obtain U's authority by naming U in the payload.

After this change:

- trusted-backend writes always evaluate permissions as the authenticated link
  identity B;
- `made_by=U` remains valid independent provenance and is retained on the row;
- a legitimate role claim admitted for B still authorizes B's attributed write;
- ordinary sessions and externally verified JWT sessions are unchanged.

Example: B uploads `{ made_by: U, permission_subject: U }` without an editor
claim. The server now returns `WriteRejected` and publishes no row, even when U
has the editor claim. Giving B the editor claim admits the same attributed
write while preserving `made_by=U`.

Verification:

- forged-subject, valid-backend-claim, attributed-provenance, ordinary-session,
  and external-admission receipts passed;
- durable rejection replay and reopen/fate-repair receipts passed;
- focused async-finalization, maintained-view/reset/revocation, and bounded M3
  receipts passed during the surrounding audit;
- independent adversarial review planted both the old selector and a nearby
  payload-first variant; each caused the new forged-subject receipt to fail.

This changes authorization only. It does not change row provenance, storage
encoding, or wire encoding.

### DIFF
--- a/crates/jazz/src/db/tests/mutations.rs
+++ b/crates/jazz/src/db/tests/mutations.rs
@@ -2539,8 +2539,10 @@ fn trusted_backend_upload_applies_session_claim_assertions_for_write_policy() {
         CommitUnitTrust::TrustedBackend,
     );
 
+    // The serving link's backend identity, not an attributed user's claimed
+    // permission subject, supplies the policy claims.
     backend.set_test_provider_claims(
-        editor_author,
+        backend_author,
         BTreeMap::from([("role".to_owned(), Value::String("editor".to_owned()))]),
     );
     let write = backend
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Closing after integration receipts disproved this interpretation.

#1971 deliberately separates backend admission from provenance by allowing an
explicitly trusted backend to assert the permission subject of a terminated
request/session while storing a different `made_by` author. That backend is a
trusted security boundary, not an untrusted session link.

Replacing the transaction permission subject with the transport identity broke
the existing SYSTEM-admitted streaming receipt and a trusted-backend delete
performed for an admitted provider session. The combined Rust workspace gate
caught both failures. The release stack therefore retains #1971's established
contract and drops this PR; no code from it remains in #2406.

PR #2406, opened 2026-08-31
fix(examples): use canonical chat policy identities

aeplay:
🤖 Updates the Chat example's policy receipts to use Jazz's canonical external
session identity rather than treating a raw JWT subject as a user identity.

Before this change, the documented example constructed test sessions without
an issuer and stored raw values such as `"alice"` in policy membership rows.
Current Jazz identities are canonical `[issuer, sub]` values, so the receipt
failed before reaching its intended permission assertions.

After this change:

- every external test session has an explicit issuer;
- stored policy identities use the public `userIdentity(issuer, subject)` helper;
- the same canonical value is exposed as `session.user`;
- a negative receipt proves raw `"alice"` does not authorize canonical external
  Alice.

Example: Alice authenticated by `https://chat.example` is represented by the
canonical encoding of that issuer plus subject. Alice from another issuer, or
the bare string `"alice"`, is a different identity and receives no membership.

Verification:

- Chat policy unit receipt: 8/8 passed;
- production build passed;
- issuer-substitution and raw/canonical planted mutations failed their intended
  authorization assertions;
- independently reviewed with no production runtime changes found.

This PR changes example tests only.

### DIFF
--- a/examples/chat-react/tests/permissions/permissions.test.ts
+++ b/examples/chat-react/tests/permissions/permissions.test.ts
@@ -1,9 +1,18 @@
+import { userIdentity } from "jazz-tools";
 import { createPolicyTestApp, type PolicyTestApp } from "jazz-tools/testing";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { app } from "../../schema.js";
 import permissions from "../../permissions.js";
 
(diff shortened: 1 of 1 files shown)


aeplay:

aeplay:

PR #2408, opened 2026-08-31
fix(release): seal assembled NAPI loader provenance

aeplay:
🤖 Repairs the artifact handoff used by pkg.pr.new and alpha publishing so
the assembled `jazz-napi` package can stage its loader without relying on a
machine-local generation pointer.

Before this change, the release workflow downloaded native binaries and their
manifests into a clean assembly job, then `pnpm pack` invoked the NAPI prepack
hook. That hook required `native-binding.pointer.cjs`, but the pointer was
intentionally ignored and absent. The alpha.54 candidate consequently failed
with `ENOENT` before pkg.pr.new could publish any package.

Copying the pointer alone is unsafe: it names a local
`.native-artifacts/generation-*` directory that does not exist in assembly.
Instead, the build now transfers a sealed loader-provenance receipt alongside
the generated loader and fingerprint. No-pointer assembly validates:

- exact generated-loader bytes by SHA-256;
- the exact native artifact fingerprint;
- kind, release profile, and selected host target;
- identity with the corresponding assembled platform provenance manifest.

The local pointer/generation path remains unchanged. Internal pointers,
generation directories, and loader provenance are consumed during prepack and
remain excluded from the published tarball.

Examples of rejected assembly:

- a loader retaining the NAPI-RS marker but replacing its executable contents;
- a different valid-looking 64-hex fingerprint;
- a self-consistent old loader bundle paired with a newer Linux manifest;
- a macOS-x64 loader paired with macOS-arm64 provenance;
- missing, symlinked, dangling-pointer, or unsupported-target artifacts.

Verification:

- focused release and NAPI contracts: 28/28 passed;
- valid Linux x64, macOS x64, and macOS arm64 assembly receipts passed;
- every tamper/replay/cross-platform case above was planted and rejected;
- dry-run packing excludes all internal generation/provenance machinery;
- independently adversarially reviewed through three repair rounds.

This changes packaging provenance only, not runtime, storage, or wire behavior.


aeplay:

aeplay:

PR #2409, opened 2026-08-31
test(band-chat): seal focused browser artifact consumer

aeplay:
🤖 Ensures BandChat’s focused browser receipt builds `jazz-tools` and runs Vitest inside one sealed correctness-artifact consumer.

Before: the focused receipt could build against one admitted artifact snapshot and start its tests under another, allowing stale generated expectations in `jazz-tools/dist` to escape the intended provenance boundary.

After: one outer consumer admits a single immutable snapshot, then runs the tools build followed by the focused Vitest target in the same child process tree. Both phases inherit the same private capability, and the outer postflight verifies that snapshot after testing completes.

Example: if the producer receipt changes between the build and test phases, the single outer postflight rejects the run. Splitting the phases into separate consumers, omitting the build, or reversing their order is caught by the executable selection receipt.

Verification:
- BandChat selection receipt: 3/3
- Artifact-store contracts: 14/14, including snapshot mutation rejection
- Real focused BandChat topology: 2/2
- Independent adversarial review planted split, omitted-build, and reordered variants; each failed as intended

This changes test execution only; application/runtime semantics are unchanged.

### DIFF
--- a/examples/band-chat/apps/nextjs-betterauth/scripts/test-browser-focused.mjs
+++ b/examples/band-chat/apps/nextjs-betterauth/scripts/test-browser-focused.mjs
@@ -2,12 +2,13 @@
 import { spawnSync } from "node:child_process";
 import { isAbsolute, relative, resolve, sep } from "node:path";
 import { realpathSync, statSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2411, opened 2026-08-31
fix(create-jazz): bind previews to immutable commits

aeplay:
🤖 Makes pkg.pr.new `create-jazz` previews scaffold from the exact immutable commit that produced the package, while preserving tag-bound release behavior.

Before: the preview tarball retained the repository package version (currently alpha.53), so `create-jazz@<preview-sha>` tried to fetch `v2.0.0-alpha.53`. If that historical source tag was absent, the CLI correctly refused to fall back to `main`, but no preview scaffold could be created.

After: the same-repository preview workflow stamps the packaged CLI with a strict source receipt containing schema version, matching package version, and the exact 40-hex PR head SHA. Preview starter and raw workspace/dependency fetches use only that SHA. Normal releases without the preview receipt continue to use only their exact version tag.

Example: invoking the f035 preview package now fetches starter sources and workspace dependencies from its producing commit rather than inferring a source tag from alpha.53. A missing, malformed, version-mismatched, or unavailable receipt fails explicitly and never falls back to `main`.

Verification:
- create-jazz tests and typecheck: 85 passed, 1 skipped
- committed production-style build → pack → extract → CLI import receipt confirms the file ships and is executable
- unavailable pinned commit fails closed with the immutable-source diagnostic
- independent adversarial review planted removal of package-version binding; the negative receipt caught it

This changes preview packaging only; normal release source selection remains tag-bound.

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -98,6 +98,13 @@ jobs:
             packages/jazz-tools/bin/native/jazz-tools-linux-arm64 \
             packages/jazz-tools/bin/native/jazz-tools-linux-x64
 
+      - name: Bind create-jazz preview to this immutable commit
+        env:
+          PREVIEW_COMMIT: ${{ github.event.pull_request.head.sha }}
+        run: |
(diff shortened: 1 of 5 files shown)


aeplay:

aeplay:

PR #2414, opened 2026-08-31
fix(runtime): wait for pending remote transport coverage

aeplay:
🤖 Makes strict `Remote` reads wait for the already-started server transport and authoritative query coverage instead of completing from empty local state during connection admission.

Before: `NativeRuntimeAdapter.connect()` could have a WebSocket handshake in flight but no native transport yet. Query attachment treated that transient state as “no upstream,” so a fresh client’s first `Remote` query returned `[]`; the same query returned the server row a second later.

After: strict remote query attachment waits for the current connection attempt, then attaches and awaits exact remote coverage. If auth refresh/reconnect replaces stalled carrier A with carrier B, the read follows B. If A terminates without a real replacement, the read rejects with that terminal error rather than hanging or falling back to local state.

Example: after restarting a persistent CLI core, a fresh memory client’s first `db.all(query, { tier: ReadTier.Remote })` now returns the persisted row in one call. A stalled handshake replaced by `updateAuth()` also settles through the replacement carrier; a malformed pre-hello frame rejects.

Unchanged:
- `LocalFirst` still answers from local knowledge.
- `RemoteIfPossible` still falls back only when the high-level connection state is definitely offline.
- an explicitly offline `Remote` read remains pending until reconnect.
- no timeout was increased and no second query path was introduced.

Verification:
- runtime/read-tier/connection-manager focused suite: 184 passed, 2 skipped
- real NAPI + persistent CLI-core restart receipt passes in one strict read
- independent adversarial review planted both inverse races; each new receipt failed as intended.

### DIFF
--- a/packages/jazz-tools/src/runtime/napi.core.test.ts
+++ b/packages/jazz-tools/src/runtime/napi.core.test.ts
@@ -2067,14 +2067,16 @@ describe.skipIf(!hasJazzNapiBuild())("jazz-napi native runtime memory DB", () =>
 
       const reader = openRuntime("reader", 42, server);
       const queryJson = JSON.stringify({ table: "todos" });
-      const propagatedRow = await waitFor(async () => {
-        const rows = (await reader.query(queryJson, null, "edge")) as Array<{
-          id: string;
-          table: string;
-          values: unknown[];
-        }>;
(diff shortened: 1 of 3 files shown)


aeplay:

aeplay:

PR #2415, opened 2026-08-31
fix(edge): use wall clock for client admission

aeplay:
🤖 Uses the correct clock domain for edge-side client HLC admission without contaminating monotonic retry or permission-scope maintenance.

Before: served-client edge admission passed a process-relative monotonic retry clock into Unix-millisecond HLC skew validation. Fresh ordinary writes through an edge were therefore rejected as `client_clock_too_far_ahead`, while the same write direct-to-core succeeded.

After: edge ingress samples Unix wall time for client HLC admission. Retry/backoff and idle permission-scope TTL eviction remain on monotonic maintenance time. Deferred fates retain the wall admission time sampled when the upload arrived, while later release/eviction uses the current monotonic maintenance clock.

Example: a current Unix-time client write now reaches Edge durability and queues upstream; a deliberately far-future HLC still rejects. If the wall clock jumps forward while an idle scope is younger than its 5s monotonic TTL, the scope remains; it is evicted exactly when monotonic maintenance reaches the TTL boundary.

Unchanged:
- Core admission already samples Unix wall time.
- retry scheduling remains monotonic;
- deferred writes do not “age into” validity while policy data hydrates;
- the clock-skew security boundary and far-future rejection remain intact.

Verification:
- real edge-client topology covers current-time acceptance and far-future rejection
- strengthened actual-ingress receipt covers `< TTL` retention and exact-boundary eviction
- simulation benches compile with both clock roles explicit
- independent adversarial review planted both wall/maintenance re-conflation and swapped admission arguments; the receipts failed as intended.

### DIFF
--- a/crates/jazz-sim/benches/s1_saas.rs
+++ b/crates/jazz-sim/benches/s1_saas.rs
@@ -535,6 +535,7 @@ fn edge_acceptance_phase(
         tx,
         versions,
         u64::MAX,
+        u64::MAX,
     ))
     .unwrap();
     let updates = settle_outcome(&mut edge.node, outcome).unwrap();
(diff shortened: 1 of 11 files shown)


aeplay:

aeplay:

PR #2416, opened 2026-08-31
fix(browser): fence worker retirement during bootstrap

aeplay:
🤖 Prevents a stale asynchronous SharedWorker retirement from closing over a newly arriving foreground bootstrap, and makes physical-owner release wait for actual Web Lock relinquishment.

Before: idle close ran from an unconditional async `finally`. A new port or lease request could arrive while the old owner release was still in flight, then the stale completion called `close()` over the successor. Separately, `epoch.release()` acknowledged release when it merely queued the Web Lock callback, so an immediate reopen could still observe the old realm as busy. Under full CI concurrency this appeared as foreground-lease, query-coverage, propagation, and teardown timeouts.

After:
- idle close uses an invalidatable retirement token and treats connecting bootstrap operations as live work;
- a lease-only successor reserves liveness before any runtime context exists;
- owner teardown awaits actual Web Lock callback completion before acknowledging release/reopen;
- a bounded internal lifecycle ledger records lease/peer/query/owner epochs for diagnostics, but exposes only entries bound to the exact auth session and live scoped DB root; scope-less lease/owner events and internal auth keys are never serialized.

Example: while old owner release is held, a successor asks only for a foreground-node lease. Releasing the old owner no longer closes the worker; the successor receives its lease. Removing the bootstrap reservation makes this receipt call `worker.close()` and fail.

Security/privacy boundary: after auth-scope turnover reuses the same physical DB name, scope B sees only its own retained trace. Scope A and unscoped lifecycle entries remain invisible.

Verification:
- focused worker/epoch/inspector units: 24/24 implementation, 20/20 final independent review
- real worker-bridge browser suite: 70 passed, 1 expected failure, 2 skipped
- TS check, browser-helper typecheck, runtime build
- independent plants prove the bootstrap fence, actual Web Lock completion, auth-scope filter, and scope-less exclusion are all test-sensitive
- no timeout or end-to-end assertion was weakened.

### DIFF
--- a/packages/jazz-tools/src/dev/inspector-overlay/browser-control-registry.ts
+++ b/packages/jazz-tools/src/dev/inspector-overlay/browser-control-registry.ts
@@ -91,6 +91,11 @@ export async function openAggregatedBrowserInspectorControlPort(
       if (message.type === "terminate-worker") {
         throw new Error("Worker termination is only available on a direct browser control port");
       }
+      if (message.type === "lifecycle-trace") {
+        throw new Error(
+          "Worker lifecycle trace is only available on a direct browser control port",
+        );
+      }
       const route = routes.get(message.contextKey);
       if (!route) throw new Error("Inspector context is no longer available");
       const control = controls.find((candidate) => candidate.port === route.port)!;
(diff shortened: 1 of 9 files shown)


aeplay:

aeplay:

PR #2420, opened 2026-08-31
fix(wire): reject noncanonical WebSocket batches

aeplay:
🤖 Enforces one exact canonical spelling for Rust WebSocket frame-batch carriers.

Before: four Rust ingress/transport paths used postcard from_bytes for the outer Vec<Vec<u8>> batch. That decoder accepts a valid value as a prefix, so appending 00 was silently ignored even though TypeScript rejects the same carrier and INV-WIRE-1 requires full consumption.

After: public server ingress, loopback ingress, native handshake, and native inbound pump all use the shared exact postcard decoder. It requires full byte consumption and encoder-canonical bytes. Existing carrier size, frame count, inner-frame, and handshake checks remain unchanged.

Example: canonical 01 01 42 decodes as one one-byte frame; 01 01 42 00 is rejected instead of acquiring a second accepted spelling.

Receipts:
- frozen valid and suffixed corpus at every Rust boundary
- central suffix and overlong-varint canonicality coverage
- valid native/server/loopback and TypeScript parity checks
- planted removal of exactness checks fails the corpus

Stacked on #2418.

### DIFF
--- a/crates/jazz-native-transport/src/lib.rs
+++ b/crates/jazz-native-transport/src/lib.rs
@@ -636,7 +636,7 @@ async fn receive_server_hello(
         return Err(WebSocketClientError::UnexpectedHandshakeMessage);
     };
     let encoded: Vec<Vec<u8>> =
-        postcard::from_bytes(&bytes).map_err(WebSocketClientError::DecodeBatch)?;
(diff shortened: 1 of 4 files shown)


aeplay:

aeplay:

Issue #2419, opened 2026-08-31
Cache repository-owned Rust tool installation in hosted CI


aeplay:
🤖 The repository-owned hosted-runner fallback for pinned Rust tools removes a fragile third-party action download, but currently compiles tools from source in a job-local Cargo home.

Measured from an empty isolated Cargo home:
- sccache 0.15.0: approximately 2m47s
- cargo-zigbuild 0.20.1: approximately 18s

Because preview jobs are isolated and sccache cannot accelerate its own installation, parallel Linux/WASM jobs repeat this cost. Design a repository-owned prebuilt artifact or exact-key cache with:

- pinned tool versions and checksums
- OS/architecture identity
- no mutable state shared between concurrent jobs
- provisioned-runner bypass parity
- a cold-miss receipt and warm-hit timing receipt
- fail-closed behavior on stale, mismatched, or tampered binaries

Context: discovered while replacing a repeatedly failing codeload dependency in the alpha.54 preview workflow. This is non-blocking for alpha.54 but should be addressed before routine preview volume increases.

PR #2421, opened 2026-08-31
fix(ci): remove preview tool installer action dependency

aeplay:
🤖 Removes the repeatedly failing taiki-e installer-action download from preview package builds while preserving exact Rust tool versions.

Before: every Linux/WASM preview builder had to download the pinned installer action before compilation. GitHub codeload returned HTTP 400 across two attempts, so no Linux artifact could be built.

After: a repository-owned allowlisted installer uses cargo install --locked for exact versions. An exact immutable cache key includes installer revision, OS, architecture, and tool/version. Warm hits verify the executable version before use; corrupt hits fail closed and tell operators to bump the installer revision; misses install, validate, and save. Provisioned runners keep their existing bypass.

Examples:
- a valid cached sccache 0.15.0 is reused without invoking Cargo
- a wrong-version binary under the exact key exits 65 instead of silently running or repeatedly reinstalling
- concurrent cold saves converge on the same immutable cache key

Receipts: 43 focused workflow tests, full workflow-contract suite, YAML parsing, shell syntax, and planted corrupt-cache/action-drift mutations.

Nonblocking cold-install optimization is tracked in #2419.

Stacked on #2420.

### DIFF
--- a/.github/actions/install-rust-tool/action.yml
+++ b/.github/actions/install-rust-tool/action.yml
@@ -1,10 +1,14 @@
 name: Install Rust tool
-description: Install a pinned Rust development tool in a job-local action home.
+description: Install an allowlisted, pinned Cargo tool in a job-local action home.
 
 inputs:
   tool:
-    description: Tool name and version accepted by taiki-e/install-action.
+    description: Pinned tool name accepted by dev/ci/install-rust-tool.sh.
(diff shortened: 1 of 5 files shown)


aeplay:

aeplay:

PR #2424, opened 2026-08-31
test(jazz): replay subscription order before reset update

aeplay:
🤖 Repairs a timing-dependent maintained-view receipt without weakening its semantic assertions.

Before: the integration test hard-coded old_index = 0 for a reset-normalized update. Because it subscribes before authority/order settlement, a legitimate earlier incremental update may already move that row to index 1; two CI runs therefore failed even though the implementation reported the actual immediately prior public state.

After: the test replays public OrderedRowDelta frames using consumer-equivalent occurrence/order semantics, captures the moving row position immediately before the unique final normalized update, and requires old_index to equal that position.

The receipt still requires:
- the final delta category is updated, never added or removed
- exact new_index = 1
- exact final row content
- exact final public ordering

Verification: focused integration pass, 12 concurrent repetitions, deterministic internal normalizer coverage. Planted wrong-index capture and retained-reset-as-added both fail.

Stacked on #2421.


aeplay:

aeplay:

Issue #2373, opened 2026-08-30
Rename the first public React Native relay ABI to V1


aeplay:
🤖 The React Native native-relay ABI is still unreleased, but its generated and native constants currently identify the settled first public ABI as version 7.

Immediately before the alpha.54 release, rename that settled ABI to V1 everywhere so the first shipped contract begins at version 1 rather than exposing implementation-history numbering.

Scope:

- generated Rust/TypeScript ABI constants and exports;
- iOS and Android native hosts;
- checked-in and generated device-acceptance fixtures;
- Open-range and Probe command encoding/decoding tests;
- packaging/contract receipts and user-facing documentation that names the ABI version.

This is a numbering-only release-settlement change. Do not alter command semantics, capability handling, lifecycle behavior, or postcard encoding as part of the rename. The device acceptance suite must still prove that Open negotiates the generated version and Probe reports the same version without hard-coded stale values.

aeplay:
🤖 Completed by #2552, merged as ab976caee4. Independent audit verified the first-public ABI is V1: Rust/generated TypeScript constants equal1, native hosts query the generated version, fixtures use generated constants, and package headers agree. ABI generation and four lifecycle/restart/ABI contracts passed, with green Android/iOS device acceptance on the integrated head. The broader final numbering sweep remains tracked separately in #2461.

PR #2426, opened 2026-08-31
test(rn): require relay readback before restart

aeplay:
🤖 This strengthens the React Native device acceptance receipt above #2360.

Before this change, the device scenario proved persistence only after terminating and restarting the app process. A foreground write could therefore appear to work without proving that a fresh public foreground client could immediately rehydrate the committed value through the persistent native relay.

After this change, the scenario:

1. writes a nonce-bearing row through one foreground client;
2. shuts that client down;
3. creates a fresh public foreground client in the same installed app process;
4. reads and verifies the exact nonce-bearing row through the persistent relay;
5. shuts the fresh client down in `finally`;
6. continues into the existing physical process-restart and readback receipt.

The source contract is AST-based and rejects decoy reads, helper substitution, swallowed failures, and missing shutdown. This increment does not claim two simultaneous physical JSI runtimes; that remains tracked separately.

Verification:

- `node --experimental-strip-types --test scripts/native-host-contract.test.mjs`: 19/19 passed
- TypeScript/source-contract checks passed in the implementation lane

Release scope reconfirmed by Anselm: alpha.54 explicitly includes React Native. This draft remains on the RN integration stack, but its required acceptance work now gates the complete alpha.54 release together with #2360 and #2373. The separately merged non-RN stack is the integration base, not the complete release candidate.


aeplay:
🤖 Superseded by #2552, now merged as ab976caee4 after Anselm’s explicit approval. Independent preservation review verified the meaningful behavior from this PR at 48a450b34691, including the governing single-foreground relay scope, remains in the integrated implementation. Both platform device suites, native linkage, isolated packaged consumers and full CI passed at a0708063b1. This PR is closed as superseded; its history and recorded decisions remain available here.

PR #2428, opened 2026-08-31
Track query shape ownership by peer

joeinnes:
## Summary

- Attribute each retained query-shape registration to the peer that owns it.
- Make repeated registration by one peer idempotent while keeping shared shapes until their final owner releases them.
- Reclaim owned shapes on unsubscribe and disconnect, including shapes parked while awaiting catalogue information.
- Reannounce a reclaimed shape before a same-cycle replacement subscription.
- Enforce per-peer and global shape-registration limits before retaining hostile input.

Before:

```text
A peer could repeat a registration or disconnect and leave a compiled shape retained without a live owner.
```

After:

```text
Each peer owns one claim per shape, and the shape is removed when its final owner releases it.
```

## Testing

- `cargo test -p jazz shape_registration_ownership` — passed


joeinnes:
Decided to use **1,024** as the per-peer shape-registration limit for this PR so it remains compatible with the existing 1,000-shape coverage-group workload. This is a pragmatic bound, not a final resource-policy decision: the per-peer and global limits, worst-case retained bytes, and fairness between peers likely need further review.

aeplay (approved):

aeplay (approved):

PR #2429, opened 2026-08-31
fix(web): isolate Node lease implementation from browser bundles

aeplay:
🤖 A clean-room alpha.54 preview install found that a fresh Next app could typecheck but could not produce a browser build: the shared default runtime source made Turbopack walk the Node foreground-lease implementation and reject `node:fs/promises`.

Before:

```text
Client Component import trace
  default-runtime-source
  node-foreground-node-lease
  node:fs/promises  <- browser build error
```

After:

- `DefaultRuntimeSource` remains the single public API on Node and browsers.
- the package's standard `browser` map replaces only the internal Node lease implementation with a type-compatible, fail-loud browser stub before a bundler traverses it;
- Node ESM resolution continues to use the real filesystem-backed implementation;
- an accidental browser call fails explicitly instead of silently inventing an unsafe identity.

Examples:

- A Next Client Component importing the normal Jazz provider now builds without pulling any `node:` module into its graph.
- A Vite browser build resolves the same guarded module to the browser stub.
- A Node process still acquires a foreground node ID, returns it with a high-water mark, reopens the same ID/mark, and retires it through the real implementation.

Verification:

- fresh create-jazz Next/Turbopack production build: passed with packed fixed package
- fresh Vite production build: passed
- `pnpm --filter jazz-tools build`: passed
- `pnpm --filter jazz-tools check`: passed
- focused package-surface and Node lease receipts: 16/16 passed in implementation lane
- planted missing/wrong browser mapping: both browser builds fail on real Node imports

This is the first repair found by exercising the exact pkg.pr.new candidate as an external adopter.

### DIFF
--- a/packages/jazz-tools/package.json
+++ b/packages/jazz-tools/package.json
@@ -13,6 +13,9 @@
   ],
   "type": "module",
   "main": "dist/index.js",
+  "browser": {
+    "./dist/runtime/native-runtime/node-foreground-node-lease.js": "./dist/runtime/native-runtime/node-foreground-node-lease.browser.js"
+  },
   "types": "dist/index.d.ts",
   "exports": {
     ".": {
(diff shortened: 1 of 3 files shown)


aeplay:

aeplay:

PR #2431, opened 2026-08-31
Bound RocksDB WAL sync debt during writes

joeinnes:
🤖 Bounds acknowledged RocksDB WAL-sync debt in the default `WalNoSync` mode without changing explicit/full-sync durability.

## Before

Background WAL writeback smoothed I/O, but the number of acknowledged mutation batches since the last synchronous WAL boundary was unbounded. A column-family-expanding reopen could also reset partially accumulated debt.

## After

- Every 64 successful backend mutation batches completes a synchronous WAL flush by default.
- Direct, conditional, and batched mutations each count once per successful mutation call; conditional no-ops do not count.
- Cadence accounting is serialized, and a failed flush retains overdue debt.
- A column-family reopen carries both the configured cadence and pending debt into the replacement handle.
- Explicit durability boundaries, synchronous close, and opt-in `FullSync` behavior are unchanged.

## Worked examples

- After 63 successful mutation batches, the next successful batch performs the synchronous boundary.
- A single atomic batch containing 100 row writes advances the cadence once, not 100 times.
- A failed conditional write advances it zero times.
- With a test cadence of 5, four writes followed by a column-family reopen still leave debt at four; the next successful write performs the boundary instead of allowing five additional writes.

## Verification

- Full `jazz-storage-rocksdb` suite: 30 passed, 1 expected ignored.
- Reopen regressions cover partial debt across repeated reopens and an exact completed boundary.
- An independent adversarial review planted the former reset behavior, observed the expected failure, and completed with no remaining findings.


aeplay (approved):
🤖 Approved by Anselm after the reopen-debt repair and independent adversarial review. Preserve this approval while the native stack is completed; do not merge independently yet.

aeplay (approved):

PR #2432, opened 2026-08-31
fix: retry typed bootstrap errors before hello

aeplay:
🤖 A clean-room CLI/server run exposed two distinct pre-Hello failures as the same misleading client error. The server sent a valid structured “catalogue bootstrapping; retry later” frame, but the TypeScript transport discarded it and reported `websocket received semantic frame before server hello`.

Before:

- only pre-Hello `AuthFailed` was decoded;
- a retryable bootstrap response failed strict remote waits immediately;
- terminal structured errors lost their code and message;
- disconnecting during a would-be retry had no explicit cancellation receipt.

After:

- v1 adds the explicit `WireErrorCode::NotReady = 6` discriminant;
- `NotReady + Later` before Hello reconnects with bounded 25ms→1s backoff while remote waits remain pending;
- explicit disconnect cancels the timer and prevents another socket;
- auth and terminal errors retain their structured behavior and real messages;
- only malformed or genuinely semantic non-error frames before Hello use the generic protocol error;
- runtime bootstrap wording is topology-neutral rather than calling every server an edge.

Example: a client that reaches a server before its authoritative catalogue is deployed now waits/reconnects. Once deployment completes and the server replies with Hello, the original strict read continues. A malformed semantic frame still fails closed.

The exact v1 bytes for `NotReady=6, Later=3` are frozen in the shared Rust/TypeScript artifact corpus alongside the existing malformed-frame representative.

Verification:

- TypeScript transport/runtime tests: 156 passed, 1 skipped
- Rust wire tests: 77/77 passed
- Rust wire fixture tests: 5/5 passed
- exact Rust↔TypeScript corpus decode/re-encode receipts passed
- planted wrong error code, wrong retry field, missing retry predicate, and uncancelled retry all fail their receipts

This is stacked after #2429, the first clean-room preview packaging repair.


aeplay:

aeplay:

PR #2434, opened 2026-08-31
Bound recursive policy decoding

joeinnes:
## Summary

- Apply shared depth and total-node limits while decoding recursive policy expressions.
- Limit policy depth to 256 nodes and total expression size to 16,384 nodes, matching existing execution and protocol ceilings.
- Decode catalogue policies with explicit frames so deeply nested input does not consume the call stack.
- Use bounded Serde visitors for Postcard wire predicates so oversized trees are rejected before full allocation.
- Return consistent named protocol-limit errors from both decoding paths.

Before:

```text
A deeply nested or very wide policy could consume stack or memory before its structure was validated.
```

After:

```text
Catalogue and wire decoders reject policies as soon as they exceed the shared depth or node boundary.
```

## Testing

- `cargo test -p jazz-server permissions_decode_enforces_policy_depth_and_total_node_boundaries` — passed
- `cargo test -p jazz policy_predicate_wire_decode_enforces_depth_and_total_node_boundaries` — passed
- `cargo test -p jazz ordinary_policy_predicate_variants_keep_their_postcard_roundtrip` — passed

### DIFF
--- a/crates/jazz-server/src/server/catalogue_payload_codec.rs
+++ b/crates/jazz-server/src/server/catalogue_payload_codec.rs
@@ -3130,6 +3130,66 @@ mod tests {
         );
     }
 
+    const BUG_124_POLICY_EXPRESSION_MAX_DEPTH: usize = 64;
+    const BUG_124_POLICY_EXPRESSION_MAX_NODES: usize = 4_096;
+
+    fn nested_policy_expression(nodes: usize) -> PolicyExpr {
+        assert!(nodes > 0);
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Superseded by #2438, which preserves this PRs commits on the actual alpha.54 ancestry and adds the independently reviewed iterative relation-expression/predicate repair. The replacement closes the ExistsRel stack-overflow bypass and enforces one shared catalogue-expression budget across outer policies and inner relation trees.

PR #2436, opened 2026-08-31
fix(sync): bind relay work to originating policy context

aeplay:
🤖 Bind every relayed subscription and row-version repair to the immutable user policy context that admitted it.

Before:

```text
A persistent trusted relay connected upstream as SYSTEM. When it multiplexed Alice and Bob, the core could re-evaluate a relayed query or terminal write proof under the relay transport identity instead of the originating user. Identical query shapes from different claim snapshots could also share one maintained coverage group.
```

After:

```text
Trusted relay messages carry an explicit delegated identity plus canonical claims snapshot. The core accepts delegation only over a TrustedBackend+SYSTEM link, partitions maintained coverage by that exact binding, preserves it across rehydrate/reconnect/repair/teardown, and fails closed if it is ever missing. Direct client links retain their ordinary connection-scoped identity.
```

Worked cases:

- Alice and Bob issue the same query shape: they use distinct policy-partitioned maintained receivers and cannot share authorized results.
- A relay reconnects or repairs missing row versions: the original immutable binding is reused rather than SYSTEM.
- An untrusted client or non-SYSTEM session attempts delegated fields: admission rejects it.
- A terminal relayed commit proves permissions under the validated transaction permission subject; row authorship remains unchanged.

The wire format is unreleased v1, so Subscribe and FetchRowVersions are redefined directly and exact Rust fixtures/spec text are updated without compatibility decoding.

The original authority-fate receipt also exposed a normal 2 MiB test-stack overflow. `PeerConnection::tick` is now heap-allocated at the database tick boundary; scheduling and control flow are unchanged.

Verification:

- Default-stack browser relay authority-fate receipt: passed
- Relay missing-binding and direct repair receipts: passed
- Exact wire corpus and rejection receipt: passed
- Jazz library test inventory compile: passed
- Independent adversarial review, including planted policy-key collision: READY
- Rebased patch independently re-reviewed: READY


aeplay:

aeplay:

PR #2437, opened 2026-08-31
fix(auth): bind internal support subscriptions to policy context

aeplay:
🤖 Bind every locally allocated authorization-support subscription to the exact permission subject and admitted session claims that caused its work.

Before:

```text
Terminal proof and deferred Edge support were keyed mainly by query shape. Locally created support subscriptions could either lack a binding, read claims from shared author-keyed compatibility state, reuse support hydrated under stale claims, or hydrate refreshed-claim support before recognizing an identical retransmit.
```

This produced four concrete failure modes:

- Inspector/backend support failed closed with `served subscription is missing its immutable policy binding`;
- same-author sibling sessions could overwrite one another's claims;
- concurrent deferred fates under claims A and B could share authorization evidence;
- an identical retransmit after claim refresh could leak support state or admit under different claims from its original deferred fate.

After:

```text
Dispatch snapshots the exact subscriber (subject, canonical claims) binding. Terminal proof threads and validates that binding. Deferred Edge support derives a deterministic runtime-only usage identity from the canonical support clause plus exact writer/claims, so A and B may coexist without sharing evidence. Existing transaction identity is resolved before support hydration, preserving the original fate and claims on identical retransmit.
```

Stale terminal support is torn down and rehydrated when its immutable binding differs. Deferred support refcounts and teardown operate on the binding-qualified usage identity. Direct client admission remains connection-scoped; trusted delegated links retain their explicit immutable binding.

This is a follow-up to #2436. It does not relax fail-closed authorization, change row authorship, or fall back to SYSTEM/transport identity.

Verification:

- same-author sibling A/B terminal isolation;
- A→B→A claim refresh and stale-support replacement;
- concurrent deferred Edge A/B support coexistence and independent teardown;
- refreshed identical retransmit leaves fate count, refcounts, publication footprint, and original claims unchanged;
- conflicting retransmit still rejects;
- authority revision, missing-seed denial, retransmit, resume/reconnect, terminal-fate, and edge-support receipts;
- planted old node-global lookup, shared support key, and late retransmit check each fail their focused receipt;
- rebuilt NAPI/WASM correctness artifacts and jazz-tools;
- standalone Inspector browser suite: 11/11 substantive flows;
- independent adversarial review: READY at `802f6d261`.


aeplay:

aeplay:

PR #2438, opened 2026-08-31
fix(policy): bound the complete catalogue expression tree

aeplay:
🤖 Replace #2434 on the real alpha.54 stack and close the recursive-policy denial-of-service gap found in independent review.

Before:

```text
#2434 bounded the outer PolicyExpr decoder, but one ExistsRel node delegated to recursive relation-expression and predicate codecs. A canonical payload with 256 nested relation nodes overflowed the default Rust test stack before the advertised limit could protect it. A later repair initially gave policy and relation trees separate node budgets.
```

After:

```text
Policy, relation-expression, and relation-predicate decoding use explicit post-order frames and one mutable catalogue-expression budget from the permissions root. Canonical re-encoding is iterative too. Depth and total nodes are checked before child allocation, including across PolicyExpr -> ExistsRel boundaries.
```

Boundary examples:

- exactly 256 nested relation nodes decode and round-trip on the default stack;
- node 257 returns the named MAX_POLICY_EXPRESSION_DEPTH error;
- 8,192 outer policy nodes plus 8,191 inner relation nodes exceed the single combined 16,384-node budget and reject;
- wide Union and And payloads reject before allocating attacker-sized child vectors;
- ordinary relation and predicate variants preserve exact canonical bytes.

This branch preserves Joe commits from #2434 as separate commits, adds the iterative relation repair, and is physically based on #2437 so the release candidate has real ancestry rather than only GitHub base metadata.

Verification:

- jazz-server library: 181/181 on the combined stack
- clippy/fmt/sensitive-data: passed on the implementation
- independent adversarial review with planted fresh-budget regression: READY

Supersedes #2434.


aeplay:

aeplay:

PR #2440, opened 2026-08-31
Harden branch-view upserts

joeinnes:
🤖 ## Summary

Make branch-view upserts consistently operate on the complete head-over-base view across Rust, N-API, and WASM, while removing the obsolete JavaScript `{ branch }` selector completely.

## Before

- Inherited large values could fail during branch upserts.
- A staged delete or staged row could hide a later upsert in the same transaction.
- Untyped JavaScript objects containing `{ branch: ... }` could be misinterpreted or silently select the root table, especially when the value was `null` or `undefined`.

## After

- Branch upserts read the complete head-over-base view and preserve verified inherited large-value descriptors.
- A later upsert visibly replaces a staged delete, and session transactions can upsert their own staged branch rows.
- The only supported selector is `{ head, base? }`.
- Presence of the removed `branch` property fails clearly before target selection, including null, undefined, inherited, getter-backed, Proxy-visible, and mixed `{ branch, head }` forms.

## Example

```ts
// Canonical branch-view target.
db.upsert(table, row, values, {
  head: draftBranch,
  base: publishedBranch,
})

// Rejected rather than ignored or silently writing to Root.
db.upsert(table, row, values, {
  branch: undefined,
})
```

## Verification

- Focused branch-view tests cover inherited large values, staged delete replacement, and repeated staged session upserts.
- Real N-API and WASM receipts cover canonical selection and all removed-property presence cases.
- Type declarations make the removed shape unrepresentable.
- A planted value-based detector causes `branch: undefined` to select Root and fails the permanent regression.
- Spec and changeset contain no compatibility-alias claim.


aeplay (approved):
🤖 Approved by Anselm after review. Preserve this approval while the dependency-ordered stack is finalized; do not merge this PR independently.

aeplay (approved):

PR #2442, opened 2026-08-31
fix(wire): bound canonical WebSocket carriers before allocation

aeplay:
🤖 Enforce the WebSocket carrier contract before allocation and before emission across every maintained Rust and browser path.

Before:

```text
Inbound batches were deserialized as Vec<Vec<u8>> before the 4,096-frame limit was checked. A count flood therefore allocated/materialized the carrier before rejection. Outbound Rust batching also compared raw frame bytes with the carrier cap, so a raw 2 MiB frame could become an oversized carrier after postcard count/length overhead.
```

After:

```text
A shared canonical decoder reads count and frame-length varints first, rejects empty or >4,096 batches before outer allocation, bounds each frame and the total carrier, and requires shortest varints plus exact consumption. Senders validate the actual encoded carrier before returning or transmitting it.
```

Examples:

- a declared 4,097-frame batch is rejected before allocating its frame vector;
- a singleton raw frame of `MAX_WIRE_FRAME_BYTES - 4` produces an exactly 2 MiB carrier and round-trips;
- a raw 2 MiB frame is rejected alone and after an earlier frame flush, so no receiver is sent an invalid carrier;
- a raw handshake frame plus trailing bytes is rejected.

The server route, loopback transport, native transport handshake/pump, and browser TypeScript codec all use the bounded contract. Browser and loopback senders also split at the frame-count limit.

Verification:

- jazz-server: 186/186;
- jazz-native-transport: 12 unit + 12 integration;
- browser WebSocket tests: 28/28 plus TypeScript check;
- planted removal of Rust/TS count bounds, canonical-varint checks, and encoded-carrier checks fails the focused receipts;
- independent adversarial review: READY at `062adf359`.


aeplay:

aeplay:

PR #2443, opened 2026-08-31
fix(ts): accept documented read-tier string literals

aeplay:
🤖 Make the documented read-tier string literals valid public TypeScript inputs while preserving the existing `ReadTier.X` values.

Before:

```ts
const options: QueryExecutionOptions = { tier: "remote-if-possible" };
// TypeScript error: the string literal is not assignable to enum ReadTier
```

The runtime accepted the value and the docs recommended it, but the published declaration used a string enum.

After:

```ts
const a: QueryExecutionOptions = { tier: "local-first" };
const b: QueryExecutionOptions = { tier: "remote" };
const c: QueryExecutionOptions = { tier: "remote-if-possible" };

ReadTier.LocalFirst; // still "local-first"
```

`ReadTier` is one const object with a same-name derived literal-union type, so runtime constants and public type declarations cannot diverge. Legacy durability-tier inputs and lowering semantics remain unchanged.

Verification:

- public package-root compile receipts cover all three literal forms;
- focused runtime tests: 37/37;
- `tsc -p tsconfig.tests.json --noEmit` and package typecheck pass;
- restoring the old enum fails all six literal compile/runtime call sites;
- independent adversarial review: READY.


aeplay:

aeplay:

PR #2445, opened 2026-08-31
Preserve legacy catalogue hash migrations

joeinnes:
## Summary
- Keep Double and Bytea structural hashes distinct across runtimes.
- Connect matching legacy and corrected Bytea identities even when both already exist, and bound concurrent historical lookups.

## Before
A deploy could select the corrected identity and skip the legacy migration edge, while historical identity probes ran serially.

## After
Deploys preserve legacy-to-current connectivity and scan historical candidates in deterministic batches of up to eight concurrent requests.

## Test plan
- [x] `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/dev/catalogue.test.ts -t 'legacy Bytea catalogue identity|stored schema identity lookup'`

### DIFF
--- a/.changeset/distinct-column-hash-tags.md
+++ b/.changeset/distinct-column-hash-tags.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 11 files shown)


aeplay:
🤖 Superseded by #2460, which preserves the forward-format correction while intentionally removing legacy catalogue migration/probing behavior. It also closes the adjacent EnumPayload cross-runtime collision and adds a complete shared structural-hash corpus. No commits from this PR were force-rewritten.

PR #2446, opened 2026-08-31
Make Inspector saves retry-safe

joeinnes:
🤖 Makes Inspector table saves retry-safe when the write may have committed but its confirmation was lost.

## Before

- Retrying an ambiguously acknowledged insert could issue a second insert with the same UUID.
- Editing that staged row before confirmation retained it as an insert rather than converting later edits into an update.
- Cancel/Delete during ambiguous confirmation silently removed local staged state; if the insert had committed, the row remained remotely despite the user's cancellation.

## After

- A submitted insert keeps one stable UUID and is never submitted as a second insert.
- Confirmation retires the submitted insert identity; edits made after submission become one update against the acknowledged UUID.
- Cancel action, selected-row deletion, and Delete-key cancellation retain intent while confirmation is pending. A successful insert acknowledgement produces exactly one delete; a definitive insert rejection discards the now-unnecessary delete.
- Retry becomes synchronously disabled as `Confirming…`, preventing duplicate retry launches.

## Worked cases

- Insert succeeds but the confirmation is lost; Retry confirms the original UUID without inserting again.
- The user edits the row while confirmation is ambiguous; after confirmation, Inspector sends one update containing the later edit.
- The user deletes the row while confirmation is ambiguous; confirmation of success is followed by one delete, while confirmation of rejection sends no delete.

## Verification

- 56 focused Inspector tests pass, including real-database ambiguous acknowledgement receipts.
- Planted duplicate-insert, lost-cancellation, and rejection-cleanup regressions fail as intended.

Stacked on #2455.


aeplay (approved):
🤖 Approved by Anselm after the ambiguous insert/edit/cancel repairs and independent adversarial review. Preserve this approval while the native stack is completed; do not merge independently yet.

aeplay (approved):

PR #2448, opened 2026-08-31
Reject ambiguous policy expression inputs

joeinnes:
## Summary
- Reject unbranded policy-shaped values on tables without a matching `type` column.
- Keep schema-declared `type` fields as row data and require `raw(...)` for manually authored policy IR.

## Before
An unbranded `{ type: "True" }` value could be mistaken for policy IR and weaken a declared row predicate.

## After
Policy-shaped unbranded values fail closed unless the schema makes them valid row data. The discriminator performs only the narrow schema check it needs.

## Test plan
- [x] `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/permissions/index.test.ts src/permissions/type-inference.test.ts`


aeplay (approved):
🤖 Approved by Anselm after review. Preserve this approval while the dependency-ordered stack is finalized; do not merge this PR independently.

aeplay (approved):

PR #2455, opened 2026-08-31
Revalidate pending enum lineages

joeinnes:
🤖 Makes pending enum lineage activation and durable recovery preserve one authoritative physical case order, including wide payload-enum ordinals.

## Before

- Out-of-order activation could leave an over-capacity sibling durably pending and poison a later reopen.
- Durable scalar and payload case records carried schema/ordinal provenance that recovery trusted without binding it back to the authoritative manifest. A forged record could keep a valid UUID but reorder physical tags after restart.
- Payload ordinals wider than 255 introduced an unnecessary adaptive v1/v2 physical-mapping codec.

## After

- Invalid pending lineage is quarantined and durably removed without consuming physical IDs.
- Direct and nested scalar/payload registries validate each UUID's introducing schema and ordinal against authoritative manifest coordinates and its earliest introduction alias before descriptor rebuild.
- The physical mapping has one canonical v1 representation. Payload ordinals are always little-endian `u32`; there is no narrow, adaptive, legacy, or v2 decoder path.

## Worked cases

- A valid scalar lineage grows from 255 to 256 cases, restarts, and preserves every physical tag.
- A nested payload enum introduces case ordinal 256 and round-trips through the exact v1 fixture.
- A stored case that retains its UUID but forges its introducing ordinal is rejected before activation, rather than silently reordering existing values.
- Header `2` is rejected as noncanonical; it is not treated as an older/newer compatibility format.

## Verification

- Scalar and payload forged-reorder restart regressions pass.
- Canonical scalar 255→256 and nested payload case-256 reopen receipts pass.
- Parked-sibling revalidation and staged-lineage crash recovery pass.

Stacked on #2460.


aeplay (approved):
🤖 Approved by Anselm after the provenance-hardening, canonical v1 wide-ordinal reset, and independent adversarial review. Preserve this approval while the native stack is completed; do not merge independently yet.

aeplay (approved):

PR #2456, opened 2026-08-31
fix(auth): refresh immutable policy-bound views safely

aeplay:
🤖 Repairs the immutable policy-binding lifecycle exposed by the exact combined alpha.54 gate.

## Before

- Direct/no-waker maintained views captured empty claims or fell back to the author-keyed compatibility cache.
- `PeerState::new()` silently represented a multiplexing relay, so standalone helpers could accidentally exercise relay semantics without an admitted binding.
- Authenticated claim refresh retained old maintained-view bindings and, at an edge, the old upstream usage. A cold/yielding refresh could still acknowledge the revision.
- Current-row views retained their pre-refresh evaluator.
- Policy-partitioned maintained receivers were not retired under their exact derived key during refresh or disconnect.

For example, Alice refreshing from `{team: A}` to `{team: B}` could keep the A-scoped maintained receiver or upstream subscription alive; empty/revoked results might never be published. A delegated session carried by a trusted relay could also be accidentally conflated with the relay transport session.

## After

- Standalone `PeerState::new/default` is explicitly a SYSTEM client link; real multiplexing links opt into `PeerState::relay()` and remain fail-closed without an exact binding.
- Direct helpers snapshot the admitted `(subject, canonical claims)` once. Production current-row opens install the connection-local snapshot explicitly, never the author cache.
- Coverage groups record whether their binding is `DirectAdmitted` or `Delegated`. Only direct groups rebind on authenticated claim refresh; delegated snapshots, including delegated SYSTEM, remain immutable.
- A direct edge refresh retires the old local receiver and upstream usage, derives a fresh claims-qualified usage key, waits for fresh upstream settlement, and then publishes a fresh reset—including an empty reset on revocation.
- Cold refresh attempts leave the revision pending for retry. Current-row views are forgotten and reset under the new snapshot.
- Refresh and disconnect retire the exact policy-derived maintained receiver key in PeerState, NodeState, and Groove.

## Worked cases

1. Two sessions share a subject but have different claims: each keeps its own immutable view; refreshing one does not retarget its sibling.
2. Edge session changes A → B: the A receiver and upstream usage disappear, core settles the B usage, and downstream receives the B result (or an empty reset).
3. Trusted relay carries a delegated subject: refreshing the outer relay transport does not rewrite the delegated binding.
4. A cold maintained view yields during refresh: the claim revision is not acknowledged and the owner loop retries.
5. Disconnect after a policy-partitioned subscription: the exact derived receiver is removed and Groove returns to its prior active-subscription count.

## Verification

- Independent adversarial review: READY after two lifecycle findings were repaired.
- Planted mutations proved direct-origin refresh, old local-receiver retirement, delegated binding preservation, trusted relay admission, and policy-derived disconnect cleanup.
- Focused claim-refresh tests: 5 passed.
- Admission/fate lifecycle tests: 46 passed.
- Rust library suite at the exact tip with bounded concurrency: 1,579 passed, 2 ignored, 0 failed.
- The two tests that initially hit host file-descriptor exhaustion passed individually; bounded rerun passed the full suite.


aeplay:
🤖 Approved by Anselm in the orchestrator review queue. Preserve this approval while the dependent stack is completed; do not merge independently yet.

aeplay:

PR #2458, opened 2026-08-31
fix(sync): align trusted relay fixtures with policy binding

aeplay:
🤖 This follow-up makes the policy-binding lifecycle repair coexist with the full release topology suite.

## Before

- Several relay fixtures described trusted relay-to-authority hops as direct user sessions, so immutable delegated binding correctly rejected them.
- A durable-worker scenario asked a client link to ship complete exclusive payloads; that preference is intentionally relay-only, leaving its generic transaction reads one publication behind.
- Authoritative ingest polled a large staging future inside the admission and peer-tick stack, overflowing the normal Rust test-thread stack for a finite exclusive bundle.

## After

- Relay fixtures explicitly distinguish trusted physical admission from delegated user policy identity.
- Durable workers and foreground clients use explicit immutable per-table bindings and the correct relay/client roles.
- The large staging future has one heap-indirection boundary; stage → finalize → notify → apply ordering and cancellation behavior remain unchanged.

## Worked cases

- An Alice subscription traversing a trusted relay is admitted physically as `SYSTEM`/trusted-backend while still evaluated and delivered under Alice's delegated claims.
- A durable worker receives complete exclusive payloads only when modeled as a relay; foreground traffic remains a client link bound to the same author.
- The browser relay exclusive-sibling projection test now completes on the default stack rather than aborting with stack overflow.

## Verification

- Replica settlement: 10 passed, 1 existing ignored; planted direct-session regression fails.
- Scenario smokes: 10/10 passed; planted client-role reversal restores `ExclusiveConflict`.
- Browser relay durability: 30/30 passed; removing the heap boundary restores the stack overflow.
- The stack-boundary repair completed an independent adversarial review with no findings.

Stacked on #2456. Draft until the remaining browser delivery repair and exact combined gates settle.


aeplay:

aeplay:

PR #2450, opened 2026-08-31
Avoid resident large-value lifecycle stalls

joeinnes:
## Summary
- Defer eviction without blocking while a resident publication owns lifecycle serialisation.
- Reuse durable metadata during reclaim when no staged override exists.

## Before
Cross-receipt eviction could deadlock behind a lifecycle guard, and reclaim could read the same backend metadata twice.

## After
Eviction returns a retryable deferred result until persistence advances, and reclaim performs one required durable metadata read.

## Test plan
- [x] `cargo test -p groove db::tests::batches::cross_receipt_eviction_defers_until_resident_publication_is_durable -- --exact`
- [x] `cargo test -p groove db::tests::batches::reclaim_without_a_staged_override_fetches_node_metadata_once -- --exact`

PR #2460, opened 2026-08-31
fix(schema): distinguish Bytea structural identities

aeplay:
🤖 Establishes complete forward-only structural schema identity parity between Rust and TypeScript. It supersedes #2445 without retaining historical migration or probing behavior.

## Before

- `Double` and `Bytea` shared one structural hash tag, so schemas differing only by those types could collide.
- TypeScript omitted `EnumPayload` entirely from structural hashing, so distinct payload-enum schemas could collide and disagree with Rust.
- Cross-runtime fixtures covered only a narrow enum subset, allowing one-sided tag drift to escape tests.

## After

- `Double` remains tag `10`; `Bytea` uses distinct tag `16` in Rust and TypeScript.
- `EnumPayload` uses tag `13` and hashes case/field counts, declaration order, names, nullability, and recursive field types identically in both runtimes.
- TypeScript's portable column-type switch is compile-time exhaustive.
- A shared v11 corpus covers every portable tag plus nullable, enum/payload-ordering, `Row`, `Array`, and nested composite shapes; both runtimes verify exact hashes and uniqueness.
- There is deliberately no legacy-identity lookup, migration edge, or backwards-compatible probing.

## Worked examples

- `Record<{ payload: Double }>` and `Record<{ payload: Bytea }>` now have different identities.
- Reordering payload-enum cases or fields changes identity consistently in Rust and TypeScript.
- Changing payload-field nullability or a nested `Array<Row<...>>` type changes identity consistently.

## Verification

- Rust and TypeScript match the complete shared corpus exactly.
- Planted independent Rust and TypeScript tag changes fail their respective corpus checks.
- Planted payload ordering/tag changes and a missing TS variant are detected.

Stacked on #2450.


aeplay:
🤖 Approved by Anselm after the complete Rust/TypeScript structural-identity corpus and adversarial review. Preserve this approval while the native stack is completed; do not merge independently yet.

aeplay:

PR #2471, opened 2026-09-01
fix(policy): bypass SYSTEM authorization support

aeplay:
🤖 **Before:** `asBackend()` writes use the exact SYSTEM permission subject, and INV-RLS-2 requires SYSTEM to bypass read and write policy. Final write evaluation already honored that rule, but terminal-proof and edge-support preparation ran first. A policy containing `session.user`, `claims.*`, or recursive joins could therefore fail while preparing authorization support before the SYSTEM bypass was reached. BigLabel’s intended first-admin backend bootstrap reproduced this.

**After:** Both terminal proof and unsettled edge support return before hydrating claim/join authorization support when—and only when—the transaction writer is exactly `AuthorSubject::SYSTEM`. Non-SYSTEM writers still prepare dependencies and fail closed when required claims are absent.

**Example:** a trusted backend may seed the first organization membership even though its normal user policy asks whether `session.user` is already an admin. A user write with missing claims remains rejected; merely using a trusted connection or backend attribution does not turn a non-SYSTEM writer into SYSTEM.

**Receipts:** recursive claim+join regression, 85 peer tests, BigLabel authority/full package tests, benchmark tests, and typecheck pass. Independently adversarially reviewed.

### DIFF
--- a/crates/jazz/src/peer/repair.rs
+++ b/crates/jazz/src/peer/repair.rs
@@ -200,6 +200,13 @@ impl PeerState {
     where
         S: OrderedKvStorage,
     {
+        // SYSTEM is the trusted backend policy subject. Row-policy admission
+        // already bypasses it, so it must not try to hydrate an authorization
+        // support proof: claim and join predicates have no SYSTEM session to
+        // bind and are irrelevant to the bypass decision.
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2472, opened 2026-09-01
fix(browser): preserve authoritative worker relay views

aeplay:
🤖 Fix browser persistent-worker authorization and relay-view correctness at the release-stack tip.

Before:
- ordinary browser worker links could carry delegated-session material intended only for trusted backend transports;
- attached tabs could race auth updates or attach with claims that were not exactly the worker's canonical claims;
- strict Edge subscriptions were sometimes recomputed from an incomplete local overlay, so public unbounded and limit-only queries could publish empty or incomplete resets after relay handoff/reconnect.

After:
- delegated sessions are an explicit transport capability enabled only for authenticated SYSTEM backend/admin links, with receiver-side trust checks retained;
- worker tabs must use semantically identical canonical claims, and auth changes await upstream admission before rebinding tabs;
- only the persistent worker authority-session-owner's Query + Relay + Edge path seeds from the authoritative membership receipt, for bounded and unbounded query shapes;
- Local reads, authorization-support queries, ordinary relays, and stale/ownerless receipts keep their prior behavior.

Examples:
- A BandChat tab reconnecting through its persistent worker receives the room, profile, and messages authorized by the edge instead of an empty locally recomputed reset.
- A limit-only Edge window is rehydrated from the authoritative result just like an offset window.
- Two tabs cannot silently share a worker under different issuer-scoped claims.
- A normal client link cannot acquire backend delegated-session powers.

Verification:
- exact BandChat browser topology seeds 29 and 41;
- relay-authority unit matrix, including unbounded and limit-only queries;
- delegated-session and ordinary-session transport tests;
- reconnect, stale-receipt, local-relay, BandChat permission, BigLabel authority, artifact rebuild, Rust, and TypeScript focused gates;
- independent adversarial review found no release-critical issue.

PR #2473, opened 2026-09-01
Specify scope-isolated client relay semantics

aeplay:
🤖 Defines the authorization and topology contract for scope-isolated client relays before the implementation repair stack.

Before:

- the browser worker was described partly as an unbound `Relay/SYSTEM` hop and partly as the owner of authority-selected result membership;
- ordinary downstream publication and repair could therefore be mistaken for trusted serving and re-run policy without the authority's private support rows;
- durable authoritative result state was keyed only by the application query identity, allowing distinct policy bindings to collide;
- client-side mergeable staging could issue a definitive policy denial from incomplete local evidence.

After:

- a browser or native persistent client worker is explicitly a **scope-isolated client relay**, never an authority;
- its physical upstream hop uses an explicit relay transport capability with no permission subject;
- the server authenticates the requested relay scope, binds its immutable foreground identity and claims, and mints a fresh admission epoch for every attachment, reconnect, or transport resume;
- every foreground Edge/Global query or repair carries only that topology-admitted binding, which edge/core validates and authorizes; raw, mismatched, mutable, or stale-epoch bindings reject;
- relayed mergeable writes are proven at fate authority under the same immutable admitted binding, never under transaction authorship, relay transport, or `SYSTEM`; the terminal proof is non-wire state and claim refresh cannot widen it;
- every worker-to-foreground read, subscription, and repair uses client-local serving;
- Local uses retained authorized knowledge plus optimistic writes; authority removals narrow future Edge/Global membership without retroactively redacting Local data;
- authoritative result state is keyed by canonical query identity plus the exact immutable policy binding, while live settlement receipts remain scoped to the current authority epoch and usage-site subscription;
- same-scope repair is limited to the previously delivered authorized payload closure or same-scope pending authored versions;
- mutation staging remains structural, with definitive read/write-policy decisions at fate authority.

`SYSTEM` is reserved for actual trusted internal principals (core, edge, or an explicitly admitted trusted backend). Relay transport never gains SYSTEM's policy bypass merely by carrying traffic.

Example: when a private membership row authorizes a room at Core, Core may send the room without sending that private evidence. The worker publishes the room to the same-scope foreground from Core's exact authority membership; it does not try to reconstruct the policy locally. A different auth scope cannot attach to the store or reuse that result identity.

Failure example: a relay admitted with `role=viewer` cannot upload an editor-only commit by changing mutable session claims, replaying an earlier binding, reconnecting with an old epoch, or setting `made_by` to an editor. The current server-issued capability is the only authorization input.

This is intentionally spec-first. The following stacked PRs implement the serving context, policy-scoped authority-result state, repair capability, authority-only mutation admission, and topology matrix. An independent adversarial review found and drove removal of the remaining stale Relay/SYSTEM text plus the explicit capability lifetime and relayed-write rules.

Stacked on #2472.


aeplay:
🤖 Independent adversarial re-review passed at `e3e6f918a8`.

The review specifically verified that:

- relay transport is subjectless and never implies `SYSTEM`;
- delegated query/repair bindings are accepted only from a live server-issued scope-relay capability (or the separately documented explicit trusted-backend session path), with mismatched, raw, or stale bindings rejected;
- reconnect and process-local transport resume mint a fresh admission epoch;
- relayed mergeable writes are proven only under the immutable admitted foreground binding, with transaction authorship, persisted state, `SYSTEM`, stale epochs, and mutable claim refresh unable to substitute or widen authority.

No remaining blocker was found; the review checkout was clean.

aeplay:

PR #2476, opened 2026-09-01
Observe native transport terminal outcomes

joeinnes:
## Summary
- Retain native transport terminal futures after connection establishment.
- Route generation-tagged terminal outcomes through the existing tick driver.
- Reconnect recoverable peer failures while preventing reconnects for deliberate owner drops and shutdown races.

## Before
The native client discarded the terminal future returned by the transport connector. An idle socket closure could therefore leave the client treating a dead upstream as an idle connection.

## After
Terminal outcomes are observed through a weak generation-tagged watcher. Peer closures and recoverable failures use the existing reconnect path; deliberate owner drops, shutdown, and stale generations do not reconnect or detach replacement connections.

## Test plan
- [ ] Run the native terminal lifecycle regression tests.
- [ ] Run the Jazz library check with native transport testing features.

### DIFF
--- a/crates/jazz/src/tools/client.rs
+++ b/crates/jazz/src/tools/client.rs
@@ -3563,10 +3563,148 @@ mod tests {
     use crate::groove::storage::{Error as StorageError, StorageFactory, StorageFuture};
     use crate::ids::NodeUuid;
     use crate::tools::AppId;
+    use crate::tools::native_transport_connector::{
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Superseded by #2529, which adapts the native terminal recovery repair to current main. Exact-main reproduction confirmed the bug remained; nine focused lifecycle tests, planted sensitivity, and independent review pass on the replacement. Full CI remains pending. The original branch has not been rewritten.

PR #2477, opened 2026-09-01
fix(sync): admit scope-isolated client relays explicitly

aeplay:
🤖 Implements the server-owned capability boundary specified by #2473 for scope-isolated client relays.

Before:

- relay transport was still partially treated as a permission identity and could drift toward `SYSTEM` semantics;
- a scope relay's admitted foreground claims were copied into mutable subscriber state, so an empty binding or later host claim refresh could bypass/widen terminal write authorization;
- generic multiplexed relays accepted arbitrary delegated bindings;
- process-local transport resume reused the previous admission epoch.

After:

- the WebSocket prelude may request a scope-isolated relay link, but only an authenticated session plus negotiated support lets the server issue the closed capability;
- the capability contains the exact immutable foreground identity/claims and a server-private attachment epoch; relay transport itself has no permission subject;
- Subscribe, repair, and mergeable terminal proof accept only that exact binding; raw `SessionClaims`, arbitrary delegation, empty mutable claims, and generic claim refresh cannot widen it;
- generic/multiplexed relays fail closed on delegated bindings until they have a future explicit per-binding capability;
- terminal admission receives a non-wire proof only after the fate authority hydrates support and evaluates the final write-policy predicate;
- detach/process-local resume preserves the authenticated scope and one-shot identity check but rotates the server-private epoch before resumed work.

Normal example: a browser worker authenticated as Alice requests its scope-relay attachment. It may aggregate Alice's queries upstream and relay Alice's pending commit, while Core evaluates both under Alice's immutable admitted binding.

Failure examples:

- a worker admitted as `viewer` cannot change mutable session claims to `editor` and upload an editor-only commit;
- a generic relay cannot attach an arbitrary `(identity, claims)` snapshot to a query;
- replaying a prior attachment after transport resume cannot reuse its old admission epoch;
- admin/backend links remain their explicit trusted roles rather than silently becoming client relays.

Verification includes server WebSocket admission, hostile delegated-binding, terminal denial, resume rotation, and browser topology suites. Independent adversarial review planted exact-binding and claim-refresh attacks and found no remaining blocker. The three browser failures remaining on this slice are the known exact AuthorityResultState lifecycle dependency repaired by the next stacked PR.

Stack: #2472 → #2473 → #2475 → this PR.


aeplay:
🤖 Independent adversarial review passed on the source implementation at `1c86e557fb`; the published branch is the same reviewed implementation replayed above #2475, with only the backend-auth conflict resolved by retaining both credential normalization and the scope-relay link request.

The reviewer verified exact immutable-binding authorization, raw/mismatched/empty/host-refreshed claim rejection, generic-relay fail-closed behavior, terminal proof timing, authenticated WebSocket admission, and process-local resume epoch rotation. A reversible delegated-binding attack rejected as expected. Focused admission, terminal denial, resume, and WebSocket gates passed; the only three browser-suite failures are the known exact AuthorityResultState lifecycle dependency in the next stack layer.

aeplay:

PR #2478, opened 2026-09-01
Validate invite codes for chat membership

joeinnes:
## Summary
- Require self-membership inserts to prove the target chat is public or the supplied private-chat invite code is current.
- Preserve public-chat membership without a join code.

## Before
A caller could insert a membership row for themselves using only a private chat ID. The caller-controlled join code was stored but never compared with the current chat row.

## After
Private membership inserts require matching chat ID, private visibility, and current join code. Forged, stale, and absent private codes are rejected; public-chat membership remains allowed without a code.

## Test plan
- [ ] Run the chat permission unit tests.
- [ ] Verify forged, stale, absent, and valid private invite codes.
- [ ] Verify public-chat membership without a code.

### DIFF
--- a/examples/chat-react/tests/permissions/permissions.test.ts
+++ b/examples/chat-react/tests/permissions/permissions.test.ts
@@ -53,6 +53,107 @@ describe("chat permissions", () => {
     ]);
   });
 
+  it("allows self-membership with the current private chat join code", async () => {
+    const privateChat = await testApp.seed((db) =>
+      db.insert(app.chats, {
+        name: "Invite-only room",
+        isPublic: false,
+        joinCode: "invite-current",
+      }),
+    );
+
+    const bobDb = testApp.as(externalSession("bob"));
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Small description correction from independent review: against the actual parent `152a62e49841ed29b572ff36d509e6a07b61e5f0`, this PR's current delta is tests only. The invite policy behavior is already present in its base.

Please describe the final change as pinning coverage for valid, forged, stale, absent private invite codes and public joins, so the release reviewer does not infer an additional production authorization change here. No correctness finding in the added cases.

aeplay:
🤖 Anselm approved #2478 in the coordinator conversation on 2026-09-09. Recorded explicitly at his request.

This records his review approval. The PR remains in Joe's post-release stack and will be restacked onto the new main after #2718 is fixed; this comment does not initiate a merge.

PR #2479, opened 2026-09-01
Fix Rust artifact cache dependency coverage

joeinnes:
## Summary
- derive each native artifact task's Rust inputs from its conservative Cargo dependency closure
- validate exact manifests and emit repository-contained, platform-independent cache paths
- make Turbo mutation expectations independent of the production closure helper

## Before

Turbo's manually maintained Rust input globs could omit transitive dependencies. Changes to crates such as `jazz-native-transport` or `benchmark-guard` could therefore leave CLI or NAPI task hashes unchanged and allow stale cache results. The shared closure helper also accepted incorrect manifest filenames and emitted host-native paths without a repository-containment guarantee.

## After

CLI, NAPI, WASM and fast-WASM inputs match their declared non-dev local Cargo dependency closures. Closure paths must come from the exact requested manifest, remain inside the repository and use Turbo's slash-delimited path vocabulary. The behavioural gate now uses independent expected task relationships, while correctness-artifact tasks remain explicitly uncached.

## Test plan
- [x] `node --test dev/artifacts/provenance.test.mjs dev/gates/test/test-artifact-pipeline.test.mjs`
- [x] `node dev/gates/test/turbo-cache-inputs.test.mjs`
- [x] `oxfmt --check` on all changed JavaScript files


aeplay:
🤖 Anselm approved #2479 in the coordinator conversation on 2026-09-09. Recorded for the review history.

The PR remains in Joe's post-release stack; this records review approval without initiating a merge.

PR #2480, opened 2026-09-01
fix(sync): isolate relay authority results by exact policy binding

aeplay:
🤖 Isolates every retained authority result and its lifecycle by the exact immutable policy binding that produced it.

This PR is published in the native stack so CI can validate the composed relay layers. Its exact-authority lifecycle implementation has completed independent adversarial review.

Before:

- two foreground sessions with the same application query but different claims could share durable membership, reset/open/defer state, generations, receipts, or known-state cursors through a `BindingViewKey` lookup;
- ambiguous lookups sometimes returned no receipt (hanging both streams) and sometimes fell back to a global clock (overstating settlement);
- a synthetic relay coverage receiver could materialize from its own key rather than the distinct upstream authority source;
- an explicit authoritative empty result did not always count as a settled receipt.

After:

- `AuthorityResultKey` contains the canonical query/read-view identity plus the exact canonical identity and claims binding;
- a fixed domain-separated digest provides the bounded ordered-store key, while a normal Groove record directory stores and validates the complete collision-free identity; collisions fail closed;
- membership, facts, generations, reset/open/defer state, terminal drains, snapshots, known state, attachment coverage, and publication lifecycle carry the exact key derived from the registered upstream subscription;
- synthetic coverage receivers explicitly retain their distinct upstream source across runtime replacement;
- settled-empty authority results are valid receipts, while disconnect/reset retires live receipt state without confusing durable retained membership;
- native storage fixtures are refreshed for the sole unreleased v1 format—there is no compatibility decoder.

Example: Alice and Bob subscribe to the same room query with different claim snapshots. Alice may receive room A while Bob receives an explicit empty result. Bob's empty settlement cannot clear Alice's reset, inherit Alice's members, reuse Alice's known-state watermark, or leave either subscription waiting on an ambiguous query key.

The local composed browser-relay suite is 31/31. Independent review proved and drove repairs for reset/defer identity collapse, exact handoff/known-state selection, and distinct upstream source association. The final review also proved deferred updates revoke only their exact live receipt until a nondeferred completion restores it; sibling scopes remain independently settled.

Stack: #2472 → #2473 → #2475 → #2477 → this PR.


aeplay:
🤖 Independent adversarial re-review approved the final exact-authority lifecycle repair at 3e0e58569d. It verified exact-key isolation during deferred updates, sibling-scope independence, non-persistence of live settlement, disconnect/ownerless retirement, and planted sensitivity across reset and delta paths. Focused known-state, browser-relay, corpus-reopen, and policy-scoped gates all passed.

aeplay:
🤖 Independent adversarial review passed on the shared Local-plus-authority reconciliation repair. Both the DB maintained-view facade and peer publication now delegate to one LocalAuthorityReconciliation state machine; the duplicate peer reconciliation path was removed. Reversible planted regressions proved same-source generation monotonicity and actual-source provenance reset. Focused evidence: shared corpus 4/4, browser relay 31/31, branch views 21/21, TypeScript tier semantics 23/23, native reopen/write corpus 1/1.

joeinnes:
Validated the BandChat persistent relay regression described by this stack.

Baseline PR #2436 fails with `guest should receive the invited room`; the PR #2480 source tree passes after rebuilding correctness NAPI/WASM artifacts and jazz-tools runtime. Focused evidence:

- `dev/t --test browser_relay_durability band_chat_owner_foreground_receives_guest_message_through_two_scope_relays`: 1 passed
- BandChat browser consumer (`tests/browser/band-chat.test.tsx`): 2 passed
- Solid package and documentation snippets: 6/6 passed

The local validation tree matches this PR exactly apart from an added five-line assertion that the delegated guest cannot see the owner-only room before the invitation arrives.

aeplay:

PR #2487, opened 2026-09-01
Validate auxiliary envelopes before routing

joeinnes:
## Summary
- route complete auxiliary envelopes through the canonical wire admission path before chunk side effects
- retain the exact negotiated protocol, features, and session context shared with each transport adapter
- keep fragments byte-exact for adapter-owned reassembly and avoid per-frame session metadata copies

## Before
The auxiliary pump decoded and consumed complete chunk messages before the canonical adapter validated protocol version, negotiated features, compression, logical size, and session metadata. A malformed envelope could therefore resolve pending chunk work despite violating the connection contract.

## After
Complete auxiliary messages use the same typed admission as ordinary wire messages before any resolver or local-read side effect. Each pump retains its paired adapter's immutable context, outbound auxiliary frames use the same values, fragments remain byte-exact, and public NAPI/WASM signatures and wire bytes are unchanged.

## Test plan
- [ ] Run the wrong-protocol pending-chunk regression
- [ ] Run the paired adapter/pump context regression
- [ ] Run the Jazz library suite with transport compression enabled
- [ ] Run the Jazz, NAPI, and WASM Clippy gate

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -632,8 +632,6 @@ pub struct Transport {
     inner: NapiTransportInner,
     queues: WireQueues,
     auxiliary_pump: jazz::db::PeerIoPump,
-    protocol_version: u16,
-    features: u64,
 }
 
 #[napi(js_name = "Subscription")]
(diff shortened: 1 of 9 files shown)


joeinnes:
Temporary local stack check: I rebased this branch onto #2429 (head `0eeefae123`) without changing the published branch. The previously failing worker-bridge late-runtime rejection test passed, but the original band-chat invited-room failure still reproduced exactly. Please re-run CI after #2429 lands so the remaining failure is evaluated against the stabiliser chain. This comment is not a restack or a claim that the complete suite is already green.

aeplay:
🤖 Superseded by #2524, which adapts this work to the newly merged main and preserves its newer async WASM connection API. Original source commits are credited in that PR; this branch has not been rewritten. Independent review and focused tests pass; full CI remains pending on the replacement.

PR #2488, opened 2026-09-01
Reject ambiguous JOIN ON columns

joeinnes:
🤖 Rejects ambiguous SQL JOIN ON operands instead of resolving each operand opportunistically against a different input. This is the focused port onto current main; historical unrelated stack commits are excluded.

Each operand resolves against the complete visible namespace of both join inputs. Missing or ambiguous references fail deterministically; resolved operands must belong to opposite inputs. Qualified, reversed, uniquely unqualified, multi-key and CTE joins retain their meaning.

For example, joining two tables that both expose id using ON id = id now fails as ambiguous. ON albums.artist_id = artists.id succeeds, as does the reversed equality. A hidden prepared-query parameter named id is not a visible SQL column and does not make artists.id ambiguous inside a prepared CTE join.

Verification: full Groove library suite 689 passed, zero failed, two ignored; integration tests and doctests passed. Independent adversarial review discovered the hidden-binding collision, which is fixed and covered through plan_prepared_shape. Restoring the original resolver makes ambiguity coverage fail; the prepared-CTE regression also failed before its correction. Coordinator and independent reviewer reran the final planner suite successfully. Invariant registry and commit guards pass. Fresh full CI passes at 28b9f87dfd.


joeinnes:
Local stack trial (no branch rewrite or push):

- Rebased this PR's head `039d39c80e57ee6a22b45d8673ed754e3e5ffe24` onto #2414 (`9929e11ec9af992fe0193fef6669e573007dde64`). The rebase was clean.
- Focused verification passed: `cargo test -p groove planner --lib` — 26 passed, 606 filtered.

This confirms compatibility with the #2414 stack and preserves the focused planner regression coverage. It does not replace required CI or establish that the earlier unrelated TypeScript-consumer/browser timeout is fixed. Required CI should be rerun after the stabiliser stack lands.

PR #2491, opened 2026-09-01
Meter shared physical storage classes once

joeinnes:
## Summary
- map logical storage families through the active Groove layout before metering
- deduplicate physical classes and meter each backend class once
- use the aggregate physical meter for Jazz edge-cache budgets

## Before
Jazz enumerated logical per-table history families and metered each independently. Under the class layout every history name maps to the same physical family, so two tables reported twice the actual history bytes and triggered premature eviction.

## After
Groove owns logical-to-physical aggregation behind the concrete Database interface. Empty, unavailable, error, metadata-exclusion, and saturating-sum behaviour is explicit, while Jazz supplies its existing logical family set once. Two logical history tables now contribute the shared physical class exactly once.

## Test plan
- [ ] Run the Groove aggregate metering contracts
- [ ] Run the two-table Jazz edge-budget regression
- [ ] Run the complete Groove and Jazz library suites

### DIFF
--- a/crates/groove/src/db/facade.rs
+++ b/crates/groove/src/db/facade.rs
@@ -428,6 +428,18 @@ impl Database {
         Ok(self.storage.approximate_class_bytes(cf.to_owned()).await?)
     }
 
+    /// Return aggregate approximate live bytes for supplied logical column
+    /// families, counting shared backing classes only once.
+    pub async fn approximate_class_bytes_for_logical_families<'a>(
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Anselm approved #2491 in the coordinator conversation on 2026-09-09. Recorded for the review history.

The PR remains in Joe's post-release stack; this records review approval without initiating a merge.

Issue #2413, opened 2026-08-31
Browser query coverage can stall under concurrent CI load


aeplay:
🤖 Canonical TS CI on #2409 failed the BandChat persistent-browser journey while other browser suites ran concurrently. The owner-side topology receipts passed, then the guest never received the invited room because five local-query coverage waits timed out at `NativeRuntimeAdapter.waitForQueryCoverage`. No policy denial or mutation rejection occurred.

Evidence:

- CI run 33365584596, job 99405429412: BandChat guest wait expired after 15s with five edge-query-coverage timeouts.
- Exact-stack focused BandChat test passed 6 consecutive runs after a correctness-artifact rebuild.
- Exact-stack full BandChat browser suite passed 5/5 with temporary tracing.
- The failure has not yet yielded a causal trace, so we did not increase timeouts or weaken the journey.

Add a trace-gated, production-safe worker event ledger for query attach, coverage request/response, inbound-frame epochs, and settlement. Reproduce under the same concurrent renderer/server load as canonical CI, distinguish scheduling starvation from a lost wakeup/coverage frame, and fix the proven cause. Keep the current end-to-end invitation assertion.

aeplay:
🤖 A second exact-stack canonical TS run reproduced a related worker failure under full concurrency (run 33367258960, job 99410519414): B→A propagation remained empty for 20s, and an unrelated IndexedDB-clear scenario’s `afterEach` cleanup hung for 30s. The same run passed 290 other worker/browser tests. Together with the prior BandChat coverage stalls and lease-admission stalls, this strengthens the hypothesis of a shared worker lifecycle/dispatch problem under concurrent load rather than one example-specific policy issue. A dedicated lane is now replaying the canonical concurrency shape with a trace-gated worker→page event ledger; timeouts and assertions remain unchanged.

aeplay:
🤖 Additional observation while verifying the Inspector clock repair on #2496 (`1cb82818e6`; same runtime source as the measured pre-commit tree): the full Jazz Tools browser suite passed 293 tests but timed out in `SharedWorker bridge with IndexedDB > can update an optional row field to null`.

[...]

Issue #2384, opened 2026-08-30
Unify artifact provenance and cache identities


aeplay:
🤖 We have accumulated several overlapping mechanisms for deciding whether generated artifacts are trustworthy and reusable:

- tracked-input fingerprints for native artifacts;
- sealed producer/consumer admission manifests;
- per-worktree correctness artifact identities;
- staged NAPI/RN package pruning and generation checks;
- cache keys in local gates and CI;
- tracked generated fingerprint files that are dirtied by every correctness-artifact rebuild.

Each guard addressed a real stale-artifact failure, but together they are difficult to explain, costly to maintain, and still produce confusing workflows. Focused browser verification currently requires a roughly two-minute rebuild and then manually restoring three tracked fingerprint files. Different tools also derive identity from subtly different input sets.

Design one coherent artifact identity/provenance model and use it across local gates, CI caches, correctness producers/consumers, native package staging, browser/WASM artifacts, NAPI, and RN.

The design should make these properties explicit:

- which source/toolchain/configuration inputs define an artifact;
- how an artifact carries its own identity and provenance;
- how consumers verify exact compatibility without tracked generated-file churn;
- when artifacts can safely be shared between worktrees or CI jobs;
- how stale or mixed-generation artifacts fail with one actionable diagnostic;
- how cache identity differs, if at all, from correctness/admission identity.

Migrate incrementally without weakening the current fail-closed guarantees. Related immediate work: #2376.

aeplay:
🤖 Another concrete workflow case from the #2496 Inspector repair:

1. The native producer plus a rebuilt `jazz-tools` passed Inspector 12/12.
2. Final Rust test formatting changed the native source fingerprint. Resealing/rebuilding the native pair correctly regenerated the three tracked expected-fingerprint files.
3. I then invoked the focused Inspector browser command without rebuilding `jazz-tools`. That command rebuilds Inspector's embedded bundle but not its `jazz-tools` dependency. The native producer/consumer preflight admitted the fresh native pair, while the previously built JS still embedded the old expected fingerprint. Standalone Inspector subsequently failed with the explicit WASM ABI mismatch diagnostic.

[...]

PR #2496, opened 2026-09-01
Derive remote results through receiver-local IVM

aeplay:
🤖 Establishes the governing model for authority-covered maintained queries, then tracks the test and implementation work needed to make the current stack conform to it.

## Before

The codebase mixed two incompatible models:

- authority updates synchronized row carriers, membership and program facts;
- peer `ViewUpdate`s could also carry authority-produced terminal resets and ordered structural operations;
- the receiver reconciled some authority membership outside its local Groove graph;
- structured authority changes could therefore reach stored rows and authority facts without ever entering the receiver's maintained query;
- applying the authority's collector index directly would be wrong whenever a local-first receiver has eligible pending data that changes the result order.

Example: the authority sees children `[A, C]` and emits “insert C at index 1”, while the receiver also has pending child B and should render `[A, B, C]`.

## After

The specification now requires one model:

1. An authority synchronizes an exact policy-scoped covered input closure: canonical row versions, typed facts, witnesses, completeness and residual-program identity.
2. The receiver stages complete carrier/fact evidence and reconciles it with only the local inputs allowed by the requested tier.
3. Those inputs drive the same receiver-local maintained Groove program used for local changes.
4. The receiver-local collector is the sole producer of application rows, ordering and structural terminal operations.
5. Authority-produced terminal rows, collector positions and structural edits do not cross peer sync as result truth.
6. A fresh strict-remote one-shot owns a transient instance of the same exact coverage and local-evaluation lifecycle, then finalizes it.
7. Settlement follows complete witness admission and local IVM quiescence.

## Tier examples

- **Remote:** waits for its fresh exact authority closure and evaluates only that closure locally.
- **Local-first:** evaluates locally known current data and pending writes for its entire lifetime, while still syncing. Remote scope withdrawal does not evict cached rows.
- **Remote-if-possible:** online, evaluates exact remote inputs with pending edits/deletes to scoped rows and matching new inserts. Existing out-of-scope rows and cached dependencies do not enter merely because of a pending edit. After confirmed explicit disconnect it uses local knowledge; reconnect requires fresh remote scope.
- **Internal local-only:** suppresses propagation but does not select another evaluator.
- **Revocation:** retracting an authority input removes it from remote-scoped results, but is not a stored revocation/deletion. Local-first and offline fallback may retain cached data. Actual authorized deletion versions suppress local reads too; clients do not evaluate read permissions.
- **Reconnect:** a fresh usage site installs and evaluates a fresh exact closure rather than reusing predecessor terminal state.

## Non-goals

- Clients do not become authorization authorities.
- Missing closure evidence is not supplemented from unrelated local history.
- No facade-side semantic scan, authority-terminal mailbox, binding-only fallback or second structured-output reducer is introduced.
- Binding-ABI terminal operations remain valid local Groove-to-application output; this decision concerns the peer sync boundary.

## PR plan

The first commit is intentionally specification-only. Follow-up commits on this same PR will:

- update and strengthen the relevant topology, structured-query, one-shot and permission-transition tests;
- feed exact authority-covered carrier/fact changes into the receiver-local maintained graph;
- ensure source content witnesses are installed before membership promotion;
- unify strict one-shot reads with the transient maintained-coverage lifecycle;
- retire authority terminal output from peer sync;
- preserve exact policy/source/generation isolation and local-first overlay semantics;
- independently adversarially review the final combined behavior and run CI-equivalent gates.

Implementation is in progress. The latest agreed tier/deletion decisions are recorded in SPEC chapters 13 and 16; broader pending dependency expansion is deferred to #2501. The Rust correctness partition passed on `4212943327`, and the subsequent canonical-handoff fix passed all 1778 Jazz library tests (2 existing skips). Browser verification remains in progress: BandChat still has an intermittent invitation-delivery timeout. No full CI-equivalent pass is claimed. See checkpoint comments for exact receipts and the provisional pagination interpretation in #2502, which awaits Anselm’s review. The coordinator is continuing single-track, without subagents.


aeplay:
🤖 Pushing the current implementation checkpoint for early CI and durable visibility; this remains WIP, not merge-ready.

Checkpoint: `308ad262de`.

Now implemented and independently reviewed: receiver-local covered-input evaluation; recursive source witnesses from evaluated steps (including depth/cycle/retraction controls); exact collector occurrence identity and ordering, including mixed remove/move/insert/update batches. Focused regressions pass on the integrated tip.

The last broad default-feature Jazz library run, before the latest recursive/collector repairs, had 1,644 passed, 36 failed, 2 ignored. That is not a CI-equivalent receipt. A fresh broad run and GitHub CI are being checked against this checkpoint; remaining failures are still being repaired. No new merge approval is implied.

WIP checkpoints will continue to be pushed as coherent repairs accumulate, rather than waiting for every gate to turn green.

aeplay:
🤖 Anselm's decision: an active subscription receives an ordered, nonduplicated stream of ViewUpdates; reconnect establishes fresh coverage. We must not invent replay tolerance or wire sequencing to conceal an impossible transition.

If an incremental update removes a fact/version the receiver does not hold, adds an already-present fact, names an unknown compiled source, or leaves conflicting source versions, reject it atomically and surface an actionable error to the client. Do not silently ignore it, automatically reset away the evidence, or leave the client waiting for a timeout. Facts within a single update remain unordered; valid remove-old/add-new replacements and explicit full resets are supported. Late detached-subscription frames retain their separate drop behavior.

The governing decision is recorded in SPEC 8.4.2. Admission and independent-review lanes are implementing/testing it, including client error propagation. This supersedes the earlier proposed idempotent absent-removal behavior.

aeplay:
🤖 Anselm approved removing the special `serve_current_rows(table)` serving API rather than introducing a second current-row subscription lifecycle.

Its canonical key construction conflicts with ordinary opaque usage-site subscription keys. Current repository callers are tests and the durable-stream benchmark. They will use ordinary whole-table subscriptions through the shared covered-input lifecycle instead, preserving resume, traffic-size, delivery, and cleanup assertions. No permissive receiver fallback or compatibility path is authorized by this change.

aeplay:
🤖 Decisions agreed with Anselm after walking through deletion, permission withdrawal, and pending-write examples. These supersede the earlier Local-first bootstrap/scoped-overlay interpretation.

- Local-first uses locally known current data and pending writes throughout its lifetime, while still propagating upstream. Receiving a remote scope does not evict cached rows.
- Remote uses only its exact current authority input scope, without pending writes, and waits offline.
- Online remote-if-possible uses that same exact scope with pending edits/deletes to scoped rows and matching pending new inserts. Merely editing an existing out-of-scope row does not admit it. Joins/includes do not expand into other cached dependencies.
- After an explicit confirmed disconnect, remote-if-possible uses local knowledge. On reconnect it requires fresh remote inputs. Subscriptions follow repeated transitions; a one-shot retains its initial choice. Transport errors/timeouts do not establish offline state.
- Scope withdrawal is not a stored revocation or row deletion. Local-first/offline fallback may show cached data whose access was later revoked. An actually authorized, synchronized deletion changes local current state and suppresses that row locally too.
- A deletion may be disclosed only under the current read policy applied to its retained row preimage and current dependencies. Previous access creates no entitlement to receive later deletions.
- Pending delete/restore/rejection follows the ordinary register winner: delete hides; newer restore shows; rejecting restore reveals delete; rejecting both reveals the accepted row. Clients do not evaluate read permissions.

Example: the current authority scope contains A, the client has cached B, and inserts C locally. Local-first can show A/B/C. Online remote-if-possible can show A/C, but an edit to B alone cannot bring B in. Remote shows A. Offline fallback may show A/B/C again.

Broader online dependency expansion is deliberately deferred to https://github.com/garden-co/jazz/issues/2501 and linked from SPEC chapters 13 and 16. Implementation and tests are being amended personally by the coordinator, single-track; no new subagents. This remains a draft repair, not a claim of green full gates or merge readiness.

aeplay:
🤖 WIP amendment pushed at `11f0212926`; not merge-ready.

Implements the newly agreed separation between ordinary local knowledge, strict remote scope, and online remote scope with a bounded pending overlay. Removes Local-first bootstrap gates and facade-level pending-row reconciliation. TS remote-if-possible subscriptions now follow repeated confirmed offline/online transitions. Specs, docs, and the PR description have been updated.

Verified:
- Focused Rust control covers untouched cached rows, scoped edits, out-of-scope edits, new inserts, pending delete/restore, and rejection. Temporarily removing the out-of-scope exclusion made this control fail at the intended assertion; restoring it passed. The planted change was not committed.
- Existing local insert/claims controls pass. Paired public permission tests show remote revocation while local-first retains cached content. Authority deletion/revocation controls pass.
- TS tier and connection-manager unit suites: 38 passed. These use a mocked native client and are not full topology evidence.
- TypeScript type-check, Jazz Clippy with warnings denied, and docs-site type-check passed.
- Broad Rust library command: `cargo test -p jazz --no-default-features --features testing,transport-compression-zstd --lib --no-fail-fast`: 1,689 passed, 67 failed, 2 ignored. It now completes rather than hanging. This is not a CI-equivalent pass. Nine previously observed failures no longer fail; no newly observed failures against the earlier same-feature timed-out run (not a clean-baseline comparison).

Remaining work includes exact authority registration/receipt paths, direct admission fixtures, maintained-view expectations, and other failures being grouped from the completed run. The pre-existing unlinked open question in SPEC 13 still fails the spec-open-question gate. Full native/browser topology gates and final review are still outstanding. All work in this amendment was performed single-track by the coordinator.

aeplay:
🤖 WIP checkpoint: `c83d3ee9d8` fixes read-tier routing and Local readiness.

Before: a host could register a Global authority view while a one-shot Edge read looked for a different Edge view. Local-first query attachments also unnecessarily waited for authority receipts. After: one-shot reads use the host-selected authority tier, while Local readiness is independent of upstream propagation/receipts. Remote reads still require fresh authority coverage after reconnect.

The membership-revocation regression now checks both remote siblings retract and Local-first siblings retain cached content, as agreed. No revocation records or client permission evaluation were introduced.

Verification: focused routing, structured membership, file materialization, Local readiness, and remote reconnect/empty-coverage controls pass. Full Jazz library command (`--no-default-features --features testing,transport-compression-zstd --lib --no-fail-fast`): **1700 passed, 56 failed, 2 ignored**, versus 1689/67/2 at the previous pushed checkpoint. This is not CI-equivalent or ready to merge. Commit hooks passed Clippy, formatting and the sensitive-data guard. Browser E2E failures remain in the preceding CI run; further repair continues single-track.

aeplay:
🤖 Design decision confirmed by Anselm: identical live usages must share one upstream wire stream within the same admitted scope, rather than receive duplicate predecessor transitions into a shared input relation. Local listeners and one-shot readers independently pin that stream. A fresh remote one-shot requests a refresh and waits for a newer receipt; final detach retires the wire identity. Different policy/session scopes must remain isolated. Strict rejection of impossible transitions remains required.

The reproduced failure was two separate subscriptions both delivering `{A} → {B}` into the same authority input state: the second valid removal of A was rejected against the first subscription's successor. We are fixing stream ownership/coalescing, not weakening removal validation or adding replay tolerance. Implementation is in progress, including refresh, final-detach, reconnect, and relay ownership review.

aeplay:
🤖 WIP checkpoint `b383929174`: identical live local usages now pin one upstream stream. A new remote one-shot refreshes that stream; dropping one pin preserves sibling receipts, and final detach releases the stream. The spec records the approved shared-stream contract.

The new regression verifies two public listeners converge while one-shot readers share the wire identity: a refresh produces one full successor reset, and a later ordinary removal is transmitted once and reaches both listeners. Temporarily restoring the old sibling-stream behavior makes this test fail at the intended wire-identity assertion; restoring the fix makes it pass. Mixed stream/read pinning, final release/reopen, reconnect, delayed authorization, and server protocol-level clone-failure controls pass.

Full Jazz library run (`--no-default-features --features testing,transport-compression-zstd --lib --no-fail-fast`): **1702 passed, 54 failed, 2 ignored**. Hooks passed Clippy, formatting and the sensitive-data guard. Still not green or CI-equivalent. Relay cross-connection stream ownership remains the next implementation step; it has not been claimed fixed by this checkpoint.

aeplay:
🤖 WIP checkpoint: `a5c027339a` follows the shared-relay-stream repair `941daf8d9d`.

The native Jazz library checkpoint is now **1,712 passed / 45 failed / 2 ignored**, down from 54 failures. The exact failure-set comparison resolves nine failures without adding a new failure. This is not a full CI-equivalent or browser pass; the PR remains WIP.

Opening-pending messages now remain progress-only: no source closure replacement, settlement, or persisted coverage receipt. An empty query still opens only after a complete source manifest. Tests cover public rejection of a malformed pending message, a subscriber attempting to impersonate an authority, and independent pending/deferred lifecycles for two policy scopes. SPEC 16 records this distinction.

Older low-level delivery fixtures now explicitly register their receiving subscriptions, and Edge fixtures preserve their original options when requesting subsequent updates. Privacy, deletion, cache retention, and counter assertions remain intact. Malformed source payloads are asserted through the public subscription rejection rather than expecting the runtime tick itself to fail.

Next: remove an obsolete aggregate-only source-selection exception and validate aggregates through the same covered-input graph as other queries. Remaining native failures, browser behavior, and full workspace gates are still under repair. No merge requested.

aeplay:
🤖 Further WIP progress: the native Jazz library checkpoint is now **1,718 passed / 39 failed / 2 ignored**, compared with 54 failures before this batch. Exact failure-set comparison shows fifteen earlier failures resolved and no new failures. Still not a full CI-equivalent or browser pass.

- `b4535b28ff` removes the obsolete aggregate exception that bypassed covered inputs. Aggregates now run over the same exact receiver source closure as other queries. Existing NULL, floating-point, and empty-to-first-insert assertions pass unchanged.
- The recursive coverage repair preserves each source's own authority routing fields until recipient partitioning. It neither strips all policy routes prematurely nor copies unrelated outer-query routes into a recursive seed/step. The existing permission/reconnect/delete controls pass, and a new two-session test confirms Alice/Bob seed rows remain separated across simultaneous usages and Alice refresh.
- Removed the obsolete claim-stripping helper and its helper-only assertion; replaced that coverage with the real two-session publication test. Total test inventory is unchanged. SPEC 16 states the authority-routing versus client-descriptor boundary. Temporary diagnostic instrumentation is removed.
- The previously failing differential oracle passes locally with CI's scenario settings: seed 11, churn depths 10 and 1,000, three interleaving steps (8.22 seconds). This is that focused scenario, not the entire CI gate.

Remaining work includes receiver batch fixtures/admission diagnostics, history/branch assertions, and end-to-end browser validation. The PR remains WIP and unmerged.

aeplay:
🤖 WIP checkpoint `bc119b4342`: native Jazz library is **1,731 passed / 26 failed / 2 ignored**. Compared with the previous 39-failure checkpoint, thirteen exact failures are resolved and none are new. This is not a full CI-equivalent result.

Malformed authority payloads now retain the precise validation reason in the subscription-bound client-visible rejection. Tests still assert exact reasons, no persisted malformed rows, atomic rejection of mixed valid/invalid batches, and unchanged hydration state after failure.

Receiver fixtures now register their actual read views, re-register after reopen, and distinguish reset manifests from incremental input additions. Repeated immutable payload bodies remain supported; repeated input additions are not made valid. Current-state coverage names the winning version while the receiver still ingests both historical bodies. Existing storage, cardinality, and payload-dedup assertions remain intact.

CI on the preceding `400aaeb375` confirms differential oracle, lint, and storage compatibility pass. Browser Todo/BandChat delivery remains red; logs have been fetched for follow-up. No readiness or merge claim yet.

aeplay:
🤖 Native checkpoint: **1,742 passed / 16 failed / 2 ignored** at `04af26265a`. The exact original failure set is down from 39 to 16, with no new existing-test failures; one new regression test passes. Still not a full CI-equivalent result.

`a0c0417896` fixes the simulation/receiver fixtures: fresh registered streams after restart, FIFO nonduplicated ViewUpdate delivery (commit/fate duplication remains tested), and complete reset closures distinct from repeated payload bodies. Both deterministic sync tests pass. Payload conflict/no-persistence assertions remain, as does the explicit absent-removal rejection control.

`04af26265a` fixes a real relay publication bug exposed by browser CI. Before: a Local-first public subscription could receive cached rows incorrectly marked as opening-pending, causing `InvalidAuthoritySourceClosure: opening-pending marker carries source data`. After: both materialization and the send boundary use the existing admission-time requirement to wait for the selected authority. Local-first still propagates, but may publish its cache immediately; strict reads wait.

The new Rust regression uses a non-durable foreground and a public subscription, reproduces that exact error without the fix, and passes with it. It also checks a simultaneous strict subscription stays unsettled before the authority opens and settles afterward. This distinction matters: a plain attached read did not reproduce the public subscription failure.

Building fresh correctness artifacts and running the TypeScript/browser consumers against this exact commit next. No claim yet that browser CI or the full stack is green.

aeplay:
🤖 Exact-commit browser check at `04af26265a` remains red. Fresh release NAPI and fast WASM artifacts were built and provenance-verified; the TypeScript public surface built successfully. The consumer runner exits 1 on both Node/Turbo and browser partitions.

The earlier `opening-pending marker carries source data` error is no longer present in this run. The next reported admission error is `reset source coverage does not exactly match compiled source set`; Todo delivery and the three BandChat browser scenarios still fail. This is not an end-to-end pass.

Continuing by preserving strict admission and exposing expected/received source manifests in the diagnostic. The native foreground regression also passes with both an initially empty relay and a prepopulated relay, including a simultaneous strict read; that alone does not yet reproduce the browser's next mismatch.

aeplay:
🤖 WIP checkpoint: `e7f8178194` fixes a native/WASM source-identity mismatch found by tracing the Todo browser failure.

Before: implicit references used `usize::MAX` in their source alias. Native emitted `include:18446744073709551615:0`, while WASM compiled `include:4294967295:0` for the same query. After: both use a stable `reference:<column>` alias, distinct from explicit includes. The spec now states the cross-platform identity requirement.

The new literal-identity regression was run before the fix (failed with the native sentinel) and after (passed). The focused native relay/include regression also passes. A fresh artifact build and focused browser rerun are in progress; this is not yet a green-browser claim. The previous run additionally reported an empty reset manifest at the foreground, which remains under investigation.

aeplay:
🤖 WIP checkpoint at `36332c8900`: fixed a late-hydration reset losing its retained source facts.

Browser tracing showed a complete worker-to-foreground reset followed by a reset with no source manifest. The late-opening publication path calculated a delta against retained facts, then labeled it a reset. It now emits the complete current closure for that initial reset, while later updates remain deltas.

The focused Rust regression passes, fails specifically on the missing manifest when the fix is disabled, and passes again after restoration. Full native library suite: **1,745 passed, 16 failed, 2 ignored**; exact remaining failure names match the previous 16-failure checkpoint. The portable-source fixture has also been updated. Fresh browser validation is still running, so this does not claim the browser repair is complete. Temporary worker tracing remains until that validation is finished.

aeplay:
🤖 Browser validation after the reset repair, with temporary browser tracing removed (`180c7f4a74`):

- Vanilla Todo: **9/9 browser tests pass**, including persistent two-instance server sync.
- BandChat: **4 passed, 1 failed**. Remaining failure: the guest does not receive a newly invited room. The other UI/topology cases now pass.
- Full TS consumer gate remains red: React Todo's persistent two-instance sync times out (also reproduced in isolation); RecordPlayer rejects during authenticated playlist setup. The gate stops early, so these are not an exhaustive TS failure inventory.

The prior empty-manifest rejection is gone from these runs. Native library checkpoint remains **1,745 passed, 16 failed, 2 ignored** with the exact same remaining failures as before the two runtime fixes. Next work is the remaining scope-change/sync failure and careful migration of stale wire assertions, preserving their correctness checks.

aeplay:
🤖 Native library checkpoint: **1,755 passed, 6 failed, 2 ignored**, down from 1,749/12/2. Exact failure-set comparison shows only the six corrected fixtures disappearing; no new failures.

This batch changes tests, not runtime admission. Lens winner/canonical-payload checks and multi-segment root deltas now inspect covered inputs instead of retired rendered result members. The physical-table-lineage rejection fixture now supplies a real covered input. The bounded-query test still verifies the exact successor row, but distinguishes an authorized deletion witness from reintroducing deleted content. Collector footprint checks require removed rows to leave memory. Alice's cold snapshot now explicitly excludes Bob's private deletion replacement witness, while retaining positive checks for Alice's own deletion and SYSTEM visibility.

Command: `cargo test -p jazz --no-default-features --features testing,transport-compression-zstd --lib --no-fail-fast`. This is a native-library receipt, **not** a full CI-equivalent gate or an all-green claim.

Remaining native failures concern contribution metadata/rejection contracts and branch behavior. Previously reported persistent React Todo, BandChat UI, and RecordPlayer browser/TS failures remain open; this fixture-only batch does not claim to fix them.

aeplay:
🤖 Native checkpoint: **1,761 passed, 1 failed, 2 ignored** (full Jazz library selection, not CI-equivalent). Five prior failures are resolved, one new regression passes, and no new failure names appeared.

Found and fixed an actual transmission bug: `commit_mergeable_unit` and automatic merge publication reconstructed transaction envelopes with `contribution_merge: None`, stripping generated branch-write intent. Both now obtain the envelope from the transaction that was actually published. A regression compares immediate publication with durable replay after reopening storage. Deliberately restoring the metadata loss makes that regression fail; restoring the fix makes it pass. The existing foreign-parent validation test now passes unchanged.

Other fixture corrections retain precise operation-validation errors while asserting the terminal rejected fate, no invalid stored provenance, and no row versions. Calculated-merge assertions include the exact generated branch intent. The recursive policy test still excludes the deleted row from public results and the hidden row from all payloads, but permits the authorized deletion witness without redisclosing its content.

**One semantic decision is open with Anselm:** the existing merge-head test creates the same row ID independently on two offline writers in the same branch and expects both inserts to merge. Current branch-intent specification/admission requires an absent exact head and rejects the second insert. I strengthened this test to check actual acceptance receipts, rather than letting a rejected write hide behind a one-head storage count. No admission behavior was relaxed. Options are to retain strict inserts and exercise merge-head convergence using concurrent updates of a shared seed row, or explicitly change concurrent-insert semantics.

Persistent React Todo, BandChat UI, and RecordPlayer browser/TS failures remain open. Nothing merged; PR remains WIP.

aeplay:
🤖 Anselm's decision: **allow concurrent inserts of the same row ID in the same branch to merge**. Implemented by removing the authority-side absent-head precondition, not by relaxing ordinary policy evaluation. The intent remains parentless and bound to its exact coordinate. Specs 11 and the type documentation now say this explicitly.

The full native library run is **1,763 passed, 0 failed, 2 ignored**. New cases verify an accepted merge with both inserts as parents, rejection when existing-content read permission is denied, and rejection when update permission is denied. The two-branch/reopen regression checks both writers' actual acceptance receipts. This is still **not** a full CI-equivalent or release-ready result.

Anselm also confirmed **edge-first merging**: edges merge concurrent mergeable inserts/edits as they arrive; core merges only the concurrency remaining across edges. Spec 4 records this. Inspection exposed an uncovered implementation gap: core admission invokes merge generation, but edge admission currently only accepts the writes and updates frontier bookkeeping. Its old test did not assert creation of a merge version.

That gap remains open in this repair. Required coverage: same-edge and cross-edge inserts/edits, no redundant core merge after an edge-reconciled publication, replay/reconnect, and Edge acceptance without an accidental Global timestamp. **Anselm approved the forwarding contract:** the edge forwards the admitted transaction together with generated merges as one coherent authority publication, and core admits the complete publication before checking remaining concurrency. This is a logical publication boundary, not dependence on physical packet batching.

Persistent React Todo, BandChat UI, and RecordPlayer browser/TS failures also remain open. Temporary gated React subscription tracing is on the branch to continue that independent investigation.

aeplay:
🤖 Edge-local checkpoint: the shared merge generator now runs during edge admission and accepts its locally generated merge only at **Edge** durability, after persistence, without allocating a global timestamp. The focused regression covers concurrent inserts and edits, exact parent identities, edge authorship, Edge-only fate, and replay. It failed before the change with two unmerged heads and passes afterward.

Full native library result: **1,764 passed, 0 failed, 2 ignored**. This is not a complete release/CI result.

**Still incomplete and not merge-ready:** the approved coherent edge-authority publication forwarding contract. Edge-generated merges must travel with the original admitted write, survive retry/reconnect, and reach core before core considers residual cross-edge concurrency. The current transport/outbox has not yet been amended for that group; cross-edge and full-topology proof are still required. Browser/TS repairs also remain open.

aeplay:
🤖 Coherent authority publication — node-level WIP checkpoint.

Anselm confirmed that edges forward an admitted write together with their generated merges as one logical authority publication. Core admits the complete publication before reconciling any remaining cross-edge concurrency. The governing contract is recorded in SPEC chapter 4.

Implemented the node-level boundary using the existing transaction admission and merge machinery, with reconciliation deferred until all publication members are present. Publication reconstruction walks accepted, not-yet-globally-acknowledged history and stops at acknowledged ancestors; it does not introduce a new persistent publication format.

Concrete regression coverage, for both concurrent inserts and updates:

- Same edge: admit A and B, create edge M(A,B), restart the edge, reconstruct and replay the publication at core. Core retains the edge's merge identity rather than creating its own.
- Different edges: E1 admits A and E2 admits B. Core admits E1, restarts, then admits E2 and creates the residual cross-edge merge. Replaying both publications does not generate another merge.
- Test sensitivity: deliberately dropping the edge merge makes the same-edge test fail directly at the no-new-core-publication assertion. The planted change was restored.

Verification: native library suite **1765 passed, 0 failed, 2 ignored**; full workspace compile covering libraries, binaries, tests, examples and benches passed with the CI-required feature selection. The first library run encountered two RocksDB fixture-open errors at the process's 1024-file soft limit; the complete rerun passed with a process-local limit of 8192. No test behavior was changed for that resource error.

**Not merge-ready:** this checkpoint is not wired into network forwarding. Authenticated authority-only admission, outbox/dependency repair/reconnect preservation of the group, wire encoding and corpus coverage, and malformed-group admission handling remain unfinished. Existing browser/TypeScript failures also remain open. Native library coverage is not a CI-equivalent or end-to-end forwarding receipt.

aeplay:
🤖 Coherent edge-publication checkpoint (still WIP; not merge-ready).

Before: the edge could admit A and B and produce merge M(A,B), but forwarding individual writes allowed core to merge A and B prematurely. A restart could also lose the process-local upload grouping.

Now: an authenticated edge sends the complete publication through a negotiated wire message. Core validates every member first, persists the canonical transactions together in one existing Groove batch, and only then reconciles remaining concurrency from other edges. Transactions retain separate IDs, fates, and global sequence numbers; this is not a new application transaction or durable journal format. The wire shape and feature bit are specified and pinned by the byte corpus.

Recovery reconstructs uploads from accepted Edge-durable history without requiring the original clients to return. An anchor-only acknowledgement does not discard the remaining publication: all members must receive eligible global receipts. Ordinary clients, relays, and backend credentials do not acquire authority-publication capability merely by claiming SYSTEM.

Verification:

- Native library: **1,766 passed, 0 failed, 2 ignored**.
- Wire fixtures: **11 passed**, including rejection of authority-produced result members.
- Server credential/capability test: **1 passed**.
- Full workspace compile including tests, examples, and benches: passed.
- Cancellation/reopen regression: never observes a persisted publication prefix. A temporary split-persistence mutation failed with exactly that symptom; the mutation was removed.
- Same-edge inserts/updates preserve the edge merge; cross-edge concurrency still produces the necessary core merge.

These are scoped receipts, not a full CI-equivalent pass. Browser/TypeScript repairs remain outstanding, and publication sizing for very large unacknowledged backlogs still needs a bounded recovery strategy.

aeplay:
🤖 Pushed a browser-relay repair checkpoint. Still draft; not a claim of full CI-equivalent green.

Before: a persistent browser worker could send an incidental admin credential alongside its user JWT, changing its server-admitted role. Separately, local and remote reads opened duplicate authority streams despite lowering to the same authority request. Their shared predecessor state then correctly rejected the duplicate additions, leaving a second foreground waiting.

After: the browser relay authenticates as its user session. Identical lowered authority requests share one upstream stream, with independent downstream connection/read-mode pins. Closing a remote reader preserves its local sibling; claims refresh retains the same sharing rule. Strict malformed-update rejection and distinct tier semantics are unchanged. Specs and a two-relay/multiple-reader regression document this.

Verification before final formatting and runtime-only feature gating: native Jazz library 1,767 passed / 0 failed / 2 ignored; exhaustive workspace compile passed; auth unit tests 7/7; Todo full suite 9/9, including the formerly failing two-client browser flow. Commit hooks, including Clippy, passed on the committed source.

Remaining measured failures: BandChat 4/5 (invited room not delivered); RecordPlayer 8/9 (now reaches accepted invitations, then fails an authority read during the offline streamed-audio scenario). Continuing those individually. Also identified a relay rejection-forwarding gap that can turn protocol errors into foreground timeouts; this must be addressed without weakening validation.

aeplay:
🤖 Second checkpoint: `fcf8e70a75`.

Two independently reproduced defects are repaired:

- A scope-isolated relay's local-first maintained query was opened in TrustedServing mode. Core correctly admitted a newly invited room, but the relay independently filtered it using its incomplete permission-support cache. Relay query opening and capability preflight now select ClientLocal; core/edge retain TrustedServing. The existing compiler/evaluator path is reused.
- After an empty read retired, a later read could reload only its durable settlement cursor and mark a source closure claimed without loading its manifest/facts. Removed that partial recovery path. Full startup recovery remains; retired usages await fresh authority closure.

Both new Rust regressions failed before their fixes and passed after. Combined native library: 1,769 passed, 0 failed, 2 ignored. Exhaustive workspace compile passed. Fresh browser artifacts: BandChat 5/5; Todo 9/9. Commit hooks passed. No full CI-equivalent claim.

RecordPlayer remains 8/9, but now passes repeated empty reads, invitation acceptance, streamed-audio upload and readback. Its next assertion exposes an older spec/test conflict: a local-first offset-8 query over cached remote positions 8–27 yields 16–17, while the older retained-window rule expects 8–9. Asked Anselm whether to preserve literal local-first semantics and use remote-if-possible for authority-relative app pagination, or retain the older local page-reuse rule. Leaving that decision unchanged while continuing independent error-forwarding work.

aeplay:
🤖 Pushed `bc4dcd3859`: upstream registration failures and invalid covered-input updates now reach every foreground reader sharing that relay stream, rather than appearing as indefinite loading. An unrelated query remains live.

Verification: 1,770 native library tests passed, 0 failed, 2 ignored; exhaustive workspace compile passed. The new regression fails when rejection forwarding is deliberately disabled, and passes after restoring it.

This is still WIP, not an all-green candidate. The broader TS run found 22 failures (1,830 passed), including missing WASM backend-read ABI methods, inspector diagnostics, browser-only type checking, and topology/lifecycle failures. Rust CI also found integration-target failures outside the native library suite. I am repairing these personally without changing the agreed read-tier semantics. Pagination semantics remain the separately raised question in #2502.

aeplay:
🤖 Pushed `c434e3818a`: WASM now implements the same explicitly guarded backend-read capabilities as NAPI. Backend opens can read and subscribe through their authority identity, including transaction and relation reads; ordinary opens cannot select that authority. This fills a missing binding boundary rather than falling back to an ordinary client read.

The focused TS batch passes **152/152** tests, including the transaction suite, all eight backend read/attachment/subscription entrypoints, inspector diagnostics, auth mapping, and real compiler validation of generated migrations. Removing a backend guard deliberately makes the new denial regression fail; the guard is restored. Workspace compile and commit-hook Clippy passed. One intermediate CLI migration-chain timeout passed both isolated and on the final full focused rerun without changing its timeout.

Also excluded runtime-generated foreground identity lease files from Git: tests were creating these files and accidentally invalidating the source-identity receipt for later consumers.

Still WIP: joined-query compilation (`schema_version` missing in a graph), topology/lifecycle integration failures, and the pagination question #2502 remain. Continuing those repairs on this PR; this is not an all-green release claim.

aeplay:
🤖 Published checkpoint `d8aeec81f4` (still draft, not ready to merge).

Fixed joined-query visibility resolving against the wrong record shape: a flat joined result is not a complete stored-version record. Witness selection now uses the exact visible row ID and transaction ID to select the normalized version witness, for both physical and covered-input queries. It does not widen the authorized input set. All four existing output-occurrence regressions pass unchanged, including joined-row deletion and add/remove within one transaction.

Current receipts: native Jazz library **1770 passed, 0 failed, 2 ignored**; output-occurrence suite **4/4**; codec-registry receipts **5/5**; server claim derivation regression **1/1**. The full local Node suite is down to **3 failed, 1851 passed**, plus one worker-process error. Browser public-flow suite is **11 passed, 1 failed**: the immediate local read after persistent reopen returns null. RecordPlayer pagination remains the separate open semantics question #2502. These are partial receipts, not a green CI-equivalent gate.

Continuing repairs without weakening validation or converting local reads to remote reads to hide failures.

aeplay:
🤖 Published `7356c5d422`: repaired existing direct-node crash/FIFO fixtures to register receiving queries before accepting their covered-input updates.

The four-tier fixtures also reused the same canonical whole-table handle for both SYSTEM upstream scope and user downstream scope. They now use the existing ordinary query APIs with separate fixture usage handles and an initial reset publication, followed by incremental updates. This is test setup only; no production validation was relaxed and no data/settlement/FIFO assertions changed.

Receipts after formatting: fate replay **3/3**, four-tier **13/13**, threaded FIFO **1/1**; commit hooks (Clippy, formatting, sensitive-data guard) passed.

CI on preceding checkpoint `d8aeec81f4`: lint and storage compatibility passed. Jazz Tools Node now has **2 failed, 1852 passed**, plus 1 expected failure and 9 skipped: reverse schema migration read and policy-graph holder read both time out waiting for coverage. The locally observed websocket timeout did not recur in that CI run. BandChat browser **5/5**; RecordPlayer **8/9**, with the same pagination decision in #2502 outstanding. Other browser inventory is not yet established green.

Continuing with catalogue/lens failures in the Rust E2E suite, while retaining the separate browser reopen/readiness investigation.

aeplay:
🤖 Checkpoint `862bc03314`: covered-input witnesses now preserve authored table names across schema lenses.

Example: the query calls a table `people`, while a row was originally written as `users`. Previously the authority shipped the canonical `users` payload but a source fact claiming a `people` payload; strict validation correctly rejected the mismatch. The fact now keeps logical source `people` and immutable version coordinate `users`, connected by the catalogue's permanent physical table ID. Initial, incremental, and resident-body lookup share the exact physical-table/row/transaction/branch/layer checks. The spec now states this distinction explicitly.

Verification:
- Native library: **1770 passed, 0 failed, 2 ignored**.
- Exhaustive workspace library/binary/test/example/benchmark compile: passed.
- Catalogue E2E: reduced **18 passed / 8 failed** to **24 passed / 2 failed**. A final confirmation run explicitly excluded the two remaining cases and passed **24/24**; it is not a full green catalogue receipt.
- Renamed known-state fetch/repair passed. Its expected request now names the authored table, with a stronger assertion that the logical source remains under its renamed identity. A temporary wrong-name mutation made the regression fail; removed exactly and restored test passed.
- Commit hooks passed; no temporary mutation remains.

Remaining catalogue cases: a joined projection asks for missing field `user_id`; an unpublished-schema test expects waiting rather than the explicit pending-catalogue rejection currently propagated. Continuing investigation without changing that readiness expectation.

PR remains draft/red. Latest preceding CI still had two Node coverage failures (migration and policy-graph holder) and the RecordPlayer pagination question #2502. Browser local persistent-reopen readiness remains separately under investigation.

aeplay:
🤖 Local repair checkpoint: `100872e807` fixes reference traversal after a flat joined result. The checkpoint is committed locally; I am letting the preceding CI Rust workspace run finish before pushing it, to preserve its complete failure report.

Example: a v2 reader joins `posts` to `people`, where the author row was originally written under the v1 name `users`. After projecting a joined result, the traversal incorrectly tried to read source-only fields from that joined tuple. It now selects the exact visible source row/version (including any prepared-binding route) before following references. The public joined tuple and its occurrence identities remain unchanged; this uses the existing Groove graph path, not an alternate evaluator.

Measured locally:
- Jazz native library: 1,770 passed, 0 failed, 2 ignored.
- Catalogue integration: 25 passed, 0 failed, 1 explicitly filtered pending the unpublished-schema readiness decision. Previously 24 of these passed.
- Join output occurrence/delta integration: 4 passed.
- Exhaustive workspace lib/bin/test/example/bench compile: passed.
- Commit hooks, including clippy and the sensitive-data guard: passed.

No test expectations were weakened. Fresh WASM/NAPI correctness artifacts are building for the next browser/Node pass. The PR remains draft and red; these are partial receipts, not a full CI-equivalent pass.

aeplay:
🤖 Repair checkpoint `9d82925228`: independently projected provenance fields now decode correctly through the existing public query path.

Before: `select(["$createdBy"])` received its row and settled remote coverage, then public decoding failed because it demanded `$createdAt`, `$updatedBy`, and `$updatedAt` too. This surfaced as repeated query errors/timeouts in local-first authentication and authorship-policy tests.

After: a shared CurrentRow accessor locates and type-checks the requested provenance field. Full-provenance access uses the same field-location rules. The public decoder no longer needs a backend lookup for these already-materialized values. There is no storage-format change, permission relaxation, or fallback to SYSTEM.

Verification:
- Jazz native library: 1,770 passed, 0 failed, 2 ignored.
- Local-first auth integration: 7 passed, including a new real-server test selecting each of the four provenance fields independently.
- The new test fails when the old complete-provenance requirement is deliberately reinstated; that planted fault was removed and the test rerun successfully.
- All five non-ignored authorship-policy tests pass, including the three that timed out in the preceding CI run.
- Exhaustive workspace lib/bin/test/example/bench compile and commit hooks pass.

Remaining failures are not hidden: the broader policy run had one array-membership delta failure (passed on focused rerun, so needs timing investigation); required array correlation still reports a root-layout mismatch; browser public-flow remains 11/12 with the immediate local-read-after-reopen issue. The filtered Node migration case still times out. The PR remains draft/red, and these are partial receipts rather than a CI-equivalent pass.

aeplay:
🤖 Checkpoint `55f90017f5` fixes the required-array-reference publication failure without changing its visibility rule.

Example: a group references Alice and a not-yet-created Bob. The group stays hidden until Bob exists. Once both members exist, the runtime proves the group's `member_ids` field non-null. The prepared output layout conservatively keeps that sibling field nullable. Previously, that harmless root-field refinement also rejected edits to the separate `members` collection, even though the collection's complete layout was identical.

The adapter now preserves the prepared root descriptor and accepts nested edits only when the addressed collection's entire subtree layout is unchanged. Child payload bytes, keys, paths, and ordering are preserved. Unknown paths and changed child layouts remain rejected. Root-level record re-encoding is unchanged.

Receipts:
- Jazz native library: 1,772 passed, 0 failed, 2 ignored.
- Query integration excluding the six pagination cases: 25 passed, 0 failed, 12 existing ignores.
- New positive and negative descriptor-boundary tests; restoring the old blanket nested-edit rejection makes the positive test fail, and the restored fix passes.
- Exhaustive workspace compilation and commit hooks pass. This checkpoint also gates the new provenance accessor with its caller's `runtime` feature for no-runtime builds.

The full PR is still red. Fresh bindings retain the two Node migration/resource-holder failures and browser public-flow's reopen failure (11/12 pass). The include-scaling canary was stopped after exceeding five minutes; stack sampling identifies repeated full-parent decode/re-encode while applying individual child edits during initial hydration, rather than a connection wait. I have raised the shared lazy-output-accumulator direction with Anselm before proceeding on that larger repair.

aeplay:
🤖 WIP repair checkpoint `345945356e`: preserve JSON/storage cell types through maintained collection.

Concrete failure: a remote whole-row query over a JSON column was rejected by the server because its prepared layout called the field a plain string, while the graph emitted a JSON storage cell. The TypeScript one-shot attachment then hid that rejection behind a coverage timeout. A new Rust integration test reproduces the rejection without WASM.

The fix keeps the existing storage-derived descriptor for ordinary current-row sources. Explicit projections remove only the row-presence wrapper; they retain the column's actual type until normal public materialization. No storage format, permissions, or read-tier semantics change. No layout validation was weakened.

Verification:
- New real-server whole-row/explicit-projection JSON regression fails before the fix and passes after it; JSON suite: 10 passed.
- Catalogue/migration: 26 passed, with the unresolved catalogue-admission case excluded. Also added passing cold, old-schema, ID-filtered array-migration coverage.
- Query integration: 25 passed, 12 existing ignores, 6 pagination cases excluded pending the recorded semantics question.
- Native library: 1,772 passed, 0 failed, 2 existing ignores.
- Exhaustive workspace compile passed. Commit hooks passed.

Not a full green gate: fresh TypeScript artifacts/tests are next. The lost one-shot rejection is a separately identified repair still to do. Other previously recorded failures and design questions remain open; this checkpoint is not merge-ready.

aeplay:
🤖 WIP checkpoint: old-schema reads of newly written rows no longer time out during deletion-input compilation.

Before: policy dependency preparation projected through the requesting client's schema view, but the IncludeDeleted source lookup reconstructed a different view. The server rejected the query with `policy dependency was not prepared before source composition`; the JS one-shot wait hid that rejection behind a coverage timeout.

After: both IncludeDeleted composition branches use the same existing read-view projection helper as dependency preparation. Current policies still apply; no authorization check is bypassed and no second query path is introduced.

Example: an old client reads `todos(title, done)` while a newer client writes `todos(title, done, tags)`. The old client now receives title/done through its lens, even when the maintained query also compiles deletion inputs.

Verification on the clean implementation (temporary instrumentation removed):
- Both existing TypeScript migration directions pass, including the previously failing old-client/new-row case.
- Jazz library: 1,772 passed, 0 failed, 2 ignored.
- Catalogue integration: 26 passed, with the previously identified pending-catalogue semantic case excluded; JSON storage: 10 passed.
- Browser public-flow gate: 11/12 pass; persistent-reopen readiness still fails.
- Full Node suite is NOT green: 1,853 passed, one websocket boundary-test timeout, plus a worker abort from native query-attachment reentry. The latter is now reproduced with a backtrace and is the next repair. An initial stale built-JS failure disappeared after rebuilding the consumer package.

The separate lost-rejection reporting defect remains to fix. This is a repair checkpoint, not a readiness or merge claim.

aeplay:
🤖 WIP checkpoint: the complete Jazz Tools Node suite now passes locally: **1,876 passed, 2 expected failures, 9 skipped**, across 144 passing files and one skipped file. No native worker aborts. This is not a full CI-equivalent or overall PR-green claim.

The native crash was reproducible in the existing Better Auth test `backend access can update and delete despite deny-all client policies`. `await waitForCoreIdle()` checked the node, then yielded to its caller; a queued tick could acquire the node in that gap before synchronous backend query attachment. The shared idle helper now runs the synchronous callback in the same continuation as its last check. Peer admission and all coverage probes use that boundary too. No new query path or read-tier semantics.

A deterministic scheduler regression queues a tick immediately before admission. It passes with the repair and fails with `admission reentered a suspended tick` when the old microtask gap is deliberately restored. That planted mutation was removed before the final run.

The other commit only optimizes an existing websocket test assertion: the two-megabyte carrier boundary still checks exact frame count and every payload byte, using byte equality instead of general indexed-property deep equality. No timeout increase or production transport change.

Additional receipts:
- Jazz Tools package build/type checks pass.
- Better Auth + migrations + websocket focused run: 59 passed, one expected failure.
- Browser public-flow gate remains 11/12 passing (persistent-reopen readiness).
- The preceding schema-projection checkpoint also fixes the Rust prepared-claim-routing regression and mixed-generation history case. Fresh full Rust CI is now **3,450 passed, 19 failed, 1 timed out, 73 skipped**, down from 20 failures and 2 timeouts on the previous head. The claim-routing case was independently rerun locally and passed.

Remaining work includes the known registration fixtures, unresolved pagination/reopen readiness semantics, the include canary, and surfacing rejected one-shot registrations as errors instead of coverage timeouts. Nothing merged.

aeplay:
🤖 Repair checkpoint: `7dcbd74254` and `80ccd6303d`.

An enum-valued prepared query could fail while opening its coverage terminal: Jazz kept the enum value but inferred an integer descriptor for the projected binding literal. The lowering now preserves the existing binding descriptor through Groove's typed-literal API. No storage/wire encoding or tier semantics changed. A regression covers three enum variants, initially empty/populated results, and updates entering the query. Deliberately restoring the old untyped projection makes it fail with the original `RecordEncoding(TypeMismatch { expected: U8 })` error.

Subscription opening now preserves the underlying Groove evaluation error instead of reporting it as a disconnected receiver.

The cold-history, sync, and validation benchmark fixtures now explicitly register ordinary whole-table subscriptions before delivery. Cold-history still asserts bounded reads and exactly one current winner at depths 1, 10, 1000, and 1025. Sync still asserts the exact permission-filtered remote result; it no longer mistakes retained local cache data for a remote-scope leak. SaaS simulation closure accounting now reads covered inputs instead of the retired authority result-member list; client/core query equivalence assertions are unchanged.

Verification:
- Native library: **1773 passed, 0 failed, 2 ignored**.
- Benchmark correctness smoke: **4 passed, 1 failed** (cold-history, sync, validation repaired; route curve still retains version witnesses unexpectedly, while all its result/delta checks pass).
- Simulator suite: **5 passed, 5 failed**; SaaS and its reconnect pass. Remaining scenarios still expose unregistered-subscription fixture paths and are under repair.
- GitHub CI on preceding `79bc03b879` independently confirms the full Node suite: **1876 passed, 2 expected failures, 9 skipped**. Browser BandChat and the previously reported Rust integration failures remain; this is not an all-green claim.

Tooling observation: preserving the original subscription error and printing the route benchmark's individual receipt fields immediately separated runtime defects from obsolete fixture assumptions. Both diagnostics are now permanent rather than lane-local logs.

aeplay:
🤖 Browser/local-owner checkpoint: `7e4d7161ac`.

Before: reopen a persistent browser database, immediately read a saved todo with `tier: "local"`, and receive `null` from the new foreground's empty memory before the worker supplied its stored rows.

After: the memory-only foreground waits for its persistent owner's answer to that query, including a confirmed empty answer. This uses the existing query-attachment receipt; it does not wait for edge/core authority coverage. A standalone durable runtime still reads its local storage immediately, and requested remote propagation remains enabled. Strict remote reads still require current authority evidence. SPEC 16 now states this local-storage-delivery versus remote-settlement distinction.

Tests:
- New Rust foreground/relay regression fails before the fix and passes after: cached and empty results, no authority connection, strict remote remains unready.
- The older authority-continuity test now exercises a standalone durable client; the new test explicitly covers the memory-only foreground instead of assuming empty foreground memory is the complete local store.
- Native library: **1774 passed, 0 failed, 2 ignored**.
- Exhaustive workspace compile (`--lib --bins --tests --examples --benches`, CI features): passed.
- Native adapter suite: **132 passed, 1 skipped**, including a new case where unrelated worker traffic cannot acknowledge this query.
- Full Node suite: **1877 passed, 2 expected failures, 9 skipped**.
- Browser public-flow suite: **12 passed**, including persistent reopen.
- BandChat browser suites: **5 passed**, including invitations, offline replay, and revoked-write checks.

These are focused/batched receipts, not a full CI-equivalent pass. GitHub CI on the preceding checkpoint independently reduced Rust failures from 19 to **15 failures plus one timeout** (3455 passing). Its TS job now reaches RecordPlayer's retained-page pagination discrepancy (#2502). The local Rust pagination reproduction still has 4 failing cases and is the next investigation; no expected page contents have been weakened.

aeplay:
🤖 Pushed `0ac0502b94`: cold remote pagination now uses the exact requested source shape before registration, rather than depending on whether RegisterShape has already reached the local registry.

Example: authority OFFSET 2 LIMIT 1 over A,B,C,D delivers C. Previously the cold receiver could apply OFFSET 2 again and publish an empty page. It now evaluates the ordinary ordered window relative to those inputs (OFFSET 0, retaining LIMIT 1). Local-first still applies its absolute offset to ordinary local current inputs. The spec records this distinction; no authority result snapshots or prefix over-fetching were added.

Local receipts: native library **1,775 passed / 0 failed / 2 ignored**; public query tests **31 passed / 0 failed / 12 ignored**; edge-server tests **12 passed / 0 failed / 1 ignored**; permission-filtered pagination passed. The new compiler regression catches a deliberately planted broken shape-identity comparison; plant removed. Existing public pagination assertions are unchanged.

The strict-remote overlay fixture also passes after explicitly setting Deferred local updates, matching public remote lowering. It previously inherited Immediate and thus requested the pending overlay that its assertions deliberately prohibit.

Not all green: the preceding CI run had 15 failures and one timeout; seven of those failures now have passing local receipts. Remaining work includes simulator fixture ports to ordinary registered queries, the retained-page/local-first question, catalogue admission, the incremental tree canary, and the timing-sensitive BandChat invitation flow (local pass, CI fail). Continuing single-track; nothing merged.

aeplay:
🤖 Checkpoint: `c623c1789f` fixes a shared-collector hydration defect and ports the canvas simulator to explicitly registered, scoped query subscriptions.

Example: Alice opens a subscription, then Bob opens the same query after the disposable output cache has been evicted. Previously the second hydration could add the complete snapshot to already-populated collector state. Updating a row then retracted only one copy, leaving a phantom old value and no public subscription update. Hydration now replaces collector state from its complete snapshot. Normal incremental ticks remain incremental; no new query path or polling workaround was added. The contract is recorded in Groove SPEC chapter 4.

Verification:
- Groove library: 648 passed, 0 failed, 2 ignored.
- Jazz library with canonical testing features: 1775 passed, 0 failed, 2 ignored.
- New forced-cache-eviction regression passes for three subscribers, update, and deletion. Disabling the fix makes it fail; the planted mutation was removed.
- Complete canvas scenario passes, including concurrency and failure scenarios, with original public-Db polling and assertions unchanged.
- Simulator suite: 6 passed, 4 failed. Remaining failures are S3/S4/S5/S9's obsolete unregistered current-row fixture setup; repairs continue.

This is still a red/WIP checkpoint, not full CI-equivalent readiness. Existing remaining integration/browser failures are not being claimed fixed. Pre-commit clippy, formatting, and sensitive-data checks passed; the committed worktree is clean.

aeplay:
🤖 Simulator repair checkpoint: `11842ad22b`.

All ten scenario smokes now pass (10 passed, 0 failed), together with all 15 jazz-sim library tests. This is a focused receipt, not a full CI-equivalent pass.

Before: permissions, durable streams, order processing, and durable execution injected row updates without registering their query receivers, or addressed a Db using a manufactured canonical handle. Those updates fail the stricter covered-input admission contract.

After: direct-peer fixtures register the exact query and session; Db fixtures open ordinary query attachments and serve the actual handles emitted by their transport. Upstream and downstream uses remain distinct. Transaction messages are preserved for the existing admission driver. No production receiver checks were relaxed.

The scenario assertions remain intact: grants/revocations, exclusive order-processing schedules, competing durable-workflow advances, gap-free events, and restart/readback. A temporary planted omission of downstream query delivery made durable execution fail immediately; the plant was removed and the full simulator batch rerun successfully.

Still working on the separate native/browser failures. The local-first pagination semantic conflict remains tracked in #2502; this checkpoint does not change that decision or claim the complete PR is green.

aeplay:
🤖 Two more repairs are pushed on #2496:

- `beea272d46`: a valid subscription against a schema draft stays pending until catalogue activation, rather than receiving a permanent rejection. The connection preserves the admitted session snapshot, cancels parked work on unsubscribe/close, and bounds retention. The spec now describes this behavior. Bob's existing query resumes after the migration lens is published without a retry; withholding activation resumption made the strengthened regression fail as intended.
- `eaa93b1f6d`: ordinary local-first queries may materialize exact result payloads from local storage instead of retaining every source payload per subscription. Strict remote receivers keep their coverage witnesses, including during cold opening.

Focused receipts: catalogue integration 27 passed; Jazz library 1,775 passed, 0 failed, 2 existing ignores; route-subscription correctness/memory smoke passed. The full Rust CI partition is running locally. These are not a complete CI-equivalent pass.

The latest completed TypeScript job still passed all 1,877 Node tests and stopped at the same RecordPlayer offset assertion tracked in #2502. No expected result was changed. Separately investigating the nested-child allocation canary: one child edit currently performs accumulated parent-snapshot decoding/encoding, which must become incremental.

aeplay:
🤖 Incremental terminal-tree repair checkpoint (still WIP, not ready to merge).

Before: hydrating a parent with 5,000 children repeatedly decoded/encoded its accumulated child array. The unchanged allocation canary timed out after 180 seconds. A parent scalar replacement could also clear unchanged children from the retained snapshot.

After: the maintained receiver and facade use one shared decoded-tree reducer. Child edits update addressed state; complete snapshots encode at explicit snapshot boundaries. Same-batch replacement of an existing parent preserves independently maintained children; actual removal drops them. Ordinary array-valued columns remain ordinary scalar replacements. No storage or wire encoding changes.

Example: Alice has an edited todo. Renaming Alice so she moves after Bob preserves that todo; subsequent todo edits remain visible; deleting the todo empties Alice's collection without changing Bob's; deleting Alice leaves Bob intact. The new regression exercises these transitions against the same live receiver's retained snapshot. Disabling decoded snapshot materialization makes it fail (`first` instead of `edited`); that planted mutation was restored.

Verification:
- Full Rust workspace partition: **3,474 passed, one failed, 73 skipped; no timeouts**. The sole failure is the already tracked local-first pagination decision in #2502. This is not a fully green or full CI-equivalent result.
- The original 250-versus-5,000-child canary passes with unchanged sizes/allocation limits: 7.10 seconds focused; 12.301 seconds under the concurrent full workspace run.
- All three incremental-delivery canaries passed together (10.67 seconds) before the final parent-replacement correction; the final workspace run also passed those canaries.
- SPEC 16 records the decoder and batch-lifetime contract, and clarifies that these are local terminal operations—not authority result transport.

Next: rebuild the exact committed WASM/NAPI artifacts and validate browser consumers. Pagination expectations remain unchanged pending the decision in #2502.

aeplay:
🤖 Exact-commit binding validation for `f92a8651a765a60d0a60be298756ad59e81cdf4f`:

- Fresh release NAPI + fast WASM produced and their sealed provenance verified.
- Full Node/Turbo consumer partition: **27 tasks successful**. Jazz Tools' Node suite: **1,877 passed, two expected failures, nine skipped**.
- BandChat browser: **five passed**.
- Focused alpha public-flow browser gate, run separately: **12 passed**.
- RecordPlayer browser: **eight passed, one failed**, at the same #2502 pagination assertion (expected positions 8–9, received 16–17). The scenario stops there, so later offline/reconnect phases of that scenario are not claimed as passing. The parallel browser command is not fully green; the focused gate was rerun independently because a sibling failure can stop that command early.

Generated local fingerprint changes were restored after the consumers finished; the checkout is clean. No new runtime failure was observed in these runs. GitHub CI is still running; storage compatibility has passed. The remaining known Rust/browser disagreement still needs the pagination decision in #2502 before changing its expectations or implementation.

aeplay:
🤖 Two follow-up repairs after the decoded-tree checkpoint:

1. **Nested scalar updates:** Groove's recursive differencer can address an update at `[collection, child-key]` (and deeper alternating paths). The decoder previously rejected that valid ending, logged a refresh error, and left the public snapshot stale. It now accepts the keyed scalar-update case only when the path key, edit key, and decoded row key agree. It reuses the same retained-record updater, preserving independently maintained descendants. Example: rename Alice's todo, edit its comment, rename Alice, then delete the comment; each change is visible in the same live receiver and unchanged descendants survive. The regression also checks the todo's new title. Removing the keyed-update case makes the test fail on the stale title; the plant was restored. SPEC 16 documents the distinction between a collection-membership path and a keyed scalar path.

2. **Inherited-policy test readiness:** a busy full run exposed three setup races. These tests waited 100 ms for their initial subscription and then assumed it had arrived. One counted a late initial snapshot as a forbidden reorder delta; two expected removal from streams that had not yet shown the row. They now wait for the expected initial rows plus settled state before mutating. All existing reorder/revocation/retarget assertions remain unchanged.

Final local verification on the combined changes:
- Rust workspace partition: **3,475 passed, one failed, 73 skipped**, no timeouts. The sole failure is still #2502's pagination choice.
- Policy suite: **59 active tests passed**, including all three formerly racing setups; 47 existing tests skipped.
- Separate CI benchmark correctness smokes: **5 core + 10 simulator scenarios passed**. These were run explicitly because the known workspace failure otherwise prevents that gate's later smoke stage from executing.

Still WIP, not fully green or claimed CI-equivalent. No pagination assertion or semantics changed. The preceding `f92a8651a7` CI finished with exactly the same sole Rust and RecordPlayer pagination failure; lint, storage compatibility, and Rust differential jobs passed.

aeplay:
🤖 Pushed checkpoint `4212943327` (still draft, not ready to merge).

The full local workspace Rust run now reports **3,476 passed, zero failed, 73 existing skips**. Its subsequent deterministic benchmark scenarios are still running. RecordPlayer's focused unit suite passes 5/5.

Pagination interpretation is explicitly recorded in [#2502](https://github.com/garden-co/jazz/issues/2502#issuecomment-5526316518) for Anselm review: Local offsets count cached rows; authority-relative online paging uses remote/remote-if-possible. With cached positions 8–27, Local offset 8/limit 2 yields 16–17. The Rust regression preserves remote window/shift assertions and now checks local offsets before and after shift/detach. RecordPlayer opts into remote-if-possible without changing its expected online page.

The previous head's CI revealed a separate BandChat browser failure: the guest never receives the invited room (the same scenario passed locally). This is still under investigation; no timing-only explanation or green-browser claim. Fresh exact-head artifacts and browser validation follow.

Historical DB API status notes were preserved in [#2514](https://github.com/garden-co/jazz/issues/2514) and replaced by a link; the offline spec-question gate now passes.

aeplay:
🤖 WIP checkpoint `cec6fc54e5`: recover stale canonical query handles during late-usage handoff.

Before: if catalogue/runtime invalidation removed a canonical query handle just as a second usage arrived, cloning cleared the stale handle and then tried to drain the missing handle forever. Ordinary updates already had a reopening path; cloning bypassed it.

After: both paths share canonical query recovery, retaining the admitted read/policy context. Existing subscribers receive the recovered complete input closure before the new usage receives its reset. The spec now states that lifecycle requirement.

The new regression `peer::tests::duplicate_usage_reopens_stale_canonical_query_before_cloning` failed before the fix (no canonical reconciliation) and passes afterward, checking both existing and new usages. The Jazz library suite passed **1778 tests, 0 failures, 2 existing skips** on the behavior change; an earlier invocation without the usual raised descriptor limit hit four RocksDB “Too many open files” setup errors. Commit hooks also passed. This is not a full CI-equivalent receipt.

Also aligned RecordPlayer's revocation scenario with the agreed deletion distinction: the actually deleted invitation disappears locally, while the undeleted cached playlist and child remain readable locally. Its focused end-to-end scenario passes, including persistent reopen.

**Still red/WIP:** BandChat passed its browser file without instrumentation, then the invitation flow passed three isolated repeats and timed out on the fourth. The stale-clone defect is fixed, but does not fully explain the intermittent browser failure. Continuing instrumentation and diagnosis single-track; not ready to merge. All temporary tracing was removed from this commit.

aeplay:
🤖 Repair checkpoint: exclusive worker writes and branch-aware large-value browser reads now pass.

Before/after:

- An exclusive todo/check/note bundle arriving through a scope-isolated worker skipped the immutable admitted-session permission proof and was rejected. Both transaction kinds now use that same proof; exclusive read-set validation remains additional. A forbidden sibling still rejects the entire bundle, with no authority rows admitted.
- A worker could re-encode a branched document's version witness with an empty physical branch and its own zero settlement clock. It now preserves the supplying branch and selected authority receipt's cut. Initial and incremental input materialization share one encoder.
- A new query usage could receive a full source manifest mislabeled as incremental because its payload cursor was current. It now gets a complete reset; known bodies remain deduplicated. Warm-reader, reopen, and missing-body repair tests retain their data/payload assertions.

Validation:

- Jazz library: 1,778 passed, 0 failed, 2 existing ignores (repository CI's existing 4 MiB test-thread setting). An earlier default-stack run aborted with a stack overflow; it is not counted as passing.
- Browser relay topology integration: 39 passed, 0 failed. New coverage includes allowed/forbidden exclusive bundles, branch isolation, late attachment, exact forwarded settlement cut, and incremental branch edits. Planted permission bypass and dropped branch witness each failed their corresponding new regression before restoration.
- Clean rebuilt browser bindings: both affected Jazz Tools files passed, 4/4 tests, including the automatic large-value multi-row transaction and concurrent exclusive sibling queries.
- Spec open-question and invariant-registry checks passed.

Still WIP: the focused Inspector test still shows an empty grid; the earlier intermittent BandChat failure needs repeated clean confirmation; malformed closures on one-shot attachments can still become timeouts rather than immediate useful errors. Continuing these repairs personally, without subagents. This is not a full CI-equivalent green receipt. The provisional pagination interpretation tracked in #2502 is not newly approved by this checkpoint.

aeplay:
🤖 Repair checkpoint pushed as `4c205d24a3` (no history rewrite).

Before: a trusted backend reading at Local could repeatedly rebuild its maintained graph from propagation resets; Inspector briefly showed rows and then emptied. After: the read tier, not the authorization host, chooses the input frontier. Local keeps its storage-backed graph and still propagates. A real-peer Rust regression fails before the fix (15 spurious events) and passes after it.

Also: malformed authority closures now reach one-shot owners as explicit errors instead of disappearing into a timeout. The fast-resume benchmark now preserves the agreed distinction between mandatory new-usage input manifests and deduplicated known row bodies.

Verified on this checkpoint: Rust workspace 3,482 passed, 73 existing skips; the complete local Rust CI partition including deterministic benchmark scenarios passed; exact ignored-test inventory passed; spec links and invariant registry passed; Inspector browser suite 12/12 passed using freshly produced, provenance-checked NAPI/WASM artifacts.

Still WIP: focused Jazz Tools browser run has 16 failures (two restart-order tests and 14 worker-bridge tests), 73 passes, one expected failure, two skips. Most worker failures explicitly report a tab/worker session-claims mismatch. Investigating that next without relaxing the scope-isolation check. This is not a claim that the full TS gate is green.

aeplay:
🤖 Checkpoint `64c64d758b` is pushed. The previously failing worker/restart browser group is now green: 89 passed, one expected failure, two existing skips. The final queued-auth-error refinement additionally passed all 49 session/worker unit tests and the real-browser auth-recovery regression.

Two changes:

1. Provisional decision for Anselm's morning review, recorded in #2518 and SPEC chapter 7: `jazz_pub_key` remains verified self-signed proof metadata, not an application policy claim. Client/server projections agree; anonymous tabs can reuse their documented shared persistent scope. Verified issuer/subject authorship, signature/audience/expiry checks, external provider claims, and the worker's exact policy-claim rejection are retained.
2. A rejected token refresh broadcasts auth failure but keeps the foreground attached, so a sibling tab can restore the shared session. Error classification is scoped to the serialized refresh attempt. A new negative test failed against the initial fix and now proves an earlier auth rejection cannot hide a later unrelated runtime failure.

The Rust workspace partition passed again (3,482 tests, 73 existing skips), including deterministic benchmark smokes and partition coverage. Spec/invariant/ignored inventories passed. Commit hooks passed, and the committed worktree was clean before push. This is a visible WIP checkpoint, not a request to merge the provisional decision without review.

Now running the complete TypeScript consumer gate on the exact committed source and regenerated sealed artifacts. The previous CI's actual Rust jobs passed; its outer aggregate failed because the TypeScript job was cancelled after reaching the buffered consumer phase. No full CI-equivalent or full-TS success is claimed yet.

aeplay:
🤖 Harness repair pushed as `8da14a9902`.

The full Node/Turbo partition passed on the preceding checkpoint (including Jazz Tools' 1,880 main tests and its React/Svelte/Solid suites). The parallel browser partition instead failed before RecordPlayer could run: another Vitest project had already bound its hard-coded strict port 63318.

The alleged registry reservation was only a static uniqueness check; default-port probing by another process could still take it. RecordPlayer now uses Vite's normal EADDRINUSE retry behavior. The brittle uniqueness lint has been replaced with a retry-configuration check. No tests are skipped and no application behavior changes.

Planted contention receipt: held 63318 open with an independent TCP listener, observed RecordPlayer retry, and completed all 9 real topology tests. Thirteen harness contract/selection tests pass. Full TypeScript consumer rerun is now in progress on the exact committed tip.

aeplay:
🤖 Pushed f5377fa431: remove supplemental local-only subscription seed reads, fabricated empty openings, and the framework empty-result refresh. The native maintained stream now owns the initial result and subsequent deltas; the spec and regression tests record that contract.

Verification before diagnostic cleanup: Node suite 1,883 passed (2 expected failures, 9 skipped); subscription browser suite 27 passed; BandChat suite 5 passed. Planted fabricated-opening and supplemental-read regressions both failed as intended and were removed. A traced BandChat repetition batch passed 30/30 runs.

Still WIP: an earlier repeated BandChat run failed on iteration 14 while the full Node suite ran concurrently (owner did not display invited member, no seed timeout). Contention is a hypothesis, not an established cause. Two existing Rust deferred-persistence tests remain unresolved around retained persistence versus tick cancellation/progress. Clean post-diagnostic verification is being restarted against this committed SHA, with CI requested in parallel. Not claiming the complete stack green.

aeplay:
🤖 Inspector save/reload repair: the acknowledged-but-invisible edit was a clock-ordering defect, not a CI-only timeout.

Before: bulk snapshot ingestion installed the seeded rows without advancing the receiver's transaction-clock high-water. The queued binding commit also used an internal counter instead of a host clock sample. In the reproduced browser failure, `done: false` was encoded correctly and globally accepted, but its transaction timestamp was 1 ms after the Unix epoch. Last-write-wins correctly retained the seeded `done: true` version stamped today.

After: bulk snapshots advance the same high-water as incremental ingestion, and WASM/NAPI queued commits pass the current Unix-millisecond sample to the shared reservation path. Row provenance timestamps remain distinct. For example, even a client whose clock lags the snapshot it just read now authors its next edit after that observed version. No Inspector runtime workaround, supplemental seed query, tier change, or permission change was added.

Local verification:

- Inspector browser suite: 12 passed, including the original edit/reload reproduction with a stronger save-completion assertion.
- Rust library via the CI test launcher: 1,782 passed, 2 existing skips.
- Exhaustive Rust workspace compile (`--lib --bins --tests --examples --benches`, CI features): passed.
- Full Node suite: 1,883 passed, 2 existing expected failures, 9 existing skips.
- New queued-commit regression covers mergeable and exclusive transactions, reservation before staging runs, unchanged returned identity, separate row provenance, and a subsequent backward clock sample.
- Regression sensitivity: the bulk-ingestion test failed before the fix; separately restoring the bad counter behavior in mergeable and exclusive reservations made the new queued test fail. All planted changes and temporary tracing were removed.

The broader Jazz Tools browser run had 293 passes and one coverage timeout in `SharedWorker bridge with IndexedDB > can update an optional row field to null`. That exact case passed all three isolated reruns; that does not establish that the full-suite timeout is fixed. Inspector then passed all 12 tests again.

This remains a WIP repair PR, not a claim that the entire stack is green. The two previously reported deferred-persistence tests and the intermittent browser coverage timeout remain separate outstanding work.

aeplay:
🤖 Final pushed tip: `0d7a4d2bbb` (the `1cb82818e6` clock repair plus refreshed generated binding fingerprints).

Rebuilt the matching native artifacts and JavaScript dependency, then verified the exact committed tip with a clean worktree:

- Inspector: **12/12 passed**, including edit, local confirmation, frame reload, and persisted readback.
- BandChat browser journeys: **5/5 passed**.
- The queued-clock Rust regression also passes on the committed repair; Clippy and formatting passed at commit.

The separate intermittent SharedWorker query-coverage timeout is recorded on #2413; the focused artifact dependency/build-order friction is recorded on #2384. No timeout or tier semantics were weakened. The PR remains draft pending the other known repairs and full tip CI.

aeplay:
🤖 Anselm approved the persistence ownership clarification: cancelling a `tick()` waiter must not abandon an accepted write. The runtime retains and resumes the same storage operation; actual failure/abandonment after submission remains fail-closed.

Updated the two failing deferred-persistence regressions and the transaction spec accordingly. No production runtime behavior changed.

Examples now exercised:

- Admit insert → delete → restore while storage is paused; cancel the tick twice; accept another write; resume without starting the paused operation again; drain and reopen to verify the restored value and following write survived in order.
- Start an earlier write and pause a later update on cold reads. Drive bounded owner turns and open an independent view of the durable backing to prove the earlier write persisted while the later row is still unchanged. Resume reads, assert the runtime-owned wakeup fires, and verify the later update reaches Local durability with the expected value.

The focused suite passes all 15 tests. Deliberately dropping a retained persistence continuation and separately omitting earlier-publication polling each made the relevant regression fail. Both planted changes were restored exactly. Underlying Groove tests remain responsible for actual atomic-write abandonment poisoning; the invariant registry now reflects that distinction.

Checkpoint `bf6e5ad7d0`: the full workspace test run passed **3,485 tests, 73 skipped**, including the underlying Groove abandonment/poisoning regressions. The deterministic benchmark correctness smokes and all four Nextest partition-coverage self-tests passed too: the complete local `rust-workspace` CI partition exited 0. Generated native artifact production/provenance, Clippy, formatting, and the sensitive-data guard passed. This is one CI partition, not a full CI-equivalent result. The pushed tip should receive fresh CI.

aeplay:

PR #2498, opened 2026-09-02
Clean up staged N-API uploads on close

joeinnes:
## Summary
- Abort staged streaming uploads when N-API streams, views, or databases close or are dropped.
- Serialise cleanup drains and preserve deterministic cleanup and close-error semantics.

## Before
Dropped initialised streaming uploads could retain staged large-value claims, and wrapper or database close could discard shared state before cleanup ran.

## After
Binding-private owner and view lifecycle code queues non-blocking drop cleanup. Close admits and drains queued cleanup before core close. Concurrent drains are FIFO, cleanup failures are bounded and attributed, and explicit abort remains idempotent.

## Test plan
- [ ] Run the nine targeted N-API lifecycle regressions.
- [ ] Run `dev/t close_fails_before_clean_marker_when_rejection_acknowledgement_fails`.

The `jazz-napi` patch changeset is included.

### DIFF
--- a/.changeset/streaming-upload-cleanup.md
+++ b/.changeset/streaming-upload-cleanup.md
@@ -1,0 +1,5 @@
+---
+"jazz-napi": patch
+---
+
+Clean up staged streaming uploads when N-API streams, views, or databases are closed or dropped.
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -4902,6 +4902,63 @@ mod tests {
     use serde_json::json;
     use std::cell::RefCell;
 
+    /// An initialized raw streaming handle must relinquish its durable
+    /// retainer when JavaScript drops it. The zero-age maintenance pass is
+    /// deliberately run only after a binding tick, so this distinguishes
+    /// asynchronous Drop cleanup from the TTL backstop.
+    #[test]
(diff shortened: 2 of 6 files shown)


aeplay:
🤖 Nonblocking hardening finding from independent review: the private cleanup drain bookkeeping is not cancellation-safe.

At `crates/jazz-napi/src/lib.rs:680–709`, the active/waiter counters are manually changed around awaits. A synthetic test polls a `PendingNativeRead` wrapping `drain_cleanups` until the existing test gate suspends it, then drops that pending operation. The drain-active flag remains set, so `cleanup_drain_blocks_tick` never clears. An RAII guard would make this private path robust to future cancellation callers; abandoned close completion also deserves an explicit policy.

This is **not a demonstrated public `NapiDb.close()` regression**: its `completePollableClose` wrapper retains and polls the future even when a caller stops awaiting the JS promise. The normal concurrent drain test passed. I would not block this PR or the release on the synthetic private-cancellation case, but wanted the finding recorded for future lifecycle changes. No production change was made during review.

aeplay:
🤖 Anselm approved #2498 in the coordinator task. The nonblocking private cancellation hardening follow-up is tracked as technical debt in #2720. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2499, opened 2026-09-02
Bound native prepared-query cache retention

joeinnes:
## Summary
- Bound native prepared-query retention with lease-aware eviction and shutdown cleanup.

## Before
Full literal-byte query keys and native prepared objects were retained for the runtime lifetime.

## After
A 256-entry/1 MiB inactive LRU is pinned by active read, subscription, and coverage leases. Acquisition closes safely after runtime shutdown, and failed coverage detaches without retaining the prepared query.

## Test plan
- [ ] Run `pnpm --filter jazz-tools check`.
- [ ] Run the focused prepared-query Vitest tests.

The `jazz-tools` patch changeset is included.

### DIFF
--- a/.changeset/bounded-prepared-query-cache.md
+++ b/.changeset/bounded-prepared-query-cache.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Bound native prepared-query retention with lease-aware eviction and cleanup during runtime shutdown.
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Anselm approved #2499 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2504, opened 2026-09-03
Allow atomic durable unique-index replacement

joeinnes:
## Summary
- Consolidate unique-index ownership changes by durable key and record within one tick.
- Permit an atomic owner replacement only when the previous owner is fully retracted and at most one positive owner remains.
- Align `INV-STORAGE-23` and `INV-TICK-13` and register the replacement regression.

## Before
A batch that deleted the current unique-index owner and inserted its replacement under the same key was rejected against the still-visible durable owner.

## After
The persistence seam resolves final net ownership before its atomic write. Genuine competing owners still fail closed. The common one-record accumulator remains inline and does not allocate a hash table per key.

## Test plan
- `cargo test -p groove durable_unique_indices_allow_atomic_replacement_within_one_batch --lib` — 1 passed, 649 filtered.


aeplay:
🤖 Superseded by #2533, which preserves this work and combines it with the related unique-index persistence/registration repair on current main. #2533 has been independently adversarially reviewed, approved, and merged.

PR #2505, opened 2026-09-03
Transform included and hopped relation rows

joeinnes:
🤖 ## Summary

- Carry each table's column transforms through root, included, and hopped query results.
- Apply partial selection before transforms.
- Share a schema-aware, collision-free include output-name resolver between planning and row materialisation.

## Before

Included and hopped rows could expose raw stored values, and a source column sharing a relation or hidden-include name could overwrite or impersonate the materialised relation.

For example, if `items.score` is stored as `3` but its application transform exposes `30`, a root item read returned `30` while `orders.include("item")` and `orders.hopTo("item")` could expose the raw `3`.

## After

Every materialised row uses its owning table's transforms. The same example returns `30` through direct reads, includes, and hops. Include aliases avoid all schema-column and relation-name collisions while preserving public relation property names.

This PR defines current QueryBuilder behavior only. It does not add a compatibility contract for older internal QueryBuilder object shapes.

## Test plan

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts tests/ts-dsl/transformed-columns.test.ts`
- exact combined stack CI at `21c85a861567134023514a3322670e84446e12e0`


aeplay (approved):

aeplay (approved):

PR #2506, opened 2026-09-03
Apply relation column aliases during lowering

joeinnes:
## Summary
- Lower CTE and table column aliases as positional projections.
- Normalise compatible `UNION ALL` arms to one logical field layout before graph construction.
- Preserve prepared-query binding fields, reject real output-name collisions, and reserve the internal binding relation name.

## Before
Relation column aliases were not consistently represented in the graph, and union arms with compatible types but different field names could not be combined. Initial aliasing also risked stripping or spoofing hidden prepared-query fields.

## After
Authored aliases apply only to public fields; internal binding fields retain their identity and routing role. Every same-name public field is checked for parameter collisions, and `__bindings` cannot be authored as a relation name or alias.

## Test plan
- `cargo test -p groove preserves_hidden_binding_fields_through_cte_aliases --lib` — passed.
- `cargo test -p groove rejects_public_parameter_name_collisions_after_a_preserved_cte_binding_field --lib` — passed.
- `cargo test -p groove rejects_the_reserved_prepared_binding_relation_name --lib` — passed.
- `cargo test -p groove applies_cte_column_aliases_to_union_all_graphs --lib` — passed.

### DIFF
--- a/crates/groove/src/ivm/planner.rs
+++ b/crates/groove/src/ivm/planner.rs
@@ -353,6 +353,15 @@ impl<'a> Planner<'a> {
         match table_ref {
             TableRef::Named { name, alias } => {
                 let table_name = single_name(&name.0)?;
+                if table_name == BINDING_QUALIFIER
+                    || alias
+                        .as_ref()
+                        .is_some_and(|alias| alias.name == BINDING_QUALIFIER)
+                {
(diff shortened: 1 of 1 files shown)


aeplay (approved):

aeplay (approved):

PR #2507, opened 2026-09-03
Apply lens renames in operation order

joeinnes:
## Summary
- Apply table and column rename operations in declared lens order.
- Preserve the intermediate names needed by later operations in the same lens.
- Cover mixed table/column rename sequences.

## Before
Grouping transformations by kind could apply a later rename against the wrong intermediate table or column identity.

## After
Lens translation follows the operation stream, so chained and mixed renames observe the state produced by every preceding operation.

## Test plan
- `cargo test -p jazz lens_translate_table_and_column_applies_mixed_renames_in_operation_order --lib` — 1 passed, 1712 filtered.

### DIFF
--- a/crates/jazz/src/tools/schema_lens/lens.rs
+++ b/crates/jazz/src/tools/schema_lens/lens.rs
@@ -534,6 +534,114 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


aeplay (approved):

aeplay (approved):

PR #2508, opened 2026-09-03
Bound WebSocket connection establishment

joeinnes:
🤖 Bounds the complete WebSocket connection-establishment sequence with one deadline.

## Before

The transport could wait indefinitely when a peer accepted TCP but stalled the HTTP WebSocket upgrade. Treating each later handshake phase as a fresh timeout would also let total establishment time exceed the configured deadline.

For example, a peer could spend almost the entire deadline completing the HTTP upgrade and then receive another full deadline to withhold its Jazz server Hello.

## After

One deadline begins before `connect_async` and covers:

1. TCP/HTTP WebSocket upgrade;
2. client prelude and Hello sends;
3. receipt of the server Hello.

A stalled establishment reports `HandshakeTimeout`. Immediate TCP refusal remains a `Connect` error. No message pump is installed before the handshake succeeds, and owner drop or shutdown cancels establishment without a later reconnect resurrecting it.

This changes neither the wire format nor successful connection behavior.

## Verification

- A peer that accepts TCP and never upgrades times out.
- A peer that spends three seconds upgrading and then withholds Hello still exhausts the original five-second deadline; resetting the Hello timeout makes this receipt fail.
- Immediate TCP refusal, successful handshake, retry-later admission, owner-drop cancellation, and server shutdown remain covered.
- `cargo test -p jazz-native-transport`: 12 unit tests and 15 integration tests passed.
- Independent adversarial review found no remaining correctness blocker.
- Canonical CI is green.


aeplay (approved):
🤖 Approved by Anselm at `976736b43d7c9cc210dac10f0cd3a560cd9be792` after the direct port, shared-deadline receipt, independent adversarial review, and green canonical CI.

PR #2511, opened 2026-09-03
Cancel stale Inspector connection submissions

joeinnes:
## Summary
- Associate each connection-form submission with its own cancellation lifetime.
- Ignore schema-hash completion from a superseded or unmounted submission.
- Keep an already-open connection stable while the cancelled edit unwinds.

## Before
A slow schema-hash lookup could complete after cancellation and replace or close connection state that belonged to a newer UI lifetime.

## After
Only the current live submission may publish connection state; stale asynchronous completions are inert.

## Test plan
- `pnpm --filter inspector exec vitest run --config vitest.config.ts src/App.test.tsx` — 5 passed.


aeplay:
🤖 Anselm approved #2511 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2513, opened 2026-09-03
Close active SSE streams during shutdown

joeinnes:
## Summary
- Enter draining state before authentication or body parsing and reject new todo/SSE requests with 503.
- Track and end active SSE responses before waiting for HTTP and Jazz shutdown.
- Prevent late broadcast writes and preserve admitted-request/persistence ordering.

## Before
Active SSE responses could keep HTTP shutdown open, and requests crossing shutdown could continue into authentication or mutation work.

## After
Existing streams observe EOF promptly, new requests fail at the admission gate, and Jazz shutdown still runs after HTTP close outcomes.

## Test plan
- `pnpm --filter todo-server-ts test` — 2 files, 29 tests passed.

### DIFF
--- a/examples/todo-server-ts/src/main.ts
+++ b/examples/todo-server-ts/src/main.ts
@@ -61,6 +61,13 @@ export interface TodoServerOptions {
   backendSecret?: string;
   adminSecret?: string;
 }
+interface ServerLifecycle {
+  draining: boolean;
+  beginDrain: () => void;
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved #2513 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2515, opened 2026-09-03
Keep example backend credentials server-only

joeinnes:
## Summary
- Keep generated backend credentials in the server process and make managed startup cleanup retry-safe.

## Before
Example configuration could expose privileged admission material, and a failed managed startup could retain resources needed by a retry.

## After
Client-visible configuration contains no backend secret; failed startup releases its resources without changing successful startup behaviour.

## Test plan
- [ ] Run the config-security example test and focused `jazz-tools` Next development tests.

### DIFF
--- a/.changeset/quiet-secrets-hide.md
+++ b/.changeset/quiet-secrets-hide.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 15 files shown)


aeplay:
🤖 Superseded by #2525: current-main adaptation with the original changes and independent-review fixes for explicit blank-secret handling and the security test's TypeScript diagnostic API. Focused tests/typechecks and independent review pass; full CI remains pending. The original branch has not been rewritten.

PR #2524, opened 2026-09-03
Validate auxiliary envelopes on the current main

aeplay:
🤖 Adapts #2487 to the newly merged main, preserving Joe's admission fix and the newer async WASM connection path. Independent review found no correctness blocker; this remains a draft pending full CI and Anselm's review.

## Before / after

Previously an auxiliary chunk response could affect pending chunk work before the regular transport adapter checked its negotiated protocol, features, and session metadata. Both paths now use the same canonical complete-envelope admission checks before routing effects.

For example, a response carrying the correct chunk request ID but the wrong protocol version must not resolve that request. A correctly admitted response still resolves it. Fragments remain byte-exact for the adapter's existing reassembly path.

The adapter and auxiliary pump share immutable connection metadata, also used for outbound auxiliary frames. Low-level Rust `PeerIoPump` methods therefore no longer accept independently supplied protocol/features/session arguments. This is an intentional Rust signature change under the new-core no-backcompat policy; the user-facing TS/NAPI/WASM API is unchanged. No new storage or wire encoding is introduced.

## Verification so far

- Wrong-protocol pending-chunk regression: passed.
- Paired-context incoming/outgoing regression: passed.
- Existing auxiliary completion regression: passed.
- `cargo check -p jazz-wasm -p jazz-napi`: passed.
- Independent review: no correctness blocker. Fragment admission and compressed-backpressure tests passed; a planted metadata-validation bypass made the new regression fail, then was restored.
- Coordinator reran the wrong-protocol regression successfully on the clean committed tree.
- Full CI remains pending.

Source work: `f61076e96f` and `7dd64846b0` from #2487; adapted commit `cdf3b2a815`. The original published branch has not been rewritten.

### DIFF
--- a/crates/jazz-napi/src/lib.rs
+++ b/crates/jazz-napi/src/lib.rs
@@ -655,8 +655,6 @@ pub struct Transport {
     inner: NapiTransportInner,
     queues: WireQueues,
     auxiliary_pump: jazz::db::PeerIoPump,
-    protocol_version: u16,
-    features: u64,
 }
 
 #[napi(js_name = "Subscription")]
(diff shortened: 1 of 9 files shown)


aeplay:
🤖 Anselm approved #2524 in the orchestration conversation after independent review and successful CI. Approval applies to `cdf3b2a815`.

PR #2525, opened 2026-09-03
Keep example backend credentials server-only on current main

aeplay:
🤖 Keeps privileged backend credentials server-only in the Next integration and example, with retry-safe managed-runtime cleanup. Supersedes #2515.

Previously, returned Next env configuration could expose backend credentials to client bundling. The example now obtains trusted credentials through its server-only module; client configuration contains public connection settings only. Explicitly supplying an empty backend secret remains invalid rather than silently selecting an ambient or generated secret.

Failed managed-runtime cleanup resets retained handles/configuration and restores an injected environment secret only while it still owns that value, allowing later startup to retry. Normal successful startup and authentication semantics are preserved.

The E2E fixture generates a fresh per-run backend secret in the web-server environment. Its previous empty values correctly triggered the newly preserved blank-secret rejection and prevented startup; the fixture now exercises valid server setup without weakening validation.

Validation: 39 focused Next/managed-runtime tests, configuration-security test, actual admitted Next E2E, example typecheck, and sealed artifact/Tools builds passed. Independent prior review included planted credential exposure, blank fallback and cleanup regressions. Fresh full CI is running at c77f935bc1.

### DIFF
--- a/.changeset/quiet-secrets-hide.md
+++ b/.changeset/quiet-secrets-hide.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Keep managed development backend credentials in the server process instead of returned Next.js configuration. Applications using `jazz-tools/dev/next` should read `BACKEND_SECRET` from the server environment.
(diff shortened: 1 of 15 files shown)


aeplay:
🤖 Anselm approved #2525 in the orchestration conversation, pending successful CI. Approval applies to the reviewed adaptation at `7ad5f9a77c`. Superseded #2515 is already closed and links here. No merge has been performed.

PR #2526, opened 2026-09-03
Reject blank configured server secrets on current main

aeplay:
🤖 Supersedes #2484 with a current-main adaptation and stronger edge-boundary tests. Independent review found no blocking issue; CI and final review remain pending.

## Before / after

An explicitly configured empty or whitespace-only admin/backend secret previously counted as configured, allowing an empty credential to match it. Server construction now rejects those values.

- Omitted credentials remain valid for an unconfigured standalone deployment.
- `backendSecret: ""` or a whitespace-only secret is invalid.
- A nonblank secret such as `" configured-secret "` is preserved byte-for-byte, not trimmed.
- An edge with a blank credential fails before connector validation, bootstrap, or connection work. Existing upstream URL parsing may report an invalid URL first; either invalid configuration is rejected without transmission.

No new auth mode, credential format, storage format, or wire format is introduced.

## Verification

- Builder suite: 15 passed, including independently rerun checks.
- Edge probe verifies zero connector validation/bootstrap/connect calls for blank credentials.
- Removing both validators made the blank-secret tests fail; restored afterward.
- Formatting and diff checks pass.
- Full CI pending.

Original #2484 commits are preserved in the adapted series; strengthened tip is `16d441e8db`. The original published branch has not been rewritten.

### DIFF
--- a/crates/jazz-server/src/server/builder.rs
+++ b/crates/jazz-server/src/server/builder.rs
@@ -809,6 +809,20 @@ fn validate_server_config(
         _ => {}
     }
 
+    if auth_config
+        .admin_secret
+        .as_deref()
+        .is_some_and(|value| value.trim().is_empty())
+    {
+        return Err("admin secret cannot be empty".to_owned());
+    }
+    if auth_config
+        .backend_secret
+        .as_deref()
+        .is_some_and(|value| value.trim().is_empty())
+    {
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved #2526 in the orchestration conversation after independent review and successful CI. Approval applies to `16d441e8db`.

PR #2483, opened 2026-09-01
Make recursive evaluation failures atomic

joeinnes:
## Summary
- Stage recursive operator, arrangement and evaluator state per tick.
- Roll back direct iteration-limit failures without changing committed state.
- Discard staged resident failures without leaking closure rows, notifications, memo state or frontiers.

## Before

Recursive evaluation could mutate retained state before discovering an iteration-limit failure. Resident failures could then leave partial recursive state or derived work behind.

## After

Direct-tick failures preserve the last committed closure and logical state. Resident scoped failures remove staged recursive state and preserve live graph retainers, notifications and frontiers. Grouped state uses bounded overlays rather than cloning the full retained history.

## Test plan
- [ ] `cargo test -p groove recursive_ --lib`
- [ ] Verify direct rollback and resident-failure teardown through the recursive runtime and Database surfaces.

PR #2527, opened 2026-09-03
Make recursive evaluation failures atomic on current main

aeplay:
🤖 Supersedes #2483 with a reimplementation against current main. This is an early visible WIP while independent adversarial review runs; it is not ready to merge.

## Before / after

A recursive evaluation can discover its iteration limit only after mutating retained evaluator state. Before this repair, the error could leave partial recursive closure, arrangements, memo/frontier state, tick advancement, or publication state behind.

Evaluation now stages its touched state and publishes it only when the relevant evaluation succeeds:

- A failed direct recursive tick leaves the previously committed runtime state unchanged.
- A maintained subscription that exceeds its recursion limit is closed and its partial derived state is discarded.
- The database write that triggered that subscription failure remains committed.
- An independent healthy subscription evaluating the same write still publishes its result; one failed owner does not suppress unrelated publication.

For example, if committed storage gains edges `2→3` and `3→4`, a deliberately limited recursive subscription may fail while deriving transitive results. Reopening the database still sees those two physical edges, but it does not inherit the failed subscription's partial derived closure. A simultaneous non-recursive edge subscription receives both inserts.

## Design constraints

The implementation uses bounded touched-state overlays rather than copying complete retained histories, keeps the current selective recursive-frontier optimization, and stays within the shared evaluator. A transformed recursive step that cannot use raw table field names falls back to its seed frontier rather than failing with `GraphFieldNotFound`.

No storage/wire format or public API changes are introduced. Current `BindingSourceKey` and terminal-hydration behavior are preserved. Independent review is specifically checking owner sharing, cancellation/hydration, notification atomicity, memory boundedness, and whether the descriptor-safe fallback is too broad.

## Verification so far

- `cargo test -p groove recursive_ --lib`: 51 passed.
- Direct rollback, resident scoped failure, persisted-base reopen, transformed-step, selective scale, and healthy co-resident publication regressions pass.
- Prior versions of the repair naturally failed the transformed-step test and stranded the healthy subscription; the final regressions cover both gaps.
- Full gates and independent review remain pending.

The original #2483 branch has not been rewritten.

### DIFF
--- a/crates/groove/SPEC/4_incremental_maintenance.md
+++ b/crates/groove/SPEC/4_incremental_maintenance.md
@@ -41,7 +41,7 @@ Invariant digest:
 - `INV-TICK-15`: A recursive positive incremental tick MUST emit each newly discovered recursive fact at weight `+1` at most once and MUST collapse duplicate derivations.
 - `INV-TICK-16`: The reference implementation selects recompute for negative table deltas, cached recursive state with table deltas, empty unbound state, or unhydrated step arrangements. This trigger set is broader than the minimum necessary; the contractual result is the minimal diff required by INV-REC-8.
 - `INV-TICK-17`: Recursive recompute and incremental recursion MUST reject non-positive recursive frontier facts instead of assigning bag-recursive semantics.
-- `INV-TICK-18`: Fixpoint recursion MUST stop with `RecursiveIterationLimit` when the frontier remains non-empty after its safety `max_iters`; semantic depth bounds MUST truncate instead.
+- `INV-TICK-18`: Fixpoint recursion MUST stop with `RecursiveIterationLimit` when the frontier remains non-empty after its safety `max_iters`; semantic depth bounds MUST truncate instead. The limit is part of recursive-node identity: equal recursive nodes share one success or failure, while distinct limits remain distinct. Failure aborts that node and its downstream closure, including dependent terminals and staged durable writes; independent graph branches may still publish and valid shared upstream state is not duplicated.
 - `INV-TICK-19`: Hydrating or querying a graph MUST NOT perturb an existing subscription stream's future tick deltas.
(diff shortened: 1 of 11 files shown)


aeplay:
🤖 Design direction from Anselm: recursive execution limits are semantic operator parameters and therefore part of recursive node identity. Identical recursive nodes share one success/failure outcome; nodes with different limits are distinct.

The failure boundary should follow graph dependencies rather than artificial per-subscription ownership:

- if a recursive node fails, that node and its affected downstream dependency closure do not publish staged state, derived writes, or notifications;
- every terminal subscription depending on that failed node observes the same failure;
- independent affected branches may still commit and publish;
- valid shared upstream state is not duplicated per subscription merely to achieve isolation.

The lane will encode this model in the Groove spec first, then reimplement sparse affected-node staging against current main. It must preserve the current selective-frontier optimization, same-tick read-your-writes for staged derived storage, and work proportional to the affected graph. It will stop and report rather than invent semantics if shared-upstream commits, durable-write ownership, cancellation, or hydration expose a conflict.

This records implementation direction, not approval of the current draft diff or permission to merge it.

aeplay:
🤖 Anselm approved trying the shared asynchronous persistence phase described above. The implementation should remain one `IncrementalEvaluation` path: evaluate against sparse staged state and a read-your-writes storage overlay; asynchronously flush derived writes for successful dependency regions; only then install retained state and notify. Failure or cancellation discards the failed region without suppressing independent successful regions.

The lane must stop and report a concrete graph if cancellation, shared upstream state, or derived-write ownership cannot satisfy this model without another semantic decision. This is implementation direction, not approval to merge the draft.

aeplay:
🤖 Anselm approved exact head `7570387b95f88f9d6edb25accfe63bcf7b7c9d5a`, pending green CI.

PR #2489, opened 2026-09-01
Invalidate eviction state before body removal

joeinnes:
## Summary
- clear durable and in-memory fast-known-state facts before publishing eviction body deletes
- invalidate both transaction-version cache layers before manual batches and budgeted per-candidate persistence
- cover definite fail-before and ambiguous write-through error outcomes

## Before
Eviction invalidated transaction caches and durable known-state facts only after its fallible deletion loop. A persistence error could leave deleted bodies covered by a surviving fast declaration or stale transaction cache.

## After
Candidate selection remains unchanged, but known-state clearing and cache invalidation occur before any body deletion can be published. Manual and budgeted eviction remain conservative across both definitely uncommitted and possibly committed storage errors.

## Test plan
- [ ] Run the four manual/budgeted persistence-failure regressions
- [ ] Run the Jazz library suite with transport compression enabled

PR #2528, opened 2026-09-03
Invalidate eviction state before body removal on current main

aeplay:
🤖 Supersedes #2489 with its eviction repair adapted to current main and a review-driven test correction. Independent adversarial review found no blocker; CI and Anselm's review remain pending.

## Before / after

Eviction previously removed row-version bodies before clearing durable and in-memory declarations that said those bodies were available. A storage error during deletion could therefore leave a node claiming it possessed data that was already missing.

Eviction now crosses the safety boundary in this order:

1. Clear durable and in-memory fast known-state facts before any body deletion.
2. Invalidate both transaction-version caches before the corresponding deletion can have a fallible or ambiguous persistence outcome.
3. Delete the selected bodies.

Manual eviction invalidates all affected transaction caches before its batch. Budgeted eviction invalidates each candidate before that candidate's persistence attempt. Candidate selection and byte-budget policy are unchanged. If candidates exist but the node is already below its low-water target, the known-state clear can conservatively force a refetch without deleting a body; it never produces a false possession claim.

## Failure examples

- If known-state clearing fails, no body is deleted and existing caches remain intact.
- If body persistence definitely fails before writing, the body remains but its fast declaration and caches are already gone.
- If persistence writes through and then returns an error, the body may be absent and its declaration/caches are still safely absent.

## Verification

- Five focused regressions pass independently: manual and budgeted fail-before/write-through outcomes, plus failure of the initial known-state clear.
- Review found the initial reopen assertion was vacuous because a reopened node lacks a live authority receipt. The tests now inspect the durable known-state store directly.
- Removing the production clear call makes the durable assertion fail; the mutation was restored.
- Diff checks pass; full CI pending.

Original #2489 commits are retained in this adaptation, followed by the review-driven durable receipt at `b1371d1c76`. The original branch has not been rewritten.

### DIFF
--- a/crates/jazz/src/node/eviction.rs
+++ b/crates/jazz/src/node/eviction.rs
@@ -206,16 +206,39 @@ where
             }
         }
         evictable.sort_by_key(|candidate| candidate.tx_id);
+        if !evictable.is_empty() {
+            // INV-SYNC-27: once local row-version bodies may be removed, no
+            // persisted or in-memory fast known-state cursor may survive.
+            // Clear these facts before publishing any body deletion so a
+            // clearing failure leaves every candidate body intact.
+            self.clear_all_known_state_facts().await?;
+        }
+
+        if low_water_bytes.is_none() {
+            // Manual eviction publishes one batch, so every affected cache
+            // must be invalid before that batch can have an ambiguous outcome.
+            let mut last_invalidated_tx_id = None;
+            for candidate in &evictable {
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Anselm approved #2528 in the orchestration conversation after independent adversarial review and successful CI. Approval applies to `b1371d1c76`.

PR #2529, opened 2026-09-03
Observe native transport terminal outcomes on current main

aeplay:
🤖 Supersedes #2476 with its native terminal-lifecycle repair adapted to current main. Independent adversarial review found no blocker; canonical CI is green. Anselm's review remains pending.

## Before / after

The native client installed a connected transport but dropped the connector's terminal future without polling it. An idle peer could close its socket while the client continued treating the dead upstream as an idle live connection indefinitely.

The client now retains a weak, generation-tagged terminal watcher and feeds its outcome into the existing tick/recovery driver:

- `PeerClosed` and recoverable failures detach the matching connection and retry through one generation-owned recovery flow.
- Deliberate owner drop and shutdown do not reconnect.
- A terminal event from an older generation cannot detach, poison, or replace a newer connection.
- Losing ownership during a retry prevents the late connection from being installed.
- Recovery backoff remains asynchronous, so local database ticks continue.

For example, if an otherwise-idle server disappears, the terminal watcher wakes the client and starts recovery without requiring application traffic. If the app deliberately closes while that recovery is pending, the pending attempt cannot resurrect the connection.

This reuses the existing connector, scheduler, and reconnect machinery rather than creating a second transport path. It does not change storage/wire formats or public APIs.

## Verification

- Nine focused lifecycle tests pass: idle close, recoverable terminal classes, owner drop, shutdown race, stale admission/exhaustion, ownership loss, terminal/tick coalescing, and local-tick responsiveness.
- Native transport terminal tests: 2 passed.
- On exact pre-fix main, the idle-close regression timed out because the terminal future was never polled.
- Suppressing `PeerClosed` recovery makes the regression fail; restored afterward.
- Independent review found no correctness blocker.
- All nine tests have the required contract/actor documentation; formatting, Clippy, and diff checks pass. A review-found placement error was corrected and rechecked.
- The lifecycle tests use explicit in-memory storage so parallel execution cannot contend over a process-global default RocksDB path; the complete parallel client partition passes 37/37.
- Canonical CI is green across lint, Rust workspace/focused/differential, storage compatibility, and TypeScript jobs.

Original #2476 behavior is preserved in adapted commit `dd16b00d8a`, followed by test documentation and isolation corrections through `5af63890f7`. The original branch has not been rewritten.

### DIFF
--- a/crates/jazz/src/tools/client.rs
+++ b/crates/jazz/src/tools/client.rs
@@ -3,7 +3,7 @@
 #[cfg(test)]
 use std::cell::Cell;
 use std::cell::RefCell;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeSet, HashMap, VecDeque};
 use std::future::Future;
 use std::ops::Deref;
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved this exact revision (`5af63890f79c60819ef5f00f89cb886f7879a10d`) after independent review and green canonical CI.

PR #2530, opened 2026-09-03
Bind Next correctness builds to sealed WASM artifacts

aeplay:
🤖 Correctness-artifact repair discovered while validating #2525.

## Problem

The correctness producer seals an admitted WASM package, but a later Jazz Tools worker build can still read the mutable `crates/jazz-wasm/pkg` output when the sealed environment is not carried through Next's resolver. A concurrent package build can therefore give Turbopack two different `jazz_wasm_bg.wasm` contents for one output path. The Next example then fails before its web server starts.

This reproduced locally and in #2525's CI. Reverting #2525's `server-only` import and externalizing Jazz Tools did not fix it; the credential behavior is unrelated.

## Intended repair

During an admitted correctness-artifact run, both Webpack and Turbopack resolve `jazz-wasm` to the exact sealed snapshot. If the run marker is present without an admitted package, Next configuration fails closed before reading mutable output. Normal application builds without the correctness-run marker remain unchanged.

For example, the worker bundle and browser runtime must both receive `/sealed/<fingerprint>/jazz_wasm.js`; neither may fall back to a concurrently changing workspace `pkg` directory.

## Verification

- Unit tests pin existing aliases and both Webpack/Turbopack sealed resolution.
- A missing admitted package under the correctness-run marker fails closed.
- Full correctness-artifact production completed with fresh admitted NAPI and WASM snapshots.
- The exact TypeScript consumer partition passed: 20 browser files (294 tests, one expected failure, three skipped) plus the Node suite, with both suite exit statuses 0.
- A real sealed Next application build passed with Turbopack and Webpack; emitted artifacts resolve the sealed snapshot rather than mutable workspace output.
- Independent adversarial review found no correctness blocker and planted/targeted contract tests passed.
- Canonical CI is green across lint, Rust, differential, storage-compatibility, and TypeScript jobs.

### DIFF
--- a/packages/jazz-tools/src/dev/next.test.ts
+++ b/packages/jazz-tools/src/dev/next.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { readFile, writeFile } from "node:fs/promises";
-import { join } from "node:path";
+import { join, relative, resolve } from "node:path";
 import { createTempRootTracker, getAvailablePort, todoSchema } from "./test-helpers.js";
 import * as devServer from "./dev-server.js";
 import * as catalogueProject from "./catalogue-project.js";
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved this exact revision (`9a7f85db90771e266c81cfe70cc914c752a4b6d1`) after the independent review and green canonical CI.

PR #2533, opened 2026-09-03
Unify atomic unique-index ownership handling

aeplay:
🤖 Combines the coherent correctness work from #2500 and #2504 into one unique-index ownership repair while preserving Joe's commits and authorship.

## Before

Two related cases disagreed about when a uniqueness conflict is fatal:

- Registering an index over existing conflicting data could leave the in-memory database poisoned even when no durable state had been submitted.
- One atomic batch that deleted the current unique-key owner and inserted its replacement could be rejected because intermediate operation order was mistaken for final ownership.

For example, replacing `(id=1, email=a)` with `(id=2, email=a)` in one transaction should succeed in either delete/insert order. Inserting two final owners of `email=a` must still fail.

## After

- Durable unique ownership is resolved from the batch's final net state before the single `write_many` submission.
- A pre-submission `UniqueIndexViolation` during registration leaves the facade usable because no state was written.
- Storage, I/O, encoding, or possibly-committed failures remain fatal/poisoning; this does not broaden recoverability.
- Both operation orders produce one final owner, while competing final owners are rejected.

This reuses the existing durable ownership and registration paths rather than adding a second index implementation.

## Verification

- `cargo test -p groove`: 654 passed, 2 ignored.
- Invariant registry and formatting/diff gates passed.
- A planted suppression of the competing-owner guard made the strengthened test fail and was restored.
- Independent adversarial review found no blocker. A planted suppression of the multi-owner guard failed the expected conflict receipt and was restored.
- Canonical CI is green.


aeplay:
🤖 Approved by Anselm at exact head 338342d00501c7740fd176bbe7a037b8d18e08d5 after independent adversarial review, planted regression proof, and green CI.

PR #2537, opened 2026-09-04
Test maintained indexed hydration through browser workers

### LINKED ISSUE #2077: Perf thesis: maintained indexed subscriptions must preserve async-storage delivery

aeplay:
🤖 Closes #2077 by exercising maintained indexed hydration across Chromium, SharedWorker, IndexedDB and the sync server.

The first global snapshot must contain every settled matching row exactly once. An initially empty two-equality indexed subscription must receive its first matching insert once. Subsequent remote updates must produce the complete ordered sequence empty → matching row → empty → matching row → empty as each indexed predicate changes.

Each transition checks row identity, title and done values, and consumes a callback after its corresponding write. Comparing the complete sequence rejects duplicate, out-of-order and unrelated snapshots that historical filtered counts could accidentally accept. No product implementation or browser-only query path is added; the existing native selective_global_hydration receipt remains the scaling measurement.

Validation: sealed artifacts and Jazz Tools build passed; implementation and coordinator independently ran the focused Chromium test (one passed, 74 skipped). An exact-test-body fault harness accepts normal delivery and rejects duplicate opening, missing first insert, duplicate-empty/missing-removal and extra-invalid-snapshot cases. Fresh full CI runs on the strengthened test head.

### DIFF
--- a/packages/jazz-tools/tests/browser/worker-bridge.test.ts
+++ b/packages/jazz-tools/tests/browser/worker-bridge.test.ts
@@ -1796,9 +1796,13 @@ describe("SharedWorker bridge with IndexedDB", () => {
       }),
     );
     const seededTitle = `indexed-seeded-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
-    await writer
+    const seededFirst = await writer
       .insert(maintainedIndexedTodos, { title: seededTitle, done: false })
       .wait({ tier: "global" });
+    const seededSecond = await writer
+      .insert(maintainedIndexedTodos, { title: seededTitle, done: false })
+      .wait({ tier: "global" });
+    const expectedSeededIds = [seededFirst.id, seededSecond.id].sort();
 
     const fresh = track(
       await createDb({
@@ -1825,8 +1829,11 @@ describe("SharedWorker bridge with IndexedDB", () => {
       15_000,
       "fresh IndexedDB worker must receive an authoritative settled indexed snapshot",
     );
-    expect(settledSnapshots[0]).toHaveLength(1);
-    expect(settledSnapshots[0]?.[0]).toMatchObject({ title: seededTitle, done: false });
+    // The first global callback is authoritative, not merely nonempty: a
+    // partial hydration that delivers either matching row is a failure.
+    expect(settledSnapshots[0]).toHaveLength(2);
+    expect(settledSnapshots[0]?.map((row) => row.id).sort()).toEqual(expectedSeededIds);
+    expect(settledSnapshots[0]?.every((row) => row.title === seededTitle && !row.done)).toBe(true);
     stopSettled();
 
     const emptyTitle = `indexed-empty-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved exact head cc63eb876a.

PR #2542, opened 2026-09-04
fix(ci): normalize NAPI declaration line endings

aeplay:
🤖 Fix Windows-only NAPI declaration rejection caused by host line endings.

Generated and checked-in TypeScript declarations compare after CRLF-to-LF normalization only. Substantive declaration changes, lone carriage returns, whitespace changes and missing final newlines remain fatal; diagnostics identify the changed line.

The three patches approved at a81bd63b89 are unchanged at d74c85d090, restacked onto merged RN main. Independent review and coordinator each pass16 contract tests with planted guard regressions. Normal PR CI passes.

Exact-head official Windows native build passes in https://github.com/garden-co/jazz/actions/runs/33919017680, as do packaged Android and iOS consumers. Subsequent npm assembly fails on a separate pre-existing provenance issue: CRLF checkout conversion changes Windows package-input and ABI hashes despite the same Git tree. That follow-up is isolated from this declaration-comparison fix and does not invalidate its Windows receipt.

### DIFF
--- a/dev/artifacts/build.mjs
+++ b/dev/artifacts/build.mjs
@@ -396,13 +396,15 @@ export function validateNapiStage(
   // semantic ABI guard, so Windows CRLF must compare equal to the checked-in
   // LF surface while every other generated difference remains fatal.
   const normalizeLineEndings = (value) => value.replaceAll("\r\n", "\n");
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved exact head a81bd63b89.

PR #2546, opened 2026-09-04
Test browser storage compatibility at the principal root

aeplay:
🤖 Completes #2308's browser storage compatibility corpus by installing the pinned raw IndexedDB fixture at the physical root production actually opens.

Before: the test computed a database name before local-first authentication. createDb then minted its principal and opened a different scope-named root, so the valid raw corpus was invisible and the test observed an empty database.

After: setup briefly opens through the public API to discover the authenticated principal root, shuts down completely, installs the immutable raw fixture there, verifies exact raw bytes with no bootstrap residue, and reopens normally. It exercises main and draft large-value data, lease claim/return, and a planted HLC incompatibility.

No production codec or diagnostic API is added. Independent review ran the sealed Chromium corpus receipt four times successfully and verified committed artifact provenance.


aeplay:
🤖 Approved by Anselm at exact head `c0739b50c4a56e2c8a83be1ee5927812d66e6884`. Independent adversarial review is clean and all canonical CI checks passed.

PR #2547, opened 2026-09-04
Keep shared worker reusable after idle release

aeplay:
🤖 Fixes the shared-browser-worker lease race that made unrelated TypeScript CI jobs intermittently time out while reopening a database after reset or logout.

Before: releasing the last foreground context could release the physical storage epoch and call `SharedWorkerGlobalScope.close()`. A same-name foreground could attach while that irreversible close was already in flight, leaving it connected to a dying realm that could never issue a foreground node lease.

After:
- ordinary idleness still releases contexts, lease ownership, and the physical storage epoch;
- ordinary idleness keeps the same worker realm available for a later foreground attach;
- explicit inspector termination remains the operation that closes the realm and carries the generation handoff;
- a deterministic regression releases the last context, waits through idle cleanup, then proves the same realm answers a lease probe and allocates a new lease.

Focused initialization, reset/logout, and full worker-bridge suites pass locally. Independent exact-head adversarial review is in progress.


aeplay:
🤖 Approved by Anselm at exact head `a8fca8cf1423a2dd01c78370ec5eb7c81d8cb5a7`. Independent adversarial review is clean and all canonical CI checks passed.

Issue #2461, opened 2026-08-31
Reset all unreleased numbered formats to v1


aeplay:
🤖 Alpha.54 format implementation and final source/corpus audit are complete at `a848ea51ef2fbe5e697e17e17fe13fd8dee4a2c7`. The exact settlement record and passing candidate CI are in #2160; boundary inventory and deliberate serializer decisions are in #2592.

Merged: typed identities (#2558), inactive result-persistence removal (#2580), dense JPFK V1 tags (#2596), V1 normalization (#2588), public union queries (#2582), recursive typed Postcard queries (#2597), and shared native/peer grammar references (#2601). #2548 is abandoned.

Rows/cells retain native encoding. Ordinary row identities remain UUIDs; joined/arm-discriminated occurrences use ResultKey V1. RocksDB namespace/value markers, schema canonical hashing, Node foreground lease and migration-publication journal use their selected V1 representations. IndexedDB generation 3 is a browser-managed schema generation, distinct from storage epoch and wire/binding versions. The existing JSON WebSocket admission prelude is retained with explicit contract and fixtures (#2599); superseded JSON relation payloads and JRQ are removed.

All candidate CI partitions passed at release metadata head `193f5894d1d1f305fd2045b4a149c0305a93ab13`: https://github.com/garden-co/jazz/actions/runs/33998110029. Independent final review found no additional codec defect. The settlement source storage-compat partition also passed.

Published-release fixture provenance remains #2309. Packaging, installed-device and physical iOS Safari acceptance are independent release gates. Format completion does not authorize publication or claim those acceptance gates passed.

aeplay:
🤖 Current-code refresh after the merge: the specific old physical-mapping V1/V2 example is already resolved. crates/jazz/src/node/codec.rs uses one PHYSICAL_MAPPING_VERSION = 1 with wide enum tags and rejects other versions. The repository-wide final numbering audit remains required for alpha.54; this note does not claim that audit complete. React Native is explicitly included in this release, so coordinate its first-public ABI rename with #2373.

aeplay:
🤖 Release decision: this remains a hard alpha.54 gate, but it is intentionally sequenced almost last. Run the repository-wide numbering sweep only after all remaining storage- and wire-changing work has landed, then complete it immediately before the final release candidate so later integration cannot reintroduce unreleased version names.

aeplay:
[...]

PR #2548, opened 2026-09-04
Distinguish stored columns from result fields in native decoding

aeplay:
🤖 Distinguish stored application columns from derived/result fields explicitly across the Rust/native boundary, and rename the internal application-column namespace from `user_` to `_app_`.

## Before

Native decoding inferred provenance from a field's spelling. A stored application column `check` and a derived field literally named `user_check` could both travel through a `user_check`-shaped carrier and collapse to the same public name.

The same inference survived in receipt validation and low-level lookup helpers, so fixing only the final TypeScript decoder was insufficient.

## After

Descriptor metadata represents the distinction directly:

```text
StoredColumn { id: <physical column id>, output_name: "check" }
ResultField  { name: "_app_check" }
```

- Stored columns retain their physical identity and explicit public output name.
- Result fields retain their literal logical name, even when it resembles an internal carrier.
- Rust terminal, projection, maintained-view, and native row layouts preserve that provenance end to end.
- Native decoding performs no prefix-based provenance inference.
- The compiler-owned application slot namespace is now `_app_`; the old `user_` storage/wire spelling is rejected rather than supported as compatibility behavior.

For example, a stored column `check` and a result field literally named `_app_check` may coexist in one descriptor and decode independently.

## Verification

- stored/result collision and descriptor-codec round trips;
- terminal and subscription provenance preservation;
- branch/current-row storage namespace consistency, including RocksDB reopen;
- immutable version receipt acceptance for `_app_` and rejection for legacy `user_`;
- exact-only native field lookup;
- focused Rust and TypeScript native-runtime suites.

This is the contained design. A separate experiment is attempting the larger principled design in which Groove field identities are typed throughout and no string prefix participates in identity.

## Current integration checkpoint

Checkpoint4c6eafaef5 includes main4b640b8e56 and the approved parked-cancellation, WASM streaming-abort, Windows declaration and browser corpus fixes. Independent review and coordinator probes previously repaired direct-read metadata, stored-name filtering, aggregate normalization and collection-alias collisions. Disambiguation moves only actual stored-column carriers and preserves clean rejection of aliases that replace internal row keys.

The combined RN readers and acceptance oracle now consume explicit tagged descriptors. Their exact row identity/value assertions remain, with result-name and prefix-collision controls. Coordinator independently passes84 RN source tests plus typecheck; a planted prefix-normalization regression fails. Implementation integrated consumers279 passed/2skipped, wire208 passed/1skipped, focused Rust integration5 passed, NAPI contract16 passed. Official artifacts sealed to final HEAD.

Earlier combined source passed Jazz1997/4skipped, native relay46 and workspace checks; coordinator independently reran46 relay tests. No production Rust change followed beyond merging the already approved cancellation changes. Standalone Groove761/4skipped passed before integration.

Remains draft until current-head CI completes. The unreleased V1 namespace/descriptor cut is governed by #2461; optional typed alternative #2558 remains separate.

Review sequencing decision: held at Anselm’s request until the shared freezable storage format with #2558 is established and clearly documented for review under #2461. Green CI does not authorize merging this PR, and no shared-format compatibility claim is made yet.

## Shared pre-freeze publication contract

Checkpoint6162941954 adds the common third native descriptor tag: postcard tag0 `StoredColumn { id: u64, output_name: string }`, tag1 `ResultField { name: string }`, and tag2 `HiddenMetadata { name: string }`, with payload fields in that order. Source and projection owners classify application cells, public provenance and hidden metadata. Consumers suppress only tag2; a valid `COUNT(*) AS schema_version` remains visible beside hidden metadata of the same name. Explicit public magic provenance projections remain tag1. Logical fields without an output-name override remain public.

The complete component contract, exact five metadata-byte golden changes and unchanged encoding surfaces are spelled out in `dev/proofs/contained-native-publication-v1.md`. This commit changes native host publication, not durable row/index encoding. It does not establish bidirectional database compatibility by itself. The separately discovered large-JSON VersionRecord descriptor correction is being reconciled across both alternatives under #2461.

Initial lane verification passes the real Db producer fixture (current rows, catalogue IDs, aggregates, public aliases, provenance, nested/join reads and resets), TypeScript codec/descriptor38/38 and typecheck. The coordinator independently passes the same38 codec/descriptor tests. A planted hidden-field suppression regression fails the strengthened collision test. Full Rust/workspace, generated-host and final independent review are still pending at this checkpoint. The user's hold remains in force.


aeplay:
🤖 Release decision: typed identities in #2558 are now the sole implementation path. Anselm explicitly chose to abandon the contained fix in #2548. The previous requirement that the two alternatives share a storage format is superseded.

Design and review the typed implementation’s durable format directly: specify authoritative encodings, keep compiler-local identities out of durable data, and pin persistence/reopen behavior with byte fixtures and real database tests. Preserve useful regression tests from the contained work, but stop reciprocal compatibility work against the abandoned implementation. Format freeze remains pending completion and independent review of the typed implementation.

PR #2553, opened 2026-09-04
fix: use last seen tx's global time for fup exclusive tx snapshots

nrainhart:
## Problem

Exclusive transactions were not reliably anchored to the current server time on partially synced clients.

The transaction’s global base was taken from the local database’s `committed_global_time`. On downstream client/edge databases, that value only advanced when the complete history through a timestamp was stored locally. With partial sync, clients intentionally receive only relevant data, so `committed_global_time` could remain at `0` even after the server had settled much later transactions.

As a result, an exclusive transaction could start with a global base of `0`. Reads requiring data outside the client’s local subset—especially relation/include reads—were then evaluated against an incorrect or incomplete snapshot. They could omit data, fail, or change when older data arrived later.

## Solution

Downstream databases now advance `committed_global_time` when they receive an authoritative `settled_through` timestamp as part of a subscription result. For a client or edge database, the field therefore means:

> The newest point in core history that this database knows the authoritative server has settled.

It does not mean the client has downloaded every row through that point.

When an exclusive transaction begins, it pins this authoritative global timestamp as its base. Transaction reads request their snapshot from upstream at that fixed base, allowing the server to supply relevant data even when it was never present locally. Staged transaction writes are still overlaid on that snapshot.

The core server’s invariant remains unchanged: its own `committed_global_time` still means history is complete through that point.

The Rust↔TypeScript read API was also consolidated to avoid method explosion:

- `attachQuery(query, opts, openTransactionId?, author?)`
- `all(query, opts, openTransactionId?, author?)`
- `allRelationSnapshot(query, opts, openTransactionId?, author?)`
- `allRelationQuery(queryJson, opts, author?)`

## Result

Exclusive transactions now point to a real, stable location in server history without requiring clients to fully synchronize that history. Relation snapshots and ordinary reads inside the transaction observe the same fixed base, while partial synchronization remains intact.


aeplay (approved):
🤖 Approved by Anselm at exact reviewed head . Independent adversarial review found and repaired the test-only-helper compile failure; focused Rust and browser snapshot gates are green.

aeplay:
🤖 Approval clarification: Anselm approved exact head `15c84a4daa5393315116e9aefff32eb12ba28826`.

aeplay (approved):

Issue #2557, opened 2026-09-04
Complete the typed Groove field identity migration


aeplay:
🤖 Complete typed Groove field identities in #2558, now the selected release implementation. Anselm retired contained #2548; reciprocal compatibility with it is no longer required. One implementation lane owns compiler, publication and durable-format work together.

Latest reported full core run: 1,771 passed, 45 failed, 2 skipped. The grouped publication repair fixed a real path but did not reduce that remaining failure set. Known clusters include stored provenance conversion, aggregate identity propagation, collector output matching and policy types. These failures prevent release readiness.

Repair complete identity-loss paths using compiler-enforced interfaces, source inspection and representative failing tests, then rerun the full suite at coherent checkpoints. Do not replace failing assertions merely to make the implementation pass.

Specify and verify durable encodings under #2461. Keep node-local physical aliases distinct from portable UUID identities and execution-only compiler slots. Audit VersionRecord, ResultCurrent and durable hash domains so execution serializer layout cannot silently define persistence. Pin same-implementation reopen, recovery, schema evolution, JSON and collision behavior. Independent adversarial review and integration gates remain required.

aeplay:
🤖 Typed column identities landed in #2558, merge 04128dde4d6873723853beaa396831a82e566eb2. The old 45-failure status in this issue is superseded: the PR records the repaired identity-loss paths, independent adversarial review and passing final GitHub gates on 3954e24c8eeb, including Rust workspace/differential, TypeScript, storage compatibility and RN scaffold checks.

Closing the implementation issue. Required inactive-result persistence cleanup remains tracked separately in #2578 / PR #2580, and final V1 settlement in #2461 / PR #2583. Aggregate pagination is independently fixed and green in #2579 / PR #2581. Those release/freeze obligations are not discharged by this closure.

PR #2558, opened 2026-09-04
WIP: propagate typed Groove field identities

aeplay:
🤖 Typed column identities are the selected release implementation, tracked by #2557. Anselm retired contained #2548 and removed the requirement to preserve compatibility with that abandoned implementation. One implementation lane owns compiler correctness, publication and durable format together. The focused typed-only checkpoint is being prepared to land; the combined RN work remains separate. Required pre-release cleanup of inactive result persistence is tracked in #2578 (Freeze storage format milestone). Landing this PR does not freeze those dormant formats.

The implementation carries explicit logical and slot identities through query planning and execution instead of reconstructing identity from field spelling. A user column `_app_score` must remain distinct from an internal carrier for `score`; joins, projections and aggregates must preserve their actual inputs. Existing collision expectations remain intact.

## Wire format: before → after

“Before” is this PR's merge-base `f27db85a8853`; “after” is the main-integrated typed checkpoint `3954e24c8eeb` (including the aggregate, join-publication and native-reader corrections). This compares actual code, not the abandoned #2548 alternative. The old VersionRecord comment claiming schema-only decoding did not match its actual OwnedRecord serializer.

| Boundary | Before | After | What changes |
| --- | --- | --- | --- |
| Network VersionRecord row | Postcard serializes OwnedRecord as descriptor + raw row bytes. A complete descriptor is already repeated per version. | The outer message is still postcard, but the row is an explicit JVRR v1 blob containing descriptor + raw row bytes. The descriptor is still repeated per version. | New explicit inner framing and a persisted-descriptor grammar that excludes compiler identities. **No descriptor deduplication is introduced.** |
| VersionBundleRun | Shares transaction/table context, with VersionRecords inside each body. | Same batching arrangement. | Does not eliminate the per-VersionRecord descriptor in either version. |
| Rust → TypeScript publication | Generic row descriptors carry field names/types; public meaning relies on producer conventions and carrier names. | Explicit StoredColumn(id, output_name), ResultField(name), and HiddenMetadata(name) variants, with explicit recursive types. | Native ABI changes; public provenance stays visible and hidden metadata is suppressed by role rather than spelling. |
| TypeScript → Rust named cells | Names/types descriptor plus raw row bytes, decoded through generic Rust descriptor serde. | Same intended names/types transport shape, decoded by a dedicated canonical reader shared by NAPI, WASM and the native relay. | Stops runtime compiler identities from changing the input ABI. Existing all-value-types native row corpus stays byte-identical. |

JVRR is **a wire blob, not the representation of a stored history row**:

```text
JVRR | version:u8=1 | descriptor_length:u32 little-endian
     | canonical persisted-descriptor bytes | canonical row bytes
```

The outer VersionRecord still carries table, schema UUID, branch and authored-column names. The receiver validates the descriptor against that schema before local catalogue translation. Per-row descriptor repetition is real overhead both before and after; this PR changes its encoding, not its frequency.

## Storage format: before → after

**Active storage versus retained machinery:** receivers always derive public results, including remote counts, through local IVM. Active persistence retains source rows, ProgramSourceCoverage/CoveredInput facts and closure receipts for reopen. In-memory ResultPayload values still serve the local aggregate pipeline, but the retained ResultPayload/JRPD and result-member storage codecs below have no identified current production write path: both the base and this PR filter result outputs from peer publication and reject them on admission. Their codec changes must not be mistaken for a currently used remote-aggregate cache.

| Stored data | Before | After | Storage effect |
| --- | --- | --- | --- |
| Application history, deletion registers and current-row indexes | Ordinary Groove records under their declared table schemas; no complete descriptor or JVRR blob inside each application row. | Still ordinary Groove records under declared table schemas; still no per-row descriptor/JVRR wrapper. | The record codec is retained. Schema-owned application carrier names change from `user_<name>` to `_app_<name>` and generated current-index names from `by_user_<name>` to `by_app_<name>`; this is not a promise of byte-identical catalogue/schema metadata. |
| Retained result-output storage codec (**not an active output-persistence path**) | JPFK v1 ResultPayload keys contain member identity, an encoded descriptor, and raw row bytes. The descriptor is already repeated per payload. | Same outer JPFK v1 fact framing, but the descriptor is now JRPD v1 with an explicit current/aggregate role and canonical role/name/type tree. Raw values must match that exact schema and member identity. | **Nested descriptor/key bytes change.** Descriptor repetition remains. Old experimental descriptor payloads are rejected, not silently reinterpreted. |
| Retained result-member/source storage codecs (**no active peer-result persistence**) | Explicit JRME/JRSE envelopes and Groove records; dynamically typed values contain descriptor + value bytes. | Same envelope families; dynamic descriptors use the persisted-only grammar, recursively excluding compiler Name/Slot/NamedSlot bindings. | Existing envelope families remain; typed descriptor/identity contents are governed by explicit roles. Do not infer blanket byte compatibility from unchanged magic/version bytes. |
| Terminal publication-layout hash | Derived from the prior compiler/publication layout representation. | BLAKE3 over domain `jazz terminal root publication v1` and canonical publication roles/names/types/layout positions/carrier kinds. | Hash identities change; compiler allocations and node-local physical IDs are excluded. No old hash alias is accepted. |

Actual history records contain branch/row/transaction/schema aliases, parents, provenance and application cells; authored-column presence uses node-local physical-column IDs. Deletion state is a separate register record. Stored values are not postcard-serialized VersionRecords.

The retained ResultPayload storage codec would encode this structure; current admitted peer updates cannot write this variant:

```text
JPFK | version=1 | ResultPayload tag=0
     | encoded member identity
     | length-delimited JRPD descriptor
     | length-delimited canonical Groove row bytes

JRPD | version=1 | role=0(current) or 1(aggregate)
     | canonical Groove persisted-descriptor bytes
```

The persisted descriptor is itself a Groove record containing a flattened recursive type tree (tags, logical names, enum registry IDs, child counts and case names), not Rust serde/postcard. Current roles distinguish source/result/provenance/metadata fields and carry ordered query-scoped occurrence declarations. Joined UUIDs and union discriminators are checked against the member identity before execution bindings are rebuilt; these are normalized query carriers, not globally stable table identities. Aggregate group/value roles have separate ordinals, so two fields named `count` need not collide. Readers validate exact schema, member identity, bounds and canonical encoding before rebuilding local execution bindings.

**Descriptor repetition:** each transmitted version repeats a descriptor. The retained result-output codec also repeats one per encoded payload, but that is not evidence of active disk usage. Remote aggregates rebuild from persisted covered inputs; they do not require persisted output rows or this dormant output codec.

These are pre-freeze format changes. Same-implementation persistence/recovery is required; reciprocal compatibility with #2548 is not. The table describes the touched boundaries, not a completed audit of every durable family. Exact fields, bounds and fixtures are in Jazz SPEC16 and `dev/proofs/typed-identity-durable-roles.md`.

## Verification and remaining work

The current candidate incorporates main at `d23a151557d0`, resolving all four conflicts while retaining both appended read tests. The ordering repair passed independent adversarial review, including all 24 permutations of a four-row result with occurrence sidecars preserved, and planted regressions failed as intended.

Full-workspace verification exposed aggregate alias collisions, join nullable-layout/contributor gaps, branch payload duplication and a remaining native reader using generic descriptor serde. These were repaired at their compiler/publication boundaries, retaining exact schema validation. The coordinator independently passed 80 affected testkit tests and 27 aggregate/role checks at the preceding `4ca89c507a35` checkpoint. The mixed-nullability projection canary now covers both alias orders, one/two nullable layers, null-preserving outputs and required-raw filtering. Exact `a8b1702dd998` full-workspace runtime: **3,641 passed, 73 skipped**, with deterministic benchmark smoke green. The coordinator independently passed 89 branch/native/projection checks at `e0295dff49ac` and 20 projection checks at the final head, after the retained-contributor layout correction. Independent adversarial delta review is complete with no remaining findings. Extended source-depth/join-prefix checks pass, and deliberately weakening wrapper restoration makes the test fail. The official local TypeScript partition is green on this exact head: 28/28 Node tasks; Jazz browser 298 passed, 1 expected failure and 3 skipped; BandChat 5/5 and record-player 9/9. The same-checkout producer manifest was verified and preserved, and generated fingerprints restored. GitHub CI is running. The RN source-scaffold jobs exposed a harness field lookup still using raw descriptor names. Commit `3954e24c8eeb` shares the production role-to-public-name mapping with that harness, preserving hidden-metadata exclusion. The coordinator independently passed all 84 harness tests and typechecking; independent review passed and restoring the old lookup reproduced the failure. Rust sources and formats are unchanged from the fully verified `a8b1702dd998` checkpoint. All final GitHub checks passed on `3954e24c8eeb`, including TypeScript, storage compatibility, Rust workspace/differential, lint and both RN scaffold jobs. Merged as `04128dde4d6873723853beaa396831a82e566eb2`. The required #2578 cleanup lane was dispatched immediately from that merge.

The retained recovery proofs cover actual JVRR wire input with relocated catalogue IDs, arbitrary nested execution-slot rebinding, and nested/aggregate results rebuilt after RocksDB reopen with no authority available. Negative controls fail as intended. Details are in `dev/proofs/typed-identity-durable-roles.md`.

The separately measured pre-existing aggregate subscription pagination defect is tracked in #2579. The broader RN API work remains separate. Landing this focused PR is not the final integrated release or a storage-freeze declaration.

## Required follow-up before release/freeze

#2578 removes inactive result-related persistence paths/codecs while preserving runtime identities and active covered-input recovery. It is in the Freeze storage format milestone and blocks the final V1 freeze and release consolidation. It must not be deferred merely because this PR lands.


aeplay:
🤖 Release decision: typed identities in #2558 are now the sole implementation path. Anselm explicitly chose to abandon the contained fix in #2548. The previous requirement that the two alternatives share a storage format is superseded.

Design and review the typed implementation’s durable format directly: specify authoritative encodings, keep compiler-local identities out of durable data, and pin persistence/reopen behavior with byte fixtures and real database tests. Preserve useful regression tests from the contained work, but stop reciprocal compatibility work against the abandoned implementation. Format freeze remains pending completion and independent review of the typed implementation.

Issue #2430, opened 2026-08-31
Verify IndexedDB fallback in iOS Safari private mode


husseinraoouf:
🤖 Alpha.54 acceptance gate: verify the current IndexedDB-backed browser runtime in real iOS Safari private mode.

The original alpha.53 report concerned legacy OPFS startup returning UnknownError. Main has removed that backend. Do not implement another legacy OPFS workaround or treat the old source diagnosis as current implementation work.

## Remaining receipt

- Pin the release candidate and exercise authenticated persistent startup, a local query/write and reload in a real iOS Safari private tab; compare regular mode.
- Record the actual storage behavior or typed unavailable result. An offline-authoritative application must not silently receive a durability downgrade.
- Exercise shutdown/retry and verify errors remain actionable.
- Record device/iOS/browser versions and the exact candidate revision without adopter identifiers.

Injected browser failures and desktop Playwright WebKit are useful automated coverage, but do not establish real iOS private-mode behavior. No dedicated completion PR/device receipt is currently identified. Keep open until this verification or an explicitly scoped follow-up is recorded.

aeplay:
🤖 The original alpha.53 OPFS failure path no longer exists on main: browser persistence now uses IndexedDB. For alpha.54 this issue is a real-device verification gate, not a request to port the OPFS workaround. Verify current main/preview in iOS Safari private mode and record whether IndexedDB opens for the private session or fails through the existing typed `SecurityError` → ephemeral fallback. If Safari returns another exception, Jazz must surface a typed storage-unavailable outcome rather than silently claiming durability. Playwright WebKit is useful for injected fallback coverage but is not evidence of real iOS private-storage behavior.

aeplay:
🤖 A public HTTPS manual check is ready: https://thee-screenshots-attempts-applied.trycloudflare.com/

The disposable local-only page is built at receipt-overlay commit `61bec5bb1fc1fba991bc919bfd02ba3b0bf886dd`, whose Jazz core/runtime source is unchanged from `bea3674ea70539017f3058a680fb5136fc9562d7`. The sealed producer receipt and served WASM hash are archived locally. No account or remote sync service is needed.

Coordinator verified startup, local write/read, same-tab reload retention, and explicit shutdown/reopen through the public URL in Chromium. This is setup verification, **not** real Safari acceptance. The user is running the physical iPhone check next.

[...]

PR #2593, opened 2026-09-05
Diagnose persistence with runner-built actual Safari tests

aeplay:
🤖 Add an on-demand actual-Safari persistence test for #2430. The existing public test showed optimistic task creation followed by missing rows after reload, but its ready label and cleared input did not establish worker attachment or a confirmed local save.

The new `safari-local` workflow mode checks that Safari WebDriver can open a session on Blacksmith macOS, builds the selected commit with the existing pnpm/Rust caches, and serves a diagnostic page on loopback. It compares reload after optimistic visibility with reload after `.wait({ tier: "local" })` resolves. The acknowledged-save path must retain the row. The optimistic path records whether an acknowledgment had already arrived, so timing is not inferred from visibility alone.

A dedicated diagnostic worker records lease admission, IndexedDB and durable-runtime opening, peer attachment, frame counts/bytes, and shutdown-flush phases. Page traces record local-read and local-write settlement. Worker frames are transport evidence, not a durability receipt. The loopback collector retains these events across navigation; failures retain screenshots, SafariDriver/system logs, and the exact source/artifact receipt. The hook is disabled in production entrypoints and cannot throw into normal worker execution.

The diagnostic page waits for a local read before reporting ready and for a local save before clearing submitted input. Startup or mutation errors remain visible. The existing public URL and its stored state are unchanged. This does not substitute for a physical iOS private-browsing check or claim the Safari product bug is fixed.

The artifact producer also handles macOS directory rename semantics: it seals descendants, keeps only the private staging root writable until rename, then seals that root before validation and publication. Existing sealed snapshots are never made writable. The Safari job opts into workspace Rust caching while continuing to build and seal artifacts against the selected source.

Validation: coordinator reran 50 transport/follower/worker tests and the TypeScript check; independent review verified opt-in traces and unchanged scheduling/routing. The artifact-store and workflow contract checks pass 64/64. Actual Safari 26.3 on macOS 15.7.4 reaches the diagnostic page after successful native/WASM builds in [run 33987146059](https://github.com/garden-co/jazz/actions/runs/33987146059), but its initial local read stalls after worker initialization and follower attachment. Two frames reach the worker with no return frames or completion/error before the readiness deadline. Reload scenarios therefore have not passed.

The next [actual Safari run](https://github.com/garden-co/jazz/actions/runs/33988218274) uses source `49cdcd7b95b184bae8ec8ba9a9b6b37713843f42`. It adds payload-free tab/worker microtask, routing, core-tick and outbound-drain phase events to distinguish the unresolved boundary. This PR remains a diagnostic draft; #2430 stays open.


aeplay:
🤖 Closing this diagnostic draft after the Safari investigation and acceptance completed. The minimal evaluator fix and regression coverage landed in #2598; actual desktop Safari regular/private receipts and the user's physical iPhone regular/private persistence receipts are recorded in #2430 (latest https://github.com/garden-co/jazz/issues/2430#issuecomment-5555520173).

This branch is preserved as the diagnostic history. It is not a release integration candidate: its old base now produces an obsolete reverse/duplicate diff against the current release branch. The fresh physical-iOS probe used the merged release runtime plus an isolated page, not this older diagnostic build. No additional diagnostic instrumentation is required to close the Safari release gate.

Issue #2591, opened 2026-09-05
Use Postcard consistently for all query payloads before release


aeplay:
🤖 Use Postcard consistently and recursively for all supported query payloads before alpha.54. Anselm selected this approach after reviewing #2597's hybrid Postcard/JRQ representation; a custom query/wire grammar is deferred until wire semantics have matured across multiple format generations.

Cover ordinary queries and nested relations, predicates, references, literals, includes, aggregates, traversals and query options wherever supported. Keep TypeScript/native, WASM, NAPI, RN and both peer Query.relation and ShapeBody::Relation producer/consumer paths aligned. No JSON AST wire payload or embedded JRQ custom query codec remains.

Rows, cells and their existing native data payloads are explicitly excluded from this conversion: preserve native row encoding/storage, carrying those bytes inside envelopes as before.

Preserve semantic query and ShapeId identity, numeric/literal distinctions, permissions and execution behavior. Use Postcard-compatible typed representations for recursive values rather than relying on unsupported deserialize-any behavior or manually embedding a second binary grammar. Document the selected layout and bounds; pin shared Rust/TypeScript bytes, malformed cases and real native/browser/RN/peer behavior.

Implementation is #2597. Final integration and complete producer/consumer receipts are required before closure; #2461 and #2160 own the final format settlement revision. Earlier JRQ green tests do not establish acceptance of the selected Postcard implementation.

aeplay:
🤖 #2591 binary-query audit — revision `cfde79aca865fd317baa0e5c00fafd61e494bd00`

**Replace before freeze (proven).** The host/WASM relation-query entry points still accept JSON: `crates/jazz-wasm/src/lib.rs:2204,2305,2322,2338` call `relation_query_from_json` (`:3777-3784`), which extracts `relation_ir` and deserializes it with `serde_json`. This is the temporary AST escape described by #2591. It crosses TS → WASM/NAPI and eventually server/peer query execution, so it needs one shared binary grammar, not a JSON parse on one side.

[...]

PR #2597, opened 2026-09-05
Use Postcard consistently for query payloads

aeplay:
🤖 Encode relation queries with typed Postcard throughout the native and peer query paths (#2591).

Previously, native relation entry points accepted JSON text, and Query.relation embedded a JSON string inside its Postcard envelope. A relation query now uses a recursive Postcard tree directly, including predicates, references and literal arrays/objects. Query.relation and ShapeBody::Relation embed that typed tree; direct native relation calls carry its standalone Postcard bytes. There is no custom query header or nested JSON AST fallback.

For example, a filter comparing a column with a nested object carries typed object, array, null, boolean and numeric variants. Numeric representations preserve signed/unsigned integer and floating-point distinctions, including negative zero. Shared Rust-produced fixtures check TypeScript byte parity and Unicode object ordering. Ordinary query fields and existing command/peer envelopes continue to use Postcard.

Rows and cells retain their existing native encoding, including when their bytes travel inside a Postcard envelope. Persisted row/storage codecs are unchanged. This changes query wire/binding payloads and requires matching runtime builds; it does not introduce a compatibility decoder for an unreleased intermediate format. Human-readable query JSON remains available at the API/diagnostic boundary.

Deserialization enforces recursive depth, total items, collection and string limits while constructing the tree. Direct and embedded carriers share the relation byte limit; native prepareQuery paths require exact canonical decoding. Malformed/trailing inputs reject, and a subsequent valid query can decode on the same thread. Tests cover exact boundaries, one-over raw wire payloads, nested Query/Shape paths and decode recovery.

Independent adversarial review approved the Rust implementation and decoder boundary tests. Coordinator verification at ef5a035ba18b10acb1f5fe2e4863c15c1aac6527 passed: source-bound NAPI/WASM producer; six Rust codec tests; 146 focused TypeScript tests (one skipped); TypeScript checking; 302 browser tests (one expected failure, three skipped); and 1,933 Node tests (two expected failures, nine skipped). The Node rerun used fresh test lease state after preserving stale V2 receipts from earlier checkout revisions. Generated fingerprints were archived and restored, and the committed source tree is clean.

BandChat passed 5/5 in isolation. A concurrent full-suite run hit the previously tracked invitation timeout (4/5); evidence remains in #2413 and no assertion or timeout was weakened. CI on the published revision is pending.

This implements the decision to use Postcard for query structure while wire semantics mature. #2461/#2160 still own final integrated freeze acceptance.


aeplay:
🤖 Anselm has reopened the encoding design: queries should use one encoding consistently and deeply, rather than interleaving Postcard and a custom relation codec. Hold this PR from landing.

Current code still serializes ordinary Query fields with Postcard and embeds JRQ in its optional relation field; direct relation payloads also remain inside Postcard command/peer envelopes. JRQ itself is fully custom recursively, including literals, but the complete Query representation is not. Earlier wording describing this only as an outer-envelope distinction was incomplete.

The coordinator recommends extending the explicit codec to the complete query grammar and removing the Postcard Query representation. Any conversion of enclosing command/peer envelopes is a further scope boundary to settle explicitly. Existing green CI validates the hybrid implementation, not the revised uniform-format requirement. #2591 and #2461 remain open.

aeplay:
🤖 Anselm's final direction for #2597: use Postcard consistently and recursively for queries now. Defer a custom query/wire grammar until the wire semantics have matured over multiple format generations.

Explicit exception: rows, cells and their existing native data payloads retain their native encoding. Carry those native bytes through envelopes; do not re-encode row contents or change the native storage formats.

The active lane is replacing JRQ through the complete query paths, including nested relations/predicates/literals, both Query.relation and direct ShapeBody::Relation, and native/WASM/NAPI/RN ingress. Existing command/peer Postcard framing remains. Remove the unreleased JRQ fallback and magic; preserve query identity and literal semantics. No JSON AST payload is introduced on the wire. The shared corpus, specification and cross-language checks will describe the actual Postcard representation.

This supersedes the prior proposal to expand custom JRQ. The current published implementation is an obsolete checkpoint and is not approved for landing under this decision. #2591 remains the implementation tracker and #2461/#2160 retain final integrated verification.

PR #2606, opened 2026-09-06
Restore NAPI runtime artifacts in starter CI

aeplay:
🤖 Restore the generated NAPI loader and fingerprint in the Starters E2E artifact handoff.

All twelve starter jobs in run33999681640 fail before app tests: the clean matrix checkout has the tracked NAPI bootstrap and restored native binary, but lacks native-loader.cjs. The prepare job stages this loader while packing jazz-napi, yet its upload list omits it.

The handoff now includes that generated loader, its matching fingerprint, and native manifests. A clean consumer therefore loads the same staged native generation without relying on the prepare runner's machine-local pointer. No runtime, storage format, starter behavior, or acceptance assertion changes.

A post-restore NAPI import preflight now reports incomplete handoffs immediately, before scaffolding.

Validation: independent adversarial review passed, including32 focused contract/provenance tests and clean-checkout restoration. The restored package imports successfully; deleting its loader, fingerprint receipt, or native binary causes failure. Coordinator independently reran6 release-gate tests, formatting, clean import and loader/fingerprint deletion negatives. No findings.

Full CI34000981589 passed after an exact TypeScript retry; the recurrent BandChat test-input failure was separately diagnosed and fixed in merged #2610. Matrix34000984904 verified this NAPI handoff and reached all12 app harnesses, then exposed a separate tools/WASM fingerprint mismatch. Stacked #2608 repairs that build ordering. The combined matrix34002971096 is running. User approval remains conditional on green CI and the full starter matrix; this PR is not merged yet.

PR #2608, opened 2026-09-06
Build coherent native artifacts and TypeScript packages for starters

aeplay:
🤖 Build Jazz Tools against the fingerprints of the native artifacts packed with it.

After #2606 restored the NAPI loader, the starter matrix reached browser startup and failed before rendering: Jazz Tools expected WASM fingerprint2351d808… while the packed WASM returned aeb2708f…. Auth signup succeeded; the missing todo input was downstream of this ABI rejection.

This change orders build:core as native builds → generated fingerprint staging → TypeScript package builds, then executes the WASM package to check its runtime fingerprint against the manifest and compiled expectations before packing. A mismatched pair should fail during preparation rather than twelve browser journeys later. The published0394059 preview tools/WASM pair independently passes the same runtime/expectation comparison; this defect was observed in starter CI preparation.

Stacked after #2606. No storage or wire format changes.

Normal worker builds now select workspace WASM even if old correctness pointers or an ambient sealed-package path are present. Explicit correctness mode requires and uses its admitted sealed package; a missing path still fails closed. This fixes the mixed glue/binary failure seen when switching from tests back to a normal build.

Validation:
- Independent adversarial review approved the final changes. Native producers ran serially; after the selector repair, fingerprint staging → package build → executable verifier passed.
- Coordinator passed19 artifact/release/correctness contract tests and3 worker bundle tests, including poisoned ambient package selection and coherent-but-wrong runtime pairs. Formatting passes.
- Full standard CI passed on a229c5e1b9 before the worker-selector follow-ups; fresh CI is running on the final head.
- Full clean CI preparation34002971096 passed, including actual WASM/NAPI fingerprint comparison. Eight starter app jobs passed. Three later failures concern todo persistence/identity handoff and are being diagnosed separately; one job was still running when this description was updated. Those failures no longer report ABI mismatch.
- Local packed ts-localfirst scaffold/install/build passed; that receipt intentionally skipped browser E2E.

The stack remains unmerged while the remaining starter failures are addressed. No npm release or release-PR merge is authorized by this change.

### DIFF
--- a/.github/workflows/starters-e2e.yml
+++ b/.github/workflows/starters-e2e.yml
@@ -75,6 +75,9 @@ jobs:
       - name: Build core workspace
         run: pnpm run build:core
 
+      - name: Verify starter package artifact hand-off
+        run: node dev/artifacts/verify-starter-e2e-artifacts.mjs
+
       - name: Pack workspace tarballs
         # Mirrors PACKAGES_TO_PACK in run-starter.ts. The harness picks these up
         # via `--tarball-dir` and pins them as `pnpm.overrides` in each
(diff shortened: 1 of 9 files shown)


aeplay:
🤖 The remaining user approval condition is now satisfied: the combined [full production starter matrix](https://github.com/garden-co/jazz/actions/runs/34083853771) passed all 12 scenarios at `ab69ee731a`, which contains the exact approved parent heads. Both parent PRs also have green standard CI at their current heads. Merging only #2606 and #2608 under their existing approvals; draft #2613 remains open for the separate browser cancellation and Android acceptance work.

Issue #2611, opened 2026-09-06
Preserve hybrid account ownership across issuer-scoped signup


aeplay:
🤖 Restore the hybrid starter's intended anonymous → provider signup ownership journey under canonical issuer-scoped identities.

In starter run34002971096, Next and SvelteKit hybrid create a todo locally, successfully sign up with a verified local-first proof, then cannot read that todo from the authenticated client. The proof preserves the subject ID, but the issuer changes from the reserved local-first issuer to the app's external issuer. Under SPEC7 the canonical session.user identities are different. The owner-only $createdBy policy correctly denies the external principal, and persistent storage scopes correctly stay separate.

Commitce9fd4e2e0 replaced the old raw session.user_id usage in starter policies with canonical session.user. The existing hybrid journey and explanatory docs still assume same-sub continuity. React hybrid's passing test may retain its old local-first client, so it must not be treated as proof of a valid cross-issuer handoff.

The release coordinator recommends an explicit app-owned account link established by verified signup proof, preserving canonical authorship and storage isolation. The product decision is pending. Do not fix this by equating issuers, reusing a differently scoped physical store, weakening owner-only access, or rewriting the test to accept data disappearance.

Required acceptance: anonymous todo remains authorized after signup/reload/logout/login for the linked account; unrelated users and unproved identifiers cannot obtain it. Preserve honest offline/pending-write behavior and distinguish local acknowledgement from remote durability. Cover all hybrid frameworks and update their docs consistently.

aeplay:
🤖 Combined starter matrix at account checkpoint `2ccfb4e241` finished 9/12 passing: https://github.com/garden-co/jazz/actions/runs/34080695721 . Next/React/Svelte hybrid, Next/React/Svelte Better Auth except plain TypeScript, and TypeScript/Svelte local-first passed as listed in the run.

Remaining failures:
- Next Local First and React Local First fail recovery-phrase round-trip interactions (Restore remains disabled or the dialog detaches). Their ordinary persistence tests pass. A dedicated lane is investigating the shared recovery UI/lifecycle; no assertion or timeout is being weakened.
- TypeScript Better Auth fails to receive rows after subscription teardown/reopen. Its redundant initial-render fix is now published at `5fbde15e08`, but the underlying public cancellation case remains separately tracked in #2622.

Installed iOS acceptance passed at `2ccfb4e241`; Android writer-read failure is recorded on #1974. These outstanding gates still block release readiness and the parents' conditional merge approvals.

[...]

PR #2612, opened 2026-09-06
Retain the admitted session for direct subscriptions

aeplay:
🤖 Retain the Db's admitted auth session when installing a direct subscription.

The TypeScript Better Auth starter creates a todo but cannot display it after reload. Its auth cookie and JWT remain valid; the subscription is rejected with ServerFailure/Internal. Framework adapters pass their Db session through the subscription orchestrator, while a direct public Db.subscribe() call can reach native subscription installation without that context.

Use the Db's admitted internal session when no read-context, context-session or explicit session was supplied. Preserve those existing precedence rules. This brings the direct public API in line with framework subscriptions; it does not merge identities or alter persistent storage scope.

WIP verification: coordinator reran25 focused runtime tests successfully. Independent adversarial review and actual packed TypeScript starter reload/signin reproduction are running. No storage or wire format changes.

Stacked after #2608 so starter verification includes the repaired native/TypeScript artifact handoff. The separate hybrid ownership decision is tracked in #2611 and is not fixed by this change.

### DIFF
--- a/packages/jazz-tools/src/runtime/db.read-tier.test.ts
+++ b/packages/jazz-tools/src/runtime/db.read-tier.test.ts
@@ -131,6 +131,30 @@ async function settle(): Promise<void> {
 }
 
 describe("Db ReadTier.RemoteIfPossible", () => {
+  it("uses the Db's admitted session for a direct subscription", async () => {
+    const client = makeClient();
+    const session = {
+      issuer: "https://issuer.example",
+      user_id: "alice",
+      claims: { role: "reader" },
+      authMode: "external" as const,
+    };
+    const db = await createDbWithRuntimeSource(
+      {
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Closing as superseded by the account-system integration in #2613. Independent review traced the real production path: `JazzClient.subscribeInternal` already falls back to the admitted session resolved from the copied Db configuration. This extra Db-level fallback is redundant there and can make custom runtime sources disagree with one-shot reads. Its added test checks a mocked call argument, rather than reproducing the reported Better Auth reload failure.

We will retain #2606 and #2608 as the starter-artifact prerequisites and stack #2613 directly above #2608. The full packed starter matrix remains required to verify the original persistence/authentication journeys; closing this PR does not claim those gates have passed.

Issue #2363, opened 2026-08-29
Browser edge read can fail with Protocol: conflicting commit unit for transaction


tobiaslins:
🤖 Alpha.54 verification gate for conflicting commit-unit errors after live authoritative inserts with overlapping relational subscriptions.

## Current understanding

The original stale-replica hypothesis was disproved by a reproduction on a genuinely fresh physical browser root. The merged stack includes canonical maintained bundle-version handling and receiver-local covered-input repairs, but general green CI does not establish the exact reported sequence.

## Remaining replay

On one pinned release candidate, start a fresh persistent browser replica, retain two overlapping subscriptions (including referenced rows on one path), insert a row through the authoritative backend and wait for global durability, then verify the browser consumes that live update and unrelated subsequent reads remain usable. Repeat across fresh roots and reopen/reconnect as appropriate.

Preserve a repository-owned regression if this reproduces. Do not bisect or attribute it to stale assets without new evidence. No dedicated completion PR for this exact replay is currently identified; no new storage-format change is currently proposed.

aeplay:
🤖 I traced this exact invariant. The conflict is what happens when two foreground runtimes mint the same TxId (HLC time + node) for different commit payloads. Preview commit `9ad56bd` already contains #2340, which replaces the historical deterministic foreground node with broker-issued exclusive leases plus high-water handoff.

I verified the preview source with two repository-owned receipts: a real Chromium two-tab same-millisecond write/global-settlement run, and a real server plus independent browser-context disconnect/recovery run. Both pass. A planted duplicate active lease identity reproduces this exact `ConflictingCommitUnit`, so the coverage is sensitive.

That makes a stale browser/SharedWorker asset or previously opened local namespace the leading explanation. Could you retry once in a fresh browser profile (or after closing every tab for the origin and clearing its Jazz site data)? If it still reproduces, please attach whether it happens on the first write/read in that fresh namespace and the browser worker asset URL/fingerprint from the error context; I will keep this issue open until that distinction is proven.

tobiaslins:
I can reproduce this on the newer `c08ec7bb3485abd2e43deeedf6d07799f911cd8d` alpha.54 candidate with a genuinely fresh physical browser replica, so this does not appear to be only a stale pre-lease IndexedDB/SharedWorker asset.

Environment:

- macOS arm64, Electron + Vite renderer
[...]

PR #2613, opened 2026-09-06
feat: account registry and handle-based contexts

aeplay:
🤖 Adds application-scoped accounts, permanent identity assignments, and opaque handles for context creation. Tracks #2611.

**Draft: release blockers remain.** Current head `c275ecabf1` has green CI and a published preview. The coordinator full TypeScript/browser gate passed (Node/browser exit 0, 304 browser tests plus five transport regressions), and installed iOS acceptance passed. Android completed the seed/core observation but failed during restart fixture metadata collection before database admission; #2645 and diagnostic PR #2646 track that investigation. Intermittent persistent subscription/hop failures (#2621) and Local-wait timeout (#2639) remain unexplained. Joe’s offline pending-write recovery fix (#2643 / #2633) has been rebased onto this stack, independently reviewed, and passed a separate coordinator full run; merge awaits approval and green CI. The inherited schema-admission recovery case (#2644) now has a locally passing, independently reviewed fix being tested with Joe’s changes. No merge of #2613 is requested.

The `95cfc8e` pkg.pr.new preview is published and verified from a fresh external install: exact scaffolder snapshot, matching preview SDK, native binding load, app build, both generated Chromium tests (recovery and persistence), and the complete inspector HTML/JS/CSS asset graph passed. The prior `04fbe5f` preview was verified as follows: [package URLs](https://github.com/garden-co/jazz/pull/2613#issuecomment-5567058898). Verification installed the published SDK and scaffolder, loaded its native binding and CLI, built the exact starter snapshot, and passed persistence/recovery in development and production modes. Production requires build-time `VITE_JAZZ_APP_ID` and `VITE_JAZZ_SERVER_URL`; full recovery needs a reachable relay to satisfy graceful synchronization before switching accounts. Offline persistence also passed with the relay unreachable. Explicitly install the matching preview `jazz-tools` dependency in generated starters, whose default dependency remains the normal release.

### Preview and starter configuration

Missing production configuration previously reached account-manager URL construction before the starter's validation, leaving a blank page. Validation now runs first and renders actionable configuration errors. The README distinguishes the dev plugin's injected URL from the explicit production build environment. Tests verify both missing variables and fail if validation is bypassed. Account switching and synchronization semantics are unchanged.

Preview packaging now mirrors normal alpha releases by building and staging the private inspector frontend into jazz-tools before publication. A real packed-tarball gate requires its embedded HTML; deleting that payload makes the gate fail. The earlier preview's inspector route returned 404; normal alpha packaging was already correct. This changes no native artifact identity or storage/wire encoding.

### Subscription source failures (#2642)

A failing readable stream or native read previously rejected a detached reader task without delivering the subscription's terminal error. Both reader paths now route active source failures through the existing once-only error callback. Rejections arriving after cancellation, shutdown, or subscription replacement are suppressed. Terminal resource-cleanup failures are reported without replacing the original failure or preventing later sources from closing; normal unsubscribe/close behavior is unchanged.

The affected runtime suite passed 147 tests with one existing optional performance skip. Five regressions cover readable/native failures, a secondary close failure, and late rejection after cancellation/shutdown. Removing the catches produces missed error callbacks and unhandled rejections. Independent review passed those cases plus existing callback-throw isolation, and the coordinator's full consumer gate passed. This fixes a confirmed error-reporting gap; it is not yet proven to cause the intermittent browser timeout.

### Inspector test isolation (#2641)

Concurrent worktrees previously bound the same relay port, and Playwright could silently reuse another checkout's frontend. Test setup now owns OS-assigned relay and HTTP listeners, mounts Vite middleware and HMR on the owned HTTP server, and passes actual URLs into both inspector fixtures. Existing UI/authentication assertions are unchanged. Normal teardown and setup failures close owned resources while preserving the primary error; the manual dev command retains its default port. Independent review exercised concurrent listeners, HMR shutdown, and planted setup/seeding failures. Both actual concurrent 12-test suites passed, with all four ports closed afterward.

### Remote browser test transport (#2618)

The failing Alpha run now exposes Chromium's actual error: the evaluated promise was garbage-collected while its page remained open without navigation. Harness commands now execute once in a page-owned registry that retains their import, operation, completion and result. Playwright reads synchronous completion snapshots instead of awaiting a long-lived browser promise across CDP. Success, rejection and transport timeout remove the registry entry; page closure destroys its realm. Existing 20/45-second title deadlines remain unchanged within the 60-second transport bound. Removing ownership on timeout does not claim to cancel the underlying operation.

A forced-GC regression runs the same weakly producer-held operation through the old and new transports: the old evaluation is collected, while the retained operation completes once with the expected rows. Planted removal of retention fails the new regression. Timer-retained delivery, error details, missing methods/modules, timeout cleanup and page closure are also covered. These synthetic cases establish the ownership boundary, not the exact GC topology of the original Alpha waiter; the real full browser suite passed after integration, and further validation remains in progress. Product APIs, query assertions, storage and wire formats are unchanged.

### Native foreground transaction clock (#2638)

Before this fix, clean native foreground reuse advanced the storage-owning transaction clock but left the shared synchronous reservation clock behind. A new C ABI write could therefore return exactly its predecessor's transaction ID, leading to a conflicting commit unit. The handoff now advances both clocks monotonically before exposing the foreground. Existing higher reservations are preserved, and overflow fails before either clock is changed. This changes no wire or storage encoding.

The C ABI regression commits before and after clean handoff and requires distinct public transaction IDs. Restoring the old code fails that assertion. Independent review passed, and both the implementation lane and coordinator independently passed all 102 native-relay tests. A separate sibling callback canary passes and fails when delivery is suppressed. The installed test's failure marker also covers a throwing tick, so the previous enabled-only wake trace is consistent with this bug; both installed Android and iOS acceptance subsequently passed at `04fbe5fcb3`.

Browser hook tests now unmount React before closing their databases. Previously nested teardown closed the database before the outer hook unmounted consumers, allowing Suspense retries against a closed database. Query behavior and assertions are unchanged; a mounted-consumer teardown regression fails under the original hook ordering.

### Reserved system authorship

All durable row authors have a non-null UUID account. System writes use account
`00000000-0000-0000-0000-000000000000`, issuer `urn:jazz:system`, and the originating
node UUID as subject. Account equality groups all system edits; whole-author
equality distinguishes their originating nodes. Replication and reopening
preserve that origin rather than substituting the relay's node.

The internal SYSTEM policy-bypass capability remains separate from recorded
authorship. Decoding system provenance cannot mint that capability. External
JWT admission, registration, and linking reject the reserved account/issuer.
Anonymous readers retain a null session account; account-ownership comparisons
fail closed (null equality does not grant ownership), while explicit anonymous
read policies remain available.

This revision includes native author descriptors/codecs and corpus updates,
public TypeScript row-author types, and documentation. Current validation is listed below; earlier installed-device receipts apply only to their stated checkpoints.

### Behavior and invariants

Previously, contexts accepted raw credentials and ownership followed an exact issuer/subject. Switching providers could lose access to existing rows. Contexts now require an `AccountHandle` with a stable account ID and immutable acting identity. One account can admit several identities; an exact `(issuer, subject)` can belong to only one account permanently. Revocation preserves that assignment. Linking never merges accounts or rewrites authorship.

The core is the registry and orders registration, linking, and revocation. Authenticating an external JWT does not register it. Local-first founding is deterministic from the application and verified signing identity and works offline. Linking uses ordinary provider JWTs: admitted A requests an expiring nonce committed to B; B authenticates acceptance. Acceptance checks A's current admission, management authority and generation, B's exact identity and availability, and expiry. Successful retries cannot resurrect revoked admission.

The shared helpers are `createLocalFirst`, `getLoggedIn`, `registerJWT`, `loginJWT`, `linkJWT`, and `logout`. Failed operations preserve selection. Logout fences late asynchronous results. Credential refresh must retain the exact acting identity. Rust owns registry transitions and native identity proof checking; shared TypeScript owns credentials and selection; host adapters own persistence/crypto; React, Vue, Solid, and Svelte observe the shared state.

Public backend `forRequest` verifies the original bearer, resolves its active account through the core registry, and only then initializes a requester context. Unregistered/revoked external identities fail without implicit registration. Explicit `forSession` remains trusted backend impersonation: a configured backend secret may choose an account/principal, but the resulting session still evaluates that chosen author's row policies. Transport credentials and policy trust are separate.

Local-first backup/recovery uses `exportLocalFirstSecret(handle)` and `accounts.restoreLocalFirst(secret)`. Recovery roots stay out of observable snapshots, survive selection changes, and cannot be exported through logged-out handles. Refresh uses the account credential's `getToken` callback in the shared context layer; provider-specific `onJWTExpired` callbacks are removed. A hung credential callback times out after 30 seconds without letting a late result overwrite a later attempt.

### Context lifecycle and examples

Linking happens outside contexts:

1. Stop admitting application work and await `oldContext.shutdown({ waitForSync: true })`.
2. Call `accounts.linkJWT(externalCredential)` independently of any context.
3. Open a new context from the returned handle.

The synchronization barrier is a general shutdown facility. The account manager has no context inventory or special upload scheduler. Ordinary shutdown remains usable offline and promises local durability, not global synchronization.

- A writes under account X, then gracefully closes and links B. B/X can read those rows under `$createdBy.account == user.account`. Exact whole-author equality still distinguishes A from B.
- A failed synchronization barrier leaves the old context and selection usable. If linking fails after shutdown, the unchanged handle can open a new context.
- An account handle retains its registry independently of active transport. Omitted context `serverUrl` means local-only; a supplied URL must match the registry. Identical account UUIDs from different registries cannot alias storage.
- RN same-identity contexts share one persistent relay with independent foreground leases. Closing one preserves its siblings. Rotated credentials update the shared snapshot only after schema/path/transport validation; malformed attachment preserves the live session. Replacement socket resources are prepared before publication and activated only after the old worker stops.
- RN defaults to Local reads, matching browser behavior, even with a configured server. Explicit Remote/Edge/Global requests retain their requested authority and waiting behavior.
- Linked identities share the durable root but cannot own it concurrently under different live identities. The old identity closes before the new one opens.
- Revocation rejects later admissions and operations. Work admitted before revocation may finish; revocation does not erase already disclosed data.

### Wire and storage before / after

| Surface | Before | After |
| --- | --- | --- |
| Row and transaction authors | Canonical author text | Native structured record `{ account: UUID, identity: { issuer, subject } }`. Whole-structure interning is memory-only; native bytes, never intern IDs, persist and travel with rows. |
| Account registry | No account assignment registry | Dedicated durable ordered journal, profile `jazz.account-command.v1`, explicit `JACC` v1 encoding with closed tags, bounded UTF-8, UUID bytes and fixed-width integers. Publication follows flush; recovery rejects malformed/noncontiguous journals. |
| Account control HTTP | No account endpoints | Explicit bearer-authenticated JSON DTOs for registration/login/link/revocation. Edges forward the original user credential. These control DTOs are separate from row/query transport. |
| Author query predicates | Principal-oriented author values | Whole-author equality lowers to account/issuer/subject predicates in the existing query IR; inequality negates the complete conjunction. The nonnull row-author revision removes nullable-account special cases. Dotted `.account` ownership remains a UUID comparison. Query transport remains postcard; row record payloads retain native encoding. |
| Policy-binding directory | Typed claim-node array | Native record `derived_v1: U8, claims_v1: Array<Record>`. Eight presence bits replace only claim values equal to the separately stored identity's canonical derived values. Other portable scalar bindings remain verbatim. Recovery preserves the exact original snapshot, including partial internal bindings. |
| Local account selection | Environment-specific auth state | Versioned JSON preferences retain local signing roots and selected-root metadata via atomic `AccountStore.update`; provider credentials are not persisted. Browser storage uses Web Locks; Expo uses SecureStore with a native OS lock for atomic updates across runtimes/processes. |
| RN account storage root | Principal-scoped native setup | Platform-selected directory plus versioned BLAKE3 filename derived from canonical registry, application, environment and account UUID. JS supplies no path. Exact derivation and a filename fixture are specified. |
| RN foreground ABI | Existing command and metadata variants | General pending-write wait is appended without renumbering existing commands; native session metadata carries foreground node, registry, account and exact principal. Internal account setup/refresh JSON is a process-local JS/native handoff, not a server wire format. |

The eight directory bits, root derivation, and lifecycle contract are spelled out in `crates/jazz/SPEC/ACCOUNT_HANDLES.md`. Runtime policy comparison/window grouping encodes a record's canonical descriptor and payload; durable arrangement/index keys and migration-lens defaults still reject records.

RN uses the shared JS projection for local advisory provider claims. Native admission reconstructs reserved identity bindings; the server independently verifies the bearer and constructs its own claims. No native advisory claim grants remote admission.

The current positive SQLite/RocksDB and browser corpora were regenerated through producer write/reopen checks for the account and nonnull-author representations. Historical negative archives remain untouched.

### Browser persistence and diagnostic attachments

The positive browser corpus now carries a real retained local-first account and its canonical registry owner. The pinned fixture restores that handle, opens without a network connection, verifies historical main/draft rows and large values, appends a current transaction, closes normally, then opens a fresh foreground and reads both generations offline. Historical negative corpora remain unchanged.

This exposed a storage deadlock: a cold query could retain the browser adapter's mutation lock while parked, then the same owner waited on a second storage read. Read operations now release the lock during cold page hydration, discard the speculative result, and retry against the current tree under the lock. A reset epoch prevents stale old-tree errors from escaping after a failed writer replaces its cache. Conditional writes and flush remain serialized. **This changes scheduling only; it adds no wire or storage encoding.**

The inspector attaches through a private, worker-minted account-scoped port. The worker verifies application, physical root, account/registry author fingerprint, and durable owner; the same worker allocates the inspector foreground lease. Parent closure cancels unclaimed allocation, and idle attachments expire. This adds no context inventory or inspector lifecycle machinery to the account manager.

### Installed RN account lifecycle

The installed fixture now creates/restores two genuine local-first AccountHandles through the production Expo manager and SecureStore. Its native adapters supply only validated endpoint/control metadata and receipts; they no longer receive bearer credentials or implement a second admission path. The private test helper uses the same account resolution and native begin/attach functions as public contexts, holding the lease until raw and public aliases close. A raw sibling remains open across public client shutdown and must still tick afterward.

The Rust-generated scope policy compares `$createdBy.account` with `session("user.account")`; row cells contain only the title. Both platforms consume the same generated schema. The seed phase may create missing retained selections; the process-restart phase must restore them and opens without `serverUrl`. No fresh identity or upstream read may satisfy that persistence assertion. Historical logout receipt names describe lease revocation, while credentials remain retained for restart.

RN context setup also uses native OS entropy when Web Crypto is absent. This supplies the existing reserved-session token mechanism; it does not change JWT proof or server admission.

### Immutable request scopes and shared transport

Request-scoped backend reads and subscriptions now retain the exact account,
acting principal, and provider claims in their prepared query and coverage
identity. Concurrent requests from the same principal with different claims do
not mutate one ambient claims slot. Relation reads/subscriptions use the same
immutable preparation path. Changing claims invalidates any earlier compiled
plan; the plan cache distinguishes absent claims from present ones.

A trusted backend may delegate that scope upstream over its authenticated
client link. Ordinary session, authority, and admin links cannot supply this
backend-only binding. Scoped backend permission advice now carries the same immutable account and claims snapshot for each request. Concurrent requests with the same identity and different tenant claims remain independent across reconnect. Support views and receipt evidence bind to the delegated subject; transport authority and connection epochs remain bound to the authenticated backend link. Ordinary callers cannot inject a delegated scope. This addresses #2614.

The unreleased v1 postcard authorization-scope intent adds an optional delegated-session binding, pinned by a byte fixture. Both peers must use this matching release build; the feature bit alone is not a compatible-build guarantee. This adds no storage format and leaves native row encoding unchanged.

All Db facades of a backend context now share actual transport ownership and
explicit offline state. Disconnect stops the shared transport; creating another
facade does not reconnect it. Reconnect uses the original backend credential.
Closing a context is terminal: waiting remote reads reject and existing facades
cannot begin new work. A failed native close retains its failure barrier.

Borrowed React Native contexts retain their verified handle JWT and advisory
claims. Entropy comes from Web Crypto or the native OS source; the native lease
and admitted account remain the authority. This fixes public-client creation
without weakening native admission or enabling identity swaps.

### Subscription continuity during local plan replacement

A live query could publish an empty reset while its replacement execution plan was still loading its first local terminal batch. The following batch restored the rows. In BandChat, this briefly removed the room and profile, unmounted the invitation editor, and discarded the typed guest account before submission. A direct Db subscription reproduces the same disappearing room without React.

The replacement graph now owns its runtime generation while loading and publishes nothing until its first local terminal batch is ready. It then emits one complete successor reset with canonical occurrence identities and removals. A genuinely empty successor still retracts the old rows. This holds only the result already delivered to the existing subscription; fresh subscriptions do not inherit it. It does not wait for remote settlement, change authorization, or add storage/wire formats. Warm replacement error recovery retains its retry behavior.

The original BandChat component and UI assertions are unchanged. A new raw-browser regression fails on the preceding implementation and passes with the fix. A controlled-storage Rust test verifies deferred publication, a mixed add/remove successor, a genuinely empty successor, and a fresh subscription afterward. Removing the cold-readiness gate makes that test fail; restoration passes. Independent adversarial review covered replacement lifetime, reset ordering, occurrence sidecars, and warm-error recovery.

### Partial transaction forwarding (#2619)

A relay may retain only one selected row from a larger transaction. Its stored row count then describes that fragment. Previously, forwarding every locally held row could incorrectly label the fragment `CompleteTransaction`, because the producer compared only row counts. A downstream replica then correctly rejected later rows from the same transaction as a conflicting commit unit.

The producer now also requires the stored transaction to be known complete before emitting `CompleteTransaction`. Partial fragments remain `ViewScoped`; receiver conflict checks, authorization, and durable/wire encodings are unchanged. The regression first ingests a real partial payload and checks its forwarded scope; restoring the old condition fails. The original 51-write RecordPlayer transaction now passes its complete browser scenario. Independent review also verified legitimate complete payloads and reopen-to-complete upgrades.

This prevents new false-complete records. A replica already containing a false-complete record from the old bug cannot repair that record through ordinary re-admission; it requires fresh replica state unless a separate repair is implemented. #2619 links the separately resolved earlier report #2363 and includes Tobi in the discussion.

### Starter server bootstrap (#2620)

The starter harness previously started the server without its schema. That server had no authoritative runtime, so browser workers retried `NotReady/Later` indefinitely. Writes remained locally durable in the upload outbox, and graceful global-sync shutdown correctly kept waiting.

The harness now loads the installed scaffold's schema before server construction. `startLocalJazzServer` accepts supported app/query/raw schema inputs through the existing schema resolver and encoder, while retaining encoded-byte compatibility. A real client/Edge-write test covers all four input forms and startup cleanup. No synchronization guarantee is bypassed. Coherent local Next and Svelte Better Auth/hybrid production flows pass after this fix.

Plain TypeScript Better Auth also had two synchronous session-callback initialization errors. Both are fixed with regression tests that fail under the old ordering, including a real DOM mount. The initial synchronous session replay now also waits for the single final initial render, avoiding a redundant todo subscription teardown; later profile and pending-state notifications still render. Focused tests and independent review cover both behaviors. A separate public subscribe/cancel/resubscribe reproduction exposed a queued evaluation error followed by `ServerFailure: Internal`; #2622 tracks the underlying defect. Three clean production replays pass with the graph-lifetime fix described below.

### Cancellation during subscription refresh (#2622)

Previously, refresh could take a subscription's local maintained handle out of shared state and then wait for the node mutex. Cancellation marked the stream closed and queued finalization. If finalization acquired the mutex first, it removed that handle's sender; refresh later drained the disconnected receiver. Existing refresh-error routing then rejected other live subscriptions, including a replacement, with `ServerFailure: Internal`.

Both drain sites now recheck closed state after acquiring the mutex. A cancelled receiver is skipped if finalization already won; otherwise mutex ownership prevents finalization during the drain. Live-stream error handling is unchanged. A deterministic handoff regression passes with the guard and fails with it removed. This changes no wire/storage format. The separate initial-render optimization is not being used as proof that cancellation works.

The Next/React Local First recovery tests also waited for an input already visible in the outgoing client. Trace timing showed the next action began before replacement/reload. They now wait for actual Next navigation or React outgoing-subtree detachment before asserting the new app and retained data. Product account code and recovery assertions remain unchanged. Both production suites pass 2/2.

Earlier installed Android runs failed at a writer read and at a sibling wake after commit. Allowlisted diagnostics preserve the original assertion and wake-driven polling. Two later installed Android runs and one iOS run passed at `954f9cf9e4`; this improves acceptance evidence but does not establish a causal fix for those earlier intermittent failures.

### Cancellation-safe policy compilation (#2623)

A cancelled policy compiler could leave its recursion marker or temporary point-policy graph in shared compiler state. A later query could then report a false policy cycle or inherit a stale temporary graph. Compiler turns now own leases; expired proof entries and graph replacements are restored before the next cache access or compilation. Genuine live policy cycles still reject. Nested graph replacements restore in stack order, and cache invalidation clears their restoration records.

Independent review audited every production graph-cache reader and invalidation path. Deterministic tests cancel after the proof-stack push and after a temporary point graph is installed. The next request succeeds, reusable graphs recover, and another user remains denied. Removing cleanup makes the graph-restoration canary fail. These are controlled cancellation tests, not an explanation of #2622's distinct browser failure. No wire or storage encoding changes.

Android diagnostics now opt in only on the affected sibling foreground after its initial reset, with fixed bridge-request, scheduling, delivery and callback markers. Tracing is off by default; failed trace toggles cannot replace acceptance results or cleanup. The parser retains only allowlisted markers in that test epoch. This is diagnostic instrumentation, not a wake-delivery fix.

### Graph lifetime during queued evaluation (#2622)

Cancellation previously released subscription retainers and immediately collected ephemeral graph nodes, even when an incremental evaluation still referenced them. Polling also temporarily moves its queue out of the shared slot, making that work invisible to lifecycle collection. The runtime now retains queued nodes and their graph dependencies, defers collection while polling owns that queue, and retries requested collection after restoration. Once pending work finishes, unused nodes are reclaimed. Matching cancelled hydration is removed before releasing its retainers. No wire or storage format changes.

A controlled checksum test proves retention while work is parked and reclamation after drain. On old code, the missing-node failure is scoped and externally invisible when there is no live subscriber, so this test alone does not explain the browser symptom. Three clean production TypeScript Better Auth replays with immediate cancel/reopen passed both persistence tests. A permanent generated-app-only probe now adds a thirteenth starter CI case, preserving the twelve ordinary cases. It asserts replacement row delivery and reports subscription errors; deliberately suppressing replacement rendering makes both tests fail. The coordinator independently reran the permanent probe successfully. Independent adversarial review and the full Groove library suite pass (703 passed, two ignored). The public test checks observable cancellation/reopen health; controlled Rust tests cover precise lifecycle timing.

### Validation and remaining gates

- `04fbe5fcb3`: [full CI passed](https://github.com/garden-co/jazz/actions/runs/34142245107), [installed Android and iOS passed](https://github.com/garden-co/jazz/actions/runs/34142244855), [all 13 production starter scenarios passed](https://github.com/garden-co/jazz/actions/runs/34142284786), and [preview publication passed](https://github.com/garden-co/jazz/actions/runs/34142244984).
- Combined local Node/browser verification also passed: 304 browser tests, one expected failure and three tracked skips, plus WorkOS 4/4, BandChat 6/6 and RecordPlayer 9/9. A bounded additional run subsequently failed the worker Local-wait case described below.
- Native foreground clock: independent review accepted the monotonic update and overflow behavior; coordinator and lane each passed 102 native-relay tests. The C ABI regression fails on the original code.
- Starter configuration and preview packaging changes passed separate independent review. Coordinator checks include nine starter unit tests, five release-config tests, an actual inspector frontend build/stage, and packed HTML verification. Missing configuration and missing packed HTML have planted failing cases.
- #2639: a worker Local wait after disconnect timed out in a combined workload, while its unchanged focused suite passed 16/16. New failure-only diagnostics report submission/Local-fate state and actual timer delay; no timeout or assertion is relaxed and no product fix is claimed.
- Alpha (#2618) and aggregate pagination/FIFO (#2621) remain tracked. The Playwright patch preserves an underlying collected-promise error instead of rewriting it as navigation; it is diagnostic only. A bounded retained-promise experiment did not reproduce Alpha's cause.
- Earlier account and policy gates include full workspace 3,773 passed / 73 skipped, Groove 703 passed / two ignored, schema-bootstrap forms, policy cancellation, and queued-graph retention sensitivity tests. Independent reviews cover account authorization, partial-scope forwarding, artifact provenance, starter startup, schema bootstrap, policy cancellation, and graph lifetime.

#2606 and #2608 merged after their green gates satisfied existing approvals. This draft is based on `main`; #2613 still requires user approval after final integration and gates. Preview publication is authorized, but a release/main merge is not.


aeplay:
🤖 Installed-device acceptance passed on account checkpoint `f48cb4092dfe3373a8e24ad28733c9159b2ce8fe`.

Both the iOS simulator device acceptance and Android device acceptance jobs completed successfully in [run 34051965160](https://github.com/garden-co/jazz/actions/runs/34051965160). These are the installed-app jobs, in addition to the source-scaffold checks.

This PR remains draft: browser relation/policy regression diagnosis and the remaining example/starter account migrations are still in progress. These device receipts apply to the named checkpoint, not subsequent uncommitted changes.

nrainhart (approved):
Storage format changes look reasonable to me 👍

aeplay:
🤖 Review followup from Anselm: replace the account registry's direct OrderedKvStorage journal operations and bespoke JACC command encoding with RecordStore access and Groove typed record encoding. Keep the ordered command journal and account/linking semantics; preserve atomic append, stale-owner fencing, durability barriers, and poison-on-ambiguous-storage-error behavior.

RecordStore needs a small atomic insert-if-absent operation backed by the existing storage primitive. The registry entries will use the existing record/enum value algebra, not wrap the old custom bytes in a record. The separate registry codec/profile and golden fixtures will be updated, with a clear before/after format description. Ordinary row storage and wire encoding are outside this change.

Implementation is assigned and will land on the unified release tip #2595. Anselm also relayed Nico's approval of the storage-freeze-relevant portions of this PR as reviewed so far. That approval predates this new registry encoding request; the new layout will receive its own explicit review.

aeplay:
🤖 The requested account RecordStore migration is now on the unified release tip #2595 at `d3bab317a832f58670e5dd90d325e56136197b7e`.

The registry uses actual Groove RecordStore reads/scans and an atomic insert-if-absent method backed directly by the existing conditional storage primitive. The bespoke JACC values are replaced by the ordinary Groove record/enum encoding. The separate registry journal is v2; old registry roots fail closed. Ordinary row storage and wire encodings are unchanged. #2595's description and ACCOUNT_HANDLES.md spell out the exact storage before/after, closed command descriptor, and pinned corpus.

Independent High review accepted the migration, including atomic winner handling, durability, stale-owner fencing, nonce replay, and old-root rejection. The complete local CI-equivalent gate at the integrated tip passed: 3,793 Rust workspace tests, differential oracle, exhaustive target compile, full concurrent TypeScript/browser suites, 94 RN bridge tests, and native/browser storage compatibility.

Fresh external CI, installed-device acceptance and preview publication are running on that tip. This comment supersedes older integration-status snapshots; it does not claim user merge approval or that every historical intermittent observation has a proven cause. Further review should use #2595 for the final combined result.

PR #2625, opened 2026-09-07
Schedule simulation writes at the requested rate

joeinnes:
## Summary
- Schedule deterministic simulation writes from their absolute index and requested rate.
- Reject a zero rate instead of silently treating it as one edit per second.

## Before
`WriteStream` truncated `1,000 / rate` to whole milliseconds. A rate of 600/s scheduled 1,000/s, rates above 1,000/s put every write at 0ms, and zero was silently clamped.

## After
Each timestamp is `floor(index × 1,000 / rate)`. Fractional and sub-millisecond cadences retain their requested average rate, while zero is rejected. Row and author selection and the public constructor shape are unchanged.

## Test plan
- [ ] `cargo test -p jazz-sim --lib fixture::tests::write_stream_ -- --nocapture`

### DIFF
--- a/crates/jazz-sim/src/fixture.rs
+++ b/crates/jazz-sim/src/fixture.rs
@@ -750,13 +750,21 @@ impl WriteStream {
             !authors.is_empty(),
             "write stream authors must be non-empty"
         );
+        assert!(
+            edits_per_sec > 0,
+            "write stream edits per second must be positive"
+        );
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved #2625 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2628, opened 2026-09-07
Fix bundled DSL schema collection isolation

joeinnes:
## Summary
- Capture side-effect DSL tables inside the bundle evaluating the schema.
- Isolate concurrent and retrying loads with unique temporary directories.
- Preserve explicit-export precedence and deterministic cleanup errors.

## Before
A public `jazz-tools` side-effect import could register tables in a bundled collector while the loader read a different process-level collector, making a valid schema appear empty. Overlapping or retried loads could share state.

## After
A private entry returns module exports and its in-bundle collected schema together. Explicit exports still win, concurrent loads remain isolated, retries start fresh, and cleanup has defined error precedence. Public schema types and storage formats are unchanged.

## Test plan
- [ ] `corepack pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/schema-loader.test.ts`
- [ ] `corepack pnpm --filter jazz-tools build:runtime && corepack pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/schema-loader.test.ts`

### DIFF
--- a/.changeset/bundled-dsl-collection.md
+++ b/.changeset/bundled-dsl-collection.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Isolate bundled TypeScript schema collection between loader runs.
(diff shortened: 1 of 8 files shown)


aeplay:
🤖 Anselm approved #2628 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2630, opened 2026-09-07
Fix typed scalar object where filters

joeinnes:
## Summary
- Classify object-shaped typed `where` values using their column type.
- Lower direct Date, byte-array, JSON-object and array values as equality predicates.
- Preserve recognised operator-map validation.

## Before
Every non-array object was treated as operators. Date filters could disappear, byte arrays became numeric operators, and JSON properties became operator names.

## After
Direct timestamp, byte-array, JSON-object and array values lower to one typed `Eq` condition. Provenance timestamps follow the same path. Recognised operator maps keep existing validation and errors.

## Test plan
- [ ] `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/query-adapter.test.ts`
- [ ] `pnpm --filter jazz-tools check`

### DIFF
--- a/.changeset/quiet-scalar-filters.md
+++ b/.changeset/quiet-scalar-filters.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Classify typed where values from their declared column types so Date, Uint8Array, JSON-object, and array equality filters lower as a single `eq` condition while operator maps retain their existing validation.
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Anselm approved #2630 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

Issue #2497, opened 2026-09-02
c08ec7b: browser WASM query coverage can re-enter a suspended operation


tobiaslins:
🤖 The reported ordinary-query/subscription re-entry bug is fixed by merged #2556 (7e6fccec3ea24e24b907edc53e74fc26854e8eb3), following the earlier #2496/#2553 repairs.

Ordinary NAPI/WASM query preparation and attachment now wait for the existing async owner when another operation is suspended in storage. Subscription handles remain synchronous while pending admission is cancellable. Each operation retains its admitted identity/claims; cancellation and shutdown release provisional resources. This is a runtime coordination change with no storage-format change.

The final repair received independent adversarial review and sealed generated NAPI/WASM/transaction coverage, plus actual Chromium overlapping query/subscription/cancellation receipts and planted regressions. During the overnight release audit on combined candidate f489f375bfaf039e736c823823143d079fb6d340, the core suspended-storage concurrent-read/subscription regression and original synchronous-path reproduction both pass again. The merged fix is an ancestor of that candidate.

Nico's broader request-queue/API simplification remains separate work; this closure addresses the reported high-level ordinary read/subscription failure, not every synchronous low-level API under arbitrary contention. RN's additional API/lifecycle acceptance remains tracked in #2567/#1974 and is included in the release stack.

tobiaslins:
Additional trace from the application renderer confirms the same invariant is also violated while preparing ordinary queries, not only during background coverage refresh.

The first entry path is a normal React subscription:

```text
jazz::db::reads::<impl Db>::prepare_query_bound
wasmdb_prepareQuery
WasmDb.prepareQuery
NativeRuntimeAdapter.prepareQuery
NativeRuntimeAdapter.createSubscription
JazzClient.subscribe
startNativeSubscription
subscribeDelta
subscribeEntry
getCacheEntry
use-all.ts:79
React passive-effect mount
```

The second occurrence is the corresponding one-shot query path:

```text
jazz::db::reads::<impl Db>::prepare_query_bound
wasmdb_prepareQuery
WasmDb.prepareQuery
NativeRuntimeAdapter.prepareQuery
NativeRuntimeAdapter.query
JazzClient.query
```

Both panic at the same `db.rs` invariant:

```text
synchronous node operation ... reentered a suspended operation
```

[...]

PR #2631, opened 2026-09-07
refactor: simplify TS<>Rust DB API

nrainhart:
## Description

Reduces the surface area of the TS<>Rust DB API by unifying similar methods and removing unused ones.

The motivation is two-fold:
- we should strive to keep this API simple and consistent, since future language bindings will need to integrate with it
- reducing the number of cross-runtime operations makes it easier to avoid bugs such as https://github.com/garden-co/jazz/issues/2497
	- will fup with a more comprehensive fix to https://github.com/garden-co/jazz/pull/2556


aeplay (approved):
🤖 Independent review completed at 0c433b9126ba0f836983bfb4460ff02b97bde87a. I checked the unified read/preparation/subscription paths, transaction ownership and cancellation, backend attribution versus permission identity, streaming mutations, and the matching React Native command codecs. No remaining blocking findings from that review.

Fixed the undefined `canMutateMessage` predicate in the simple-auth example; CLI schema/permission validation now passes. The symmetric predicate checks both old and new rows, with announcement changes requiring admin permission.

Validation: 204 focused TypeScript tests passed (1 skipped), and 234 targeted Rust tests passed using the canonical runner's 4 MiB stack setting. The four default-stack overflows reproduced on the PR base as well. A temporary local mutation substituting provenance for the exclusive transaction's permission subject failed the attribution regression as expected; the mutation was fully restored and the same test passed again. The review checkout is clean.

Anselm authorized merging after review and green checks. Holding merge only for CI on this exact head.

PR #2632, opened 2026-09-07
Fix payload enum defaults in schema loading

joeinnes:
## Summary
- Preserve payload-enum column and case-field defaults loaded from `wasmSchema`.
- Reject malformed nested tags and nulls for required payload fields.

## Before
The loader did not decode tagged Enum defaults, losing the column default, and left case-field defaults as tagged runtime values.

## After
Defaults decode recursively according to declared types and re-encode losslessly. Nested tags and nullability are checked; existing scalar, array, JSON and byte default shapes remain unchanged.

## Test plan
- [ ] `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/schema-loader.test.ts`

### DIFF
--- a/.changeset/quiet-enum-round-trip.md
+++ b/.changeset/quiet-enum-round-trip.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Preserve payload-enum column and case-field defaults when loading schemas from `wasmSchema` exports.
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Anselm approved #2632 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2635, opened 2026-09-07
Reject unsupported session claim range operators

joeinnes:
## Summary
- Reject `gt`, `gte`, `lt` and `lte` in `session.where(...)` at definition time.
- Preserve ordinary row range predicates.

## Before
Session range expressions compiled in TypeScript despite core accepting only equality and inequality, then failed later during schema conversion.

## After
Defined session ranges throw a targeted synchronous DSL error. Undefined optional values remain omitted; ordinary row ranges and supported session predicates remain unchanged.

## Test plan
- [ ] `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/permissions/index.test.ts -t "session.*range|ordinary row range"`

### DIFF
--- a/.changeset/reject-session-claim-ranges.md
+++ b/.changeset/reject-session-claim-ranges.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reject unsupported range operators in session claim rules.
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Anselm approved #2635 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2637, opened 2026-09-07
Keep Vue clients stable for equivalent configs

joeinnes:
🤖 Equivalent Vue config replacements retain the pending or ready client, avoiding unnecessary shutdown and fallback flicker. Failed attempts can still be retried with equivalent configuration.

Previously, every deep-watch notification recreated the client, even when only object identity or property order changed. The provider now compares the shared canonical client identity. A genuinely different configuration still replaces the client serially: clear the rendered client, shut down the old client, then create its replacement.

Examples:
- Replace a ready configuration with a new object containing the same values: retain the client and rendered content.
- Replace an equivalent configuration while creation is pending: keep the same pending attempt.
- Creation or handover shutdown fails: an equivalent replacement may retry. The current failed attempt clears its remembered identity.
- An older request fails after a newer request was queued: it must not clear the newer request's identity or publish an obsolete error.

The last two cases address the independent review finding. Public Vue interfaces and caller-owned client behavior are unchanged.

Validation:
- Three real Vue/jsdom regressions cover initialization failure, handover failure, and stale failures preserving newer pending/ready identity.
- Before the fix, two retry tests failed; after the fix, all ten focused provider tests passed, including an independent coordinator rerun on the integrated stack.
- Temporarily clearing identity outside the current-generation guard makes the stale-failure regression fail; the mutation was restored.
- The prior integrated stack passed the full local CI-equivalent gate. This amendment received focused provider verification; fresh CI checks the published revision.

The regression fixtures use an account handle. Both the package check and `tsc --project tsconfig.tests.json` pass; the latter is required because the ordinary package check excludes test files and Vitest transpilation alone does not typecheck them.

### DIFF
--- a/.changeset/vue-config-stability.md
+++ b/.changeset/vue-config-stability.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Keep Vue provider clients stable while mutable configuration changes.
--- a/packages/jazz-tools/src/vue/provider-config-retry.test.ts
+++ b/packages/jazz-tools/src/vue/provider-config-retry.test.ts
@@ -2,6 +2,7 @@
 import { afterEach, beforeEach, expect, it, vi } from "vitest";
 import { createApp, h, nextTick, shallowRef } from "vue";
 import type { AccountDbConfig } from "../accounts/context.js";
+import { makeFakeAccount } from "../react-core/test-utils.js";
 
 const mocks = vi.hoisted(() => ({ createJazzClient: vi.fn() }));
 vi.mock("./create-jazz-client.js", () => ({ createJazzClient: mocks.createJazzClient }));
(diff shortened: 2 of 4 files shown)


aeplay (changes requested):
🤖 Requesting changes: allow an equivalent config replacement to retry after initialization failed.

The new canonical-config guard in `packages/jazz-tools/src/vue/provider.ts` records `lastRequestedConfigKey` before calling `createJazzClient`. If creation rejects, the catch at lines 184–188 sets the error but leaves that key latched. The equality guard around line 213 then suppresses a subsequent equivalent config object even though there is no pending or successfully initialized client.

Concrete example: initial client creation fails transiently; the app retries by assigning `{ driver: { type: "memory" }, appId: "retry" }` again as a new object. The provider never calls the factory again and remains failed. Avoiding needless handovers for an existing/pending client should not prevent recovery from a failed attempt.

Independent review reproduced this with real Vue + jsdom: the factory is called once instead of twice on this stack tip. The same test passes when only provider.ts is replaced with the preceding #2636 version. Please clear the remembered key on failure of the current generation (without clearing a newer request's key), or otherwise distinguish failed requests from pending/successful ones, and add this regression test.

aeplay:
🤖 Anselm approved the intended #2637 change conditional on fixing the reported equivalent-config retry failure directly in this PR. The coordinator is implementing that fix, with focused regression coverage and independent verification before publishing and propagating it through the stack.

aeplay (approved):
🤖 The reported retry regression is fixed in 9e95952276. I independently reviewed the current-generation guard and reran all ten provider tests on the integrated tip. Both initialization and handover failures now permit equivalent-config retries; stale failures preserve newer pending/ready identity. The fail-before and guard-removal mutation receipts confirm test sensitivity. Anselm's conditional approval is satisfied for this fix, with normal CI/integration gates still applying.

PR #2643, opened 2026-09-07
Fix browser pending-write recovery after offline restart

### LINKED ISSUE #2633: Browser IndexedDB: synchronous acceptSubscriber deadlocks on pending-write replay after restart
## Summary

A persistent browser client can acknowledge a local transaction and read it correctly, but hang on the next browser start while recovering pending local writes. The SharedWorker enters the synchronous WASM `acceptSubscriber` export and spins while replay needs an asynchronous IndexedDB page read. The first public query never becomes ready.

This reproduces on **`d964d35186f12625b6c31480e0b0acc946794815`**, package version **`2.0.0-alpha.54`**, using matching prebuilt outputs from [Actions run 34085309289](https://github.com/garden-co/jazz/actions/runs/34085309289). This is not a report against the older `0394059` preview or npm alpha.53.

## Environment and artifact identity

- Linux x64; Chromium **151.0.7922.34**, Playwright **1.62.1**, Bun **1.4.0**, esbuild **0.27.7**.
- Persistent browser driver, IndexedDB, real local Jazz server, signed external test JWT from `startTestJwtIssuer()`, explicit permissive policies for synthetic tables.
- No framework or application code is required. No production data or credentials are involved.
- Matching CI archives: `pkg-jazz-tools` artifact **10005528832**, `pkg-jazz-wasm` **10005529894**, `pkg-jazz-napi` **10005534972**. No local SDK compilation or functional SDK patch was used for the baseline.
- Tools archive SHA-256: `0ee5d83fb1d7e70e8caad2bc2198d8ab1d681c4eb85aac6b76a6357c5cf01025`.
- WASM archive SHA-256: `4693104925f8d93eac27ebdef84f22e996925e0e9f4f3f7268d5a989bb924130`.
- NAPI archive SHA-256: `6060a9e0529947744065d0bf3784278183f277582a7df5e6b1bb39fd7af825ce`.

These directory artifacts contain development/workspace manifest references. For the isolated install, only package metadata was normalized to link the matching prebuilt packages and omit development/lifecycle dependencies; runtime JavaScript, WASM, and native binaries were unchanged. The artifacts expire on 2026-12-06. Use the same source revision and matching runtime cohort when reproducing, since the version string alone does not identify these bytes.

## Reproduction

1. Open a fresh persistent browser profile with the signed test JWT and seed a marker online.
2. Stop the sync server.
3. In one mergeable transaction, insert 500 small integer rows and update the marker. Await `write.wait({ tier: "local" })`.
4. Query locally: marker and all 500 values are correct.
5. Close Chromium and reopen the same profile, keeping the server stopped.
6. `createDb` resolves, but the first `db.all(..., { tier: "local-first" })` does not complete within 30 seconds. Worker admission never finishes.

**Expected:** locally acknowledged data is readable after offline restart, and storage I/O yields to the worker event loop.

**Actual:** the SharedWorker spins inside subscriber admission; the public read waits indefinitely (bounded by the test's external 30-second timeout). This is a recovery/readiness failure; the timeout alone does not demonstrate data loss.

## Controls and measurements

| Case | Same-session read | Read after restart |
|---|---|---|
| One insert, local acknowledgment, offline restart | Passed | Passed |
| Five separate inserts, each locally acknowledged | Passed after each insert | Passed |
| Five-row transaction plus marker update | Passed | Passed |
| 500-row transaction plus marker, pending local writes | Exact values 0–499 and marker verified | Timed out |
| Same 500-row workload, explicit clean `db.shutdown()` | Passed | Still timed out |
| Same 500-row workload, server kept online without explicitly waiting for global acknowledgment | Passed | Still timed out |
| Same 500-row workload, explicitly await **global** acknowledgment before stopping server and restarting offline | 500 rows and marker verified | 500 rows and marker recovered |

The fully synced control suggests the pending-local replay path is decisive. It is not a general workaround for offline use or an unexpected browser close. The small cases do not establish a fixed row threshold: a cold page during replay, rather than a documented 500-row limit, is the relevant trigger.

Two additional 500-row reruns, with **no quota override**, reproduced the failure:

| Origin storage accounting | Unmodified baseline | Page-stage diagnostic run |
|---|---:|---:|
| Before writes | 125,979 bytes | 125,477 bytes |
| After writes | 12,851,343 bytes | 13,272,765 bytes |
| After restart | 12,854,039 bytes | 13,275,461 bytes |
| Browser-reported quota | 7,379,447,808 bytes | 7,379,447,808 bytes |

This is roughly 0.18% of quota. The failure is not explained by exhausting origin capacity.

## Captured failure and independent storage check

A five-second SharedWorker CPU profile and a paused debugger stack show this path (Rust symbol hashes omitted for readability):

```text
handleTabMessage
  acceptPeerWhenIdle
    runWhenCoreIdle
      acceptPeer
        wasmdb_acceptSubscriber
          jazz_wasm::WasmDb::accept_subscriber_with_admitted_identity
            jazz::node::NodeState::commit_unit_for
              jazz::node::currency::query_versions_for_tx
                groove::db::primary_storage::index_scan_raw
                  decode_raw_index_entries_with_storage
                    RecordStore::get_raw / OrderedKvStorage::get
                      idb_tree::IdbTree::hydrate
                        IndexedDbPageStore::read_page
```

The worker remains CPU-active in these storage futures; it is not simply waiting idly for a network connection. Example self-times from the 5.016-second profile: `index_scan_raw` 626 ms, IndexedDB `read_page` 610 ms, WASM subscriber admission 488 ms, `LayoutStorage::get` 473 ms, and `IdbTree::hydrate` 364 ms. These are self-times, not additive inclusive timings for separate operations.

In a separate diagnostic run, messages were inserted immediately before/after subscriber admission and the JavaScript page read, without changing control flow:

```text
probe-accept-start
probe-read-start(pageId=227)
// no probe-read-resolved or probe-accept-done before the timeout
```

While that worker remained stuck, an independent readonly IndexedDB request from the tab retrieved **the same page 227**, returning **8,527 bytes in 2.4 ms**. Thus the page exists and IndexedDB can serve it; completion is stranded on the blocked worker. The diagnostic worker was restored to the unmodified prebuilt bytes afterwards.

## Source-supported diagnosis

1. The browser branch of [`WasmDb::accept_subscriber_with_admitted_identity`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/jazz-wasm/src/lib.rs#L3074) calls the synchronous `db.accept_subscriber_with_claims(...)`.
2. [`accept_subscriber_with_peer`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/jazz/src/db/node_runtime.rs#L2276) calls `crate::db::block_on(self.subscriber_startup(...))` at line 2295.
3. Startup invokes [`restore_local_subscriber`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/jazz/src/db/node_runtime.rs#L817), reconstructing pending commit units and causal ancestors. This matches the captured transaction-version reads.
4. A cold page enters [`IdbTree::hydrate`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/idb-tree/src/lib.rs#L264), then [`JsFuture::from(read_page_js(...)).await`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/idb-tree/src/web.rs#L61).
5. [`block_on`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/jazz/src/db.rs#L1461) polls repeatedly with a no-op waker and `std::thread::yield_now()` on `Pending`. This does not return control to the browser worker event loop, so the IndexedDB promise cannot settle and the loop cannot complete.

The fact that cached tree reads can finish on their first poll explains why resident/small cases can pass while a cold replay hangs.

## Suggested correction and regression coverage

Rust already exposes [`accept_subscriber_with_claims_async`](https://github.com/garden-co/jazz/blob/d964d35186f12625b6c31480e0b0acc946794815/crates/jazz/src/db/node_runtime.rs#L2310). Please route browser admission through an async WASM/JS boundary while preserving identity validation, owner coordination, replay ordering, cancellation, and error propagation. Merely awaiting the existing synchronous export from JavaScript cannot yield while that export is spinning.

The self-signed admission export enters the same synchronous helper and should also be audited; the reproduction here uses an external JWT, so that variant is not independently proven.

Suggested regression cases:
- Cold IndexedDB page during recovery of pending local writes; offline browser reopen; exact recovered rows and marker; worker remains responsive.
- Yielding page-store admission test that cannot pass solely through resident cache hits.
- Multi-tab admission and both external-JWT and self-signed identity paths.
- Keep storage-quota rejection testing separate from recovery liveness.

Related context: #2370 discusses synchronous bridges over suspendable mutation work; #1857 describes a different synchronous-spin trigger. Neither appears to cover this pending-subscriber recovery path.

## Standalone reproduction files

Use an isolated directory containing the matching prebuilt `jazz-tools`, `jazz-wasm`, and `jazz-napi` packages, plus `esbuild@0.27.7` and `playwright@1.62.1`. Install the corresponding Playwright Chromium. The files below use Bun only to serve the fixture; no application code is involved.

Prepare the runtime assets and execute:

```sh
mkdir -p public/jazz-runtime/worker
cp node_modules/jazz-tools/dist/worker/jazz-broker-worker.js public/jazz-runtime/worker/
cp node_modules/jazz-wasm/pkg/jazz_wasm_bg.wasm public/jazz-runtime/
bun repro.mjs
```

The default run stops the server before the 500-row transaction, uses unrestricted storage, and closes the browser without `db.shutdown()`. `JAZZ_CLEAN_SHUTDOWN=1` adds an explicit SDK shutdown before closing; `JAZZ_KEEP_SERVER=1` keeps the server online as a separate control. All identifiers and records below are synthetic.

<details>
<summary>schema.mjs</summary>

```js
import { schema as s } from 'jazz-tools';

export const app = s.defineApp({
  headers: s.table({ version: s.int() }),
  items: s.table({ version: s.int() }),
  summaries: s.table({ version: s.int() }),
});

```

</details>

<details>
<summary>browser.mjs</summary>

```js
import { createDb } from 'jazz-tools';

import { app } from './schema.mjs';

const config = await (await fetch('/connection')).json();
const db = await createDb({
  ...config,
  driver: { type: 'persistent' },
  runtimeSources: {
    baseUrl: `${window.location.origin}/jazz-runtime/`,
    wasmVersion: 'd964d35186f12625b6c31480e0b0acc946794815',
  },
});
const targetId = '018f0000-0000-7000-8000-000000000099';
const local = { tier: 'local-first' };
if (!localStorage.getItem('recovery-seeded')) {
  await db.insert(app.headers, { version: 0 }, { id: targetId }).wait({ tier: 'edge' });
  localStorage.setItem('recovery-seeded', '1');
}

await db.all(app.items, local);
window.probe = {
  shutdown: () => db.shutdown(),
  async read() {
    return {
      version: (await db.one(app.headers.where({ id: targetId }), local))?.version,
      items: (await db.all(app.items, local)).length,
    };
  },
  async write() {
    let timer;
    try {
      const write = await db.transaction((batch) => {
        for (let index = 0; index < 500; index++) batch.insert(app.items, { version: index });
        batch.update(app.headers, targetId, { version: 1 });
      });
      await Promise.race([
        write.wait({ tier: 'local' }),
        new Promise((_, reject) => {
          timer = setTimeout(() => reject(new Error('local wait exceeded 5000ms')), 5000);
        }),
      ]);
      return {
        acknowledged: true,
        ms: performance.now() - started,
        transactionId: await write.txId,
      };
    } catch (error) {
      return { acknowledged: false, ms: performance.now() - started, error: String(error) };
    } finally {
      clearTimeout(timer);
    }
  },
};

```

</details>

<details>
<summary>repro.mjs</summary>

```js
import assert from 'node:assert/strict';
import { mkdtemp, readFile, writeFile } from 'node:fs/promises';
import { tmpdir } from 'node:os';
import { resolve } from 'node:path';

import { build } from 'esbuild';
import { definePermissions } from 'jazz-tools';
import { deploy, startLocalJazzServer, startTestJwtIssuer } from 'jazz-tools/testing';
import { chromium } from 'playwright';

import { app } from './schema.mjs';

// This isolated SDK contract test does not modify a developer or production workspace.
const directory = await mkdtemp(resolve(tmpdir(), 'jazz-recovery-'));
const issuer = await startTestJwtIssuer();
const jazz = await startLocalJazzServer({
  inMemory: true,
  port: 0,
  enableLogs: false,
  jwksUrl: issuer.jwksUrl,
  jwtIssuer: issuer.issuer,
  jwtAudience: issuer.audience,
});
const permissions = definePermissions(app, ({ policy }) => {
  for (const name of Object.keys(app.wasmSchema)) {
    const table = policy[name];
    table.allowRead.always();
    table.allowInsert.always();
    table.allowUpdate.always();
    table.allowDelete.always();
  }
});
const clientToken = issuer.jwtForUser('018f0000-0000-7000-8000-000000000098');
let http;
let browser;
try {
  await deploy({
    schema: app,
    permissions,
    appId: jazz.appId,
    serverUrl: jazz.url,
    adminSecret: jazz.adminSecret,
  });
  const bundle = await build({
    entryPoints: ['browser.mjs'],
    bundle: true,
    format: 'esm',
    platform: 'browser',
    target: 'es2022',
    write: false,
  });
  http = Bun.serve({
    hostname: '127.0.0.1',
    port: 0,
    async fetch(request) {
      const path = new URL(request.url).pathname;
      if (path === '/connection')
        return Response.json({
          appId: jazz.appId,
          serverUrl: jazz.url,
          jwtToken: clientToken,
        });
      if (path === '/probe.js')
        return new Response(bundle.outputFiles[0].text, {
          headers: { 'content-type': 'text/javascript' },
        });
      if (path.startsWith('/jazz-runtime/') && !path.includes('..'))
        return new Response(Bun.file(resolve('public', path.slice(1))));
      return new Response('<script type="module" src="/probe.js"></script>', {
        headers: { 'content-type': 'text/html' },
      });
    },
  });
  browser = await chromium.launchPersistentContext(resolve(directory, 'profile'), {
    headless: true,
    args: ['--no-sandbox'],
  });
  const page = await browser.newPage();
  page.on('console', (message) => process.stderr.write(message.text() + '\n'));
  page.on('requestfailed', (request) =>
    process.stderr.write(request.url() + ' ' + request.failure()?.errorText + '\n'),
  );
  page.on('pageerror', (error) => process.stderr.write(`${error.message}\n`));
  await page.goto(`http://127.0.0.1:${http.port}`);
  await page.waitForFunction(() => Boolean(window.probe));
  const session = await browser.newCDPSession(page);
  const origin = `http://127.0.0.1:${http.port}`;
  const before = await session.send('Storage.getUsageAndQuota', { origin });
  const quotaBytes = null;
  if (quotaBytes !== null)
    await session.send('Storage.overrideQuotaForOrigin', { origin, quotaSize: quotaBytes });
  if (process.env.JAZZ_KEEP_SERVER !== '1') await jazz.stop();
  const result = {
    diagnostic: true,
    sourceCommit: 'd964d35186f12625b6c31480e0b0acc946794815',
    serverKeptOnline: process.env.JAZZ_KEEP_SERVER === '1',
    cleanShutdown: process.env.JAZZ_CLEAN_SHUTDOWN === '1',
    sdk: JSON.parse(await readFile('node_modules/jazz-tools/package.json', 'utf8')).version,
    browser: browser.browser()?.version(),
    before,
    quotaBytes,
    initial: await page.evaluate(() => window.probe.read()),
    write: await page.evaluate(() => window.probe.write()),
    sameSessionRead: await page.evaluate(() => window.probe.read()),
    after: await session.send('Storage.getUsageAndQuota', { origin }),
  };
  await writeFile(resolve(directory, 'evidence.json'), JSON.stringify(result, null, 2));
  if (process.env.JAZZ_CLEAN_SHUTDOWN === '1') await page.evaluate(() => window.probe.shutdown());
  await browser.close();
  browser = await chromium.launchPersistentContext(resolve(directory, 'profile'), {
    headless: true,
    args: ['--no-sandbox'],
  });
  const reloaded = await browser.newPage();
  reloaded.on('console', (message) => process.stderr.write('reload: ' + message.text() + '\n'));
  reloaded.on('pageerror', (error) => process.stderr.write('reload: ' + error.message + '\n'));
  await reloaded.goto(process.env.JAZZ_KEEP_SERVER === '1' ? origin : origin + '/?offline=1');
  try {
    await reloaded.waitForFunction(() => Boolean(window.probe));
    result.offlineReload = await reloaded.evaluate(() => window.probe.read());
  } catch (error) {
    result.reloadError = String(error);
  }
  result.acknowledgedWriteSurvived =
    !result.write.acknowledged ||
    (result.offlineReload?.version === 1 && result.offlineReload?.items === 500);
  await writeFile(resolve(directory, 'evidence.json'), JSON.stringify(result, null, 2));
  process.stdout.write(JSON.stringify(result, null, 2) + '\n');
  assert.equal(
    result.acknowledgedWriteSurvived,
    true,
    'A locally acknowledged write must remain readable after offline browser restart',
  );
} finally {
  await browser?.close();
  await http?.stop();
  await issuer.stop();
  await jazz.stop();
}

```

</details>

joeinnes:
🤖 Fixes #2633: a persistent browser client could acknowledge a 500-row transaction locally, read it correctly, and then hang after a full offline browser restart. Synchronous WASM subscriber admission blocked the event loop while pending-write recovery needed IndexedDB I/O.

Both WASM admission exports now await the existing core recovery operation. The adapter reserves evaluator ownership throughout admission; worker attachment stays within the auth transition queue, validates current account/storage/Inspector binding, and retires late transports after close, replacement, or invalidation. A failed pump constructor removes its callbacks so a later wake cannot disrupt a successful retry. Direct jazz-wasm callers must await acceptSubscriber and acceptSubscriberWithSelfSignedProof. Storage and wire encodings do not change.

This candidate is rebased onto the current account-system stack (#2613), preserving Joe's separate regression and implementation commits. Account API adaptations are test-only:

- The local-first cold-restart case restores a real opaque AccountHandle and uses public createDb.
- The external-identity protocol case performs genuine online enrollment and captures the genuine registry login response. A strictly scoped internal account-manager fetch stub replays only that exact login URL, POST method, and bearer credential; normal loginJWT then mints the opaque handle used by public createDb. This isolates database recovery from account-registry availability. It does not claim supported public external login while the registry is offline, or test offline revocation freshness.
- Both cases stop the real database server and issuer completely, restart the entire browser with the same profile, and require exactly values 0–499 plus marker 1. No row/query/transport response is stubbed.

Validation:

- Independent High review of the admission/ownership changes: 87 focused tests passed; four planted ownership, publication, auth-ordering, and callback-cleanup defects failed. Adapted worker review: 88 tests passed; final registry-stub source review clear.
- Final lane: both cold-restart cases passed; 235 adapter/worker tests passed (one existing optional skip); pinned-storage corpus passed.
- Restoring the original synchronous WASM implementation makes the public local-first recovery case fail its unchanged 30-second read deadline. Restored fixed implementation passes.
- Coordinator independently built matching artifacts and ran full TypeScript/browser consumers at a6c62ef9eb: Node and browser exit 0; 306 browser tests plus five transport regressions passed. The inherited expected-failure case is tracked/fixed separately in #2644.
- Exact rebased-head CI pending. The old branch's CI failed the pinned offline corpus on its older base; the adapted corpus and full run pass. No claim that this fix resolves the separately tracked intermittent hop or Local-wait failures.


aeplay:
🤖 Additional combined verification caught an unresolved failure. Coordinator full run at 0b17c853f0 (this recovery candidate plus #2648) passed Node tests and 306 browser cases, but the external registry-stub recovery case failed with “Timed out waiting for edge query coverage”. The first helper progress marker is emitted only after offline writes and their same-session read, so its absence does not identify whether this was seed, initial read, or offline same-session read. A lane is adding exact phase evidence without changing deadlines or issuing extra queries/wakes.

The preceding a6c62ef9eb candidate passed the coordinator full run, and its rebased CI partitions have passed. Neither observation establishes the cause of the combined failure. Merge readiness remains held while this is investigated; no timeout increase or retry-until-green is proposed.

PR #2682, opened 2026-09-08
Align profiled read materialization

joeinnes:
## Summary
- Keep profiled reads aligned with ordinary reads when the query is not selected for profiling.
- Preserve the same materialised rows and metadata across the two read paths.

## Before
The profiled read path could materialise an unselected query differently from the ordinary read path.

## After
Unselected profiled reads follow the ordinary materialisation semantics while retaining profiling support for selected queries.

## Test plan
- [ ] Review `profiled_read_matches_ordinary_read_for_unselected_query` in `crates/jazz/src/db/tests/reads.rs`.
- [ ] Run the focused Jazz read tests.

### DIFF
--- a/crates/jazz/src/db/tests/reads.rs
+++ b/crates/jazz/src/db/tests/reads.rs
@@ -631,6 +631,22 @@ fn filtered_root_prepared_query_still_reads_without_preinstalled_plan() {
         vec![row(1)]
     );
 }
+#[test]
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved #2682 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2683, opened 2026-09-08
Reject variable spans inside record headers

joeinnes:
## Summary
- Reject record headers whose variable-field span starts before the variable payload.
- Keep structural validation aligned with full record decoding.

## Before
A malformed record header could point a variable-field span into the fixed header area and reach lookup logic with an invalid payload boundary.

## After
Record validation rejects the impossible span before variable-field lookup, matching the full decoder's safety boundary.

## Test plan
- [ ] Review `lookup_rejects_offset_before_variable_payload_start` in `crates/groove/src/records/tests.rs`.
- [ ] Run the focused Groove records tests.

### DIFF
--- a/crates/groove/src/records/mod.rs
+++ b/crates/groove/src/records/mod.rs
@@ -1667,7 +1667,7 @@ fn record_value_span_for_layout(
             } else {
                 u32_to_usize(read_u32_at(record, fixed_size + variable_idx * 4)?)?
             };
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved #2683 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2686, opened 2026-09-08
Return empty reversed storage ranges

joeinnes:
## Summary
- Treat reversed storage scan bounds as empty ranges across memory, IndexedDB, SQLite, and RocksDB adapters.
- Preserve the ordered-storage contract for index-backed reads.

## Before
Reversed start and end bounds could reach backend scans with inconsistent behaviour instead of returning an empty range.

## After
Storage adapters reject impossible reversed ranges consistently before scanning, and index reads inherit the same ordered semantics.

## Test plan
- [ ] Review `reversed_range_apis_return_empty` and the storage scan coverage.
- [ ] Run the focused Groove storage and index tests.


aeplay:
🤖 Anselm approved #2686 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2687, opened 2026-09-08
Serve SPA deep links from Hono starters

joeinnes:
## Summary
- Serve the SPA shell for root and deep-link requests from the React and TypeScript Hono starters.
- Keep API and static-asset routing ahead of the fallback.

## Before
Production starter servers could return a missing page for a client-side route that was not the root path.

## After
The four affected starters return the SPA shell for valid client-side deep links while preserving API and asset responses.

## Test plan
- [ ] Review the production-server deep-link smoke test in each affected starter.
- [ ] Run the starter production E2E checks with `JAZZ_E2E_PROD=1`.


aeplay:
🤖 Anselm approved #2687 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2688, opened 2026-09-08
Singleflight server JWKS refreshes

joeinnes:
🤖 Concurrent authentication requests share one server JWKS refresh rather than independently fetching the same public keys. Refresh budgeting and cooldown remain shared, and completion is retained for every request that joined the refresh.

Before, simultaneous authentication misses could start duplicate fetches and consume refresh capacity independently. The middleware now lets one request own the fetch while others wait for its result. Successful fetches, failed fetches, and owner cancellation all resolve the joined requests through the same completion path.

A joined request may be paused before it begins listening. If the owner finishes during that pause, completion must remain available: the waiter must immediately observe the fetched keys or error when it resumes, rather than wait for a notification that already happened. The implementation retains the outcome even when no receiver is subscribed at publication time. Multiple late waiters receive the same result. This fixes the independent review finding tracked in #2719.

Authentication verification rules, refresh budget, cooldown, and stale-key policy are unchanged. There is no new storage or wire format.

Validation:
- Three deterministic regressions force join-before-subscribe ordering through real owner HTTP success, HTTP failure, and cancellation paths. All three time out before the fix and pass after it.
- Each regression checks repeated late subscriptions observe the same completion, the active flight is cleared, and only one HTTP fetch occurred.
- The full server-auth module passes all 31 tests, including existing coalescing and cooldown checks, both in the implementation lane and an independent coordinator run on the integrated stack.
- Fresh integrated CI remains the final gate for the amended stack.

### DIFF
--- a/crates/jazz-server/src/middleware/auth.rs
+++ b/crates/jazz-server/src/middleware/auth.rs
@@ -317,7 +317,9 @@ impl JwksRefreshFlight {
     }
 
     fn publish(&self, result: Result<JwkSet, String>) {
-        let _ = self.result.send(Some(result));
+        // A caller can join before subscribing. Retain completion even when no
+        // receivers exist yet, including fetch errors and owner cancellation.
+        self.result.send_replace(Some(result));
     }
 }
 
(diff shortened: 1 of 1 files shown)


aeplay (changes requested):
🤖 Requesting changes: retain refresh completion when a waiter has joined but has not subscribed yet.

`JwksRefreshFlight::new` drops the initial watch receiver. In `load`, joining the shared flight and subscribing in `wait()` are separate operations. Another thread can finish the owner between them. `publish()` currently uses `watch::Sender::send` (auth.rs:320–321), which returns an error without retaining the value when there are no receivers. The late waiter then subscribes to `None` and waits indefinitely; its Arc also keeps the sender alive, so channel closure cannot rescue it.

A deterministic independent-review test, `review_jwks_completion_survives_late_waiter_subscription`, publishes after joining but before subscribing and fails by timeout on eeccaa11. Root independently checked the implementation and the channel behavior.

Please retain the completed value regardless of receiver count (e.g. `send_replace`) and add regression coverage for success/error/cancellation delivered before late subscription, alongside ordinary concurrent waiters. This is a coordination bug, not a reason to relax JWT validation. Tracked in https://github.com/garden-co/jazz/issues/2719.

aeplay:
🤖 Anselm approved the intended #2688 change conditional on fixing the reported late-waiter completion loss directly in this PR. The coordinator is implementing the fix with a deterministic regression and independent verification before publishing and propagating it through the stack.

aeplay (approved):
🤖 The late-waiter finding is fixed. I independently reviewed retained completion and ran all 31 auth tests on the integrated stack. Success, fetch failure, and owner cancellation now remain observable to late subscribers; all three regressions fail before the fix. Existing refresh-budget checks pass. The complete local CI-equivalent gate also passed on the combined Vue/JWKS/Next-test tip. Anselm's conditional approval is satisfied for this fix; normal CI and stack merge gates still apply.

PR #2692, opened 2026-09-08
Preserve SQL NULL join semantics

joeinnes:
## Summary
- Preserve SQL NULL semantics for nullable inner-join keys.
- Apply the same rule to prepared parameter equality joins.

## Before
SQL-lowered inner joins and prepared nullable-parameter equality could match rows whose join keys were both `NULL`.

## After
Rows with `NULL` in any nullable join-key position are excluded before equality matching; non-`NULL` equality remains unchanged.

## Test plan
- [ ] Review the nullable inner-join and prepared nullable-parameter regressions.
- [ ] Run the focused Groove graph and parameterised-subscription tests.


aeplay:
🤖 Anselm approved #2692 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2701, opened 2026-09-09
fix: hydrate JSON storage cells in relation includes

### LINKED ISSUE #2689: Browser relation includes fail with graph output descriptors do not match on c31f623
## Observed failure

On published `c31f62306fbb169ad7ab42de3d276f2bc00d4211` browser artifacts, a local `db.one()` succeeds for a check header but throws `Storage: graph output descriptors do not match` when adding either its summary relation or reverse item relation. Projecting fewer fields does not avoid it.

This artifact is browser-equivalent to inspected main `6d352663f8e03278b0007752e27213a6062d5917`; the intervening changes are server WebSocket diagnostics/tests. This report pins the tested artifact, not a moving latest-main claim.

## Runnable reproduction

[Standalone reproduction, complete schema, synthetic seed, and pinned packages](https://gist.github.com/husseinraoouf/b97b2d28374d099b98deeb486bb0900a). Clone/download the gist, then on Linux x64 with Bun:

```sh
bun install
bunx playwright install chromium
BENCH_VERSION=main BENCH_MODE=flat BENCH_QUERY_PROBES=1 BENCH_CYCLES=3 bun run.mjs
```

Read `queryShapes` in the emitted temporary `result.json`. This portable reproduction was run successfully outside the application checkout and reproduced all four relation failures. It uses Chromium 151.0.7922.34, persistent browser storage, a local-first account, and permissive policies on the fixture tables. Seed/hydrate online first, then stop the matching fixture server and query offline.

The serialized schema is the complete application schema; 25 synthetic rows occupy seven tables. No production data/credentials are included. The schema trigger has **not** been minimized, and ancillary references are placeholders. Keeping the full schema is intentional until the trigger is understood.

Equivalent query shapes in `probeQueryShapes`:

```js
const q = app.check.where({ id: checkId });
await db.one(q, { tier: 'local' }); // succeeds
await db.one(q.include({ latestCheckSummary: true }), { tier: 'local' }); // fails
await db.one(q.include({ check_itemViaCheck: true }), { tier: 'local' }); // fails
```

Combined includes and projected combined includes also fail. These probes run outside a write transaction, so the trigger is not limited to transaction-local reads. Four flat queries return the expected check, ten items, and summary; repeated in-place updates and offline restart pass using that workaround.

## Expected / investigation

Includes should return the available local related records, or a specific supported API/schema validation error if this shape is invalid, rather than a storage descriptor mismatch. Please isolate descriptor/schema/graph lowering and add a browser regression retaining the relevant schema/column trigger. No root cause is claimed.

Deduplication: searched open/closed issues for the exact error and descriptor/include reports; reviewed related historical descriptor PRs including #2558 and #1591. No matching existing issue was found. #2379 concerns transaction-local asynchronous reads; this failure also occurs outside transactions.

### LINKED ISSUE #2699: Native reverse include fails when parent has an unselected JSON column (two-table repro)
## Minimal regression

Adding an unselected JSON column to a parent table makes a reverse include fail in the native backend SDK. One parent and one child are sufficient; all referenced rows exist. Flat reads of both tables succeed. Removing the JSON column makes the include succeed.

Error: `Protocol: invalid stored value: binding value does not match its descriptor`.

Tested main source: `90af64c1231beda253970aff5edf6082da71ae9e`. Matching SDK/native/server artifacts: `2.0.0-alpha.55` from [preview run 34280043435](https://github.com/garden-co/jazz/actions/runs/34280043435), build commit `1886e065d23b166ac689735a7739aeb0817ec9c2`. That commit is one version/changelog-only commit ahead of the source; runtime code is identical. This pins the tested build, not a moving latest-main claim. Baseline: `2.0.0-alpha.53`. Linux x64, Node 26.2.0, shared host with 8 available CPUs.

| Version | Server storage | Parent JSON column | Result |
|---|---|---|---|
| alpha.53 | persistent | present | passes |
| alpha.53 | memory | present | passes |
| pinned main artifact | persistent | present | fails |
| pinned main artifact | memory | present | fails |
| pinned main artifact | memory | absent | passes |

## Standalone reproduction

Save as `jazz-include-repro.mjs`. It imports no application code and creates only synthetic records in a disposable local server with generated credentials. Exit 1 indicates the error; normal failure cleanup closes the session and server.

Download matching `pkg-jazz-tools`, `pkg-jazz-napi`, and `pkg-jazz-wasm` from that run, and resolve the packages together (including the Linux x64 GNU NAPI binary). Do not mix a new SDK with an older native package. The preview package is a CI artifact, not an assumption that alpha.55 is on npm.

```sh
# Installed alpha.53 control:
node jazz-include-repro.mjs --memory
# Matching main SDK/native packages from the pinned CI artifacts:
node jazz-include-repro.mjs /path/to/pkg-jazz-tools --memory
# Same main build; removing only the JSON column makes it pass:
node jazz-include-repro.mjs /path/to/pkg-jazz-tools --memory --without-json
# Omit --memory to repeat with persistent server storage.
```

```js
// Standalone upstream reproduction: no Cosmos data, credentials, or imports.
// node jazz-include-repro.mjs /path/to/jazz-tools [--memory] [--without-json]
import { randomUUID } from 'node:crypto';
import { dirname, resolve } from 'node:path';
import { fileURLToPath, pathToFileURL } from 'node:url';

const root =
  process.argv[2] && !process.argv[2].startsWith('--')
    ? resolve(process.argv[2])
    : dirname(fileURLToPath(import.meta.resolve('jazz-tools/package.json')));
const { schema: s } = await import(pathToFileURL(`${root}/dist/index.js`).href);
const backend = await import(pathToFileURL(`${root}/dist/backend/index.js`).href);
const { deploy, startLocalJazzServer } = await import(
  pathToFileURL(`${root}/dist/testing/index.js`).href
);
const withJson = !process.argv.includes('--without-json');
const app = s.defineApp({
  parents: s.table({ name: s.string(), ...(withJson ? { metadata: s.json() } : {}) }),
  children: s.table({ parentId: s.ref('parents'), name: s.string() }),
});
const server = await startLocalJazzServer({
  appId: randomUUID(),
  inMemory: process.argv.includes('--memory'),
});
let close;
const timer = setTimeout(() => {
  console.error('Reproduction timed out');
  process.exit(1);
}, 30000);
try {
  await deploy({
    serverUrl: server.url,
    appId: server.appId,
    adminSecret: server.adminSecret,
    schema: app,
    permissions: {},
  });
  const common = {
    appId: server.appId,
    serverUrl: server.url,
    app,
    permissions: {},
    driver: { type: 'memory' },
    tier: 'local',
    env: 'test',
  };
  let db;
  if (backend.createJazzSession) {
    const session = await backend.createJazzSession({
      ...common,
      initial: { backendSecret: server.backendSecret },
    });
    close = () => session.close();
    db = session.getSnapshot().client?.db;
  } else {
    const context = backend.createJazzContext({
      ...common,
      adminSecret: server.adminSecret,
      backendSecret: server.backendSecret,
    });
    close = () => context.shutdown();
    db = context.asBackend(app);
  }
  if (!db) throw new Error('Session not ready');
  const parent = await db
    .insert(app.parents, { name: 'Parent', ...(withJson ? { metadata: {} } : {}) })
    .wait({ tier: 'global' });
  await db.insert(app.children, { name: 'Child', parentId: parent.id }).wait({ tier: 'global' });
  console.log('flat parents:', (await db.all(app.parents, { tier: 'edge' })).length);
  console.log('flat children:', (await db.all(app.children, { tier: 'edge' })).length);
  const rows = await db.all(
    app.parents.select('id').include({ childrenViaParent: app.children.select('id') }),
    { tier: 'edge' },
  );
  if (rows.length !== 1 || rows[0].childrenViaParent?.length !== 1)
    throw new Error('Incorrect included rows');
  console.log('included children:', rows[0].childrenViaParent.length);
} catch (error) {
  // This fixture is entirely synthetic; the diagnostic is intended for an upstream issue.
  console.error(error instanceof Error ? error.message : 'Unknown error');
  process.exitCode = 1;
} finally {
  await close?.();
  await server.stop();
  clearTimeout(timer);
}
process.exit(process.exitCode ?? 0);

```

## Expected and investigation

The projected parent ID and its child ID should be returned, regardless of an unrelated JSON column. The error emerges from [native binding hydration validation](https://github.com/garden-co/jazz/blob/90af64c1231beda253970aff5edf6082da71ae9e/crates/jazz/src/node/state/lifecycle.rs#L1391). The JSON trigger is isolated; the exact descriptor-construction defect is not yet established. Please add a native SDK regression covering JSON parent columns, including projections omitting JSON, in both storage modes.

Related: #2689 reports browser includes failing with a **different** graph-output descriptor error on an earlier pinned build and a larger schema. Both have successful flat reads and failed includes; this native two-table reproduction may help narrow that investigation. We have not proved a shared root cause; feel free to merge/link if appropriate. The separately measured native latency regression also occurs with successful product graph queries, so it should not be assumed to follow from this exception.

husseinraoouf:
JSON relation includes now return their requested rows instead of `Protocol: invalid stored value: binding value does not match its descriptor`, including when a parent JSON column is omitted by `select("id")`.

Collectors retain JSON's physical storage descriptor. Binding hydration handled logical strings but rejected that descriptor. This change accepts inline JSON and materializes indirect JSON through the existing asynchronous hydration path. Other descriptor/value mismatches still fail; query semantics and storage/wire encodings are unchanged.

For example, one parent with `metadata: {}` and one child now returns both IDs through a projected reverse include. Full includes also return a 100 KB JSON value. Browser forward/reverse includes continue to work with the server stopped and after persistent reopen.

Fixes #2699. Fixes #2689.

Related: #1844, #1966, #2402.

The older browser report used a different graph-descriptor error. With the matching pre-fix packages from preview run 34280043435 (runtime source `90af64c12`), its original full-schema fixture instead fails all four include probes with the JSON hydration error. The same fixture passes all four probes, update cycles, and offline restart with this change. This does not claim that the older graph error had the same underlying cause.

Validation:

- Jazz core library: 1,914 passed, 2 existing ignored tests.
- All three incremental-delivery canaries passed; the ignored differential oracle also passed with seed 11 and churn depths 10/1,000.
- Native SDK regression passed with memory and persistent server storage, projected/full includes, and large JSON.
- Persistent browser regression covers forward, reverse, and projected includes offline and after reopen.
- Original browser reproduction compared against matching published pre-fix packages.
- Jazz Tools build, TypeScript check, Clippy (`-D warnings`), and formatting/lint hooks passed. Full CI-equivalent gate was not run locally; the optional private sensitive-data guard was unavailable.

## Post-Deploy Monitoring & Validation

On the next preview build, rerun the two linked standalone reproductions. Includes should return the expected rows; either reported descriptor error is a failure signal.

### DIFF
--- a/crates/jazz/src/db/tests/reads.rs
+++ b/crates/jazz/src/db/tests/reads.rs
@@ -1989,6 +1989,75 @@ fn relation_snapshot_reverse_array_skips_deleted_children_with_camel_case_ref() 
     );
 }
 
+/// Alice reads a projected reverse include whose parent has an unrelated JSON cell.
+/// The binding hydration boundary must expose the parent and child without physical types.
+#[test]
+fn relation_snapshot_json_parent_binding_hydration() {
+    assert_relation_snapshot_json_binding_hydration(false);
+}
+
+/// Alice includes a child with JSON, so collector anchor and child arms must
+/// agree on the physical JSON descriptor before binding hydration.
+#[test]
+fn relation_snapshot_json_child_binding_hydration() {
+    assert_relation_snapshot_json_binding_hydration(true);
+}
+
+fn assert_relation_snapshot_json_binding_hydration(child_json: bool) {
+    let mut children = PublicTableSchemaBuilder::new("children").fk_column("parentId", "parents");
+    if child_json {
+        children = children.column("metadata", PublicColumnType::Json { schema: None });
+    }
+    let schema = build_public_db_test_schema(
+        PublicSchemaBuilder::new()
+            .table(
+                PublicTableSchemaBuilder::new("parents")
+                    .column("name", PublicColumnType::Text)
+                    .column("metadata", PublicColumnType::Json { schema: None }),
+            )
+            .table(children),
+    );
+    let db = open_db(0xc2, AuthorSubject::for_test_bytes([0xc2; 16]), &schema);
+    // Db's binding-facing API accepts core cells; row_input! is for JazzClient's
+    // public Value algebra and cannot represent this lower-level input type.
+    let parent = db
+        .insert(
+            "parents",
+            BTreeMap::from([
+                ("name".into(), Value::String("alice".into())),
+                ("metadata".into(), Value::String("{}".into())),
+            ]),
+            Default::default(),
+        )
+        .unwrap()
+        .row_uuid();
+    let mut child_cells = BTreeMap::from([("parentId".into(), Value::Uuid(parent.0))]);
+    if child_json {
+        child_cells.insert("metadata".into(), Value::String("{}".into()));
+    }
+    let child = db
+        .insert("children", child_cells, Default::default())
+        .unwrap()
+        .row_uuid();
+    let children = ArraySubquery::new("childrenViaParent", "children", "parentId", "id");
+    let children = if child_json {
+        children
+    } else {
+        children.select(["id"])
+    };
+    let query = Query::from("parents")
+        .select(["id"])
+        .array_subquery(children);
+    let prepared = db.prepare_query(&query).unwrap();
+    let mut snapshot = block_on(db.all_relation_snapshot(&prepared, ReadOpts::default())).unwrap();
+    block_on(db.hydrate_rows_for_binding(&mut snapshot.rows)).unwrap();
+    assert_eq!(row_ids(&snapshot.rows), vec![parent]);
+    assert_eq!(
+        terminal_nested_values(&snapshot, parent, "childrenViaParent", "row_uuid"),
+        vec![Value::Uuid(child.0)]
+    );
+}
+
 #[test]
 fn relation_snapshot_reverse_array_reads_local_nullable_ref_child() {
     let schema = build_public_db_test_schema(
--- a/crates/jazz/src/node/state/lifecycle.rs
+++ b/crates/jazz/src/node/state/lifecycle.rs
@@ -1399,6 +1399,26 @@ where
 
         Box::pin(async move {
             match value_type {
+                // JSON remains string-shaped to callers, but query collectors
+                // retain its distinct storage codec through the binding boundary.
+                ValueType::Internal(_)
+                    if *value_type == groove::large_values::physical_storage_value_type(
+                        groove::large_values::LargeValueKind::Json,
+                    ) =>
+                {
+                    match value {
+                        Value::String(_) => Ok(()),
+                        Value::Large(value_ref)
+                            if value_ref.kind == groove::large_values::LargeValueKind::Json =>
+                        {
+                            *value = self.materialize_large_value(value_ref).await?;
+                            Ok(())
+                        }
+                        _ => Err(Error::InvalidStoredValue(
+                            "binding JSON value does not match its descriptor",
+                        )),
+                    }
+                }
                 ValueType::String => match value {
                     Value::String(_) => Ok(()),
                     Value::Large(value_ref)
--- a/packages/jazz-tools/src/backend/json-include.integration.test.ts
+++ b/packages/jazz-tools/src/backend/json-include.integration.test.ts
@@ -1,0 +1,63 @@
+import { randomUUID } from "node:crypto";
+import { describe, expect, it } from "vitest";
+import { schema as s } from "../index.js";
+import { deploy, startLocalJazzServer } from "../testing/index.js";
+import { createJazzSession } from "./index.js";
+
+describe("JSON parent includes through the native backend", () => {
+  for (const inMemory of [true, false]) {
+    it(`hydrates projected and full includes with ${inMemory ? "memory" : "persistent"} server storage`, async () => {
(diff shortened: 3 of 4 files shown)


aeplay:
🤖 We have adopted this patch, preserving your authorship, onto current main in #2724 so our full same-repository CI can run for the release. The complete diff passed our CI execution-safety inspection and initial correctness review. Focused regression-sensitivity checks and full gates are running now. We will report any findings here; #2724 is not merged yet.

aeplay:
🤖 Adopted and merged via #2724, preserving your commit authorship, after independent review, the full local CI-equivalent gate, and green CI. The included-child JSON regression fails when the fix is removed and passes after restoration. On current main the projected parent-only Rust case already passes without the fix, so we documented it as compatibility coverage. A fresh preview build and packaged end-to-end acceptance are next. Closing this PR as incorporated; thank you for the focused fix and reproductions.

PR #2703, opened 2026-09-09
Require trusted transport for JWKS discovery

joeinnes:
## Summary
- Require an authenticated trusted transport for backend JWKS discovery.
- Reject insecure JWKS redirects instead of silently following them.

## Before
Backend JWKS discovery could cross an untrusted transport boundary or accept an insecure redirect.

## After
JWKS discovery is admitted only through the trusted backend transport contract and redirect policy fails closed.

## Test plan
- [ ] Run the focused request-auth and server JWKS suites.
- [ ] Run repository CI.

### DIFF
--- a/.changeset/trusted-jwks-transport.md
+++ b/.changeset/trusted-jwks-transport.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Anselm approved #2703 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2704, opened 2026-09-09
Align streamed JSON admission with buffered validation

joeinnes:
## Summary
- Make streamed JSON admission enforce the same syntax and numeric boundaries as buffered validation.
- Keep validation bounded across chunk boundaries.

## Before
Buffered and streamed JSON admission could disagree on malformed, overflowing, or boundary-split values.

## After
Both paths accept and reject the same JSON inputs while retaining bounded streaming state.

## Test plan
- [ ] Run the focused Groove large-value admission suite.
- [ ] Run repository CI.

### DIFF
--- a/.changeset/streamed-json-admission.md
+++ b/.changeset/streamed-json-admission.md
@@ -1,0 +1,9 @@
+---
+"jazz-tools": patch
+"jazz-wasm": patch
+"jazz-napi": patch
+"jazz-rn": patch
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Anselm approved #2704 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2707, opened 2026-09-09
Roll back failed multisink shape preparation

joeinnes:
## Summary
- Roll back binding-source and graph state when prepared multisink shape admission fails.
- Preserve active shared binding sources across failed preparation.

## Before
A failed multisink preparation could leave new or shared binding-source state behind.

## After
Failed preparation is atomic for new state and leaves existing active shared state usable for retry.

## Test plan
- [ ] Run the focused Groove prepared-shape suite.
- [ ] Run repository CI.

### DIFF
--- a/.changeset/atomic-multisink-rollback.md
+++ b/.changeset/atomic-multisink-rollback.md
@@ -1,0 +1,9 @@
+---
+"jazz-tools": patch
+"jazz-wasm": patch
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Anselm approved #2707 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2708, opened 2026-09-09
Reject duplicate public column declarations

joeinnes:
## Summary
- Reject duplicate public column declarations during public schema compilation.
- Preserve existing validation precedence.

## Before
Duplicate public column declarations could pass compilation and become ambiguous downstream.

## After
Compilation rejects duplicates with a path-specific error before lowering.

## Test plan
- [ ] Run the focused public schema compilation suite.
- [ ] Run repository CI.

### DIFF
--- a/.changeset/public-column-name-uniqueness.md
+++ b/.changeset/public-column-name-uniqueness.md
@@ -1,0 +1,9 @@
+---
+"jazz-tools": patch
+"jazz-wasm": patch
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved #2708 in the coordinator task. This records review approval; the PR remains part of the post-release stack pending integration gates.

PR #2715, opened 2026-09-09
Restore local Edge query serving while retaining authorized repair

aeplay:
🤖 Restore the server Edge's responsibility for evaluating queries and read policies over its local data, on top of #2696. Browser/native client relays retain their separate role of consuming exact authority-selected inputs.

Anselm approved merging this stack on 2026-09-09 after local, hosted, packaged, and Cloud acceptance. The focused scope-exit suite passes 16 cases, including new grant-only revocation, shared-cache isolation and same-connection readmission checks. A raw-transport test also proves that an ordinary or delegated reader can receive an authorized cached row from an Edge with no Core connected, while another reader's row is withheld. The 146 peer-protocol tests and nine inherited-policy/revocation tests also pass. Mutation checks prove the cached-serving test detects a return to Core-selected serving and the exclusion test detects ignored Edge denials. The full Rust library suite passes: 1,957 passed, two ignored, zero failures. The final topology rerun also passes all 25 selected tests. Broad acceptance is now complete; results and the remaining intermittent-test investigation are recorded below. The release stack is approved for merge.

## A new query can use the Edge's cache

Alice asks an Edge for unfinished tasks. The Edge has the relevant task data and evaluates the query and Alice's read policy locally. It still forwards the subscription to keep inputs synchronized, but it does not require Core's exact result for this new query before opening its local evaluator.

In #2696, non-SYSTEM scopes on partial server Edges instead take the same selected-source path as a browser worker. This PR removes that restriction for server Edges. Ordinary serving follows the policy inputs received from Core, with the existing network/processing lag, rather than requiring a Core roundtrip for every answer. A verified Core repair denial supersedes stale local policy evidence for that reader. It does not change the browser worker: a client relay cannot infer permissions from its incomplete client cache.

## A task stops matching

Alice has task T with `done = false`. Bob changes it to `done = true`, perhaps while Alice is disconnected. The existing query-driven reconciliation from #2696 notices the extra cached T and obtains its current readable version. The Edge computes the new answer locally and T leaves the unfinished list. Readable repair remains authorized under Alice; it does not blindly fetch fresh shared-cache content as SYSTEM.

## Access disappears without a filter change

Alice can read T through grant G. Bob revokes G and changes T's title, but T remains unfinished. A stale local G must not make an explicit repair disclose T's new title.

The repair still asks Core under Alice's admitted identity and claims. Core returns current-unavailable without the new task content. The Edge now honors that verified decision in its ordinary serving graph, excluding T for Alice. Previously only client-local graphs consumed that decision, so simply restoring Edge evaluation left the old authorized task visible indefinitely.

The shared row is retained. SYSTEM and other readers do not inherit Alice's exclusion. When Core later confirms that T is readable again, the exclusion clears and the same Edge/subscription can readmit T.

A trusted transport can carry this delegated Alice decision; it does not mean the Edge's own SYSTEM account lost access. SYSTEM reconciliation remains query-driven refresh without a user availability marker.

## Responsibilities and boundaries

- Ordinary Edge queries: local evaluation of query predicates and read policies, plus verified upstream reader-specific exclusions.
- Explicit extra-row and missing-body repairs: current Core authorization remains required for non-SYSTEM readers.
- Fresh permission probes and raw policy-proof subplans: ignore the cached exclusion, so a previous denial cannot prevent its own later correction.
- Client relays: retain exact selected-authority inputs and existing scope isolation.
- Remote requests cannot disable upstream propagation. Connection/claims replacement, cancellation, unknown replies and receipt ordering retain #2696's rules.

This does not claim that an Edge can independently authorize arbitrary repair from cached grants. Doing that requires coverage of all relevant authorization inputs. The spike restores ordinary serving while retaining the smaller, conservative repair boundary.

## Reader lifetime and scale

Opening and closing queries must not exhaust a fixed lifetime quota of reader identities. Serving graphs now retain a shared owner for their reader-specific exclusion inputs. Closing the final graph queues those inputs for retirement on the next input allocation; another live query for the same reader keeps them alive. Abandoned capability checks are reclaimed too. Existing unavailable-row markers remain available to seed a later reopening.

Tests keep more than 256 different readers live at once, churn another 260 readers while checking existing subscriptions still update, and repeatedly abandon preflight checks. Removing retirement makes both churn tests fail. This removes the accidental 256-reader ceiling without increasing the request-only advice-cache budget. It is lifecycle coverage, not a throughput benchmark.

## Wire and storage

No wire or storage format changes. The existing postcard control messages, native row encoding and reader-scoped RecordStore availability records are unchanged. The new serving mode is internal compiler/runtime state.

## Validation at `ab40af746a`

The new tests target the distinctions that a simple filter-exit test cannot prove: the task can remain in the filter while access changes; another reader can populate the shared cache; revocation can later be reversed; and a new query must actually evaluate without an exact Core source. Verification:

- Full Rust library suite: 1,957 passed, two existing ignored tests, zero failures (includes all 146 peer-protocol tests and 20 availability-input tests).
- Real native Core/Edge/client topology suites: 16 readable-scope-exit cases, eight inherited-policy cases, one scope-revocation case; all passed.
- Reversible mutation checks: restoring selected-Core serving fails the cached-query test; ignoring Edge denials fails the reader exclusion test; omitting retirement fails both query/preflight churn tests. All mutations restored before final positive runs.
- Clippy, formatting, sensitive-data guard, invariant registry and spec issue-link gates passed. Final committed worktree is clean.

Release acceptance is complete. The full local CI-equivalent gate now passes: 3,869 workspace Rust tests (73 skipped), differential/scenario checks, all target classes, TypeScript/browser consumers, RN bridge tests and native/browser storage compatibility. The producer's three generated fingerprint-literal updates were preserved separately and restored after the gate; implementation source was unchanged.

Hosted and packaged acceptance completed on the unchanged commit:

- [CI](https://github.com/garden-co/jazz/actions/runs/34383800392): all partitions passed. The first React todo canary timeout remains tracked in #2716; the unchanged hosted retry, full local consumer gate, ten focused repetitions and eight full-suite repetitions restricted to four CPUs passed. This is not a causal fix for the original timeout.
- [Android and iOS device acceptance](https://github.com/garden-co/jazz/actions/runs/34384001380): passed. iOS needed an unchanged retry after React Native's dependency archive failed to unpack before app launch.
- [Exact preview publication](https://github.com/garden-co/jazz/actions/runs/34384001869): passed, including freshly installed Android/iOS package-consumer builds. Publication needed an unchanged retry after a multipart-upload whitelist rejection/HTTP 413; the repository was already on the provider's whitelist.
- Four standalone apps installed the exact published `ab40af746af828eb9ffa384c6832f4c4d8af9402` packages. Backend inventory passed prior-preview data reopen, fresh validate/deploy/transaction/delete/restore and server restart.
- External JWT enrollment, login, linking, account isolation, offline access loss and readmission passed through a separate packaged CLI Edge connected to Core.
- Chromium and Linux WebKit passed persisted offline predicate-exit repair, disconnected writes, reconnect upload and reload through that Edge. Backend mutations went directly to Core.
- Installed Android passed prior persisted data after upgrade, immediate readback, offline done-only query exit through Edge, offline deletion surviving force-stop/reopen while Edge was unavailable, then upload after Edge restart.

Fresh Cloud backend, Chromium/Linux WebKit, external-JWT, and Android acceptance passed on the exact ab40 preview; see the acceptance comment below. No physical iPhone Safari test was performed for this spike. Existing related-input/negative-evidence reconciliation limits documented in #2696 remain. All owned adopter services were stopped after acceptance, preserving their test data and receipts. The source worktree is clean. Anselm has approved merging #2678 → #2696 → #2715. Core-restart admission recovery remains separately tracked in #2718 and will be fixed on the new main.

### DIFF
--- a/crates/jazz-testkit/tests/readable_scope_exit.rs
+++ b/crates/jazz-testkit/tests/readable_scope_exit.rs
@@ -299,6 +299,19 @@ fn revocation_schema(dependency: bool) -> Schema {
 }
 
 async fn run_revoked_exit(dependency: bool, relayed: bool) {
+    run_revoked_exit_case(dependency, relayed, true).await;
+}
+
+async fn run_revoked_exit_case(dependency: bool, relayed: bool, changes_filter: bool) {
+    run_revoked_exit_shared_case(dependency, relayed, changes_filter, false).await;
+}
+
+async fn run_revoked_exit_shared_case(
+    dependency: bool,
+    relayed: bool,
+    changes_filter: bool,
+    shared_cache: bool,
+) {
     let schema = revocation_schema(dependency);
     let authority = JazzServer::start_with_schema(schema.clone()).await;
     let relay = if relayed {
@@ -340,6 +353,38 @@ async fn run_revoked_exit(dependency: bool, relayed: bool) {
     }
     let (task, _, tx) = bob.insert("tasks", input).unwrap();
     jazz_testkit::wait_for_edge_txs(&bob, &[tx.unwrap()]).await;
+    // A different reader can populate this same Edge's cache with a newer
+    // task. Alice's narrowed scope must not leak that shared-cache successor.
+    let shared_reader = if shared_cache {
+        Some(
+            TestingClient::builder()
+                .with_server(&server)
+                .with_schema(schema.clone())
+                .with_user_id("shared-backend")
+                .as_admin()
+                .connect()
+                .await,
+        )
+    } else {
+        None
+    };
+    let _shared_subscription = if let Some(reader) = &shared_reader {
+        let mut subscription = reader
+            .subscribe_with_read_tier(Query::from("tasks"), ReadTier::Remote)
+            .await
+            .unwrap();
+        wait_for_subscription_update(
+            &mut subscription,
+            &mut Vec::new(),
+            TIMEOUT,
+            "shared Edge cache receives task",
+            |log| has_added_id(log, task),
+        )
+        .await;
+        Some(subscription)
+    } else {
+        None
+    };
     let alice = TestingClient::builder()
         .with_server(&server)
         .with_schema(schema)
@@ -369,7 +414,7 @@ async fn run_revoked_exit(dependency: bool, relayed: bool) {
     let tx = bob.begin_transaction().unwrap().transaction_id();
     let staged = bob.with_write_context(WriteContext::default().with_transaction_id(tx));
     let mut changes = vec![
-        ("done".into(), Value::Boolean(true)),
+        ("done".into(), Value::Boolean(changes_filter)),
         ("title".into(), Value::Text("new forbidden title".into())),
     ];
     if let Some(grant) = grant {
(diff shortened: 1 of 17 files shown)


aeplay:
🤖 Fresh Jazz Cloud end-to-end acceptance passed against the exact published `ab40af746af828eb9ffa384c6832f4c4d8af9402` package set, using four newly provisioned preview tenants and schemas deployed with that exact CLI.

- Backend inventory: fresh tenant, transactional product/order update, predicate exit, delete/includeDeleted/restore, then a separate process verifies persisted rows.
- Browser workboard, Chromium and Linux WebKit: local-first insert, persisted offline predicate exit repaired after reopening, disconnected insert proven absent upstream, reconnect upload and reload persistence. Runtime connections used the actual Cloud WebSocket endpoint; backend mutations also ran against Cloud.
- External JWT notes: new inline Ed25519 public JWK with localhost issuer; atomic enrollment, login/logout, explicit fresh-identity link, already-assigned identity rejection, account isolation, offline cached-row access loss, includeDeleted authorization boundary, and readmission.
- Android emulator using the exact preview's installed native consumer: fresh Cloud session, immediate local readback, Cloud write/read, offline narrow-query exit repair, offline durable deletion proven still present upstream, force-stop/relaunch and confirmed deletion upload.

All scenarios passed. Cloud status confirmed regional readiness. Infra's image pin is based on Anselm's rollout confirmation; the status API exposes configuration versions, not source SHA. The external-auth test uses a signing fixture rather than a live third-party provider; Linux WebKit is not physical iPhone Safari. No Cloud Core restart was injected: the separately reproduced local Core-restart admission problem remains open in #2718 and is not cleared by these passes.

Disposable test-tenant cleanup is being recorded separately. No source or package changes were needed for this run. Joe's stack remains outside this preview and, per Anselm's decision, is scheduled for post-release integration.

PR #2721, opened 2026-09-09
Retry temporary Edge account registry outages

### LINKED ISSUE #2718: Treat transient Edge account-registry outages as retryable during reconnect
🤖 A packaged ab40 adopter run found that restarting Core can permanently stop a local-first client connected through an Edge.

Reproduction: run the exact #2715 preview (ab40af746af828eb9ffa384c6832f4c4d8af9402) with one CLI Core and two CLI Edges, independent local-first native sessions on each Edge, and a maintained query. Cross-Edge insert and disconnected scalar-filter exit repair passed. Stop and restart Core while retaining both Edges and sessions. A reconnect during the outage fails with `account registry unavailable` and the native session treats it as terminal.

Mechanics: `crates/jazz-server/src/server/routes/accounts.rs::read_upstream_assignment` returns a string for a transient upstream request failure. `routes/websocket.rs::handle_ws_connection` maps all `ws_admission` errors to `AuthFailed` / `WireRetry::Never`. The native adapter appropriately does not retry that permanent rejection. Temporary registry unavailability needs a typed, retryable classification distinct from genuinely invalid/revoked credentials; client reconnect handling must support that classification without bypassing admission.

Acceptance should retain live client subscriptions across Core restart, prove automatic recovery after the registry returns, and also prove that actual credential rejection remains terminal. Cover client→Edge→Core, not only a direct client/Core transport restart. No data loss was demonstrated by this run. This was found on the release preview; no fix is claimed and whether it predates #2715 has not yet been measured.

aeplay:
🤖 A short Core outage must not permanently reject a valid client connected through an Edge. Fixes #2718.

Alice is connected to Edge A. Core restarts, and Edge A temporarily cannot resolve her account. Previously that request failure became `AuthFailed` / `Never`; an already-connected session could also be closed as “revoked.” The client then stopped recovering even after Core returned.

The server now preserves the distinction between a denied identity and an unavailable registry. Network failures, upstream 5xx/408/429 responses, and locally unavailable registry state produce the existing `NotReady` / `Later` response. Invalid credentials, assignment mismatch, malformed registry responses and actual denials remain terminal. This applies both during admission and to every existing live-session recheck. No data is sent and no operation is admitted while the check is unavailable: the connection closes and must reauthenticate.

The JS WebSocket carrier now treats a negotiated `NotReady` / `Later` response as a reconnect event; the native adapter retains its existing retry and shutdown machinery. Pre-hello readiness retry remains distinct. Native/RN transport already classifies the same structured response as retryable.

Validation:
- Existing server library suite passed (212 tests before the new regression).
- A real HTTP registry/WebSocket-boundary regression distinguishes 503/429 from permanent 403 denial.
- 78 JS runtime transport/carrier tests pass, including recovery when an already-negotiated Edge reports registry unavailability.
- The original standalone two-Edge scenario passes using the locally rebuilt CLI and JS runtime: Core restart with both Edges and clients retained, subsequent global write and live reader recovery, then Edge restart and another update. This is source acceptance, not a newly published preview.
- Independent review passed with no actionable findings: 222 server feature-test cases; 215 TS cases passed, one skipped; a supplementary pending-native-admission reconnect race passed. Three reversible mutations each failed the relevant regression (remove negotiated retry; turn unavailable into permanent denial; turn denied into retryable). All mutations were restored.
- The full canonical local CI-equivalent gate passed (exit 0): Rust workspace and target classes, differential oracle, TypeScript consumers and browser examples, 94 React Native bridge tests, and native/browser storage compatibility corpora.

No new wire variants, persisted formats, or migration. #2715's local Edge serving and Core-authorized extra-row repair remain unchanged.


aeplay:
🤖 Anselm approved #2721 in the coordinator task. Merge remains conditional on green CI; independent review, restart acceptance, and the full local CI-equivalent gate passed.

PR #2736, opened 2026-09-10
fix: preserve nullable JSON storage descriptors

### LINKED ISSUE #2733: Native optional JSON columns reject explicit null inserts and clearing updates
On published `2.0.0-alpha.54` (also inspected main `b9a3d9e1947a77c3d06e99845003c8f14e66e399`), native writes of SQL null to an optional JSON column fail with:

```
Protocol: value does not match type Internal(InternalValueType(StoredScalar(Json)))
```

A fresh local server and a public backend session reproduce this with one table; no old data or custom query extensions are needed:

```ts
const app = s.defineApp({
  documents: s.table({ name: s.string(), metadata: s.json().optional() }),
});
// After deploying app and admitting createJazzSession({ ..., initial: { backendSecret } }):
const db = owner.getSnapshot().client!.db;
const write = await db.transaction(tx =>
  tx.insert(app.documents, { name: "Alice", metadata: null }),
);
await write.wait({ tier: "global" });
```

The transaction insert fails during native core ticks. A direct update of populated JSON to null fails synchronously at the native update call. Omitting the field on insertion succeeds and reads back as null; writing a non-null JSON object also succeeds.

The likely cause is `schema::storage_column_type`: selecting JSON's distinct physical scalar descriptor drops the logical `Nullable` wrapper. The version-presence wrapper and the column-nullability wrapper represent different states and must both survive.

Expected: omitted inserts, explicit-null inserts, object writes, and clearing updates all succeed; SQL null remains distinct from JSON literal null. Cover local/remote reads, transaction confirmation, nullable predicates, and encoded descriptor receipts. Replacing SQL null with a JSON literal or omitting clearing updates would change application semantics.

husseinraoouf:
Optional JSON columns currently lose their nullable wrapper when lowered to `StoredScalar(Json)`. Inserting an explicit null or clearing a populated cell then fails during the native tick. This preserves column nullability in the shared storage descriptor conversion and makes physical table descriptors use that same conversion.

The regression covers insert, populate, clear, and JSON literal null. A frozen byte corpus distinguishes omitted cells, SQL null, JSON literal null, and populated objects. This is a pre-freeze descriptor correction: affected optional JSON schemas require matching peers and fresh data; no legacy decoding fallback is added. The existing non-nullable JSON corpus is unchanged.

Validation: the new Rust Db regression fails before the fix and passes after it. Both nullable and existing non-nullable wire corpora pass. The public native-session regression passes with memory and persistent storage in [fork CI](https://github.com/husseinraoouf/jazz/actions/runs/34439285433/job/102750752635), including mergeable insertion, exclusive clearing, remote reads, and null filtering. All three incremental-delivery canaries, the bounded 20-seed differential oracle (churn depths 10/1000, three steps), Clippy, formatting, invariant registry, and spec-link checks pass.

The CI preview combines this fix with #2737; subsequent PR changes only add release changesets. Full upstream landing gates and the unavailable private sensitive-data guard have not run. Downstream installed-package validation is pending preview package assembly.

Fixes #2733


aeplay:
🤖 Thank you for identifying this bug and for the regression coverage, @husseinraoouf! We have published the superseding implementation in #3007.

Following the design discussion, we chose to make column null and JSON root null equivalent for nullable JSON columns, including predicates and policies, while retaining the existing published storage descriptor. That lets existing databases keep their data without a reset.

Closing this PR as superseded by #3007, which is an alpha.56 integration draft. The replacement is not merged yet; it still needs integration with current main and fresh CI. The original issue #2733 remains tracked by the replacement.

PR #2745, opened 2026-09-10
fix(groove): wait for hydration temporal predecessors

husseinraoouf:
A subscription opened while an earlier, overlapping hydration is still yielding can end with `SubscriptionFailed(Evaluation(EvaluationBlocked))`. All of its storage reads may be ready while shared nodes remain blocked until the predecessor installs its private snapshot. `EvaluationSession::poll` currently treats that temporal wait as an impossible storage stall and fails the subscription.

Return `Pending` while temporal blockers remain. The predecessor retains the continuation and releases the shared nodes after installation; the existing error still covers unfinished work with neither storage requests nor temporal dependencies.

The regression uses public Groove subscription APIs: a 256-sink opening spans owner turns, then a second opening shares one source and adds an independent source. It fails with `EvaluationBlocked` without the change and receives both outputs with it. The full `async_hydration_session` target passes (29 tests), including cancellation, scoped failure, and incremental-write handoff coverage. The Groove library suite also passes (726 tests; two existing tests ignored). Both suites were rerun on this PR head. `cargo clippy --package groove -- -D warnings` and rustfmt pass.

A downstream browser diagnostic also stopped reporting internal subscription errors and delivered an employee-state update through to the next screen with the patched WASM. That diagnostic used an unoptimized build and an extended wait; normal packaged-runtime browser and native verification remain downstream work.

Validation:
```
cargo test -p groove --features test --lib --test async_hydration_session
```

### DIFF
--- a/crates/groove/src/ivm/runtime/runtime_tick.rs
+++ b/crates/groove/src/ivm/runtime/runtime_tick.rs
@@ -1541,6 +1541,11 @@ impl<'a> EvaluationSession<'a> {
             if self.requests.has_pending() {
                 return Poll::Pending;
             }
+            if !self.work_queue.temporal_waiting.is_empty() {
+                // Earlier evaluations own the continuation until their shared
+                // nodes are installed and the temporal barriers are released.
+                return Poll::Pending;
+            }
             return Poll::Ready(Err(IvmRuntimeError::EvaluationBlocked));
         }
     }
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 The reviewed fix is now carried by #2907 in this repository so it can run full integration CI and participate in the stable-fixes stack. Contributor commits are preserved; this original PR remains available for provenance while integration completes.

aeplay:
🤖 Thank you for the fix and the thorough regression coverage! We integrated this work in [#2907](https://github.com/garden-co/jazz/pull/2907) so it could run through our full integration CI alongside the other fixes. That PR has now merged into `main`.

Closing this PR as superseded by #2907. Your contribution is included—thanks again for helping improve Jazz!

PR #2749, opened 2026-09-10
fix(wasm): use exact local row reads for transaction staging

### LINKED ISSUE #2746: Browser performance: transaction updates fall back to full-table reads in WASM
## Confirmed defect

On main at `b9a3d9e194`, browser transaction updates repeatedly read and decode the entire table to obtain one row. This creates approximately quadratic work when both the table size and the number of updates grow together.

`NativeRuntimeAdapter.update` calls `mergeRowState` after staging a native transaction patch. For a row not already staged, `readRowForWriteMerge` prefers `localCurrentRow`, but that binding exists in NAPI and is absent from WASM. The fallback executes an unfiltered table query, materialises all rows, and finds the requested ID in JavaScript.

Relevant code:

- [Transaction update and row merging](https://github.com/garden-co/jazz/blob/b9a3d9e194/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts#L1423)
- [Exact-reader fallback to full-table query](https://github.com/garden-co/jazz/blob/b9a3d9e194/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts#L2452)
- [Existing NAPI localCurrentRow binding](https://github.com/garden-co/jazz/blob/b9a3d9e194/crates/jazz-napi/src/lib.rs#L3182)

## Reproduction and measurements

Headless Chromium, fresh optimised WASM profiling build from that revision, public schema/Db APIs, a local-first account, and todos with `title: string` and `done: boolean`. No React rendering. Seed in one transaction, query all rows, open and cancel a subscription, then update 90% of the rows. Wait explicitly for local durability.

```ts
const app = schema.defineApp({
  todos: schema.table({ title: schema.string(), done: schema.boolean() }),
});

// db is an initialised browser Db with driver: { type: "memory" }.
await db.all(app.todos);
const seeded = await db.transaction((tx) => {
  for (let i = 0; i < count; i++) {
    tx.insert(app.todos, { title: `Todo ${i}`, done: false });
  }
});
await seeded.wait({ tier: "local" });
const rows = await db.all(app.todos);
const ids = rows.slice(0, Math.floor(count * 0.9)).map((row) => row.id);

const tx = db.beginTransaction();
const start = performance.now();
for (const id of ids) tx.update(app.todos, id, { done: true });
const staged = performance.now();
await tx.commit().wait({ tier: "local" });
console.log({ stageMs: staged - start, commitMs: performance.now() - staged });
```

| In-memory dataset / updated rows | Stage patches | Commit + local wait |
| --- | ---: | ---: |
| 150 / 135 | 2,231 ms | 35 ms |
| 300 / 270 | 8,356 ms | 66 ms |

Doubling both sizes increased staging cost by 3.75×. The 150-row CPU profile attributes approximately 2,053 of 2,323 sampled milliseconds inclusively to `readRowForWriteMerge`. This reproduces without IndexedDB, so storage reclamation is not the cause of this particular defect.

These are diagnostic samples with profiling overhead, not statistically established release benchmarks. The full 1,500-row transaction-staging run was interrupted; no completed duration is claimed for it on this revision.

## Proposed scope

Provide WASM with an exact local row read equivalent to the existing NAPI binding, preserving authorisation, staged-row, large-value and transaction semantics. Add coverage that detects full-table reads per staged update and a maintained scaling receipt. Avoid changing storage reclamation as part of this fix; that is being investigated separately.

## Related performance observations, not yet attributed

Ordinary `db.update(...).wait({ tier: "local" })` calls do not have the same behaviour as one explicit transaction per row:

| 1,500 todos / 1,350 sequential ordinary updates | Memory | Persistent IndexedDB |
| --- | ---: | ---: |
| Batched seed + local wait | 392 ms | 16,697 ms |
| First all-row query after seed | 218 ms | 2,602 ms |
| Initial subscription after that query | 845 ms | 799 ms |
| Ordinary updates + local waits | 581 ms | 99,611 ms |

A repeat memory run measured 402 / 222 / 852 / 585 ms respectively. The persistent comparison changes both storage and browser topology; it does not establish whether storage, transport, retention or foreground replay causes the slowdown. A post-write reopen query remained unresolved when interrupted and is not a completed cold-load measurement.

A separate 1,500-row memory subscription CPU profile sampled approximately 907 ms: 769 ms beneath JS-to-WASM calls, 675 ms beneath `refresh_subscriptions_in`, and substantial record decoding/descriptor creation. JavaScript `applySubscriptionChunk` accounted for about 21 ms and `indexRowsByKey` about 1 ms. Inclusive timings overlap and must not be summed.

Duplicate reset/index/copy work exists in the adapter, but these measurements put WASM record processing ahead of that cleanup. Projection support is already present. A redesign removing foreground query reconstruction is not yet justified by this evidence alone.

aeplay:
🤖 Fixes #2746. Browser transaction updates no longer query and decode the entire table to merge each individual patch.

### Example and cause

An app loads 1,500 todos, starts a transaction, and marks 1,350 of them done. Previously, each first patch to a row called `readRowForWriteMerge`. NAPI supplied `localCurrentRow`, but WASM did not, so the adapter fell back to an unfiltered table query followed by a JavaScript search for that ID. That means roughly 1,350 complete table materializations for this example. Joe's optimized baseline measured 2.23 seconds for 135 updates and 8.36 seconds for 270 updates.

The patch exposes the existing core exact-row operation through WASM. The adapter's existing preferred path now reads just the requested local row; a second patch to that same row uses the transaction's already staged state. It does not introduce a new query or change mutation authorization. This internal local write-merge read has the same authority boundary as NAPI.

### Failed patches cannot leak into a later commit

Review found that `updateInTransaction` previously ran before the merge read. If that read threw and the application caught the error, the native patch could remain staged even though the JavaScript call failed.

Update and upsert now compute the merged state before native staging, and append JavaScript transaction bookkeeping only after staging succeeds. A caught read failure therefore cannot be committed accidentally; a native staging failure also cannot poison the cached row used by the next patch. Tests commit an unrelated successful update after the error and verify that the rejected patch remains absent.

### Coverage and limits

- The browser workload covers 150/300/1,500 rows, forbids any full-table `all` call during staging, and requires one exact read per first updated row. It checks the final values and a second patch to the same row.
- Five new browser cases passed, including failure atomicity. The coordinator independently reran them: **5/5 passed**. Existing mergeable transaction coverage passed **11 tests** (19 outside the selected scope); independent adapter lifecycle verification passed **22/22**.
- Three deliberate regressions were detected: restoring full-table fallback, staging before a failed merge read, and caching a patch before native staging succeeds.
- Independent correctness/security review found no remaining implementation blockers.

The tests assert call behavior and correctness, not a wall-clock speed threshold. Optimized before/after measurements are still in progress. Transaction bookkeeping still scans prior staged writes, so eliminating full-table reads is not a claim that total staging is asymptotically linear. IndexedDB page reclamation (#2751), cold-load CPU profiling (#2750), and ordinary updates with active subscriptions are separate work.

A pre-existing limitation remains: a transaction updating a small field on a row with a streamed large Text value can reject at commit because a physical large-value descriptor reaches the logical authoring boundary. The identical fixture fails with the original fallback and with this patch, leaving the original data unchanged. This is tracked on #1844; the tests preserve that behavior rather than claim new large-value support.

Storage and wire formats are unchanged. Fresh CI and optimized measurements remain pending on this draft.

### DIFF
--- a/.changeset/browser-exact-write-merge.md
+++ b/.changeset/browser-exact-write-merge.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Avoid full-table reads when staging browser transaction updates, and prevent failed write-merge reads from leaving patches staged for a later commit.
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,5 @@ data-multi/
 
 # Lane-local Android SDK, NDK, JDK and emulator cache.
 /.cache/android-device-acceptance/
+# Generated browser benchmark phase receipts.
+packages/jazz-tools/.vitest-browser-bench/
--- a/crates/jazz-wasm/src/lib.rs
+++ b/crates/jazz-wasm/src/lib.rs
@@ -1957,6 +1957,17 @@ impl WasmDb {
             .map_err(to_js_error)
     }
 
+    /// Exact local state for the write-merge bridge, matching NAPI. Write
+    /// authorization remains at the mutation boundary; this is not a query.
+    #[wasm_bindgen(js_name = localCurrentRow)]
+    pub fn local_current_row(&self, table: String, row_id: Vec<u8>) -> Result<Vec<u8>, JsValue> {
+        let row_id = row_uuid_from_bytes(&row_id)?;
+        let inner = self.open_inner()?;
+        let row = with_wasm_db!(&inner, |db| block_on(db.local_current_row(&table, row_id)))
+            .map_err(to_js_error)?;
+        encode_rows(&row.into_iter().collect::<Vec<_>>()).map_err(to_js_error)
+    }
+
     #[wasm_bindgen(js_name = prepareQuery)]
     pub fn prepare_query(
         &self,
--- a/crates/jazz-wasm/src/lib.rs
+++ b/crates/jazz-wasm/src/lib.rs
@@ -1965,7 +1965,7 @@ impl WasmDb {
         let inner = self.open_inner()?;
         let row = with_wasm_db!(&inner, |db| block_on(db.local_current_row(&table, row_id)))
             .map_err(to_js_error)?;
-        encode_rows(&row.into_iter().collect::<Vec<_>>()).map_err(to_js_error)
+        encode_synchronous_rows(&row.into_iter().collect::<Vec<_>>())
     }
 
     #[wasm_bindgen(js_name = prepareQuery)]
--- a/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts
+++ b/packages/jazz-tools/src/runtime/native-runtime/native-runtime-adapter.ts
@@ -1440,6 +1440,10 @@ export class NativeRuntimeAdapter implements Runtime {
     const patch = encodeCellsForPatch(this.table(table), values);
     if (tx) {
       this.assertTransactionWriteIdentity(tx, attribution ? undefined : writeIdentity);
+      // Resolve the synchronous preimage before staging. A rejected merge
+      // (for example an indirect large value) must not leave a native patch
+      // that a caller can accidentally commit after catching the error.
+      const row = this.mergeRowState(table, rowId, values, tx, writeIdentity);
       this.db.updateInTransaction(tx.id, table, rowId, patch, {
         head: branchView?.head,
         base: branchView?.base,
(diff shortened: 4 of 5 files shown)


aeplay:

Issue #2747, opened 2026-09-10
Reclaim obsolete IndexedDB B-tree pages safely


joeinnes:
🤖 ## Summary

`idb-tree` uses copy-on-write pages: each write allocates a new leaf and new ancestor pages, but replaced durable pages are not consistently reclaimed. A 1,500-row browser workload produced approximately 174,148 page records and about 1.27 GB of page payload.

This causes pathological storage growth and increases reload/hydration work.

## Correctness risk

A path-local reclamation prototype exposed a second issue: `IdbTree<S>` currently has no ownership capability at the reclamation seam. If two independent `IdbStorage`/`IdbTree` handles share a store, one handle can delete a page still needed by a cold handle. The cold handle then fails with `MissingPage` before its generation-conflict recovery can replay the logical write.

The relevant contract is `crates/groove/SPEC/2_storage_model.md` (IndexedDB page-store physical format):

- COW writes publish a complete immutable closure atomically.
- Reclamation may delete only pages proven unreachable from the published root.
- Reopening must observe either a complete old closure or a complete new closure.

## Required fix

- Make reclamation opt-in through an ownership-aware `PageStore` capability.
- Keep generic/custom stores and `MemoryPageStore` safe by default: no `deleted_page_ids` unless exclusive ownership is proven.
- Enable reclamation for the browser adapter only after the worker epoch/Web Lock ownership is established, and invalidate that capability before the handle is released.
- Use deterministic set-based retirement tracking rather than quadratic vector membership checks.
- Retire replaced leaves/ancestors and only the overflow chain owned by a replaced/deleted value.
- Preserve fresh-versus-durable partitioning, commit failure recovery, generation conflicts, and same-handle pending-read retries.
- Add public behavioural coverage for independent handles, cold reads during publication, overflow replacement, `write_many`, failed commits, and reopen.

Historical garbage is separate from stopping future growth. The agreed alpha.55 approach is an explicitly scoped IndexedDB reset followed by resynchronization for adopters who can safely discard their local cache; the reset permanently loses unsynced or local-only data and requires retaining the same account credentials and server access. An automatic historical collector (#2753) is deferred, not a release prerequisite. See #2768 for the recovery guidance.

## Acceptance criteria

1. Independent generic/custom handles cannot delete each other’s reachable pages.
2. The production browser worker can reclaim obsolete pages only while its exclusive epoch is valid.
3. Published roots never reference deleted pages under supported ownership semantics.
[...]

PR #2750, opened 2026-09-10
bench: separate browser local-load and reopen costs

aeplay:
🤖 Add an opt-in synthetic browser workload to separate local-load and update costs while investigating #2746 and #2747.

The workload uses 150, 300 and 1,500 small rows with memory and persistent drivers. It records seeding with local durability, the first query, initial subscription, persistent shutdown/reopen, and updating 90% of rows while a subscription is active. Phase reports and a chunked synthetic IndexedDB snapshot survive the run for comparison and upgrade testing.

Reopen creates a fresh runtime/tree cache in the same browser session. The browser realm, WASM module and potentially the physical IndexedDB connection remain warm. This is not a cold browser launch or download benchmark, and timings are diagnostics rather than correctness thresholds.

The profiling artifact admission route explicitly verifies optimized WASM provenance and its native pairing. It does not weaken correctness-artifact checks. All 23 staging-contract tests passed in the lane and independently at the coordinator; tampering with the artifact was detected.

Baseline, before product fixes, using optimized WASM: 1,500 persistent rows took 2.88 seconds for the first query after reopening; 1,350 ordinary updates with a subscription active took 135.95 seconds. The corresponding memory update workload took 10.40 seconds. Seeding produced 24,200 stored pages versus 336 reachable pages, approximately 210 MB versus 2.77 MB. These page counts are diagnostic traversal results, not an independent checksum-validation receipt.

Separate instrumentation measured only about 73 ms cumulative page-read await time across two cold-runtime reopens. CPU profiles instead pointed to repeated nested-record reconstruction during delivery. Thus reclaiming unreachable pages addresses storage growth but does not by itself explain the load delay. Product fixes are separate: #2749, #2751, #2757 and #2758. Combined optimized before/after results are pending.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,5 @@ data-multi/
 
 # Lane-local Android SDK, NDK, JDK and emulator cache.
 /.cache/android-device-acceptance/
+# Generated browser benchmark phase receipts.
+packages/jazz-tools/.vitest-browser-bench/
--- a/dev/artifacts/release-staging-contract.test.mjs
+++ b/dev/artifacts/release-staging-contract.test.mjs
@@ -311,6 +311,48 @@ test("release fingerprint staging verifies downloaded WASM bytes before deriving
   assert.throws(() => stageNativeFingerprints(root), /downloaded WASM artifact hash mismatch/);
 });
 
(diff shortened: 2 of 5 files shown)


aeplay:

PR #2751, opened 2026-09-10
fix(idb): reclaim obsolete pages under exclusive ownership

### LINKED ISSUE #2747: Reclaim obsolete IndexedDB B-tree pages safely

aeplay:
🤖 Fixes #2747: stop accumulating obsolete IndexedDB B-tree pages during repeated writes, while retaining complete readable trees across publication and ownership changes.

A write copies the changed leaf and its ancestor path. Before this fix, replaced durable pages accumulated across generations. This patch tracks obsolete structural pages and the overflow chain belonging to a replaced/deleted value, omits already-obsolete fresh pages from publication, and retires eligible durable pages atomically with the new root.

Reclamation requires explicit exclusive ownership. Generic/custom stores and MemoryPageStore default to no durable deletion. The production browser worker supplies its live database Web Lock proof, and the adapter admits one independently opened tree; clones share that tree. Ownership is acquired before asynchronous metadata reads. Each tree has a revocable generation so an idle runtime can retire without waiting for JavaScript garbage collection; stale reads and writes fail closed, and old transactions drain before a successor is admitted. The adapter checks the capability at commit publication, and release/close/invalidation revoke it. This is the supported adapter ownership contract, not protection against unrelated raw IndexedDB access ignoring that contract.

For example, repeatedly updating one row should retain the live tree rather than every prior copy of its leaf/path. Updating one overflow value must not delete a surviving value's overflow chain. An independent generic handle can still read its old root because generic stores do not reclaim. Under exclusive browser ownership, an old runtime must retire before its successor uses the store. For example, a tab waiting for a new schema keeps its foreground lease but does not pin the old schema runtime. When the last admitted old-schema peer leaves, the old tree generation is revoked and drained while the physical database lock remains held. A late destructor cannot release the successor generation.

No physical storage format change or reset is introduced. Existing databases remain supported. This patch stops future accumulation but does not collect historical orphan pages. For this hotfix, explicit scoped-reset/resync guidance replaces shipping a collector (#2753): unsynced and local-only data would be lost, and account credentials must be preserved. There is no automatic reset.

Validation: native IDB tests (30), full Groove tests (826 passed; 4 existing ignored), focused TypeScript checks (82), and wasm32 compilation passed. Final independent review has no remaining blockers. Mutation checks cover stale reads, pending open, stale destructors, post-await publication, worker transaction draining, and actual WASM adapter clone isolation. The coordinator independently reran the 82 TypeScript checks and the formerly failing real-browser schema handoff.

Generated production browser acceptance passed: worker/IndexedDB compatibility/recovery 84 (2 existing skipped), RecordPlayer 9, docs/main todo examples 8/9, custom JS WASM store 11, physical epoch 7. The optimized combined synthetic 1,500-row workload stores 336 pages / 2.77 MB after seeding, compared with 24,200 / 209.84 MB before the fixes. Combined seeding fell from 22.46 seconds to 1.30 seconds; other concurrent performance changes also contribute, so that timing is not isolated attribution. Reopen and ordinary active-subscription updates still have substantial CPU/runtime costs and are being investigated separately.

Final CI is pending on the changeset-only head. No storage reset is required to use the new implementation.

### DIFF
--- a/.changeset/browser-page-reclamation.md
+++ b/.changeset/browser-page-reclamation.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reclaim obsolete IndexedDB pages during writes under exclusive ownership, and safely retire browser runtime storage handles during schema changes and reconnects. Existing unused pages are not automatically collected.
--- a/crates/groove/SPEC/2_storage_model.md
+++ b/crates/groove/SPEC/2_storage_model.md
@@ -468,9 +468,18 @@ Tree writes are copy-on-write: the changed leaf and every changed ancestor get
 fresh page ids, then one IndexedDB transaction writes the new immutable closure
 and replaces `current` after checking the observed generation. A crash before
 publication leaves at most unreachable new pages; a published root never names
-a torn or missing child. Reclamation is a separate reachability operation and
-may delete only pages proven unreachable from the published root, never pages
-merely replaced by an in-flight write. Reopening observes either the old root
+a torn or missing child. Reclamation may delete only pages proven unreachable
+from the published root. Path-local retirement atomically deletes replaced leaves,
+ancestors, and the overflow chains owned by replaced/deleted values together with
+publication, only when the PageStore proves exclusive tree ownership. Generic and
+memory stores default to retaining durable pages so independent cold handles can
+finish reading their older complete closure and reach generation-conflict recovery.
+The browser capability consumes one live database Web Lock/worker epoch proof for
+one tree; tree clones share its root. It expires before release/close/invalidation,
+and deletion commits recheck it at publication. Unpublished superseded fresh pages
+are omitted from the commit regardless of ownership. A separate reachability
+collector is still required for historical garbage; this policy changes neither
+the durable page encoding nor the storage epoch. Reopening observes either the old root
 and complete closure or the new root and complete closure.
 Before persistence, one logical write—including every operation in a
 `write_many` call—is also locally atomic. If page construction or validation
--- a/crates/groove/SPEC/2_storage_model.md
+++ b/crates/groove/SPEC/2_storage_model.md
@@ -479,7 +479,11 @@ handles may not coexist with an exclusive browser worker owner.
 The browser capability consumes one live database Web Lock/worker epoch proof for
 one live tree at a time; tree clones share its root, and the last clone dropping
 releases tree admission for a fresh open. It expires before release/close/invalidation,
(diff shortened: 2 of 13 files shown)


aeplay:

PR #2757, opened 2026-09-10
fix(browser): require explicit worker handoff during slow startup

aeplay:
🤖 Prevent slow browser runtime startup from creating a competing SharedWorker generation while the original realm can still own the IndexedDB database lock.

For example, a runtime that replies after 1.1 seconds previously exceeded the initial one-second alive timeout. The caller treated silence as permission to advance generation and sent a second connection, even though its foreground lease could still belong to the first worker. That can produce a database-busy failure instead of completing startup.

The runtime now uses its existing five-minute initialization budget from the start. Silence at the deadline is a terminal BrowserWorkerUnresponsiveError, not permission to change generation. Only the explicit worker-closing response permits handoff. First or repeated alive responses do not restart the deadline; existing page suspension/resume handling remains unchanged.

The tests cover a delayed successful reply with exactly one connection, a completely silent worker with one cancellation and terminal error, and a reply at 299 seconds that still times out at the original 300-second deadline. Independent review and the coordinator each ran the complete mocked lifecycle file: 27/27 passed. Restoring the original generation retry or alive-triggered deadline extension makes the new tests fail.

This is a reproduced lifecycle bug found while investigating #2749's broad CI database-busy failure. The CI log alone does not prove it caused that particular occurrence. No storage/wire format changes; native artifact builds are not needed for the deterministic lifecycle tests. Fresh broad CI remains pending.

### DIFF
--- a/.changeset/browser-worker-startup-budget.md
+++ b/.changeset/browser-worker-startup-budget.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Avoid opening a competing browser worker generation when runtime startup is slow; preserve a bounded initialization deadline and explicit worker handoff.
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2758, opened 2026-09-10
perf: avoid repeated canonical record reconstruction

### LINKED ISSUE #2756: Avoid repeated nested-record reconstruction in browser query hot paths
🤖 CPU profiling of a synthetic 1,500-row browser reopen found repeated nested record reconstruction during canonicality checks on the hot path.

Optimized baseline: reopened first query about 3 seconds. Page diagnostics across two reopen operations totaled 372 reads, approximately 3 MB, and about 73 ms summed IndexedDB wait. That is insufficient to explain the observed query delay. Main-page and SharedWorker CPU profiles each attribute roughly 1.1 seconds inclusively to RecordDescriptor::create, with allocation, to_values, ensure_value_type and encode/decode work underneath. These profiles overlap in wall time; their inclusive times must not be added together.

In records/values.rs, nested Record and Enum checks decode values, recreate the record, and compare bytes. This repeats recursively during creation/encoding. Investigate replacing reconstruction with an equivalent allocation-light canonical validation path while preserving exactly the accepted byte language, descriptor identity, error boundaries, canonical offsets/null padding/tags, NaN rejection and large-value semantics.

The storage spec's decode/recreate/compare requirement protects canonical byte identity and weighted consolidation. It must not be weakened to generic structural validation. Keep the old path as a differential test oracle; conservatively retain it for type families where equivalence is not established. No storage or wire format change, migration, or acceptance broadening is intended.

Acceptance: sensitive malformed-byte/corpus tests, independent correctness review, focused native allocation/time evidence, and a fresh optimized browser reopen measurement. Related browser investigations: #2746, #2747, #2750.

aeplay:
🤖 Browser profiling found repeated reconstruction of nested records consuming substantial CPU while delivering locally stored query rows. Validating a record previously decoded its children, rebuilt them, then compared the resulting bytes. With nested records, those steps repeatedly validated the same descendants.

This change checks canonical bytes directly for tuple-free embedded records and enum payloads. For example, validating a row with nested authorship now walks its existing encoding instead of allocating and reconstructing every nested value repeatedly. Descriptors containing tuples retain the previous implementation because their historical decoding and encoding rules differ. Invalid inputs also fall back to the old path, preserving diagnostic errors as well as rejection behavior.

The storage acceptance rule and persisted bytes remain unchanged. Canonical encoding still matters for byte-based consolidation and deterministic ordering; this is an equivalent fast path, not weaker validation. Decode paths are unchanged.

Validation so far: full Groove tests passed; independent correctness review found no blockers, including exhaustive short-input comparisons and nested tuple/NaN cases. Deliberately bypassing validation or removing tuple fallback makes the differential tests fail. A focused native benchmark reduces allocation counts from 63/591/16035 to 21 at nesting depths 1/3/6. These are mechanism measurements, not a claim about browser wall time: combined optimized browser acceptance is pending.

Fixes #2756. Part of the investigation of #2746 and #2747; exact-row write lookup and IndexedDB page reclamation are separate PRs.

### DIFF
--- a/.changeset/canonical-record-validation.md
+++ b/.changeset/canonical-record-validation.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reduce CPU work and allocations when validating nested records during local query delivery, without changing storage encoding or accepted values.
--- a/crates/groove/Cargo.toml
+++ b/crates/groove/Cargo.toml
@@ -40,3 +40,7 @@ harness = false
 [[bench]]
 name = "micro"
 harness = false
+
+[[bench]]
+name = "record_validation"
+harness = false
--- a/crates/groove/SPEC/2_storage_model.md
+++ b/crates/groove/SPEC/2_storage_model.md
@@ -508,10 +508,16 @@ bytes of each element descriptor; no second outer-record layout is introduced.
 Record values are admitted only when their embedded descriptor equals the
 declared `ValueType::Record` descriptor and their raw bytes are canonical for
 that descriptor: decode every child value, recreate the record, and require
-byte equality. Validation alone is insufficient because `OwnedRecord::new`
-currently accepts arbitrary raw bytes (`src/records/mod.rs:1577-1582`). The
-recreate-and-compare rule is required for byte-based weighted consolidation and
-deterministic final tie-breaking.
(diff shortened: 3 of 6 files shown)


aeplay:

PR #2760, opened 2026-09-10
Reclaim obsolete IndexedDB B-tree pages

### LINKED ISSUE #2747: Reclaim obsolete IndexedDB B-tree pages safely

joeinnes:
## Summary
- Reclaim pages replaced by browser IndexedDB B-tree copy-on-write writes only while the owning worker epoch is held.
- Closes #2747.

## Before
Copy-on-write updates left replaced pages durable, allowing browser storage to grow indefinitely. Reclaiming unconditionally would risk readers using another or stale store handle.

## After
- Browser worker-epoch ownership gates path-local page deletion; generic and non-browser stores retain old pages.
- Replaced leaves, internal ancestors, and owned overflow chains are retired atomically, with deferred retry when capability is lost.
- Reads that race publication retry against the current generation; historical reachability cleanup remains separate.

## Test plan
- [ ] `cargo test -p idb-tree`
- [ ] `cargo test -p groove storage::idb --lib`
- [ ] `cargo check -p idb-tree --target wasm32-unknown-unknown`
- [ ] `pnpm --filter jazz-tools check`
- [ ] Run the focused browser IndexedDB adapter tests

### DIFF
--- a/.changeset/idb-page-reclamation.md
+++ b/.changeset/idb-page-reclamation.md
@@ -1,0 +1,6 @@
+---
+"jazz-tools": patch
+"jazz-wasm": patch
+---
+
+Reclaim obsolete IndexedDB B-tree pages after copy-on-write updates while preserving safe reads and non-browser storage behaviour.
--- a/crates/idb-tree/src/lib.rs
+++ b/crates/idb-tree/src/lib.rs
@@ -9,12 +9,11 @@ mod page;
 mod store;
 #[cfg(target_arch = "wasm32")]
 mod web;
-
 use std::cell::RefCell;
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::rc::Rc;
 
-pub use store::{BoxFuture, Commit, MemoryPageStore, Metadata, PageStore};
+pub use store::{BoxFuture, Commit, MemoryPageStore, Metadata, PageReclamationToken, PageStore};
 #[cfg(target_arch = "wasm32")]
 pub use web::IndexedDbPageStore;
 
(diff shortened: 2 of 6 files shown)


joeinnes:
Closing at the author's request; comparison work continues separately.

PR #2763, opened 2026-09-10
fix: await cold pending-upload recovery in browser startup

### LINKED ISSUE #2762: Browser reopen can spin forever restoring pending uploads from IndexedDB
🤖 A synthetic optimized-browser acceptance run seeded 1,500 rows, updated 1,350 with local durability, then shut down and reopened. The first query after reopening never completed; the shared worker consumed a CPU core. The exact physical database and profiles were preserved without resetting data.

The captured stack spends essentially the entire sample in open_scope_isolated_relay_db → Node::restore_pending_uploads → pending-transaction index scan → IndexedDB read. restore_pending_uploads and restore_browser_relay_pending_uploads synchronously call crate::db::block_on. That helper repeatedly polls Pending with a no-op waker and thread::yield_now. In WASM this does not yield the JavaScript event loop, so an uncached IndexedDB read cannot deliver the callback that would complete the future.

Convert restoration and its already-async startup callers to async/await, preserving which identity's pending uploads are restored and ensuring storage errors propagate. Cover both ordinary reopen and the browser relay path with truly asynchronous storage, not only immediately-ready memory storage. The preserved post-update database must reopen and retain every selected and untouched row after the fix. This requires no storage-format change and does not imply corruption.

Found while validating #2746/#2747. The earlier short browser suites did not force this cold pending-upload index path. Also audit other production block_on callsites for asynchronous storage reachability rather than assuming they are ready-immediate.

aeplay:
🤖 Fixes #2762. Reopening browser storage with pending uploads can freeze the shared worker when restoring those uploads needs an uncached IndexedDB read.

For example, create local rows while offline, update them, close the database, then reopen it. Startup scans pending transactions so they can upload later. Previously that asynchronous scan was driven by a synchronous polling loop. On WASM the loop never yielded the JavaScript event loop, so the IndexedDB callback could not resolve its own read. A live profile of the reproduced failure spent essentially its entire sample in this loop.

Pending-upload restoration now awaits storage. The normal and catalogue-backed open paths, plus both browser relay bootstrap variants, propagate that await. Identity selection, deduplication, relay recovery markers, queue ordering and storage encodings are unchanged. A failed scan still fails initialization before recovered entries are queued.

Regression tests force the restore scan to return Pending, verify that the executor receives its real waker, then resume and check the upload queue. They cover normal client and browser relay restoration. The old busy loop is contained by a subprocess watchdog so a regression cannot hang the entire suite.

Focused regressions, host clippy and the WASM-target compile passed. Independent review found no correctness or security blockers. Both deliberately restored blocking loops fail their watchdog tests; the restored fix passes. The coordinator independently reran both recovery regressions on the isolated PR branch. The captured post-update browser database will be reopened through the fixed runtime; that end-to-end acceptance is pending. This fixes a startup deadlock, not the separate ordinary-update CPU cost being investigated alongside it.

### DIFF
--- a/.changeset/browser-pending-upload-reopen.md
+++ b/.changeset/browser-pending-upload-reopen.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix browser startup freezing when restoring pending uploads requires a cold IndexedDB read. Recovery now yields to asynchronous storage instead of blocking its callbacks.
--- a/crates/jazz-wasm/src/lib.rs
+++ b/crates/jazz-wasm/src/lib.rs
@@ -1831,6 +1831,7 @@ impl WasmDb {
             .await
             .map_err(to_js_error)?;
         db.restore_browser_relay_pending_uploads()
+            .await
             .map_err(to_js_error)?;
         db.set_deferred_local_persistence(true);
         Ok(Self {
@@ -1866,6 +1867,7 @@ impl WasmDb {
             .await
             .map_err(to_js_error)?;
         db.restore_browser_relay_pending_uploads()
+            .await
             .map_err(to_js_error)?;
         db.set_deferred_local_persistence(true);
         Ok(Self {
--- a/crates/jazz/src/db/lifecycle.rs
+++ b/crates/jazz/src/db/lifecycle.rs
@@ -142,7 +142,7 @@ where
         let node =
             NodeState::new(config.identity.node, config.schema.clone(), config.storage).await?;
         let node = Node::new(node);
-        node.restore_pending_uploads(config.identity)?;
+        node.restore_pending_uploads(config.identity).await?;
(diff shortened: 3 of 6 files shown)


aeplay:
🤖 Combined optimized browser acceptance now confirms the original failure is fixed. The exact synthetic physical database captured from the hanging run reopens through the production reader in 3.378s and returns all 1,500 unique rows with the expected 1,350 updated / 150 untouched states. Its 702 stored pages remain intact.

The seven-case primary acceptance also passes, including ordinary-update shutdown/reopen and an older physical fixture without reset. A separate checksum-corrupted copy is rejected with `invalid IDBTree page: checksum mismatch`; the original fixture is preserved. Standalone CI for this PR is green. Combined runtime receipt: `479d040eb5688ef1422dc93fa09f860f1437e7da`; integration gate #2771 tracks the reviewed changes together.

aeplay:

PR #2764, opened 2026-09-10
perf: read transaction status without decoding its payload

aeplay:
🤖 Checking whether a transaction is durable currently materializes its entire audit record, including author identities, contribution payloads and user metadata. Browser bulk-update profiling spent about 72 of 113 sampled worker seconds under this status path; these samples are inclusive and are not additive with its child functions.

This change reads only fate, global time and durability from the transaction row. For example, forwarding a locally persisted update can determine its status without reconstructing row authorship or decoding unrelated metadata. The shared lookup still verifies the exact transaction time and node alias, supports the existing durable alias fallback, and validates the required status fields. Pending persistence still forces durability to None.

Status is deliberately a projection, not a full payload audit. It no longer rejects a status request solely because an unrelated contribution payload cannot be decoded. Full transaction/recovery readers keep their existing contribution-identity validation before those payloads can influence merge calculations. No stored encoding, authorization decision, or settlement threshold changes.

Focused coverage checks alias restoration and missing IDs, fate/time/durability combinations, pending persistence, malformed rejection fields, and repeated status reads with small versus large metadata without full-payload decoding. Full payload readers still reject malformed contribution identities. Independent review found no correctness or security blockers. Six focused tests passed, including persistence-failure and authority-publication boundaries. Mutations removing the durability clamp, restoring the full reader, or redundantly decoding payloads were all detected. The coordinator independently reran the expanded status regression on the isolated PR branch. End-to-end performance validation and final CI remain pending.

Related: #2746 and #2747. This addresses the remaining transaction-status CPU cost after the exact write lookup and IndexedDB reclamation fixes.

### DIFF
--- a/.changeset/transaction-status-projection.md
+++ b/.changeset/transaction-status-projection.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reduce transaction-status polling cost by reading fate and durability fields without decoding unrelated transaction payloads. Preserve pending-persistence and rejection checks.
--- a/crates/jazz/src/node/currency.rs
+++ b/crates/jazz/src/node/currency.rs
@@ -8,6 +8,11 @@
 use super::*;
 use crate::schema::RuntimeSchema;
 
+#[cfg(test)]
+thread_local! {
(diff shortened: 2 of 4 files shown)


aeplay:

PR #2765, opened 2026-09-10
perf: probe exact parent history before scanning sibling versions

aeplay:
🤖 Stacked on #2764. Updating one row from a large seed transaction can decode every history row in that transaction merely to validate that one parent version exists.

For example, seed 1,500 todos in one transaction, then update a single todo. The cold path previously queried all 1,500 versions before filtering to the requested physical row, branch and layer. This change first probes the existing immutable history primary key for that exact coordinate. Content and deletion history use their respective existing keys; the returned row is decoded using its authored schema and checked against the expected transaction and full coordinate.

The full parent transaction audit still runs first. The resident-cache path is unchanged. A missing exact witness still uses the existing transaction-wide completeness rules: missing or partial evidence remains inconclusive, while a complete transaction with the wrong row/branch/layer is rejected. Noncanonical branch keys retain the existing fallback rather than introducing a new encoding panic. No index, storage format, wire protocol or authorization rule changes.

The new canary uses both 1-row and 1,500-row transactions, including content and deletion witnesses. The valid cold lookup decodes exactly one history row, reads at most two records and performs no history index range scan. It also pins the full transaction audit and checks wrong/missing coordinates. Independent correctness/security review passed with 15 focused tests. Temporarily restoring the old scan made the canary fail with 1,500 decoded rows instead of one; all mutations were restored and the tests passed again. The coordinator independently reran the canary. Final combined browser timing is pending.

This addresses another measured bulk-update CPU cost alongside #2746/#2747. It does not replace the more expensive completeness check when an exact witness is absent.

### DIFF
--- a/.changeset/exact-parent-history-lookup.md
+++ b/.changeset/exact-parent-history-lookup.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Avoid scanning every sibling version in a large transaction when validating an existing exact parent row. Preserve branch, layer, identity and incomplete-parent checks.
--- a/crates/jazz/src/node/currency.rs
+++ b/crates/jazz/src/node/currency.rs
@@ -10,6 +10,7 @@ use crate::schema::RuntimeSchema;
 
 #[cfg(test)]
 thread_local! {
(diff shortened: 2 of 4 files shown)


aeplay:

PR #2768, opened 2026-09-10
docs: clarify scoped storage reset and resync

aeplay:
🤖 Clarifies the scoped browser reset workaround for #2753 without adding a historical page collector.

The FAQ now warns that resetting permanently deletes unsynced writes and local-only data, explains that resync requires server-held data and continued access by the same account, and preserves default credentials and unrelated browser storage. It also documents that the console helper needs a live registered client. The auth lifecycle page now has the missing Storage reset anchor and explains how reset differs from logout.

Validation: documentation formatting and type generation passed; the coordinator independently reviewed the two-page diff and checked the corrected link targets. No runtime behavior changes.

Issue #2753, opened 2026-09-10
Collect historical unreachable IndexedDB pages without resetting existing databases


aeplay:
🤖 Follow up #2747 / #2751 with safe collection of unreachable pages already present in existing IndexedDB databases.

Path-local retirement stops future accumulation but does not remove historical garbage. A synthetic 1,500-row seeded baseline contains 24,200 stored pages, of which 336 are reachable from the published root (approximately 210 MB stored versus 2.8 MB reachable). The baseline fixture is retained privately for upgrade acceptance. This is storage-volume evidence, not attribution of the measured cold-load delay.

Implement or design a bounded reachability collector under the same proven exclusive browser/tree ownership as normal retirement. Mark the complete structural and overflow closure from a stable published root; never delete reachable pages or race root publication, pending operations, ownership release, or another supported handle. Retain existing physical format and data; no reset or clear-storage workaround. Bound scan/memory/transaction work so cleanup does not impose a long startup pause, and make interruption/restart safe.

Acceptance: an existing bloated synthetic database preserves every row and value across upgrade, collection, crash/reopen and further writes; stale ownership cannot collect; reachable-page equality and reduced stored bytes are measured. Separately measure cold-load performance before/after instead of assuming reclaiming disk space removes its CPU/IO bottleneck.

aeplay:
🤖 Followup design from the performance investigation: keep historical collection separate from the immediate cold-load fixes. Reclamation prevents new garbage; recovering old disk space is worthwhile but does not address the measured record-processing CPU cost.

A bounded collector can mark the complete live structural/overflow page graph under an unchanged durable root, generation and ownership epoch. Abort incomplete marking on mutation or publication. Capture the allocation high-water H. Once marking completes, enumerate physical keys below H in small batches and retire unmarked IDs through the existing atomic tree commit path. Do not read orphan payloads or issue separate unfenced JS deletes.

Ordinary COW writes after marking allocate IDs at or above H and cannot resurrect an unmarked historical page below H. That permits writes between sweep batches, provided reload, ownership loss and reset invalidate the collector. This invariant requires explicit coverage; pending commits must be handled at the initial stable boundary.

[...]

Issue #2655, opened 2026-09-07
Diagnose BandChat guest membership stall after owner observes invitation


aeplay:
🤖 Full TypeScript consumer execution at `0fa92d2b21b9c02a68c39fc2050007e5f5561ad8` failed the BandChat persistent-browser journey at the guest invited-room wait.

Diagnostics showed both previews connected, one room and an active conversation each, with owner memberships=2 and guest memberships=1. The invitation was submitted, reached the root and cleared its input. This differs from the earlier empty React input occurrence fixed by #2610: the owner now observes the second membership. It still does not establish the remote write/coverage/React boundary where progress stopped.

The same candidate passed all six focused BandChat tests and three concurrent docs/BandChat pairs. Those runs do not explain the full-workload failure or prove a harmless timeout. No measured failing-base comparison exists for this occurrence. The original full log is retained; first-failure tracing is added to the release tip.

Capture write local/remote settlement, guest query coverage and UI delivery in the failing full release workload, then fix the demonstrated boundary. Keep the current membership, message and removal assertions and timeouts. This tracks the exact release-candidate recurrence separately from broader deferred #2413 history.

aeplay:
🤖 Release disposition from Anselm: keep this issue tracked and monitor it, but do not treat it as an alpha.54 release blocker unless new evidence materially changes the assessment. Existing assertions and failure capture remain enabled. Passing focused runs are not claimed as a causal fix.

aeplay:
🤖 The same failure recurred in #2724 CI at 1db3abd67a4a37b3ce10c7c77a19fef936c55854: owner memberships=2, guest memberships=1, invitation submitted/reachedRoot/inputCleared=true, followed by the guest invited-room timeout. [Failing job](https://github.com/garden-co/jazz/actions/runs/34417725654/job/102686219175).

The exact candidate's full local concurrent TypeScript/browser gate passed, including all 11 BandChat tests and 318 SDK browser tests. Matching failures also appear in preserved earlier stack CI logs, before the JSON hydration patch. This is not evidence of a causal fix or of a harmless timeout.

Diagnostic clarification: this test's `connected` field means DOM attachment, not transport connectivity, and empty UI alerts do not establish that the underlying subscription has no error. A bounded lane is examining those evidence gaps. The previously recorded release disposition remains unchanged; no new causal evidence has emerged yet.

aeplay:
[...]

PR #2771, opened 2026-09-10
Integrate browser storage and update performance fixes

aeplay:
🤖 Base integration PR for the browser performance fixes behind #2746/#2747. Review and merge this together with its follow-up stack: #2772 → #2774 → #2775 → #2778 → #2779. The complete stack passed fresh CI and the full local TypeScript consumer gate at 6b9f0a2a9c; later optimization work remains separate.

This base combines the independently reviewed changes from #2749 (exact row lookup), #2751 (safe IndexedDB reclamation), #2757 (worker startup deadline), #2758 (record validation), #2763 (asynchronous pending-upload recovery), #2764 (transaction-status projection), #2765 (exact parent-history lookup), #2768 (reset guidance) and #2769 (the earlier cancellation-log diagnostics; its final periodic-log amendment is included in #2778), plus the #2750 profiling harness.

Before these fixes, repeated writes accumulated obsolete IndexedDB pages, transaction staging could read a whole table for one row, and reopening a database with pending writes could block the browser event loop. The stack stops future page growth under proven exclusive ownership, reduces repeated decoding/history work, and keeps recovery asynchronous. Its follow-ups fix deferred transaction subscription delivery (#2774), remove synchronous transaction-staging preflight reads entirely (#2778), and stop rereading already locally acknowledged transaction history (#2779).

For an app with 1,500 tasks, the optimized synthetic Chromium measurements were:

- Seed plus local durability: 22.459s → 1.123s.
- Reopen first local all-row query: 2.883s → 1.424s.
- Subscription-first reopen: 1.697s → 0.686s.
- Update 1,350 specific tasks, awaiting local durability after each write with an active subscription: 135.948s → 49.112s. This remains a performance target; it is not a claim that bulk latency is solved.
- Fresh seed storage: 24,200 pages / 209.845MB → 336 pages / 2.77MB; the earlier integrated post-update receipt retained 702 pages / 5.926MB instead of 61,017 pages.

These are synthetic single-run measurements, not guarantees for the reporter’s app or device. A separate clean batched-update receipt staged 1,350 updates in 85ms and committed locally in 7.764s, but reopening after that large batch took 6.628s. The 1.424s reopen result is for the initial seeded state; large batched history remains a cold-load performance target. Correctness checks verify exact changed and untouched row identities, titles, subscription delivery and reopen results. The seven-case compatibility run also loaded an old physical database without resetting it. The memory transaction subscription bug found during batching is fixed by #2774 and verified in the final browser run.

Historical orphan pages are not collected. As agreed, #2753 is deferred: scoped reset/resync guidance explicitly warns that unsynced or local-only data is permanently lost, and preserves account credentials and unrelated databases. Existing databases remain readable without a reset.

Final integrated validation: all CI categories passed in run 34517812150. The local consumer gate passed all 28 Node tasks, 2,245 SDK Node tests, 326 SDK browser tests, five transport checks, framework suites and BandChat/WorkOS/RecordPlayer examples. Existing skips remain explicit. The intermittent BandChat invitation symptom remains tracked under #2655 with the previously agreed nonblocking disposition; passing runs are not claimed as its causal fix.

### DIFF
--- a/.changeset/browser-exact-write-merge.md
+++ b/.changeset/browser-exact-write-merge.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Avoid full-table reads when staging browser transaction updates, and prevent failed write-merge reads from leaving patches staged for a later commit.
--- a/.changeset/browser-page-reclamation.md
+++ b/.changeset/browser-page-reclamation.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reclaim obsolete IndexedDB pages during writes under exclusive ownership, and safely retire browser runtime storage handles during schema changes and reconnects. Existing unused pages are not automatically collected.
--- a/.changeset/browser-pending-upload-reopen.md
+++ b/.changeset/browser-pending-upload-reopen.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix browser startup freezing when restoring pending uploads requires a cold IndexedDB read. Recovery now yields to asynchronous storage instead of blocking its callbacks.
--- a/.changeset/browser-worker-startup-budget.md
+++ b/.changeset/browser-worker-startup-budget.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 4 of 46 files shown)


aeplay:
🤖 Anselm approved #2771 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates.

PR #2772, opened 2026-09-10
perf: avoid copying sibling values during browser point reads

aeplay:
🤖 Stacked on #2771. The remaining bulk-update profile spends substantial time on point reads: looking up one transaction-status record clones every value in its B-tree leaf, then clones the selected value again. A status poll therefore copies unrelated sibling payloads even though it never decodes them.

Borrow the resident leaf entries during descent and pass the requested value by reference to the existing reader. The returned value remains owned. Writes still copy their leaf only after validating every retained value edge. Existing cycle/shared-page checks, overflow hydration, missing-page retries, and corruption errors remain in place. There is no storage-format change.

A public-API allocation canary proves a warm hit copies one 1,000-byte payload and a miss copies none, with sibling records in the leaf. Restoring the previous implementation produces 12 allocations and fails the canary. It also checks cold overflow hydration, warm overflow reads, and sibling-preserving writes/deletes. Independent storage correctness/security review passed all 31 IDB tests and reproduced the old-clone mutation failure. The coordinator independently reran the allocation canary. Combined browser measurement is pending.

### DIFF
--- a/.changeset/browser-point-read-copy.md
+++ b/.changeset/browser-point-read-copy.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reduce browser storage lookup overhead by copying only the requested row value instead of unrelated values in the same page.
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Anselm approved #2772 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates.

PR #2774, opened 2026-09-10
fix(browser): refresh subscriptions after deferred transaction commits

### LINKED ISSUE #2770: Browser memory subscription stays stale after a batched transaction
🤖 Found during synthetic browser acceptance of the combined performance candidate `479d040eb5688ef1422dc93fa09f860f1437e7da` for #2746/#2747. Whether this predates the candidate is not established yet.

Reproduction:
1. Use the browser memory driver and seed 1,500 tasks with `done: false`.
2. Start a local subscription and explicitly wait for its full 1,500-row baseline snapshot.
3. Begin one transaction, update 1,350 specific tasks to `done: true`, commit, and wait for local durability.
4. A direct local `all()` returns those exact 1,350 changed rows and the 150 untouched rows correctly.
5. The subscription delivers no updated snapshot within 30 seconds and reports no error. Its history contains only the initial empty placeholder and the original 1,500 false rows.

The immediate-subscribe variant also reproduces. The equivalent persistent-driver transaction passes subscription delivery and exact-state reopening. Ordinary individual updates pass both drivers. This isolates the investigation to memory transaction/subscription delivery; it is not evidence of lost stored writes.

Expected: an established subscription converges to the committed state without requiring an unrelated read or later write. Investigating the existing notification/invalidation path before changing behavior. Add a focused regression that checks selected row identities and values, not just result size.

aeplay:
🤖 Stacked on #2772; fixes #2770. Browser memory runtimes defer local persistence. When an app updates several rows through one queued transaction, that path published resident state and scheduled persistence but omitted the local-subscription refresh used by ordinary writes. Direct reads therefore saw the changes while an established subscription stayed stale indefinitely.

For example, subscribe to 20 unfinished tasks, then mark 18 specific tasks done in one transaction. The subscription must receive those 18 updates without requiring another write or a direct query. Route queued mergeable commits through the existing deferred-publication helper: first hand the publication to the runtime so persistence remains owned, then refresh local subscribers. This preserves immediate local visibility and does not grant durability before persistence completes. No new tick ordering, storage format or wire behavior is introduced.

A focused authenticated, history-complete owner/schema-view test reproduces the browser lifecycle, including unsubscribe/re-subscribe and a waiting stream reader. The old implementation failed the 1,500-row diagnostic; the smaller 20/18 canary passes with the fix. The blocked/failed-persistence canary verifies resident delivery before persistence, durability None while blocked, retained ownership after cancelling a tick, successful Local durability, and failure without a Local receipt. Independent correctness/security review passed 12 focused tests; restoring the old queue-only call fails both new native canaries. A one-row public browser regression is included; fixed-artifact browser confirmation and the complete integrated browser/Node consumer gate passed at 6b9f0a2a9c.

### DIFF
--- a/.changeset/deferred-transaction-subscriptions.md
+++ b/.changeset/deferred-transaction-subscriptions.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix browser subscriptions remaining stale after a transaction commits with deferred local persistence.
--- a/crates/jazz/src/db/tests/transactions.rs
+++ b/crates/jazz/src/db/tests/transactions.rs
@@ -2891,3 +2891,128 @@ fn mergeable_tx_emits_one_subscription_delta_for_many_writes() {
     assert!(removed.is_empty());
     assert!(subscription.try_next_event().is_none());
 }
+
+// Internal binding-contract coverage: the WASM memory owner defers persistence
+// and drives queued synchronous staging. A pending stream must receive its
+// wake without an unrelated peer, query, or write triggering a refresh.
+#[test]
+fn deferred_queued_mergeable_commit_wakes_existing_subscription() {
+    let schema = doctest_support::schema();
+    let empty = build_public_db_test_schema(PublicSchemaBuilder::new());
+    let families = empty.column_families();
+    let refs = families.iter().map(String::as_str).collect::<Vec<_>>();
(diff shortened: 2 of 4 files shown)


aeplay:
🤖 Anselm approved #2774 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates.

PR #2775, opened 2026-09-10
perf: validate inline scalar bytes without owned round trips

aeplay:
🤖 Stacked on #2774. Fresh browser cold-load profiles identify inline scalar access as repeated native CPU work: the accessor decodes owned values, builds a field map, clones the payload, and recreates the record to validate it, even though its result is a borrowed byte slice.

Primitive scalar tag 2 has exactly one raw bytes/string field in slot 1. Its canonical record payload is therefore the field itself, with no offset table, padding or reserved slots. After the existing variant framing check, validate the same logical Bytes/String/JSON rules and return that borrowed payload. Indirect tag 3 remains on the existing decoder; error mapping and validation precedence are preserved. No wire or storage bytes change.

Tests pin the schema premise, compare exact results/errors with the previous round-trip implementation over malformed framing, UTF-8/JSON, truncation and per-byte mutations, and check zero-allocation borrowed bytes/string access for several payload sizes. JSON keeps its existing logical parser and allocations. Removing logical validation fails the oracle; restoring the old implementation fails the allocation canary. Independent codec correctness/security review passed, including 57 large-value tests, the allocation canary, and 24 query tests. Supplemental checks covered indirect descriptors, error precedence and JSON/UTF-8 boundaries. The coordinator independently reran the oracle and allocation tests. A clean optimized browser run at bf942bfb5b measured reopen first-all at 1.418s versus 1.733s on the preceding measured stack, subscription-first at 0.681s versus 0.902s, and post-update reopen at 2.349s versus 2.979s. Exact identities, titles and the selected 1,350 completed / 150 untouched rows survived reopen. Ordinary updates measured 50.751s versus 52.292s; this small single-run difference is not evidence of a substantial bulk-write improvement. Final integrated correctness acceptance remains in progress.

### DIFF
--- a/.changeset/borrowed-inline-scalar-validation.md
+++ b/.changeset/borrowed-inline-scalar-validation.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reduce validation overhead when reading inline scalar values without changing their storage encoding.
--- a/crates/groove/SPEC/9_large_values.md
+++ b/crates/groove/SPEC/9_large_values.md
@@ -92,6 +92,10 @@ parameterized by the immutable kind supplied by schema lowering, so neither
 physical arm duplicates that context. Inline payloads are interpreted as the
 primitive selected by that schema; the same raw UTF-8 content can therefore be
 a valid string or JSON value when stored under the corresponding column kind.
+The primitive arm contains only slot 1, a trailing raw field: its payload is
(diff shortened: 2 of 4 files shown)


aeplay:
🤖 Anselm approved #2775 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates.

PR #2779, opened 2026-09-10
Avoid rereading locally acknowledged transaction history on every tick

aeplay:
🤖 Stacked on #2778. Removes repeated transaction-status storage reads after a downstream listener has already received its local durability acknowledgment.

When an offline browser writes 1,500 rows, its relay retains routes for those writes so it can later forward server confirmation or rejection. Previously, every peer tick reread the status of every retained transaction, even when all listeners had already received local acknowledgment. As writes accumulated, each new write triggered increasingly many reads of old transaction history. A counter reproducer measured exactly 150 and 1,500 storage reads in one pass over 150 and 1,500 already-acknowledged transactions.

The acknowledgment pass now reads status only when a transaction has at least one live listener still awaiting local acknowledgment. The same counter reports zero reads for both retained-history sizes. For example, a second listener subscribing to an old transaction still gets its own acknowledgment: that transaction is read once, while the previously acknowledged listener receives no duplicate. Dropped listeners are pruned without a read. All live routes remain registered for eventual global acceptance or rejection and are removed by the existing terminal-fate path.

This does not cache transaction status, change durability meaning, or alter storage/wire formats. The cheap scan of retained routes remains; the expensive repeated storage probes are removed. A clean optimized browser run at 23595340dd measured 49.112s for 1,350 ordinary updates with a local-durability wait after each row, versus 50.751s previously. This small single-run difference is not a substantial end-to-end speedup claim. Exact completed/untouched rows and reopen checks passed; reopen first-all was 1.424s and subscription-first 0.686s. Further attribution of serialized durability waits and batched-write behavior continues.

Validation: the focused counter and new-listener/dead-listener/global/rejection boundaries pass; the old implementation fails the count assertion with [150, 1500] instead of [0, 0]. Independent correctness/security review passed with nine focused tests covering the counter, new observers, bounded fate retries, authority-epoch handling, cancellation, blocked durability and global/remote waiters. Its independent restoration of the old all-keys loop failed with exactly [150, 1500]; restoring the fix passed. The coordinator independently reran the counter successfully. The full integrated TypeScript consumer run passed at 6b9f0a2a9c: all 28 Node build/test tasks, 2,245 SDK Node tests, 326 SDK browser tests, five transport checks, framework suites, and BandChat/WorkOS/RecordPlayer examples. Ten browser tests remain explicitly skipped as before. Fresh CI is still running; this remains draft pending the complete gate result.

### DIFF
--- a/.changeset/skip-settled-local-ack-probes.md
+++ b/.changeset/skip-settled-local-ack-probes.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Avoid repeatedly reading transaction status for writes already acknowledged locally while waiting for server confirmation, improving offline bulk-write performance.
--- a/crates/jazz/src/db.rs
+++ b/crates/jazz/src/db.rs
@@ -1800,7 +1800,19 @@ async fn queue_local_acknowledgements<S>(routes: &LocalFateRoutes, node: &Shared
 where
     S: OrderedKvStorage,
 {
-    let tx_ids = routes.borrow().keys().copied().collect::<Vec<_>>();
+    // Acknowledged routes remain registered for later global/rejection fates,
+    // but no longer need a local durability probe. A new queue on the same
+    // transaction still gets its own acknowledgement; dead queues need none.
+    let tx_ids = routes
+        .borrow()
+        .iter()
+        .filter(|(_, pending)| {
+            pending
+                .iter()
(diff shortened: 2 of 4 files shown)


aeplay:
🤖 Anselm approved #2779 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates.

PR #2781, opened 2026-09-10
Reuse the complete subscription closure during publication bookkeeping

aeplay:
🤖 Stacked on #2779 as a separate performance follow-up. Publishing a one-row change in a large subscription already computes the complete current supporting-row set. Previously the hot path unnecessarily rebuilt that BTreeSet, then traversed the maintained view again to reconstruct the same set for predecessor bookkeeping.

Reuse the existing complete set for bookkeeping after successful update construction. For example, updating one of 150 subscribed tasks still sends all 150 supporting-row identities, including the updated task’s exact new version. Only the duplicate internal traversal disappears. The wire builder still constructs the complete manifest in the same way; resets, negative evidence and authorization behavior remain unchanged. No storage or wire format changes.

The captured set is stable: update construction receives only a shared borrow of the maintained view, and bookkeeping advances at the same successful point as before. Failure or cancellation does not advance the predecessor. Other publication paths retain their existing helper.

Independent correctness/security review passed 12 focused tests covering resets, transient changes, self-joins, permission grant/revocation, cancellation and stale-closure repair. Restoring the old bookkeeping helper makes the new traversal canary fail with three traversals instead of two. The coordinator reran the canary; additional assertions pin complete cardinality and the changed row’s exact transaction version.

This remains draft: the final amended canary passed, while integrated artifact acceptance and optimized browser measurement are pending. The already-green stack through #2779 remains independently reviewable; no end-to-end speedup is claimed for this follow-up yet.

### DIFF
--- a/.changeset/reuse-publication-source-closure.md
+++ b/.changeset/reuse-publication-source-closure.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Reuse the complete supporting-row set during subscription publication bookkeeping instead of reconstructing it again.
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Anselm approved #2781 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates.

PR #2782, opened 2026-09-10
perf: remove canonical round trips from trusted row decoding

aeplay:
🤖 Trusted local reads were repeatedly decoding and re-encoding native row bytes to prove canonicality: inside nested-field getters, in the immutable-row wire codec, in message encoding/decoding, and again during authority receipt ingestion. Loading 1,500 simple rows therefore performed substantial encoding work before returning the query.

This PR removes those round trips from trusted paths. Nested records and enum payloads are lazy byte-backed values. Our outgoing encoder preserves native row bytes. Connection installation selects direct trusted decoding for upstream and authenticated backend/authority/admin links; session and relay links retain the checked decoder. The sender cannot select trust with a message flag. Native host queues and auxiliary I/O carry the same locally established distinction.

For example, an edge sends an accepted supporting-row set to a browser: the browser decodes the message and applies the authoritative inputs without re-encoding every row. A session sending an upload to an edge still goes through checked admission. A valid row referring to an unavailable authored schema still causes atomic semantic rejection. Policy evaluation, transaction identity/fate checks and catalogue compatibility remain distinct from representation validation.

The specs make encoders responsible for correct representation and storage/wire responsible for preserving bytes. The eventual untrusted-input contract is bounded decoding with connection-local failure containment. This PR does not relax the existing checked untrusted admission path before containment is established. It removes the unnecessary checks on trusted paths now, without changing wire/storage bytes or the ancestry algorithm.

Validation:
- Full Groove library for the nested-read slice: 729 passed, 2 ignored.
- Final Jazz library: 1,994 passed, 2 ignored; native relay library: 102 passed.
- Nested getter work counter: old decoder 7 record encodes; candidate 0. Mutation restored.
- Trusted encode/decode receipt counter: 0. Restoring checked decoding produces 2 and fails. Session decoding still validates. Mutation restored.
- Connection-role test covers upstream, session, relay, backend, authority and admin selection.
- Trusted corrupt-byte rejection tests now cover unavailable-schema atomicity; malformed-byte rejection remains at untrusted admission. Frozen JVRR bytes remain unchanged.

Performance on the same synthetic 1,500-row persistent-browser workload:

| Operation | Original | Nested-only slice | Complete change |
|---|---:|---:|---:|
| Initial cold read | 1.41 s | 1.05 s | 0.98 s |
| 1,350-row commit + local durability | 7.85 s | 7.55 s | 7.49 s |
| Post-batch reload | 6.48 s | 6.11 s | 6.09 s |

All four benchmark sizes passed, plus a separate CPU-profile run. Fresh profiles contain no receipt-validation samples in the trusted cold-read or reload paths; the unit counter independently proves zero on the trusted codec route. Batch/reload time is still dominated by the unchanged per-row ancestry scan (6.38 s worker / 5.25 s foreground). These are single-run measurements of a runtime reopened inside an already-loaded browser, not cold-browser startup or latency percentiles.

Draft: coordinator implemented and self-reviewed under the user's single-track instruction. No independent review or merge approval is claimed.


aeplay:
🤖 Anselm approved #2782 in the coordinator conversation. Merge remains subject to the final performance checkpoint passing its integration gates. The broader untrusted-decoder failure-containment design is tracked separately.

Issue #2784, opened 2026-09-10
Audit row-specific callers of transaction-wide version reads


aeplay:
🤖 Audit callers of query_versions_for_tx and replace broad transaction reads where the caller only needs specified row coordinates.

The content ancestry walker loads, clones/decodes and sorts every version in a transaction before filtering to one physical table, branch, row and content layer. Repeating that for wide transactions causes approximately quadratic work. The immediate ancestry bottleneck is being fixed separately.

Other candidates: row_availability.rs; policy preimage lookup in policy.rs; witness resolution in views.rs and query_eval/materialization.rs; repair coordinate/payload construction in state/read_payload.rs; cold fallback in query_versions_for_tx_physical_coordinate in currency.rs.

Keep genuinely transaction-wide operations: complete commit reconstruction, transaction fate cleanup, complete write-policy evaluation and complete replay comparison. Fragment assembly and missing-parent completeness checks need transaction-wide membership/count knowledge, but may not need every decoded body. Preserve the distinction between a missing parent witness in an incomplete transaction and an invalid coordinate in a complete transaction.

Audit each caller's semantics before changing it. Preserve physical identity across schema renames, branches, content/deletion layers, resident/unpersisted caches, and partial transactions. Add deterministic work-count tests proving unrelated rows are not materialized for coordinate lookups, with mutation sensitivity checks. Track timing separately from correctness.

This issue records the broader follow-up; do not expand the immediate ancestry optimization into a general rewrite.

aeplay:
🤖 The new native local-batch profile confirms a large remaining cold-miss case in this issue.

At 1,500 rows, a fresh receiver gets 150 unchanged versions from transaction A and 1,350 newer versions from transaction B. The B versions name parents in A that are absent from the current scope. After each exact miss, validate_known_parent_coordinate calls query_versions_for_tx_physical_coordinate (cold fallback loads all retained siblings), then query_versions_for_tx again to decide completeness. The cold path caches table names, not those materialized versions.

Measured MemoryStorage receiver ingestion: 10% updated 844 ms; 50% 2,151 ms; 90% 870 ms; 100% 135 ms. At 50%, counters record 2,259,793 total reads/probed entries, including 1,126,500 history-index entries. The 100% case has no known fragment of A and returns inconclusive immediately. At 90%, doubling the row count from 1,500 to 3,000 takes 887 ms to 3,376 ms.

[...]

Issue #2783, opened 2026-09-10
Specify and enforce one final row write per mergeable and exclusive transaction


aeplay:
🤖 Specify and enforce the authoring invariant for both mergeable and exclusive transactions: a transaction produces at most one final write per row. Repeated edits while it remains open must be collapsed by the authoring node into the final accumulated row state; intermediate edits must not become separate committed versions.

For example, inserting a row, updating its title and then updating its done flag in one transaction must produce one final write preserving both updates. The write count must describe emitted writes rather than API calls.

Work:
- State this explicitly in the transaction and public API specs, with a numbered invariant.
- Audit both mergeable and exclusive authoring paths and enforce the rule consistently.
- Define the precise physical table/branch row coordinate and document content/deletion behavior for insert, update, delete and restore sequences.
- Cover both transaction modes with tests asserting final state, emitted versions and n_total_writes; include repeated patches to different fields and repeated overwrites of one field.
- Verify intermediate open edits do not escape as independent committed versions.

Existing mergeable coverage includes mergeable_tx_coalesces_insert_then_update_for_same_row, mergeable_tx_coalesces_restore_then_update_for_same_row, mergeable_tx_coalesces_repeated_same_row_updates and mergeable_tx_coalesces_update_then_delete_for_same_row in crates/jazz/src/db/tests/transactions.rs. INV-API-26 currently describes grouping and version counts but does not explicitly state this collapse rule.

Deferred follow-up requested during the performance investigation; not part of the current ancestry-lookup optimization.

PR #2785, opened 2026-09-10
perf: scope merge ancestry lookups to one physical row

aeplay:
🤖 A row ancestry check currently loads every row version in each visited transaction, sorts them, and then discards all but the requested physical row. For a transaction updating 1,350 different rows, repeating that lookup for each row amplifies shallow ancestry into approximately quadratic work. Profiles attributed 6.38 seconds of the batch and 5.25 seconds of its reload to this walk.

This change keeps the ancestry algorithm and merge semantics, but retrieves only the requested row. Resident transactions use the existing row-addressable cache. Cold transactions use the immutable history primary key: physical table, branch, row, transaction time and node alias. The transaction-presence check remains, and a resident cache remains authoritative rather than silently falling back to stored history. Authored-schema decoding preserves table identity across renames.

For example, after A -> B -> C for one row, receiving older B must not introduce a false concurrent head alongside C. Conversely, two children B and C of A must remain concurrent until merged. Both cases still follow that row's parent links; other rows written by the same transaction do not participate. Missing history retains the existing reachability behavior. Transaction-wide completeness checks and full commit reads are unchanged.

No wire or storage format changes. Broader caller audit: #2784. Separately deferred authoring collapse invariant: #2783.

Validation so far: a resident/cold regression test proves ancestry answers and zero transaction-wide helper calls. Restoring the previous walker fails with 3 calls versus 0; mutation restored. This test is internal because unnecessary materialization is not observable in query results. Full Jazz library: 1,995 passed, 2 ignored. Wire fixture integration suite: 10 passed. CI exposed an old outgoing-encoder rejection expectation from the preceding trusted-decoding slice; this PR also moves that assertion to untrusted decode. Routing it through trusted decode fails as expected; mutation restored.

Same optimized persistent-browser workload, 1,500 rows / 1,350 updates:

| Operation | Before this PR | After |
|---|---:|---:|
| Initial runtime-reopen read | 976.41 ms | 965.07 ms |
| Batch commit + local durability | 7,491.53 ms | 1,166.78 ms |
| Post-batch runtime-reopen read | 6,094.77 ms | 1,010.54 ms |

Four-size sweep passed. Batch times at 150/300/750/1,500 rows: 118.50 / 216.20 / 531.19 / 1,166.78 ms, replacing the prior approximately quadratic growth with approximately linear growth over this range. The harness asserts updated row values, subscription delivery and values after reopen. Single-run measurements on synthetic data, no concurrent local builds/tests, not cold-browser startup or latency percentiles.

Separate passing CPU-profile run: ancestry falls from 6,379 ms to 9.02 ms in the worker batch and from 5,252 ms to 7.87 ms in the foreground reload. Remaining costs include version ingestion/current-index maintenance and supporting-view publication; ancestry is no longer dominant. Inclusive profile totals overlap and must not be added.

Draft, coordinator-only implementation and self-review under the user's single-track instruction. No independent review or merge approval claimed.

### DIFF
--- a/crates/jazz/src/node/ingest/view_updates.rs
+++ b/crates/jazz/src/node/ingest/view_updates.rs
@@ -743,7 +743,7 @@ where
         Ok(())
     }
 
-    async fn content_version_reaches_tx(
+    pub(super) async fn content_version_reaches_tx(
         &mut self,
         table_id: PhysicalTableId,
         branch_key: &BranchKey,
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Anselm approved #2785 in the coordinator conversation. Recorded for integration; merge remains subject to the final checkpoint gates.

aeplay:

PR #2786, opened 2026-09-10
perf: batch immutable history ingestion with ensure_exact

aeplay:
🤖 This changes transaction ingestion from loading and comparing every stored row in a known transaction to checking only the incoming immutable row versions, through a batch-scoped `ensure_exact` operation.

### Behavior, with examples

Suppose a transaction contains 1,500 rows and a peer sends 10 of them again. Previously, known-transaction matching loaded the transaction's stored versions and searched that collection for each incoming row. Now each of those 10 coordinates is compared directly with its stored bytes:

- Missing coordinate: stage its insertion and the corresponding derived-state changes.
- Identical bytes: accept the replay without inserting the row or emitting another insertion delta.
- Different bytes at the same immutable coordinate: reject the commit unit. Other insertions prepared in that batch are not published.

Repeated coordinates within the same batch observe earlier staged writes. The batch is tied to one Database and its resident publication revision: intervening publication or applying it to another Database rejects the stale preparation. This preserves the existing serialized Database ownership contract; it is not a cross-writer storage compare-and-swap.

Jazz still owns transaction identity, authorization, partial-transaction cardinality, fate, visibility and persistence. Groove owns staged exact-match insertion and atomic application. Existing fate processing may still read a whole transaction; this change removes that read specifically from immutable-version matching.

### Storage implementation

`compare_value` reports absent / identical / different without requiring an owned copy of the existing value. Memory compares borrowed bytes, RocksDB uses pinned values, IndexedDB's tree compares inline bytes or overflow chunks while hydrating pages as needed, and SQLite compares the blob inside its query. Wrappers preserve backend dispatch. Other implementations have a correct `get`-based fallback.

There is no wire or persisted-format change. Alias allocation occurs before batch preparation, including multi-transaction authority publications and receiver snapshots containing several authors. Known complete transactions that need independent fate updates are explicitly ingested after the prepared receiver batch commits. For example, a snapshot can insert one new row while changing another known transaction from pending to accepted, without invalidating the batch or losing the fate update.

### Validation and measurement

Tests cover identical replay, conflicting replay, atomic rejection, stale preparation, cross-Database use, later overwrites, and cold inline/overflow comparisons. A Jazz regression checks that known-transaction matching no longer adds a whole-transaction history read. The final Jazz library suite passes (1,998 passed, 2 ignored). Memory/SQLite/RocksDB/IndexedDB comparison tests and mutation checks passed. [Hosted CI passed](https://github.com/garden-co/jazz/actions/runs/34539216641) on implementation revision `e05e400dd5`. The quiet Groove suite passed 732 tests (2 ignored). The optimized browser four-size sweep, repeat and Rust-function profiling run also passed.

### Measured outcome

Same optimized WASM harness, 1,500 rows and 1,350 updates:

| Operation | Before | After, two quiet runs |
| --- | ---: | ---: |
| First read after runtime reopen | 965 ms | 778–808 ms |
| Batch commit through local durability | 1,167 ms | 1,116–1,181 ms |
| First read after post-update reopen | 1,011 ms | 856–865 ms |

This improves the first read by roughly 16–19% and the post-update read by 14–15%; bulk writes are effectively unchanged in this sample. These are runtime reopens in a loaded Chromium session, not browser-process cold starts.

The targeted foreground matching path improved about 5× (224 → 43 ms on initial read; 223 → 46 ms during the batch). Worker exact insertion/comparison now accounts for about 7 ms during the batch. Remaining costs include supporting-row publication, repeated record-field access/descriptor construction and page encoding; these need separate work to reach the larger end-to-end target. Detailed measurements and the next performance thesis are in `dev/benchmarks/post-ancestry-profile.md`.

### DIFF
--- a/crates/groove/SPEC/2_storage_model.md
+++ b/crates/groove/SPEC/2_storage_model.md
@@ -836,3 +836,35 @@ The former hybrid columnar-base proposal is rejected and is not part of Groove's
 - 🔶 [#1775](https://github.com/garden-co/jazz/issues/1775) — Current-base selection, compaction handoff, scan exclusion, and compaction-quality receipts.
 - 🔶 [#1774](https://github.com/garden-co/jazz/issues/1774) — Portable storage guarantees, reopen normativity, async persistence, serverless backends, row encoding, and compression policy.
 - 🔶 [#1776](https://github.com/garden-co/jazz/issues/1776) — Explicit index declarations and stale-index behavior.
+
+### Batch-scoped immutable record insertion
+
+`DatabaseBatch::ensure_exact(database, table, key, record)` resolves an immutable
+record against the resident database and preceding operations in that batch.
+It returns `Inserted`, `AlreadyIdentical`, or `Conflict`. `Inserted` stages the
+record but does not publish or persist it. `AlreadyIdentical` adds no physical
+write or query delta. `Conflict` invalidates the whole batch; attempting to
+apply it must fail before publishing any of its writes, without poisoning the
+database. Later operations must not contradict an ensured record.
+
+Equality compares the same locally encoded record representation, including
+its variant discriminator. It does not decode and reconstruct logical fields.
+The operation retains its absence decision for delta computation, avoiding a
+second storage read for inserted records. Repeated keys observe earlier staged
+writes. An empty duplicate-only history batch may still accompany mutable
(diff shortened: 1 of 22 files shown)


aeplay:
🤖 Anselm approved #2786 in the coordinator conversation. Recorded for integration; merge remains subject to the final checkpoint gates.

PR #2787, opened 2026-09-10
perf: index supporting row witness selection

aeplay:
🤖 Supporting-row publication searched a transaction's retained versions twice for every requested row. For a transaction containing 1,500 requested content rows, that can require about 2.25 million candidate comparisons, repeatedly decoding record identities.

This change builds a content-coordinate presence index once per transaction. For example, publishing all rows from a large insert now checks each requested coordinate in the index instead of scanning the transaction again. A missing coordinate still consults the maintained replacement winner; a matching-transaction replacement is appended as before and its content coordinate enters the index. If no content witness is found, the existing storage fallback still runs.

The index only answers whether a content witness exists. It does not select, reorder, or deduplicate the retained version vector. Content and deletion/restore versions remain distinct, and complete exclusive payload handling, permissions, storage, and wire formats are unchanged.

Validation: full Jazz library suite passed (1,998 passed, 2 ignored), Clippy and formatting passed. The suite needs a 4 MiB test-thread stack and a raised file-descriptor limit in this shell; initial attempts hit those environment limits. This is a draft, not yet independently reviewed.

Stacked on #2786; this PR isolates the indexed selection change from the previous ingestion work.

Optimized Chromium receipt (1,500 synthetic rows; 1,350 updates in one transaction, including local durability): two runs measured initial reopen/read at 603.0/603.3 ms (previous slice 778.4/808.2 ms), bulk update at 1062.0/1089.2 ms (previous 1181.2/1115.7 ms), and post-update reopen/read at 685.7/696.6 ms (previous 864.6/855.6 ms). Both browser runs passed. These are fresh runtimes inside a warm browser, not a cold browser-process launch; timings are two-run observations, not confidence intervals. The measured Rust source differs from committed a7c85d2e7b only by rustfmt line wrapping.

### DIFF
--- a/crates/jazz/src/node/views.rs
+++ b/crates/jazz/src/node/views.rs
@@ -69,18 +69,6 @@ fn maintained_view_tx_versions_contain_winner(
     })
 }
 
-fn maintained_view_find_content_witness<'a>(
-    tx_versions: &'a [VersionRow],
-    entry_table: &str,
-    row_uuid: RowUuid,
-) -> Option<&'a VersionRow> {
-    tx_versions.iter().find(|version| {
-        version.table() == entry_table
-            && version.row_uuid() == row_uuid
-            && version.deletion().is_none()
-    })
-}
-
 fn merge_receiver_version_bundle_ref(
     bundles: &mut BTreeMap<TxId, VersionBundle>,
     bundle: VersionBundleRef<'_>,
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved #2787 in the coordinator conversation. Recorded for integration; merge remains subject to the final checkpoint gates.

PR #2788, opened 2026-09-11
bench: native local batch profiling and overhead analysis

aeplay:
🤖 Add a native counterpart to the synthetic 1,500-row browser performance receipt so profiling iterations do not require a WASM build.

The fixture sends a Local supporting-row snapshot from a persistent worker to an in-memory foreground, authors 1,350 updates in the foreground, uploads them to the worker, delivers the updated scope, and repeats the read after reopening. It runs against memory and RocksDB, reports separate phase times and storage/IVM metrics, and asserts exact row-ID and completed-row sets. A raw 128-byte storage baseline provides context, not an equivalent database operation.

The report traces native profiles back to code. The largest finding is a quadratic cold parent-miss path: at 50% updated, a 1,500-row snapshot causes about 2.26 million read/index entries and takes 2.15 seconds to ingest. At 100% updated it takes 135 ms. It also identifies trusted internal encode/decode/re-encode validation and repeated descriptor/row reconstruction.

No production runtime or storage/wire behavior changes. This is a research/benchmark PR on #2787. It excludes JS, IndexedDB, external auth, remote fate assignment and facade transaction staging; native timings are not predictions of browser timing. The report documents those boundaries and profiling methodology.

Validation: focused benchmark compile and deterministic protocol smoke passed. A temporary mutation that kept updated rows incomplete made the exact completed-row assertion fail; it was restored. Optimized native scaling and percentage sweeps passed within the existing 4,096-version transaction limit. Final optimized smoke also passed.

### DIFF
--- a/crates/jazz/Cargo.toml
+++ b/crates/jazz/Cargo.toml
@@ -199,3 +199,8 @@ required-features = ["runtime"]
 [[test]]
 name = "fate_replay"
 required-features = ["testing"]
+
+[[bench]]
+name = "local_batch_phases"
+harness = false
+required-features = ["testing"]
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Anselm approved #2788 in the coordinator conversation. Recorded for integration; merge remains subject to the final checkpoint gates.

Issue #2789, opened 2026-09-11
perf: remove trusted internal row round-trips and repeated representation work


aeplay:
🤖 Native profiling of the 1,500-row Local worker/foreground workload shows substantial trusted internal conversion work above storage. Reference: dev/benchmarks/native-local-batch-profile.md, branch perf/native-local-batch-profile, commit 2941c30fe3.

Concrete targets:

- node/descriptor_roles.rs::encode_current_payload_record still serializes field-type descriptors per row, encodes a row, decodes it, re-encodes it to compare canonical bytes, then emits another descriptor. This is trusted internal output. It appears in about 18% of batch-publication samples, underneath decode_typed_terminal_record (~38%, overlapping).
- maintained_subscription_view::decode_typed_version_witness expands typed cells/authors into values and constructs another history record. Share schema plans and preserve exact authored-row provenance instead of unnecessary reconstruction.
- groove/large_values.rs::materialize_record_attempt decodes all values and copies the original bytes even when no indirect value changes. Its path appears in about 14% of worker-ingest samples for the small inline-value fixture. Establish a safe unchanged/inline fast path from trusted descriptors/records.
- views.rs::preflight_view_bundle_conflicts still rebuilds every stored version into wire records before ingest_known_transaction performs its own work. The preflight appears in about 37% of already-authoring foreground receipt samples. Reuse exact identity/batch evidence while preserving complete/view-scoped transaction checks.
- normalize_public_current_rows/CurrentRow::project rebuild descriptors and metadata per row even for the simple all-column query (~50% of initial foreground query samples).

Do not sum these inclusive percentages. Allocation/free and byte-copy self samples together account for at least 33–39% of the main phases. Native memory and RocksDB Jazz timings are similar; raw storage batches are much faster.

Design direction: compile schema/descriptor work once and carry owned immutable bytes plus typed identity/provenance through processing. Preserve authorization, authored-row fidelity, branch/schema semantics and malformed-input resource boundaries. Remove trusted encoder round-trip validation in accordance with the established trust contract. This is runtime work; no format change is proposed.

The separate quadratic partial-parent miss is tracked in #2784 and should be addressed first. Keep native benchmarks/counters as the inner loop, with fully optimized browser checkpoints.

aeplay:
[...]

PR #2791, opened 2026-09-11
perf: compile current-result payload projection per batch

aeplay:
🤖 Compile current-result payload projection once per terminal delta batch, then copy encoded fields instead of repeatedly decoding and encoding every row.

For a 1,350-row update, the old publisher rebuilt the same role descriptor and field-type comparison for every row, materialized values, encoded the selected fields, decoded that output, and re-encoded it to check itself. The new plan validates the compiled source/target schemas once per source descriptor and uses Groove's existing byte-copy projector. For example, when the input carries an unrelated field before the requested payload and row ID, the mapping selects only the correct fields and rebuilds their framing. Nested enum registry rebinding, nulls, arrays, join occurrence identities and reordered fields retain their encoded meaning.

The cache belongs to one terminal's delta batch, so it cannot reuse a plan against another query schema. It uses interned source descriptor handles for lookup, never for durable identity. No storage or wire format changes; this does not remove the runtime descriptor attached to each publication payload or change the protocol's supporting-row semantics.

With the quadratic parent fix already applied, native batch publication improves from 163 ms to 136 ms in memory and 167 ms to 137 ms with RocksDB. Initial publication improves from 216 ms to 195 ms in memory. The encoder accounts for 113 samples before and 3 after across five runs. These are synthetic native phase measurements, not browser timing predictions. The accompanying report includes the parent fix's results, methodology, limitations and remaining ingestion/IVM conversion bottlenecks.

Validation: 2,000 active library tests pass, 2 ignored. The new byte-equivalence test covers 32 reused rows, reordered fields, nested enum registry rebinding, arrays and nulls. A temporary wrong-field mapping makes it fail; restored. Existing publication role/occurrence tests pass. Optimized memory/RocksDB acceptance and fresh Rust profiles completed. Draft pending broader integration and independent review.

Builds on #2790. First slice of #2789.

### DIFF
--- a/crates/jazz/src/node/descriptor_roles.rs
+++ b/crates/jazz/src/node/descriptor_roles.rs
@@ -316,78 +316,98 @@ fn current_role_schema(
     )?)
 }
 
+/// One compiled field mapping per source descriptor and publication schema.
+/// The caller scopes reuse to a single terminal's delta batch.
+pub(super) struct CurrentPayloadEncodePlan {
+    projector: records::RecordProjector,
+    descriptor: Vec<u8>,
+}
+
+impl CurrentPayloadEncodePlan {
+    pub(super) fn new(
+        descriptor: RecordDescriptor,
+        schema: &super::query_engine::ResultMembershipSchema,
+    ) -> Result<Self, Error> {
+        let selected = std::iter::once(schema.row_field.as_str())
+            .chain(
+                schema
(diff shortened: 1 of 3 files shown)


aeplay:
🤖 Anselm approved #2791 in the coordinator conversation. Recorded for integration; merge remains subject to the final checkpoint gates.

Issue #2792, opened 2026-09-11
Permissioned relay reopen fails scalar enum provenance validation


aeplay:
🤖 The anonymized `customer_cold_start` benchmark successfully primes a permissioned relay, then cannot reopen that relay's fresh RocksDB state:

`invalid stored value: scalar enum registry case provenance disagrees with authority identities`

Reproduced at full scale and at `JAZZ_CUSTOMER_SCALE=0.01`, on the current performance stack plus benchmark compatibility repairs in branch `bench/current-permissioned-load`. The repairs compare UUID membership against `user.account`, seed the corresponding account IDs, and finalize core seed transactions. The dataset is generated locally; the relay database is created and primed during the same invocation, not reused from an old release.

Full-scale cold loading passes all 39 per-table checks: 27,518 expected/observed rows, including 23,831 in the dominant child table. The warm-prime phase also passes before `Db::open` fails at the reopened relay. Thus no valid warm timing can currently be reported.

Reproduce after building `cargo build -p jazz-sim --bench customer_cold_start --profile perf` by invoking its executable with `JAZZ_CUSTOMER_IDENTITY=member JAZZ_CUSTOMER_PHASES=warm JAZZ_CUSTOMER_SCALE=0.01 JAZZ_CUSTOMER_MAX_TICKS=200000`.

Failure originates in `validate_scalar_enum_case_provenance_for_column` in `crates/jazz/src/node/physical/bindings.rs` during `open_db_node` in the benchmark. Investigate whether authority mapping replacement and durable scalar-enum provenance remain coherent across reopen. Do not remove the provenance validation to make the fixture pass. Root cause is not yet established; this may expose a runtime persistence/reopen defect rather than fixture API drift.

aeplay:
🤖 Post-release triage (2026-09-12): Merged #2903 repairs provisional enum identity rebinding on catalogue adoption and records cold/warm acceptance of all 27,518 expected rows. Closing the relay-reopen defect; already-incoherent historical caches are not claimed repaired.

PR #2793, opened 2026-09-11
bench: restore permissioned graph loads and record current baseline

aeplay:
🤖 Restore the historical permissioned graph benchmark on the current account and storage APIs, and record its first current cold-load baseline.

The fixture now compares UUID memberships against `user.account` and seeds matching fixture account IDs. Core seed writes are finalized as accepted after persistence, following the maintained `s3_permissions` fixture; previously all downstream tables stayed empty. A new seed-cache version keeps older generated stores separate. The test-only storage byte meter now derives column families from the current schema instead of querying a removed hard-coded family.

The full-scale member cold phase reaches exactly 27,518 rows across 39 subscriptions, including 23,831 rows in the dominant child table. Subscription readiness takes 51.9 seconds; final one-shot queries take 1.55 seconds; harness wall time is 53.8 seconds. This is one completed cold phase, not a repeated median or browser measurement. Diagnostic codec probes remain included. Our iteration target is approximately 5 seconds to correct cold readiness, ideally under 1 second.

Warm priming passes, but reopening that fresh relay fails scalar-enum provenance validation. This also reproduces at 1% scale and is tracked separately in #2792. No warm timing is claimed, and the combined process exits unsuccessfully at that point. No assertions were weakened to produce the cold receipt.

Validation: optimized benchmark build, successful full-scale cold visibility/materialization checks, reproductions of the original fixture failures and reduced-size warm reopen failure; commit Clippy/format/sensitive-data checks passed. The report documents exact measurement boundaries and remaining blocker. Production query/sync/permission behavior and storage formats are unchanged. Draft pending the warm blocker and further profiling; builds on #2791.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -284,7 +284,7 @@ const CHILD_TABLES: usize = 6;
 const DOMINANT_CHILD_TABLE: &str = "res_l_child_3";
 const DOMINANT_CHILD_VISIBLE_PARENTS: usize = 36;
 const DOMINANT_CHILD_TOTAL_PARENTS: usize = 65;
-const SEED_CACHE_VERSION: &str = "customer-cold-start-seed-v5";
+const SEED_CACHE_VERSION: &str = "customer-cold-start-seed-v7";
 const SEED_CACHE_READY: &str = ".jazz_customer_seed_ready";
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Anselm approved #2793 in the coordinator conversation. Recorded for integration; merge remains subject to the final checkpoint gates.

PR #2794, opened 2026-09-11
perf: skip parent history reloads without waiting children

aeplay:
🤖 Avoid rereading downloaded transactions that have no waiting children.

When a receiver imported a complete batch, it invalidated the transaction caches and then loaded every transaction's row versions to check for pending children. That lookup probed history indexes across unrelated tables even when no child referred to the transaction. The pending-edge table was also scanned once per parent.

Now settlement scans pending constraints once for the batch, groups relevant constraints by parent, and loads history only for those parents. For example, downloading 30,000 independent rows with no waiting children needs no parent-history reloads. An unrelated waiting child does not change that. If a waiting child references an arriving complete parent at the wrong physical row, branch or layer, the existing causality rejection still applies. Missing or partial parents remain inconclusive, and accepted children are not retroactively rewritten.

The optimized native permissioned fixture delivers all 27,518 expected rows across 39 subscriptions in 41.8 seconds, compared with 51.9 seconds before this change (individual samples, not medians). This is a fresh relay and client against a populated Core using RocksDB and in-process transport; it is not a browser timing. No storage or wire format changes.

Validation: 2,001 Rust library tests passed (2 ignored), including receiver parent/reopen cases. The new mechanism regression passes; restoring the old implementation makes it fail with 32 history reloads instead of zero. The broad run used 4 MiB test-thread stacks and an increased file-descriptor limit; earlier undersized harness runs aborted or hit descriptor exhaustion. This is a draft performance slice on #2793, not a release-ready claim.

### DIFF
--- a/crates/jazz/src/node/ingest/validation.rs
+++ b/crates/jazz/src/node/ingest/validation.rs
@@ -234,10 +234,9 @@ where
     ) -> Result<(), Error> {
         for parent in complete_parents {
             self.invalidate_tx_version_tables_cache(*parent);
-            self.reject_mismatched_pending_children_for_parent(*parent)
-                .await?;
         }
-        Ok(())
+        self.reject_mismatched_pending_children_for_parents(complete_parents)
+            .await
     }
 
     /// Resolve durable constraints left by children that referenced a parent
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Anselm approved #2794 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2797, opened 2026-09-11
perf: prepare projections per field and borrow supported values

aeplay:
🤖 Make projection execution borrow supported fields independently instead of decoding an entire row when one expression needs conversion.

A projection selecting `title`, `author.account`, a constant, and a remapped enum previously fell back as a whole: it allocated owned values for every field, encoded a temporary row, and copied that row into the batch output. Simple fast-path constants also cloned their encoded Vec for every input row.

The prepared plan now resolves nested descriptor/index paths and checks copied/nullable field types once. Copies and constants borrow encoded spans, nullable wrapping writes a prefix plus the borrowed payload directly, and variable offsets are patched in the shared output buffer. Only the enum-conversion field uses the existing semantic conversion. The hydration evaluator also uses prepared field plans. This does not yet implement arbitrary recursive enum conversion directly over encoded bytes.

The batch remains immutable after publication and each output row owns a shared Bytes slice. Input lifetimes do not escape. Row order, signed weights, exact enum meanings and field identities remain intact. If an enum conversion intentionally omits a row, all bytes already appended for that row are rolled back before the next row. Constant preparation errors are cached but raised only if an input row evaluates the expression, preserving empty-input behavior. No storage or wire encoding changes.

Validation: all 738 Groove and 2,002 Jazz library tests pass (2 ignored in each). The mixed-field regression compares bytes against the existing semantic implementation and observes 12 owned field evaluations for 12 rows instead of 96. It checks output ownership after dropping input. Other checks cover nested fields, constants, nullable values, negative weights, partial-row omission, invalid copy plans and lazy constant errors. Forcing the allocating fallback fails the evaluation-count assertion; disabling rollback fails the shared-batch continuity assertion. The existing enum-parameter empty/populated coverage test caught the eager-error regression during development and now passes.

The full-scale native permissioned fixture still returns all 27,518 rows. Readiness measures 35.423s and 35.254s versus the prior 35.863s observation: a small observed change, not a statistically established speedup. Complete-run allocation requests fall from 485.6M to 483.5M (0.43%); cumulative requested bytes fall from 295.662GB to 295.523GB. This is allocation churn, not peak memory. The prior trace shows 70.7% of projection visits are pure field selection, 29.1% add constants/nullability, and only 0.16% have complex fields that disabled the old plan. This explains why broader fast-path coverage alone does not materially reduce this workload. The accompanying [measurement report and code walk](https://github.com/garden-co/jazz/blob/perf/prepared-field-projections/dev/benchmarks/prepared-field-projections.md) explains the remaining projection-stage count: compositional Jazz lowering plus exact-node deduplication, without projection fusion or byte-preserving rename elimination.

Draft on #2796. Full CI, browser acceptance and independent review are outstanding; this is not a release-ready claim.

### DIFF
--- a/crates/groove/src/ivm/runtime/operator_updates.rs
+++ b/crates/groove/src/ivm/runtime/operator_updates.rs
@@ -601,18 +601,38 @@ impl NodeState {
             .sum::<usize>();
         let mut output = BytesMut::with_capacity(estimated_output_bytes);
         let mut spans = Vec::with_capacity(input.deltas.len());
-        let mut raw_projection_scratch = RawProjectionScratch::default();
         for delta in &input.deltas {
             let span = if let Some(fields) = raw_projection {
-                output_desc
-                    .project_raw_fields_into(
-                        &input.descriptor,
-                        delta.raw(),
-                        fields,
-                        &mut output,
-                        &mut raw_projection_scratch,
-                    )
-                    .map_err(IvmRuntimeError::RecordEncoding)?
+                let start = output.len();
+                let result = output_desc.project_raw_fields_into(
+                    &input.descriptor,
+                    delta.raw(),
+                    fields,
+                    &mut output,
+                    |index, output| {
+                        let value = project_field_value(
+                            &project.expressions[index],
+                            index,
+                            output_desc,
+                            &input.descriptor,
+                            delta.raw(),
+                        )?;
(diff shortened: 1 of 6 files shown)


aeplay:
🤖 Anselm approved #2797 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2798, opened 2026-09-11
perf: compose field projections and share identical record bytes

aeplay:
🤖 Avoid copying byte-identical projections and compose adjacent field-only selections/renames.

A projection may change logical names or declaration order without changing the physical encoded layout. Its prepared plan now proves matching source/target types, fixed/variable counts and exact physical slots once. If all slots match, output rows share the input Bytes while retaining the new descriptor, ordering and signed weights. Swapping variable payload slots still takes the copying path. This shares immutable ownership rather than borrowing beyond input lifetime.

The compiler also follows adjacent MapProject nodes containing only ordinary field selections and composes their mappings directly onto the underlying source. For example `title -> label -> name` becomes one projection from `title` to `name`. A separately used `label` output remains available. This does not cross constants, enum conversions, filters, joins, winner selection or other semantic operators: an unselected expression can still be fallible or omit a row, so discarding it would change behavior.

Validation: all 740 Groove and 2,002 Jazz library tests pass (2 ignored in each). New internal checks verify physical-layout byte sharing, logical-order changes, incompatible variable-slot swaps, composed output values, shared inner nodes and preservation of an unselected fallible expression. Mechanism checks are internal because correct public row values alone cannot reveal redundant copies or redundant executing nodes. Disabling byte reuse fails the pointer-sharing test; disabling composition fails the graph-dependency test. Both mutations were restored. The full permissioned native load returned all 27,518 expected rows in 35.320 s versus 35.254 s on #2797: no measurable end-to-end gain from this small slice in these individual runs.

No storage or wire format change. Draft on #2797; full CI, browser acceptance and independent review remain outstanding.

### DIFF
--- a/crates/groove/src/ivm/runtime/compilation.rs
+++ b/crates/groove/src/ivm/runtime/compilation.rs
@@ -744,32 +744,71 @@ impl IvmRuntime {
             GraphBuilder::Project { input, fields } => {
                 let compiled_input =
                     self.add_dedup_graph_cached(input, output_memo, compiled_memo)?;
-                let input_node = compiled_input.node;
+                let mut input_node = compiled_input.node;
                 let input_output = compiled_input.output;
                 let output = inferred_output;
-                let mapping = fields
+                let mut expressions = fields
                     .iter()
-                    .filter_map(|field| {
-                        field.source().map(|source| {
-                            resolve_field_ref(&input_output, source).map(|idx| (0, idx))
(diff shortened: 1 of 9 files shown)


aeplay:
🤖 Anselm approved #2798 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2799, opened 2026-09-11
perf: reuse record metadata and owned version buffers

aeplay:
🤖 Prepare record metadata once per shape/batch instead of reconstructing it for each row version.

A cold permissioned load repeatedly constructs the same descriptor fields even though the resulting descriptors are interned. For example, reading 1,000 stored rows with two physical variants used to construct 1,000 variant descriptors. Table and index source loading now group by variant and construct two. Payload bytes and signed row semantics are unchanged; an unknown variant still fails the read.

The fixed account/identity descriptors are initialized once. Durable non-null row authors remain distinct from nullable session/capability authors. Decoding borrows the nested identity bytes rather than allocating an intermediate owned record, while preserving issuer, subject, reserved-account and SYSTEM-origin checks. This does not let persisted authorship rehydrate the in-process SYSTEM capability.

Version encoding uses a prepared history descriptor or the fixed deletion-register descriptor directly, avoiding construction of a full storage table and its indices for every version. The history cache is bounded to 128 shapes per thread and compares table name, ordered column names/types and large-value semantic kinds. A changed type, JSON/text distinction or enum binding cannot reuse a stale layout. Eviction changes preparation cost only. Caller-supplied history descriptors retain precedence.

Version decoding now resolves the physical table ID once rather than formatting a name for every catalogue candidate. It borrows the matched schema and retains its incoming owned record buffer. Physical table lookup still distinguishes history from register names and rejects noncanonical names exactly as before.

No storage or wire encoding changes. This is preparation/ownership work; other row payload copies, canonical author interning, transaction/version object construction and remaining schema clones are not eliminated.

Validation: all 741 Groove and 2,004 Jazz library tests pass (two ignored in each), plus the differential oracle at two seeds and churn depths 10/1,000. Mutation checks fail when variant descriptors rebuild per row (1,000 versus 2), history preparation bypasses its cache (500 versus 5), or padded physical names are accepted. Mutations are restored. All three incremental-delivery canaries pass. The 27,518-row permissioned native fixture improves from 35.254 s before these two PRs to 32.415 s (individual clean runs, about 8%). Allocation requests fall from 483,481,285 to 383,877,574 (20.6%); cumulative requested bytes fall only 2.1%, so larger record/value copying remains. See [the benchmark report](https://github.com/garden-co/jazz/blob/perf/prepared-metadata-descriptors/dev/benchmarks/projection-composition-metadata-preparation.md) for scope, intermediate measurements and remaining hotspots. Draft on #2798; independent review, full CI and browser acceptance remain outstanding.

### DIFF
--- a/crates/groove/src/ivm/runtime/operator_updates.rs
+++ b/crates/groove/src/ivm/runtime/operator_updates.rs
@@ -137,6 +137,41 @@ impl NodeState {
         )
     }
 
+    // Resolve each variant layout once for the batch, before projecting rows.
+    // A descriptor is interned, but constructing its fields is not free.
+    pub(super) fn group_source_rows<'a>(
+        table: &crate::schema::TableSchema,
+        rows: impl IntoIterator<Item = &'a [u8]>,
+    ) -> Result<Vec<TableDelta>, IvmRuntimeError> {
+        let mut grouped = HashMap::<u32, TableDelta>::default();
+        for stored in rows {
+            let (variant_tag, payload) = crate::records::split_variant_record(stored)?;
+            let group = match grouped.entry(variant_tag) {
+                std::collections::hash_map::Entry::Occupied(entry) => entry.into_mut(),
+                std::collections::hash_map::Entry::Vacant(entry) => {
+                    let descriptor =
+                        table
+                            .record_schema_for_variant(variant_tag)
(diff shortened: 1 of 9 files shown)


aeplay:
🤖 Anselm approved #2799 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2800, opened 2026-09-11
perf: borrow authors and encoded fields during row ingestion

aeplay:
🤖
Incoming row versions and storage commit bookkeeping repeatedly reconstructed owned record values even when the encoded data was already available. This change borrows those representations and allocates the final output instead.

For example, an incoming row with text, nullable fields and structured authors previously decoded its parents, author records and cells into owned values, validated cells by encoding them again, and then encoded the storage row. Compatible fields now copy their encoded spans into one record buffer; storage-only aliases and converted timestamps are generated directly. Differing field types retain the conversion path. Author admission and sorted-parent checks remain in place.

Author lookup now hashes borrowed `(account, issuer, subject)` components in the existing intern pool. Canonical spelling and native row/session encodings are built once per identity. Reading authors from records borrows the nested record and strings. SYSTEM row provenance remains distinct from the SYSTEM authority capability.

Storage commit accounting now traverses encoded containers to find indirect large-value references. Inline strings and unrelated fields are not decoded into values. Repeated references still count separately, including references inside arrays, nullable values, nested records and enum payloads.

Storage and wire bytes are unchanged. Test builds compare the direct incoming-version encoder against the previous Value-based encoder throughout the unit ingress corpus. Focused record tests cover byte identity, borrowed nested records, reference multiplicity and early stopping.

Measured on the same optimized native 39-subscription/27,518-row permissioned cold-load fixture: all-ready time fell from 32.415s to 29.485s (9.0%), allocation requests from 383.9M to 318.5M (17.0%), and cumulative allocated bytes from 289.4GB to 275.2GB (4.9%). A second clean run was 29.720s, confirming an 8.3–9.0% readiness gain across these two runs. These are individual clean timing runs and separate allocation runs, not browser results or peak RAM. The 5s goal remains unmet.

The new profile still shows substantial allocator/copy work. Remaining sampled callers include complete-parent version cloning, per-write primary-key descriptor creation, witness history descriptor reconstruction, and other Value-based storage encoders. See the [measurement and profile report](https://github.com/garden-co/jazz/blob/perf/borrowed-author-version-values/dev/benchmarks/borrowed-author-version-values.md).

Validation: Jazz library suite 2,005 passed/2 ignored and Groove 743 passed/2 ignored; record tests, all three incremental delivery canaries, and the two-seed differential oracle passed. Test mutations that corrupt offsets, stop indirect-reference traversal early, or disable author-record reuse each fail the corresponding new test and were restored. An initial deletion-coverage test failure passed alone on the same binary and in the full rerun; the report records it. Clippy passed, including a narrow immutable-identity exemption for derived author caches.

Draft: independent review, full canonical CI and browser acceptance remain outstanding.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,4 @@
+# Author identity is immutable. Only derived encoded-record caches use OnceLock;
+# manual Eq/Hash and stable Debug exclude them. Initializing these caches cannot
+# change an author or any containing query/authorization key.
+ignore-interior-mutability = ["jazz::ids::InternedAuthor"]
--- a/crates/groove/src/db/commit.rs
+++ b/crates/groove/src/db/commit.rs
@@ -151,12 +151,8 @@ impl Database {
                 else {
                     continue;
                 };
-                for value in descriptor.bind(record).to_values()? {
-                    if value_contains_large_ref(&value, &staged.value_ref) {
-                        found = true;
-                        break;
-                    }
-                }
+                found = descriptor
+                    .visit_large_value_refs(record, |reference| reference == &staged.value_ref)?;
                 if found {
                     break;
                 }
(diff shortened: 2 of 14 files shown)


aeplay:
🤖 Anselm approved #2800 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2801, opened 2026-09-11
perf: attribute native cold load to exclusive node phases

aeplay:
🤖 The native cold-load benchmark previously exposed whole-process CPU samples and allocation sites without a non-overlapping breakdown of the time spent in Core, relay, and client phases. Add opt-in nested timing under the existing `cold-settle-attribution` feature.

For example, a relay receiving a supporting snapshot can perform parent completion, apply storage changes, update query operators, persist writes, and publish another snapshot. The report measures each boundary and subtracts nested time from its parent. Exclusive times therefore partition the relay tick; inclusive times explain containment and must not be summed. Snapshot emission asserts the partition identity.

Async spans enter during polling and scoped cleanup. A suspended ingestion future does not remain the active phase while another node runs. Synchronous storage calls, OS preemption and work inside a poll remain elapsed time in that phase; this is not a hardware CPU-time counter. Root remainder includes uninstrumented work and executor waiting.

The benchmark captures this report at subscription readiness, before final diagnostic queries. Node roles come from its explicit sequential Core/relay/client ticks. The collector is thread-local and intended for this native driver, not a production multi-thread profiler. The benchmark transport's serialization/compression measurement has its own category.

Without the feature, annotations and the collector are disabled. No storage, wire, query or authorization behavior changes. The feature check also repairs an orphan doc comment and preserves Clippy's default immutable `bytes::Bytes` exemption alongside the existing author-cache exemption.

The full optimized 27,518-row fixture now attributes 29.55s of settling to Core 8.20s, relay 14.17s, and client 7.17s. Roughly 94.4% is classified beyond the explicit remainder. Query operators account for ~6.68s, query-output decoding ~4.44s, and supporting-row construction ~3.47s. Parent completion is only ~0.35s despite its allocation-stack prominence. All expected rows were present.

See the [phase table, definitions, and measurement limits](https://github.com/garden-co/jazz/blob/perf/cold-load-phase-attribution/dev/benchmarks/native-cold-phase-attribution.md). Projection counters record 1.89M input visits across the three nodes; these are repeated operator visits, not unique rows.

Validation: deterministic nested timing and pending/resumed future tests pass. Mutating child-time subtraction and role attribution makes the tests fail; restoration passes. The full fixture passes the runtime partition assertions. Feature-enabled and default-feature Clippy pass. Draft; independent review and full canonical CI are not claimed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1526,6 +1526,7 @@ dependencies = [
  "smallvec",
  "tempfile",
  "thiserror",
+ "tracing",
  "uuid",
  "web-time",
 ]
@@ -2349,6 +2350,7 @@ dependencies = [
  "rusqlite",
  "serde_json",
  "tempfile",
+ "tracing",
  "uuid",
  "zstd",
 ]
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,4 +1,5 @@
 # Author identity is immutable. Only derived encoded-record caches use OnceLock;
 # manual Eq/Hash and stable Debug exclude them. Initializing these caches cannot
 # change an author or any containing query/authorization key.
-ignore-interior-mutability = ["jazz::ids::InternedAuthor"]
+# Preserve Clippy's default immutable Bytes exemption as well.
+ignore-interior-mutability = ["bytes::Bytes", "jazz::ids::InternedAuthor"]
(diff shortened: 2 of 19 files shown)


aeplay:
🤖 Anselm approved #2801 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2802, opened 2026-09-11
perf: copy encoded witness cells and defer publication descriptors

aeplay:
🤖 Query-output witnesses already contain encoded user cells, but decoding them previously allocated a map of owned values, cloned those values into a second record representation, and encoded them again. Copy fields whose types agree directly into the history record. Metadata keeps its existing normalization, and differing field types retain a decoded fallback.

For example, a nullable text field or a nested record now keeps its encoded bytes while moving from the query witness layout into the history layout. A missing projected cell remains absent; this does not promote a partial query witness into a complete authored version.

Supporting-row publication still resolves the exact stored authored version before sending it. Previously it cloned the table schema and constructed a history descriptor before that lookup. Those operations now happen only if the stored version is absent and the synthesized-row fallback needs to verify completeness. Branch, physical table, schema identity, and authored-column checks remain in place.

No storage or wire format changes. An internal test oracle compares the optimized witness bytes with the previous value-based encoder. The full Jazz library suite passes: 2,005 passed, 2 ignored. All three incremental canaries and the two-seed maintained-vs-one-shot oracle (churn depths 10 and 1,000) pass. Draft, not yet independently reviewed or ready to merge.

Opt-in phase instrumentation further separates witness decoding, root-layout conversion, and canonical supporting-version preparation. A baseline on the full 27,518-row native fixture measured these at approximately 1.92s, 0.68s, and 1.60s respectively. These are nested components of the previous output/publication measurements, not additional time.

Continues #2789, on top of #2801.

Initial optimized comparison with identical phase instrumentation: settling 30.649s → 28.700s, witness reconstruction 1.923s → 1.191s, canonical supporting-version preparation/lookup 1.600s → 0.615s. Both runs produced all 27,518 expected rows. These are single-run observations; the targeted phases explain about 1.72s of the 1.95s total improvement. The 5s target remains far away. See [measurement details](https://github.com/garden-co/jazz/blob/perf/query-output-boundaries/dev/benchmarks/query-output-boundaries.md).

Sensitivity: flipping one reconstructed byte makes the byte-equivalence oracle fail in the maintained local index snapshot test. The exact mutation was restored and the test passes again.

### DIFF
--- a/crates/jazz-sim/src/phase_attribution.rs
+++ b/crates/jazz-sim/src/phase_attribution.rs
@@ -11,7 +11,7 @@ use tracing::{
     span::{Attributes, Id, Record},
 };
 
-const PHASES: [&str; 17] = [
+const PHASES: [&str; 20] = [
     "core",
     "relay",
     "client",
@@ -29,6 +29,9 @@ const PHASES: [&str; 17] = [
     "decode_query_outputs",
     "deliver_query_outputs",
     "benchmark_transport",
+    "decode_version_witness",
+    "rebind_terminal_output",
+    "canonical_supporting_version",
 ];
 const ROLES: [&str; 4] = ["outside_node_ticks", "core", "relay", "client"];
 #[derive(Clone, Copy, Default)]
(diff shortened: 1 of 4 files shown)


aeplay:
🤖 Anselm approved #2802 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2803, opened 2026-09-11
perf: remove discarded primary-key descriptor preparation

aeplay:
🤖 Groove built primary-key record descriptors for each pending write and for primary-key/index reads, then passed them to `record_store_for_table`, which immediately discarded the argument. Remove the obsolete helper and descriptor construction; construct the same RecordStore directly.

For example, a batch of 1,500 rows previously rebuilt a key descriptor once per pending physical write, despite the store receiving already encoded keys. Reads likewise paid this preparation after their keys had already been encoded and validated. The key values, key encoding, value descriptor, storage operations, validation, and IVM deltas are unchanged.

This is a net code deletion, not a new cache. No storage, wire, authorization, or query semantics change.

Validation: Groove library 743 passed, 2 ignored. All three incremental canaries and the two-seed maintained-vs-one-shot oracle pass. Initial optimized settling is 28.700s → 28.181s; the 1,500-todo native roundtrip is 425ms → 412ms. These are small, single-run gains; the change is worth retaining as a code deletion. Draft, not independently reviewed or ready to merge.

Continues the representation/allocation work in #2789, on top of #2802.

[Measurement details and todo progress](https://github.com/garden-co/jazz/blob/perf/remove-unused-key-descriptors/dev/benchmarks/unused-key-descriptors.md).

### DIFF
--- a/crates/groove/src/db/commit.rs
+++ b/crates/groove/src/db/commit.rs
@@ -176,13 +176,7 @@ impl Database {
         let stores = pending_writes
             .iter()
             .zip(&descriptors)
-            .map(|(write, descriptor)| {
-                let key_descriptor = self
-                    .table(write.table())
-                    .ok()
-                    .and_then(|table| table.primary_key.as_ref().map(primary_key_descriptor));
-                record_store_for_table(&overlay, write.table(), key_descriptor, descriptor)
-            })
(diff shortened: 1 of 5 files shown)


aeplay:
🤖 Anselm approved #2803 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2805, opened 2026-09-11
bench: allow explicit full-load allocation attribution

aeplay:
🤖 Allocation profiles now stop when subscriptions are ready, matching the cold-load phase boundary. Final one-shot verification and storage-size scans no longer inflate allocation totals. Receipts identify this scope as `connect_subscribe_settle`.

Explicit allocation-feature builds permit their sampler settings, print that their timings are not comparable, and emit `allocation_instrumented: true`. Ordinary builds retain the timing guard. Count sampling uses the configured allocation interval; a separate 16MiB volume interval attributes bytes without extrapolating one rare large allocation by the count interval. Allocations larger than that volume interval are sampled at their actual size. Both rankings are estimates; total allocator counters are direct counts.

The full anonymized fixture passes all 27,518 expected rows. The readiness-scoped run reports 241,155,830 allocations and 262,420,969,705 cumulative requested bytes, with 19,533 sampled stacks below the 50,000 cap. This is allocated volume, not resident memory. Native Zstd decompression dominates the largest byte stacks: its bulk API reserves the supplied 256MiB maximum because the dependency’s experimental upper-bound helper is disabled. This is a lead for a separately measured fix, not evidence that decompression dominates elapsed time.

Optimized allocation-feature build, scoped Clippy and formatting pass; the original clean binary rejects allocation settings with exit 2. No product/storage/wire behavior changes. Draft on #2804; continues #2789. Independent review and broad gates are not claimed.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -308,7 +308,14 @@ fn main() {
     #[cfg(feature = "cold-settle-attribution")]
     let _phase_subscriber =
         tracing::subscriber::set_default(jazz_sim::phase_attribution::Collector);
+    // Allocation attribution deliberately instruments the workload. Its timing
+    // is not a benchmark receipt, and sampler configuration must remain usable.
+    #[cfg(not(any(feature = "bench-alloc-sites", feature = "bench-alloc-metrics")))]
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved #2805 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2806, opened 2026-09-11
perf: reuse scalar join key scratch across batches

aeplay:
🤖 Scalar join input rows built a new heap-backed key buffer each time, then copied short keys into the existing inline JoinKey and discarded the buffer. Reuse one temporary encoder buffer across the batch and clear it before each row.

For example, thousands of UUID equality keys now share the temporary capacity while each retained key still owns its bytes. Composite, nullable and policy-comparison encoding remains unchanged; retained keys never borrow the scratch buffer. Longer keys still use JoinKey’s existing owned heap representation.

Groove library tests pass: 743 passed, 2 ignored. Scoped Clippy and formatting pass. The first clean optimized large-fixture run settles in 27.215s versus 27.287s on #2804, with all 27,518 rows. This difference is within noise; this is a small allocation simplification, not a demonstrated end-to-end speedup. Further attribution is in progress; draft, no broad acceptance or independent review claimed. Continues #2789 on #2805.

### DIFF
--- a/crates/groove/src/ivm/runtime/join.rs
+++ b/crates/groove/src/ivm/runtime/join.rs
@@ -799,15 +799,18 @@ fn keyed_join_deltas<'a>(
 ) -> Result<Vec<KeyedRecordDelta<'a>>, IvmRuntimeError> {
     if let Some(field_indices) = scalar_join_field_indices(descriptor, fields)? {
         let mut keyed = Vec::with_capacity(deltas.len());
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved #2806 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2807, opened 2026-09-11
perf: size native Zstd output from known frame content length

aeplay:
🤖 Native Zstd decoding reserved the caller’s maximum decoded length for every message. Jazz supplies 256MiB; the dependency’s bulk upper-bound helper is disabled without its experimental feature. Even a small ordinary frame therefore reserved a maximum-sized output buffer.

For a single frame with a known content size, use that size capped by the existing maximum. For example, a 1KiB decoded message gets a 1KiB output buffer. Unknown-size frames and concatenated frames retain the previous bounded bulk-decoder path, because the first frame’s size is not the size of an entire concatenated stream. Decoding still rejects an output beyond the limit. No wire format or accepted-message semantics change is intended.

Tests cover empty and small known-size frames, output limits, unknown-size streaming frames, concatenated frames and invalid input. Restoring the old maximum-sized allocation makes the capacity test fail; exact restoration passes. Native/LZ4 tests: 4 passed. Pure-Rust Zstd test: 1 passed. All-feature/all-target Clippy and formatting pass.

The readiness-scoped allocation profile identified this reservation pattern among the largest byte-volume stacks. Cumulative allocated bytes are not touched bytes or peak resident memory; the first clean optimized cold-load measurement is 27.394s versus 27.215s on the parent, with all 27,518 rows. That is within run-to-run noise; no elapsed-time speedup is claimed. Retain as a small resource-allocation correction, not as progress toward the cold-load time target. Draft on #2806; continues #2789. No independent review or broad acceptance claimed.

### DIFF
--- a/crates/jazz-compression/src/lib.rs
+++ b/crates/jazz-compression/src/lib.rs
@@ -49,7 +49,21 @@ pub fn compress_zstd(_payload: &[u8]) -> Result<Vec<u8>, String> {
 
 #[cfg(feature = "zstd")]
 pub fn decompress_zstd(payload: &[u8], max_decoded_len: usize) -> Result<Vec<u8>, String> {
-    zstd::bulk::decompress(payload, max_decoded_len)
+    // The stable bulk API otherwise reserves the entire caller-supplied limit.
+    // A frame content size applies to one frame, not a concatenated stream.
+    // Keep the existing bounded decoder path when that size is unavailable.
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Anselm approved #2807 in the coordinator conversation. Recorded for integration; merge remains subject to completion of the checkpoint review sequence and final merge preflight.

PR #2808, opened 2026-09-11
perf: decode only selected collection and sort fields

aeplay:
🤖 Collection rendering and sorting decoded whole input records into owned Value vectors, then cloned only the selected fields. This included sorting paths with no sort fields, where the full decode produced no useful output.

Read the selected fields directly from a borrowed record. A query sorting by one timestamp no longer reconstructs the row’s unrelated author, nested values or other columns. Collection roots, children, nested slots and tuple projections likewise decode only their requested fields. Existing nullable unwrapping, multiplicity, grouping, ordering and output encoding remain in place. No wire/storage format changes.

Groove library tests pass: 743 passed, 2 ignored. Scoped Clippy and formatting pass. All three scaling canaries and the two-seed maintained/one-shot oracle also pass. The clean optimized permissioned fixture settles in 25.693s versus 27.394s on the parent (6.2% less), with all 27,518 rows. Collection/results drops from 3.032s to 1.661s and IVM updates from 3.190s to 2.819s; those exclusive phase reductions explain the overall gain. The native todo checkpoint is 364.47ms for the measured 1,350-update roundtrip, compared with 371–387ms repeats at #2804; this comparison includes intervening changes and is not an isolated collection result. Details and limits: `dev/benchmarks/borrowed-collection-fields.md`. Draft on #2807; continues #2789. No independent review or full canonical acceptance claimed.

### DIFF
--- a/crates/groove/src/ivm/runtime/windows.rs
+++ b/crates/groove/src/ivm/runtime/windows.rs
@@ -238,13 +238,13 @@ pub(super) fn collect_by_output_value(output_type: &ValueType, value: Value) -> 
 }
 
 fn collect_by_projected_value(
-    values: &[Value],
+    record: &BorrowedRecord<'_>,
     field: &CollectByProjection,
 ) -> Result<Value, IvmRuntimeError> {
-    let value = values
-        .get(field.field_idx)
-        .cloned()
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2809, opened 2026-09-11
perf: encode record values without redundant staging

aeplay:
🤖 Ordinary table writes cloned their entire value vector into descriptor order even though descriptors now preserve logical declaration order. Record encoding then staged each encoded variable field in a separate vector before appending it to the output.

Use the supplied values directly when declaration and descriptor order match. Variant descriptors that select or reorder fields retain the existing mapping fallback. Assemble fields into the output through the existing encoded-field builder instead of staging a vector of encoded fields. Field type validation remains in the per-field encoder; invalid input still returns an error before any encoded record is returned.

For example, a title/author/nested-record write no longer deep-clones those values merely to put them back in the same order. The physical fixed-field/offset/variable-field layout remains the same. Individual variable-field encoders still use temporary buffers; this does not claim allocation-free encoding.

Groove library tests pass: 743 passed, 2 ignored, including byte fixtures and layout cases. Scoped Clippy and formatting pass. The full Jazz library also passes: 2,005 passed, 2 ignored. The first clean optimized permissioned run settles in 25.560s versus 25.693s on the parent; this is within noise. Storage apply falls from 2.993s to 2.820s. Retain as a smaller shared encoding implementation, not a demonstrated material end-to-end speedup. Allocation attribution continues to guide the next larger change. Draft on #2808; continues #2789. No independent review or full canonical acceptance claimed.

### DIFF
--- a/crates/groove/src/db/encoding.rs
+++ b/crates/groove/src/db/encoding.rs
@@ -147,8 +147,19 @@ pub(super) fn encode_record(
         }
         .into());
     }
-    // Callers provide values in SQL declaration order. RecordDescriptor stores
-    // fixed-width fields first, so we reorder here before positional encoding.
+    // Descriptors retain logical declaration order even though fixed fields
+    // come first in the physical byte layout. Ordinary table writes therefore
+    // already have the encoder's order and need no cloned staging vector.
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2810, opened 2026-09-11
bench: attribute allocation volume and counts to repository callers

aeplay:
🤖 Full allocation stacks split one caller’s work across many async and generic call paths. Add a second ranking that attributes each captured stack once to its nearest repository allocation caller, while retaining the existing full-stack detail.

Symbol resolution is cached by address and happens after allocation tracking stops. Count and allocated-volume rankings remain separate. The report covers the instrumented Rust allocator, not allocations performed internally by native C/C++ dependencies; truncated stacks without a repository frame are not attributed. These are sampling estimates, not elapsed-time percentages.

Optimized allocation-feature build, scoped Clippy and formatting pass. The full fixture passes all 27,518 rows. At the #2809 runtime tip it records 194,410,097 Rust allocations and 59,998,789,814 cumulative requested bytes through readiness. Aggregated count leaders include scalar/value encoding, storage-operation copies, large-value scalar field ordering, and schema/descriptor construction. OwnedRecord cloning is about 1.64M estimated allocations, so it is not the first ownership target. Byte-volume reduction from the earlier profile includes the Zstd reservation correction and is not an elapsed-time improvement claim. No product/storage/wire changes. Draft on #2809; continues #2789.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -233,6 +233,69 @@ mod alloc_metrics {
             ALLOCS.load(Ordering::Relaxed),
             BYTES.load(Ordering::Relaxed)
         );
+        // A caller can appear under many monomorphized/async stacks. Attribute
+        // each sample once to its nearest repository allocation site as well.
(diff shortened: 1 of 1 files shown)


aeplay:

PR #2811, opened 2026-09-11
perf: use direct primitive scalar encoding and owned decoding

aeplay:
🤖 Ordinary inline strings and bytes still passed through a field map, owned values, an enum payload record and another encoding wrapper. Owned primitive decoding also reconstructed and re-encoded its one-field record before returning the original payload. The borrowed-read path already avoided that work.

The primitive arm has one trailing raw field, so its record payload is exactly the logical bytes. Feed borrowed bytes to the ordinary variant-record encoder for tag 2, and let owned decoding copy the logically validated payload directly. For example, the primitive string `hello` remains the ordinary tag-2 record followed by its UTF-8 bytes. This is an implementation specialization of the existing format, not a new private scalar codec.

String/JSON logical validation remains. Chunked descriptors retain their existing path. The spec now explicitly describes the direct primitive implementations. A new internal byte test compares Bytes, String and JSON cases against the generic enum/record codecs. Changing the primitive tag makes it fail; exact restoration passes.

Groove library: 744 passed, 2 ignored. Jazz library: 2,005 passed, 2 ignored. Scoped Clippy and formatting pass. All three scaling canaries and the two-seed maintained-query oracle pass. Optimized native cold settling is 23.881s versus 25.560s (~6.6% faster), with all 27,518 rows correct. The 1,500-todo/1,350-update batch roundtrip is 317.238ms versus the preceding 364.470ms checkpoint (~13% faster; this comparison also includes #2809). Post-update fresh ingest is 52.444ms versus 64.094ms. These are native measurements, not browser acceptance. The committed benchmark report records scope and evidence. Draft on #2810; continues #2789. No independent review or full canonical acceptance claimed.

### DIFF
--- a/crates/groove/SPEC/9_large_values.md
+++ b/crates/groove/SPEC/9_large_values.md
@@ -96,6 +96,11 @@ The primitive arm contains only slot 1, a trailing raw field: its payload is
 exactly the logical bytes, without record headers, offsets, or reserved slots.
 Borrowed inline access validates the enum framing and logical kind directly;
 this preserves decode/recreate acceptance without constructing owned values.
+Owned primitive decoding copies that same raw payload after logical validation;
+it does not construct field maps or re-encode a record. Primitive encoding may
+pass the raw payload directly to the ordinary variant-record encoder for tag 2.
+These are specialized implementations of the same enum/record layout, not a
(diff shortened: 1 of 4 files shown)


aeplay:

PR #2813, opened 2026-09-11
perf: borrow schemas for row identity checks and serialization

aeplay:
🤖 The allocation profile traced repeated full TableSchema copies to row identity checks. Physical-table resolution and canonical maintained-witness lookup cloned the schema solely to check that the table existed, then discarded it. Version serialization also cloned a schema that it only read.

Use the existing borrowed lookup at these three sites. For example, publishing many rows from one table still checks each row's authored-schema membership and physical identity, but those checks no longer copy all column definitions, policies and references. Errors, catalogue resolution and encoded bytes are unchanged; there is no new cache or ownership abstraction.

Jazz library: 2,005 passed, 2 ignored with the established 4MiB test-thread stack. An initial invocation omitted that setting and aborted on stack overflow; the corrected invocation passes. Scoped Clippy and formatting pass. All three scaling canaries and the two-seed maintained-query oracle at depths 10 and 1,000 pass. Clean optimized cold settling is 23.541s versus 23.881s (~1.4%), with all 27,518 rows correct; this small single-run difference is not a substantial performance claim. Retained because four substitutions remove unnecessary copies using existing borrowed APIs. Draft on #2812; no independent review or full canonical acceptance claimed.

### DIFF
--- a/crates/jazz/src/node/physical/projections.rs
+++ b/crates/jazz/src/node/physical/projections.rs
@@ -7,7 +7,7 @@ where
         schema_version: SchemaVersionId,
         logical_table: &str,
     ) -> Result<PhysicalTableId, Error> {
-        self.table_in_schema(logical_table, schema_version)?;
+        self.table_in_schema_ref(logical_table, schema_version)?;
         self.catalogue
             .physical_mappings
             .get(&schema_version)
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2814, opened 2026-09-11
perf: prepare view descriptors once per validation batch

aeplay:
🤖 Incoming view validation rebuilt a table's complete wire descriptor for every row. Prepare each expected descriptor once per validation batch, keyed by authored schema version and table name. Reuse that preparation across all bundles in the receiver frame.

For example, 1,000 rows from one table under one authored schema construct one expected descriptor rather than 1,000. A row from a different schema version or table gets its own expected descriptor. Every incoming row is still compared against that descriptor; unknown schemas/tables, incomplete layouts, invalid provenance and invalid branch keys still reject before receiver mutation. The map lives only for the shared-borrow validation pass, so no catalogue invalidation or retained cache is introduced. Direct ingress retains its own full validation pass.

Jazz library: 2,005 passed, 2 ignored. Scoped Clippy and formatting pass. All three scaling canaries and the two-seed maintained-query oracle pass. Clean optimized cold settling is 22.966s versus 23.541s (~2.4%), with all 27,518 rows correct. This is a modest single-run gain, retained as a small batch-local preparation change; no persistent cache or invalidation machinery. Draft on #2813; no format change, independent review or full canonical acceptance claimed.

### DIFF
--- a/crates/jazz/src/node/ingest/validation.rs
+++ b/crates/jazz/src/node/ingest/validation.rs
@@ -921,8 +921,17 @@ where
         &self,
         versions: &[VersionRecord],
     ) -> Result<(), Error> {
+        self.validate_view_payload_versions_prepared(versions, &mut BTreeMap::new())
+    }
+
+    pub(super) fn validate_view_payload_versions_prepared<'a>(
+        &self,
+        versions: &'a [VersionRecord],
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2816, opened 2026-09-11
bench: scope external CPU sampling to cold-load readiness

aeplay:
🤖 Whole-process CPU profiles included seed checks and diagnostic queries performed after subscriptions were ready. Add an explicit bench-perf-control feature that uses perf's acknowledged enable/disable commands to collect only connection setup, subscription setup and settling to all expected rows.

Start external perf disabled with control/ack FIFOs. The benchmark enables sampling before setup and disables it at readiness, before post-readiness diagnostics. JSON labels this scope and stderr identifies a profiling-only run; ordinary benchmark builds reject the control environment through the existing contamination guard. No product runtime code changes.

The optimized profiling target compiles; benchmark-guard tests (2), scoped Clippy and formatting pass. Completed scoped captures at 199Hz and 99Hz both materialize all 27,518 rows with zero lost samples. Larger DWARF snapshots improve async caller recovery; some outer frames and kernel symbols remain unresolved. The first control trial exposed a NUL-framed acknowledgement and failed after disabling sampling; the corrected implementation completes. A committed report records the exact procedure and limits. Draft on #2815. Profile wall-clock values are not clean timing receipts.

### DIFF
--- a/crates/benchmark-guard/src/lib.rs
+++ b/crates/benchmark-guard/src/lib.rs
@@ -13,6 +13,14 @@ pub struct ContaminatingInstrumentation {
 /// wall-clock or storage-counter benchmark and receipt measurements.
 pub const CONTAMINATING_INSTRUMENTATION: &[ContaminatingInstrumentation] = &[
     ContaminatingInstrumentation {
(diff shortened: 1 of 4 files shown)


aeplay:

PR #2818, opened 2026-09-11
perf: construct record errors only on failure

aeplay:
🤖 CPU profiling showed record-error destructor routines on successful reads. Inspecting the optimized assembly confirmed that field_span constructed and dropped unused FieldIndexOutOfBounds and LengthOverflow errors: the success path called error drop glue twice. Even an enum variant without heap fields shares the error type's nontrivial destructor.

Construct these record/value access errors lazily with ok_or_else. A valid field read no longer needs to create an error merely to discard it. An invalid field index, offset or length still returns the same error at the same check; no bounds/type validation or format is removed. The module's scoped Clippy exception explains why its usual eager-error suggestion is counterproductive here. Formatting accounts for much of the diff around the changed closures.

Groove library: 744 passed, 2 ignored. All three scaling canaries and the two-seed maintained-query oracle pass. Scoped Clippy and formatting pass. Clean native cold settling is 21.757s versus 22.951s (~5.2%), with all 27,518 rows correct. Exclusive query-evaluation time is 2.409s versus 2.915s (~17%); this is part of the total, not additional time saved. The optimized field_span symbol shrank from 382 to 297 bytes and contains no calls after the change, including no error-destructor calls. The native todo batch roundtrip is 319.211ms, within the preceding 319–328ms checkpoint range; no meaningful todo roundtrip improvement is claimed. A committed report contains the phase and assembly evidence. Draft on #2817; no format change, independent review or full acceptance claim.

### DIFF
--- a/crates/groove/src/records/mod.rs
+++ b/crates/groove/src/records/mod.rs
@@ -1,3 +1,10 @@
+// Even unit variants have Error's nontrivial drop glue. Optimized field_span
+// emitted destructor calls on successful reads for eager ok_or errors.
+#![allow(
+    clippy::unnecessary_lazy_evaluations,
+    reason = "construct record errors only on failure; eager errors emit drop calls on hot successful reads"
+)]
+
 //! Compact binary record descriptors, layout, and encoded row access.
 //!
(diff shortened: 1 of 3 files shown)


aeplay:

PR #2820, opened 2026-09-11
bench: attribute allocation totals to cold-load phases

aeplay:
🤖 Extend cold-load phase timing with exact Rust allocation-request and cumulative-byte counts. The allocation-enabled benchmark assigns each request to the innermost active phase and role; work outside spans and uninstrumented threads have an explicit unscoped bucket. These are exclusive counts, so nested phases do not double-count their parent.

A constant-initialized thread-local phase marker avoids recursively borrowing the timing stack from inside the allocator. The benchmark asserts that phase totals equal its global allocator totals before reporting. This counts Rust global-allocator traffic, not C++/RocksDB internal allocations or resident memory. Counting is compiled out of normal timing builds.

Three attribution tests pass. Breaking restoration to the parent phase makes the new test fail (one Core allocation instead of two); exact restoration passes. Feature-specific Clippy and the optimized allocation-enabled benchmark build pass. The full run materialized all 27,518 expected rows from a clean committed tree. Its 143,366,980 allocation requests and 35,787,730,107 requested bytes exactly equal the sum of the exclusive phase buckets. Largest request counts: ingest 20.684M, query setup 17.277M, storage application 16.494M, supporting-row publication 13.180M; persistence is 4.121M. Query evaluation and output decoding request 5.930GB and 5.225GB respectively. Instrumented wall time is not a clean performance comparison. Draft instrumentation on #2818; no product code or format change.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -115,6 +115,8 @@ mod alloc_metrics {
             if ACTIVE.load(Ordering::Relaxed) {
                 let alloc_index = ALLOCS.fetch_add(1, Ordering::Relaxed) + 1;
                 let size = layout.size() as u64;
+                #[cfg(feature = "cold-settle-attribution")]
+                jazz_sim::phase_attribution::record_allocation(layout.size());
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2822, opened 2026-09-11
perf(jazz): project stored versions directly into wire records

aeplay:
🤖 Project stored versions directly into wire records instead of reconstructing every cell as an owned Value and encoding it again.

For a stored task with a text title, nullable priority and structured author, publication previously decoded those fields, allocated their temporary values, rebuilt the wire descriptor, and then re-encoded them. It now copies their existing field bytes into one output record. Parents and author structures are copied too. Packed storage HLC timestamps are still converted to wire milliseconds. Deletion-register rows still emit their deletion state and null user fields; branch coordinates and authored-column metadata are retained.

The existing bounded schema-shape descriptor cache now prepares both history and wire descriptors. Its key still includes table name, ordered column names/types and large-value semantic kinds, so JSON, text and table-bound enums cannot share an incompatible layout. No new long-lived cache or wire/storage format is introduced.

This addresses the sampled allocation path through `VersionRecord::from_stored` inside supporting-row publication (13.180M allocation requests across Core and relay on the parent instrumentation run). In test builds the former value-based encoder remains an independent byte oracle for every conversion. All 2,005 Jazz library tests pass (2 ignored), including wire/history byte roundtrips and timestamp boundary cases. Clippy and formatting pass. Clean cold load: 21.037s settle / 21.522s readiness versus 21.850s / 22.335s (3.7% total improvement). Supporting-row publication falls 2.022s → 1.414s (30%). The todo 1,350-row batch roundtrip measures 281.629ms versus the preceding 319.211ms receipt (~12%; also includes #2821). All expected rows pass. A deliberate packed-HLC-as-milliseconds mutation fails the wire/history byte oracle; exact restoration passes. These are individual native measurements, not browser acceptance or confidence intervals. Draft on #2821.

### DIFF
--- a/crates/jazz/src/node/codec.rs
+++ b/crates/jazz/src/node/codec.rs
@@ -1905,29 +1905,74 @@ impl VersionRecord {
         schema_version: SchemaVersionId,
         authored_columns: Option<BTreeSet<String>>,
     ) -> Result<Self, Error> {
-        // Wire records remain the replicated immutable projection. Content and
-        // register rows now live in different storage tables, so projection at
-        // this API boundary is assembled from typed row accessors.
-        let cells = table
-            .columns
-            .iter()
-            .map(|column| stored.cell(table, &column.name))
-            .collect::<Result<Vec<_>, _>>()?;
-        VersionRecord::encode(
-            table,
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2824, opened 2026-09-11
perf(jazz): move terminal operations through result handling

aeplay:
🤖 Move owned terminal operations through maintained result handling instead of cloning each operation during layout binding and cloning it again into the emitted transitions.

A batch already owns its operations. Layout binding now returns that same operation when its descriptor matches; when nullable layout conversion is needed it replaces the payload in place after the same compatibility checks. A stable partition processes root operations before nested operations, preserving order within each group. Root removals still suppress unreachable nested edits, and scalar replacements still retain independently maintained descendants. Separately owned retained query state still receives the copies it needs.

All 2,005 Jazz library tests pass (2 ignored), including layout-rebinding and nested collector cases. Check, Clippy and formatting pass. Clean optimized fixture materializes all 27,518 rows: 20.438s settle / 20.922s readiness versus 20.958s / 21.445s (2.5%). Exclusive result decoding plus layout binding falls by about 145ms; variation in other phases also contributes to the 520ms total difference, so the entire difference is not attributed to this change. No wire/storage/API change. Draft on #2823.

### DIFF
--- a/crates/jazz/src/node/maintained_subscription_view.rs
+++ b/crates/jazz/src/node/maintained_subscription_view.rs
@@ -504,22 +504,22 @@ impl MaintainedSubscriptionView {
         // whole active closure here would turn every incremental tick into a
         // snapshot-sized operation.
         let mut peer_source_fact_changes = BTreeMap::<ProgramFactEntry, (bool, bool)>::new();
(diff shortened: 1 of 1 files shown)


aeplay:

PR #2825, opened 2026-09-11
bench: distinguish phase attribution from absolute latency

aeplay:
🤖 Make the permissioned benchmark state whether phase attribution is enabled, and print a notice that its timings should be compared only with the same instrumentation. Add the measurement and profiling-method report.

On the same clean runtime tip, the 27,518-row fixture settles in 20.438s with attribution and 18.736s without it (dominant readiness 20.922s / 19.203s). The attribution build made roughly 1.28M clock reads on this HPET host and also enabled other counters; the 1.7s difference is not attributed solely to the clock. Absolute latency should use the feature-free build. Existing phase-enabled comparisons remain like-for-like diagnostic comparisons.

Scoped 99Hz CPU captures with 65,528-byte DWARF snapshots complete with all expected rows. Without phase attribution, allocation/free routines still account for about 23% and copying for 6.62%; much of the kernel clock/copy activity recedes. The report also records the inline-symbol expansion slowdown and the limits of inclusive/partially unwound stacks. Feature-specific Clippy, default Clippy and formatting pass. No product behavior or format change. Draft on #2824.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -495,6 +495,10 @@ fn main() {
     );
     #[cfg(feature = "bench-perf-control")]
     eprintln!("scoped CPU attribution run: wall-clock timings are not clean receipts");
+    #[cfg(feature = "cold-settle-attribution")]
+    eprintln!(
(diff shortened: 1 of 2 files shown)


aeplay:

PR #2827, opened 2026-09-11
perf: avoid intermediate decoded terminal event buffer

aeplay:
🤖 Decode terminal events directly into the existing net-change accumulator instead of first collecting an intermediate vector containing the entire decoded batch.

For example, a batch containing thousands of row changes previously held all decoded events in a vector before transferring them into the map that combines positive and negative weights. It now feeds that map directly. Retained subscription state is still updated only after every input has decoded successfully: a valid row followed by a decode error publishes neither row. Weight cancellation, event identity, application order and storage/wire formats are unchanged.

Validation: the full Jazz library suite passes (2,006 passed, 2 ignored), formatting and Clippy pass. The new late-error regression fails when errors are deliberately ignored, and passes after exact restoration. Optimized uninstrumented cold load materializes all 27,518 rows in 18.866s versus 18.736s before: effectively flat. This is a small buffer/ownership simplification, not a latency improvement. Draft on #2825; broad acceptance and independent review remain outstanding.


aeplay:

PR #2831, opened 2026-09-11
bench: remove misleading query attribution and observer graph walks

aeplay:
🤖 Correct operator attribution by removing the unreliable logical-name classifier and its recursive graph walks. Report hydration/incremental operator cardinalities by runtime role instead.

The prior counters attempted to recognize the dominant child table by logical names after the graph had been lowered to physical identities. Recent captures therefore put all activity into “other”; zero dominant buckets did not mean the dominant query performed no work. The classifier also allocated and traversed ancestors on every counter update, perturbing profiling. This change removes that work rather than introducing another heuristic. Actual subscription-based dominant-query readiness measurements remain unchanged.

Feature-enabled benchmark check and commit checks pass. Optimized instrumented acceptance passed with all 27,518 expected rows, 20.234s settle and 20.720s dominant readiness. Every aggregate map/join cardinality matches the preceding capture exactly after summing the former buckets. These instrumented times are not uninstrumented latency claims. Default runtime behavior is unchanged. The report records current native timings and the two discarded experiments (#2828 and #2830). Based directly on #2827; neither discarded trial is included.

### DIFF
--- a/crates/groove/src/cold_settle_attribution.rs
+++ b/crates/groove/src/cold_settle_attribution.rs
@@ -4,7 +4,7 @@
 
 use std::sync::atomic::{AtomicU64, Ordering};
 
-const BUCKETS: usize = 4;
+const BUCKETS: usize = 2;
 
 static MAP_CALLS: [AtomicU64; BUCKETS] = [const { AtomicU64::new(0) }; BUCKETS];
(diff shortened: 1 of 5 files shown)


aeplay:

aeplay:

PR #2833, opened 2026-09-11
bench: record CPU only within the selected todo phase

aeplay:
🤖 Add optional acknowledged perf control around a selected native todo benchmark phase. Set JAZZ_PERF_BACKEND and JAZZ_PERF_PHASE plus the existing control/ack FIFO variables. Sampling starts before the phase timer and stops after it; the phase JSON reports cpu_profiled. Explicit profiling mode prints that its timings are not clean receipts; ordinary runs retain the contamination guard.

This avoids recording several gigabytes of setup when the question is where roughly 90ms of worker ingest goes. Whole-process trials had sample/chunk loss, so their caller percentages remain preliminary. The report describes the method and its limits. No production/runtime behavior changes.

Optimized build and commit checks pass. Selected worker-ingest capture completed: 996 samples / 62.6MiB, with 961 samples inside the 20 phase windows, no empty in-phase stacks and no reported sample loss. Repeat fixtures inside one process using the existing row-count list; repeated child processes left later mappings unavailable in this environment. Based on #2831, which is based directly on retained runtime #2827. The rejected #2828/#2830 trials are not included.

### DIFF
--- a/crates/jazz/benches/local_batch_phases.rs
+++ b/crates/jazz/benches/local_batch_phases.rs
@@ -2,6 +2,8 @@
 //! Direct nodes intentionally expose phase boundaries that the public Db owner
 //! loop combines. This excludes JS, IndexedDB, scheduling and auth bootstrap.
 use std::{collections::BTreeMap, time::Instant};
+#[path = "support/perf_control.rs"]
+mod perf_control;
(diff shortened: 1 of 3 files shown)


aeplay:

aeplay:

PR #2838, opened 2026-09-11
perf(groove): prepare old-row descriptors once per batch variant

aeplay:
🤖 Prepare each old-row record descriptor once per table/variant in a write batch, instead of rebuilding it for every overwritten or deleted row. The cache lives only for the batch, whose schema is immutable.

For example, deleting 1,000 stored rows split between two variants now prepares two descriptors rather than 1,000. Each row still uses its own variant; an unknown stored variant still returns the same error. Same-batch write overlays, net delta grouping, record bytes and storage/wire formats are unchanged.

The focused test verifies both decoded deltas and the number of preparations, including mixed variants and an unknown variant. This is intentionally an internal test because public result equality cannot prove amortized preparation. Disabling cache reuse makes it fail with 1,000 preparations instead of 2; exact restoration passes. All 745 Groove library tests pass (2 ignored), as do commit checks and optimized benchmark builds.

Initial matched-build native measurements: permissioned cold settle 18.300s → 18.603s with all 27,518 expected rows; todo batch roundtrip 280.519ms → 259.870ms. Worker ingest itself is effectively unchanged (90.349ms → 89.655ms), so the total difference is not yet attributed to this change. Three alternating todo repeats are effectively flat: median 269.463ms before and 269.120ms after. No end-to-end speedup is claimed. This retains a small, batch-local simplification; all 2,006 Jazz library tests pass (2 ignored), along with all three scaling canaries and the two-seed maintained/one-shot comparison at churn depths 10 and 1,000. Draft on #2836; broader acceptance and independent review remain outstanding.

### DIFF
--- a/crates/groove/src/db/storage_helpers.rs
+++ b/crates/groove/src/db/storage_helpers.rs
@@ -923,6 +923,9 @@ where
     // previous path allocated a singleton TableDelta (and Vec) per old/new
     // record, then hashed every group and record again in a second pass.
     let mut by_table = HashMap::<(&str, u32, RecordDescriptor), HashMap<bytes::Bytes, i64>>::new();
+    // The schema is fixed for this batch. Old rows may use several variants,
+    // but each variant's descriptor needs preparation only once.
+    let mut old_descriptors = HashMap::<(&str, u32), RecordDescriptor>::new();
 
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2840, opened 2026-09-11
perf(jazz): name physical fields without constructing storage schemas

aeplay:
🤖 Build physical projection field names directly from the record prefix and physical column mapping. Previously, naming a history/current/rejected projection constructed complete storage-table schemas, indexes and type registries, then discarded everything except the names.

For a current-row projection, the output is still the declared system prefix, selected user columns in schema order under their physical identities, and `authored_columns`. Partial variants omit only absent user columns. A missing mapping for an included column remains an error; an excluded column requires no mapping. History and rejected-row projections follow their existing respective layouts. No storage or wire bytes, tags, descriptors or authorization semantics change.

The record macro exposes its declared prefix names as a static slice, avoiding a second manually maintained prefix list. One shared helper replaces the three schema-building implementations. An internal test compares all output names and their order against the real storage schemas for full, empty and partial variants. This is an internal contract because public query rows do not expose physical projection names. Temporarily omitting the trailing metadata field makes the test fail; exact restoration passes.

All 745 Groove and 2,007 Jazz library tests pass (2 ignored in each), along with the three scaling canaries, the two-seed maintained/one-shot comparison and commit checks. Optimized native cold settle is effectively flat: 18.603s → 18.665s, all 27,518 rows; readiness 19.072s → 19.063s. Todo roundtrip is 271.474ms, within the preceding repeat range. No end-to-end speedup is claimed; retain the smaller implementation because it removes unnecessary schema construction. Draft on #2838; broader acceptance and independent review remain outstanding.

### DIFF
--- a/crates/groove/src/records/macros.rs
+++ b/crates/groove/src/records/macros.rs
@@ -814,6 +814,8 @@ macro_rules! define_record {
             $crate::records::macros::paste::paste! {
                 $(pub const [<FIELD_ $field:upper _IDX>]: usize = $idx;)*
             }
+            /// Names of the declared fields before the dynamic user-cell tail.
+            pub const PREFIX_FIELD_NAMES: &'static [&'static str] = &[$(stringify!($field)),*];
             pub const USER_BASE: usize = 0 $(+ { let _ = stringify!($field); 1 })*;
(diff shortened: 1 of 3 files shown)


aeplay:

aeplay:

PR #2841, opened 2026-09-11
perf(groove): append variable encodings directly into record output

aeplay:
🤖 Encode variable-width values directly into the record output buffer. Previously, record encoding created a temporary field buffer; nullable values created another payload buffer; variable arrays materialized a vector of encoded element buffers; primitive scalar and enum framing could add another copy. The new append path shares the destination through these layers.

For example, a nullable string now appends its presence flag, primitive tag and string bytes directly. A variable array reserves its count/offset header, appends each element, and fills each completed offset in place. Offsets remain relative to that array, including an array nested in another array or appended after existing record bytes. Empty arrays retain their four-byte zero count. Fixed-width arrays and tuple ordering retain their existing encoding. Indirect large-value encoding keeps its existing path.

This is an encoder implementation change, not a storage/wire format change: record offsets, nullable flags, scalar/enum variant tags, primitive payloads and all decoding rules remain identical. It adds no serializer or encoding version. Existing input validation remains in place. If variable-field encoding fails after appending an earlier element, the caller's original output is restored, matching the old temporary-buffer behavior.

All 747 Groove library tests pass (2 ignored). New internal literal-byte fixtures cover nested arrays, nullable strings, empty arrays, byte payloads and a nonempty destination; these byte-level invariants are deliberately hidden by the database API. Separate sensitivity checks confirm failure when offsets become destination-relative or error rollback is removed. Both mutations are restored. Commit checks pass. The full 2,007-test Jazz library suite also passes (2 ignored), as do all three scaling canaries and the two-seed maintained/one-shot comparison. Clean native measurements are effectively flat: cold settle 18.665s → 18.546s with all 27,518 rows; todo batch roundtrip 271.474ms → 277.209ms. Refreshed cold allocation attribution across the recent slices is 115,600,178 allocations / 32,651,257,883 requested bytes, down from 128,343,865 / 34,351,374,780 at the earlier retained runtime. Temporary variable encoding buffers no longer lead allocation callers. This cross-slice allocation reduction is not an isolated latency improvement for this PR.

Draft on #2840. No material latency improvement is claimed; broader acceptance and independent review remain outstanding.

### DIFF
--- a/crates/groove/src/large_values.rs
+++ b/crates/groove/src/large_values.rs
@@ -2229,16 +2229,29 @@ pub fn encode_stored_scalar(kind: LargeValueKind, value: &StoredScalar) -> Resul
 /// Encode the primitive arm from borrowed logical bytes. Its sole raw payload
 /// field is the complete record payload, so only ordinary variant framing is
 /// needed; no Value, field map, or intermediate record is required.
+#[cfg(test)]
 pub(crate) fn encode_primitive_stored_scalar(
     kind: LargeValueKind,
     bytes: &[u8],
 ) -> Result<Vec<u8>, Error> {
+    let mut output = Vec::new();
+    encode_primitive_stored_scalar_into(kind, bytes, &mut output)?;
+    Ok(output)
+}
+
+pub(crate) fn encode_primitive_stored_scalar_into(
+    kind: LargeValueKind,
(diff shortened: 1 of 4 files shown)


aeplay:

aeplay:

PR #2843, opened 2026-09-11
bench: correlate cold-load CPU samples with phase timelines

aeplay:
🤖 Record an optional phase timeline so native CPU samples can be attributed even when async or stack-switching boundaries leave incomplete call stacks. Set `JAZZ_PHASE_TIMELINE` to an output prefix in a Linux `cold-settle-attribution` benchmark build; the benchmark writes `<prefix>.cold.json` or `<prefix>.warm.json` after readiness.

Each interval names its phase, runtime role and nesting depth and uses absolute `CLOCK_MONOTONIC` timestamps. The file identifies the owner thread; matching perf samples must use `--clockid mono` and include TID. Attribute an owner-thread sample to the deepest active interval, while keeping other threads and samples outside intervals separate. Inclusive intervals overlap; resulting innermost sample assignments do not.

Capture reserves bounded space for one million intervals before measurement, reports dropped intervals explicitly, and performs file I/O only after CPU capture and allocation counting stop. A truncated timeline must not be used for attribution. Without the opt-in variable, no interval buffer is allocated and the existing relative clock path remains. This is benchmark-only diagnostics, not a production storage or wire format.

All four phase tests pass, including allocation accounting. A sensitivity check fails when nesting depth is removed; exact restoration passes. The affected benchmark compiles, and commit checks pass. The optimized end-to-end capture materialized all 27,518 rows and recorded 641,651 intervals without truncation. Perf reported 10,164 samples, no empty stacks and no sample loss. The included analyzer preserves other-thread and unscoped buckets; synthetic nesting/thread/truncation checks pass. The report records phase shares and the next value-layout thesis. Draft on #2841; diagnostic overhead is explicit and these runs are not clean latency measurements.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2348,6 +2348,7 @@ dependencies = [
  "postcard",
  "pprof",
  "rusqlite",
+ "serde",
  "serde_json",
  "tempfile",
  "tracing",
--- a/crates/jazz-sim/Cargo.toml
+++ b/crates/jazz-sim/Cargo.toml
@@ -11,7 +11,7 @@ bench-alloc-sites = ["dep:backtrace"]
 profiling = ["dep:pprof"]
 transport-compression-lz4 = ["jazz/transport-compression-lz4"]
 transport-compression-zstd = ["jazz/transport-compression-zstd"]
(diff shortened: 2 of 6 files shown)


aeplay:

aeplay:

PR #2846, opened 2026-09-11
perf: shrink materialized value cells from 152 to 40 bytes

aeplay:
🤖 Shrink each materialized `Value` from 152 to 40 bytes on x86-64 by boxing the uncommon indirect large-value reference. Previously a boolean, integer or null paid the same inline width as the 152-byte large-value reference. A vector of 1,000 ordinary cells now needs 40 KB of cell storage instead of 152 KB, excluding its existing child allocations.

The change affects Rust ownership only. Native record encodings, existing serde variant tags and payloads, permissions and sync semantics remain unchanged. Large strings/bytes still use the same reference and materialization path; only that uncommon variant adds a box allocation. The splice loop unboxes once and keeps its working reference inline between edits. Rust callers constructing `Value::Large` now pass a boxed reference.

All Rust workspace target classes compile with CI's required features. Groove's library suite passes (749 passed, 2 ignored). A representation-budget test fails on the original 152-byte implementation, and a new compatibility test pins native scalar bytes plus the existing postcard/JSON carrier representations. Existing large-value tests retain their assertions; their constructors are updated mechanically.

Draft experiment on #2843. Clean native comparison against preserved parent binaries: cold settle 18.031s versus 18.353s, readiness 18.426s versus 18.754s, all 27,518 rows; the 1,500-row todo batch is 269.058ms versus 269.362ms. Treat this as latency-flat, not a substantial speedup. The 74% smaller inline cell representation is verified from compiler layout constants. Jazz library tests pass (2,007 passed, 2 ignored), along with all 24 large-value query tests, the unchanged large-JSON wire fixture, all three scaling canaries and the two-seed maintained/one-shot oracle at depths 10 and 1,000. Allocation capture over connect/subscribe/settle materialized all 27,518 rows: requested bytes fell from 32,651,257,883 to 28,848,569,899 (11.6%), while allocation count was effectively unchanged (115,600,178 versus 115,597,486). This explains why smaller cells did not materially reduce elapsed time. Retained as a small ownership/memory improvement, not a claimed latency win. No independent review or full release gate is claimed.

### DIFF
--- a/crates/groove/src/db/tests/batches.rs
+++ b/crates/groove/src/db/tests/batches.rs
@@ -489,7 +489,10 @@ async fn staged_large_value_is_consumed_atomically_with_its_referencing_row() {
     let mut batch = database.open_batch();
     batch.insert(
         "objects",
-        vec![Value::U64(1), Value::Large(staged.value_ref.clone())],
+        vec![
+            Value::U64(1),
+            Value::Large(Box::new(staged.value_ref.clone())),
+        ],
     );
     batch.accept_large_value(staged.id);
     database.commit_batch(batch).await.unwrap();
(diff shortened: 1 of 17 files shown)


aeplay:

aeplay:

PR #2849, opened 2026-09-11
perf: skip duplicate rendering of newly inserted roots

aeplay:
🤖 Avoid rendering every newly inserted root twice. Root collection first renders a row to publish its Insert, then later visits touched roots to publish Updates. Previously that second pass rendered the new row again before noticing that it had no previous row and discarding the work.

For example, opening a subscription with 100 new roots produced 100 Inserts but encoded 200 root records. The Update pass now skips roots without a previous record before looking up or rendering their current record, reducing this to 100 encodings. Existing-row Updates still compare before/after records; removals, ordering and inserts retain the same behavior. No storage or wire change.

The new internal work-bound test observes 200 encodings on the original implementation and 100 after the fix, while asserting the exact Insert count and materialized rows. Groove's library suite passes (750 passed, 2 ignored). The test-only encode counter is exposed within the crate; production carries no new instrumentation.

Draft on #2846. The shared-buffer experiment #2848 was discarded after a repeatable cold-load regression and is not in this branch. Clean native comparison: cold settle 17.693s versus parent 17.651s, readiness 18.089s versus 18.044s, all 27,518 rows. Todo is 278.362ms versus 272.087ms in this single pair; no improvement is claimed. Retained as removal of proven redundant work with simpler control flow, not a latency win. Jazz library tests pass (2,007 passed, 2 ignored); all three scaling canaries and the two-seed maintained/one-shot oracle at depths 10 and 1,000 pass. No independent review or full release gate is claimed.

### DIFF
--- a/crates/groove/src/ivm/runtime/windows.rs
+++ b/crates/groove/src/ivm/runtime/windows.rs
@@ -790,6 +790,11 @@ fn update_collect_by_root_terminal_state(
     }
 
     for (root_key, before_record) in before {
+        // A new root already emitted its Insert above. With no previous row,
+        // this pass cannot emit an Update, so do not render it a second time.
+        let Some(before_record) = before_record else {
+            continue;
+        };
         // `groups` can retain a root-collector maintenance group before it
         // has ever been presented to this terminal consumer. Without a prior
         // Insert, an Update would address no facade occurrence.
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2855, opened 2026-09-11
Attribute sampled allocation callers to node phases

aeplay:
🤖 This adds phase-specific allocation callers to the native cold-load profiler. We could already see that relay ingest allocated millions of times and that record clones were globally expensive, but could not reliably connect those observations.

Each sampled stack now captures the innermost node/phase labels at allocation time. Reporting groups callers within that phase, retaining an explicit unresolved bucket. For example, a clone in relay ingest is separate from the same clone in client persistence. Nested phases restore their parent label, and another thread does not inherit a node label.

This is diagnostic-only benchmark code, with no storage, wire or runtime semantics changes. Labels are static strings read without allocation; stack aggregation and symbol resolution happen after measurement stops. Counts are sampling estimates, while existing phase totals count all tracked allocations. Instrumented timings must not be treated as clean latency measurements.

Validation: all five phase-attribution tests passed; deliberately returning the unscoped label made the new nested/thread-isolation test fail. Optimized combined phase/allocation benchmark build passed. Full synthetic fixture capture passed: 27,518 rows, clean commit 900f71d786, 113,706,786 allocations and 28,521,853,022 requested bytes. The previous capture on the unchanged runtime counted 113,706,692 allocations and 28,521,779,518 bytes. Phase caller output identifies relay query setup as metadata-heavy and relay persistence as staged-write cloning plus physical-key construction. These are diagnostic totals, not clean timing claims.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -100,6 +100,7 @@ mod alloc_metrics {
         len: usize,
         count_sample: bool,
         byte_weight: u64,
+        phase: (&'static str, &'static str),
     }
 
     static ACTIVE: AtomicBool = AtomicBool::new(false);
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2857, opened 2026-09-11
Borrow pending batch rows during delta computation

aeplay:
🤖 Applying a batch temporarily encoded every pending row into a same-batch lookup map, then encoded it again for storage. The map now borrows each pending row's variant and payload. Final storage operations take ownership of table names and keys instead of cloning them. Exact-match checks also use a borrowed-key index instead of allocating lookup coordinates per write.

For example, insert row 7 as “first”, then update it to “second” in one batch. Delta computation still sees the earlier pending write and emits only the final row. It now reads that earlier payload directly rather than building another encoded copy. Delete and reinsert sequences retain their same-batch visibility; immutable exact-match conflicts still reject the batch.

No persistence or wire encoding changes, and no durability/lifecycle changes. The borrowed data stays owned by the pending batch until delta computation finishes. Storage receives the same ordered owned operations as before.

Validation: all 750 Groove tests passed (2 ignored). Deliberately dropping prior-write visibility makes the existing repeated-key test fail by exposing both versions rather than only the final one. Clean native pair: 27,518-row cold settle 17.682s versus 18.006s on the parent runtime; dominant-query readiness 18.075s versus 18.398s. The 1,350-row todo update was 258.207ms versus 252.450ms. Treat this as modest/flat timing, not a demonstrated broad speedup. Jazz library: 2,007 passed, 2 ignored. All three incremental-delivery/work-scaling canaries passed, plus the two-seed maintained-vs-one-shot oracle at churn depths 10 and 1,000. These are focused checks, not the full CI-equivalent gate or browser acceptance.

### DIFF
--- a/crates/groove/src/db/commit.rs
+++ b/crates/groove/src/db/commit.rs
@@ -122,11 +122,14 @@ impl Database {
             batch.notification_timing == NotificationTiming::AfterPersistence;
         // Later ordinary writes must not invalidate an ensure_exact result.
         let exact_keys = std::mem::take(&mut batch.exact_keys);
+        let borrowed_exact_keys = exact_keys
+            .iter()
+            .map(|((table, key), value)| ((table.as_str(), key.as_slice()), value))
+            .collect::<HashMap<_, _>>();
         let pending_writes = self.pending_writes_from_batch(batch)?;
         for write in &pending_writes {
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2859, opened 2026-09-11
Validate direct-store keys without constructing records

aeplay:
🤖 Direct-store key validation built a new prefix schema and encoded a temporary record on every read, write and range lookup, then discarded those bytes and separately encoded the actual key. It now calls the same existing value/type checker against the store's declared key fields and encodes only the ordered key.

A prefix such as `["scope"]` still selects the same records, and complete keys retain their exact stored encoding. Wrong scalar types, malformed tuple members, and excessive prefix arity still fail. No serializer, storage layout, wire format or durability change is involved.

Validation: 751 Groove tests passed (2 ignored), including existing ordered tuple-key/reopen coverage. The new public read/prefix test also checks that key validation creates zero throwaway records; restoring the old path fails with 3 encodings. Independently removing type validation makes its invalid-key assertion fail. Optimized benchmark build passed. Clean native pair: 27,518-row cold settle 17.332s versus 17.276s on the parent; readiness 17.727s versus 17.672s. The 1,350-row todo update was 264.164ms versus 255.937ms. Timing is flat; this is retained as a small code/work simplification, not a measured broad speedup. Jazz library: 2,007 passed, 2 ignored. All three incremental work/delivery canaries and the two-seed differential oracle (depths 10 and 1,000) passed. This is focused validation, not full CI-equivalent or browser acceptance.

### DIFF
--- a/crates/groove/src/db/storage_helpers.rs
+++ b/crates/groove/src/db/storage_helpers.rs
@@ -181,16 +181,9 @@ impl DirectRecordStore<'_> {
             }
             .into());
         }
-        let prefix_descriptor =
-            RecordDescriptor::new(self.key.fields().iter().take(values.len()).map(|field| {
-                (
-                    field.name.clone().expect("direct store fields are named"),
-                    field.value_type.clone(),
-                )
-            }));
-        let _ = prefix_descriptor.create(values)?;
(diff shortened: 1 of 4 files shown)


aeplay:

aeplay:

PR #2862, opened 2026-09-11
Separate lowering preparation and binding in cold-load profiles

aeplay:
🤖 The native cold-load profiler previously grouped query lowering, preparation and binding into one query-setup bucket. That could identify setup as expensive but could not tell us which stage an optimization would affect.

Add opt-in spans for Jazz program lowering, Groove shape preparation, bound subscription setup and graph compilation. The existing reporter gives inclusive timing for each stage and exclusive timing/allocation totals without double-counting nested work. Sampled callers retain the finer phase label. No query data or parameters are included in the spans.

These annotations are gated behind `cold-settle-attribution`; ordinary runtime, wire and storage behavior is unchanged. All five phase-accounting tests and the optimized combined phase/allocation benchmark build passed. Full synthetic fixture passed on clean ea66214f87: 27,518 rows, 109,953,630 allocations and 28,251,620,915 requested bytes. Graph compilation accounted for approximately 586ms across roles; relay binding was 218ms inclusive, of which 186ms was nested compilation. The larger query-setup spans include nested row execution and must not be interpreted as compiler time. Instrumented timing is diagnostic, not a clean latency claim.

### DIFF
--- a/crates/groove/src/ivm/runtime/compilation.rs
+++ b/crates/groove/src/ivm/runtime/compilation.rs
@@ -23,6 +23,10 @@ impl IvmRuntime {
         node
     }
 
+    #[cfg_attr(
+        feature = "cold-settle-attribution",
+        tracing::instrument(skip_all, name = "cold.phase.query_graph_compile")
+    )]
     pub(super) fn add_dedup_graph(
         &mut self,
         graph: &GraphBuilder,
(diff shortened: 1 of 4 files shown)


aeplay:

aeplay:

PR #2863, opened 2026-09-11
Make native benchmark allocator choice explicit

aeplay:
🤖 Native benchmark receipts did not identify their Rust allocator. They used System even when being compared with jazz-tools, which explicitly uses mimalloc. Add an opt-in `jazz-benchmark-guard/mimalloc` mode to the two active native perf targets, keeping System as the default so older receipts remain comparable. Each receipt now records `rust_allocator` and whether `LD_PRELOAD` is present.

The cold-load allocation wrappers now forward `alloc_zeroed` and `realloc` to the selected allocator while recording the request. Previously the trait defaults replaced those operations with allocation/zeroing or allocation/copy/free. Clean uninstrumented measurements were unaffected; instrumented allocation reports count allocation requests, including requested realloc sizes, rather than physical heap growth.

Linux cold-load receipts now report peak RSS in bytes (Linux getrusage reports KiB); previously this field returned zero outside macOS. No product allocator, query, storage or wire behavior changes.

Build example: `cargo build -p jazz -p jazz-sim --bench local_batch_phases --bench customer_cold_start --profile perf --features jazz/testing,jazz/transport-compression-zstd,jazz-benchmark-guard/mimalloc`. Omit the final feature for System. NAPI uses a separate mimalloc fork, and browser/WASM timing remains a separate measurement.

Validation at `37993fcda6`: benchmark-guard tests pass (2); optimized System, mimalloc and combined allocation-site/phase targets compile; the allocation-metrics variant also passes cargo check. Both clean full-fixture runs return all 27,518 rows.

| Native measurement | System | mimalloc |
| --- | ---: | ---: |
| Cold settle | 16.877s | 12.619s |
| Dominant child ready | 17.268s | 12.913s |
| Todo batch phases (1,500 rows) | 243.437ms | 228.264ms |
| Peak process RSS, including seeding | 3.73GiB | 3.96GiB |

This corrects the benchmark configuration when comparing with the CLI; it is not a newly delivered product speedup. Both clean runs have no allocator preload. The combined mimalloc phase/allocation capture also returns all 27,518 rows: 15.209s instrumented settle, 109,953,857 allocation requests and 28,251,885,251 requested bytes. Instrumented timing is not the clean latency baseline. Browser and end-to-end acceptance remain separate.

CI exposed an executable-composition case missed by the standalone checks: `legacy_benchmark_smoke` includes the todo scenario alongside another benchmark with its own global allocator. Fixed at `3f22a94680` by moving the todo allocator into a small executable entry file; the reusable scenario module no longer selects an allocator. The combined correctness target now passes all 6 tests, and the standalone mimalloc benchmark passes its compile check. Fresh CI passed after rerunning the known BandChat guest-invitation failure tracked in #2655; the allocator compile failure is fixed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2072,6 +2072,9 @@ dependencies = [
 [[package]]
 name = "jazz-benchmark-guard"
 version = "0.1.0"
+dependencies = [
+ "mimalloc",
+]
 
 [[package]]
 name = "jazz-cli"
--- a/crates/benchmark-guard/Cargo.toml
+++ b/crates/benchmark-guard/Cargo.toml
@@ -7,3 +7,9 @@ description = "Shared precondition for trustworthy Jazz benchmark and receipt me
 
 [lib]
 path = "src/lib.rs"
+
+[features]
+mimalloc = ["dep:mimalloc"]
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+mimalloc = { version = "0.1", default-features = false, optional = true }
--- a/crates/benchmark-guard/src/lib.rs
+++ b/crates/benchmark-guard/src/lib.rs
@@ -1,5 +1,17 @@
 //! Refuse benchmark and receipt runs whose instrumentation changes the numbers.
 
+/// Explicit allocator for native benchmark leaves. Opt into `mimalloc` to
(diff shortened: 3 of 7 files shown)


aeplay:

aeplay:

PR #2868, opened 2026-09-11
Store the three source witness weights inline

aeplay:
🤖 A supporting fact has exactly three independent witness origins: version, replacement and program fact. Store their signed weights inline instead of allocating a tree map per fact. Update the outer fact map through one entry lookup, eliminating the previous extra fact clone and repeated tree lookups.

Presence remains the union of positive weights, not their sum. For example, version=+1 and replacement=-1 still means present; removing the version makes it absent, but retains the negative replacement weight until it returns to zero. The entry is removed only when all three counters are zero. No storage, wire, authorization or query semantics change.

Validation: Jazz library 2,008 passed / 2 ignored; the added signed-origin test passes and fails when a temporary mutation aliases two origins to one slot. The mutation and temporary layout diagnostic are restored. Clippy and formatting pass. Clean benchmark comparison, integration canaries and independent review remain pending; this is a draft performance trial.

Clean comparison at `42ce7135e9`: all 27,518 rows, 12.115s cold settle versus 12.147s parent (flat); todo batch phases 223.266ms versus 234.440ms (about 4.8% lower in this pair). Peak process RSS including seeding is essentially flat. Retaining this as a small fixed-domain data-structure simplification, not claiming a material cold-load improvement. All three incremental-delivery canaries and the two-seed maintained/one-shot oracle at churn depths 10 and 1,000 passed.

### DIFF
--- a/crates/jazz/src/node/maintained_subscription_view.rs
+++ b/crates/jazz/src/node/maintained_subscription_view.rs
@@ -171,7 +171,7 @@ pub(crate) struct MaintainedSubscriptionView {
     /// publication diffs this set against its acknowledged predecessor so a
     /// +/− pair observed in one drain is never serialized as an ambiguous
     /// ordered operation.
-    source_fact_weights: BTreeMap<ProgramFactEntry, BTreeMap<SourceFactOrigin, i64>>,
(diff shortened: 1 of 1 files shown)


aeplay:

aeplay:

PR #2874, opened 2026-09-11
perf: share immutable version index key bytes

aeplay:
🤖 Maintained version indexes currently copy an encoded row into its identity key, copy it again into the secondary ordering key, and clone those buffers when inserting into additional indexes. This change shares immutable bytes between those private keys. Keys still compare their full byte contents; no hashes or pointer identity replace exact equality.

For example, a version appearing in both the primary identity index and its transaction index now shares one key buffer. Retracting that exact version still removes its transaction entry, while other versions remain independently addressable. The public owned record, storage bytes and wire protocol are unchanged.

The synthetic 27,518-row cold-load fixture passed in 11.806s and 11.870s versus 11.988s for the parent: effectively flat latency. Peak RSS, including seeding, fell from 4.262GB to 3.949–3.960GB (about 7%). The todo phase sum was 224.581ms versus 249.017ms in one pair; this is preliminary, not a robust 10% speedup claim. Measurements use the optimized perf profile with explicit mimalloc. This is a small ownership change retained primarily for lower memory use.

Validation: 2,008 Jazz library tests passed (2 ignored), all three incremental-delivery canaries and the two-seed differential oracle at churn depths 10/1000 passed; formatting and Clippy passed. Existing exact-version ordering, retraction and replacement tests cover behavior. No independent review or full canonical acceptance is claimed. Footprint diagnostics conservatively count referenced bytes per index entry and are not a deduplicated heap measurement.

### DIFF
--- a/crates/jazz/src/node/maintained_subscription_view.rs
+++ b/crates/jazz/src/node/maintained_subscription_view.rs
@@ -1,5 +1,6 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::mem;
+use std::sync::Arc;
 
 use groove::ivm::{
     MultisinkDeltas, RecordDeltas, TerminalEdit, TerminalOperation, TerminalPathSegment,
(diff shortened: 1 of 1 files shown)


aeplay:

aeplay:

PR #2875, opened 2026-09-11
perf: use the latest-write index for staged scans

aeplay:
🤖 Scanning a transaction's staged writes currently walks every operation, clones all matching set/delete operations, groups them into per-key lists, and replays those lists into cloned output values. This change reuses the existing ordered latest-write index and snapshots one final value or deletion per in-range key.

For example, `set(a, old), set(a, new), set(b, value), delete(b)` becomes the scan overlay `a → new, b → absent`. Earlier operations remain in the commit log; only the scan snapshot is simplified. A scan opened before another `set(a, later)` keeps its original staged snapshot. A newly opened scan sees `later` through the updated index.

The first indexed access builds the key index once; subsequent staged writes maintain it. Narrow scans then seek directly to their column-family/key interval. The base storage cursor stays lazy, reverse scans stay reversed, and the physical base limit still allows for staged deletions that hide durable rows. Empty/reversed intervals and prefixes without a finite successor remain supported. No storage or wire bytes change.

Validation: all 752 Groove tests and 2,008 Jazz library tests passed (2 ignored in each). The new storage API test covers last-write-wins, delete/restore, scan-open snapshot isolation, cross-family isolation, reverse ordering, range exclusion and an all-0xff prefix. Changing index construction to first-write-wins makes its exact-value assertion fail; mutation restored. All three incremental canaries, the two-seed differential oracle at churn depths 10/1000, Clippy and formatting passed.

Synthetic native cold load: 11.672s and 11.570s versus parent 11.753s, all 27,518 rows present. This is about 1%, not a substantial cold-load improvement. Todo phase sum was 234.236ms versus 250.390ms in one preliminary pair. The main rationale is removing full-list filtering and redundant operation histories while preserving the storage contract. No independent review or full canonical acceptance is claimed.

### DIFF
--- a/crates/groove/src/storage/mod.rs
+++ b/crates/groove/src/storage/mod.rs
@@ -1670,20 +1670,61 @@ impl StagedWriteState {
                     });
             }
 
-            let mut latest_by_cf_key: BTreeMap<String, BTreeMap<Vec<u8>, usize>> = BTreeMap::new();
-            for (index, operation) in self.operations.iter().enumerate() {
-                latest_by_cf_key
-                    .entry(operation.cf().to_owned())
-                    .or_default()
-                    .insert(operation.key().to_vec(), index);
-            }
-            self.latest_by_cf_key = Some(latest_by_cf_key);
+            self.ensure_key_index();
         }
 
         self.latest_by_cf_key
             .as_ref()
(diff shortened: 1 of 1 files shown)


aeplay:

aeplay:

PR #2878, opened 2026-09-11
perf: borrow or move preflight version bundles

aeplay:
🤖 Receiver bundle processing makes several deep copies even when it only needs borrowed views or can consume the owning map. This change removes those copies without changing bundle validation, ordering, encoding, or ingest behavior.

Bulk reset ingestion now builds `VersionBundleRef` values directly from the preflight map instead of cloning complete bundles and immediately borrowing the clones. After bulk selection, remaining bundles move out of that map; bundles needing deferred handling move into the deferred list. Stale-authority and direct-update paths likewise consume their finished preflight maps instead of cloning each payload.

For example, a complete transaction carrying many row versions no longer duplicates every encoded row and authored-column set merely to call the bulk ingest API, which already accepts borrowed bundles. A deferred transaction still follows the same per-bundle path and fate settlement; ownership is transferred instead of duplicated.

Validation: 2,008 Jazz library tests passed (2 ignored), all three incremental-delivery canaries, the two-seed 10/1000 differential oracle, Clippy and formatting passed. Native full fixture materialized all 27,518 rows. No independent review or full canonical acceptance is claimed.

Performance: cold 11.682/11.809s versus parent 11.496s, effectively flat/slightly slower in this pair; no latency improvement claim. Peak RSS including seeding fell from 3.960GB to 3.900–3.904GB. Todo single pair 231.935ms versus 241.361ms. Retained as removal of unnecessary ownership work, not a material cold-load optimization.

### DIFF
--- a/crates/jazz/src/node/views.rs
+++ b/crates/jazz/src/node/views.rs
@@ -1593,18 +1593,14 @@ where
             .iter()
             .map(|bundle| bundle.tx.tx_id)
             .collect::<BTreeSet<_>>();
-        let bulk_candidate_bundles = preflight
+        let bulk_candidate_refs = preflight
             .bundles
             .iter()
             .filter(|(tx_id, bundle)| {
                 bulk_candidate_tx_ids.contains(tx_id)
                     && bundle.scope == crate::protocol::VersionBundleScope::CompleteTransaction
             })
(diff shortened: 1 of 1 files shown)


aeplay:

aeplay:

PR #2882, opened 2026-09-12
perf: prune unused columns from query joins

aeplay:
🤖 Query execution can retain and index complete records even when a join only needs their matching fields. This draft implements consumer-local column pruning in Groove and defers Jazz witness construction until after exact-version selection.

## Division of responsibility

Jazz keeps the semantic decision: an immutable content/replacement witness must match the authorized visible row **and its exact transaction identity**. It now performs that match against the raw source before constructing the wide witness carrier. Routed publication preserves binding fields and bag multiplicity. Deletion witnesses already selected the authorized exact deletion before encoding and retain that contract.

Groove makes the general physical decision after resolving field identities:

- Semi/anti joins retain only matching columns on their right side, preserving weights rather than applying DISTINCT.
- A projection over an inner join retains only consumer columns plus matching columns in each join input. Output expressions are rebound by resolved position, preserving their declared output identities and types.
- Narrow consumers get their own deduplicated projections/arrangements. Other consumers can continue using the complete source or join.
- The transformation preserves comparison modes and does not cross filters, ordering/winner selection, recursive boundaries, or fallible upstream expressions. Joins with residual predicates or non-inner kinds are not rewritten by the projected-join transformation.

This introduces no second storage lookup, public API change, or wire/storage encoding change.

## Worked cases

For a task-existence query whose child records contain a large body, the existence arrangement needs the matching key and its multiplicity, not the body. If two children match, removing one keeps the task visible; removing the last child changes the result. Replacing the last child with another matching child in the same commit does not produce a spurious disappearance.

For a join that returns a task title and only uses the related row for matching, the related input keeps its key while dropping its unused payload. A separate query requesting the complete related row still sees the complete source. Join output positions are mapped through the existing physical input mapping, including unnamed input slots that do not appear in the join output. A new behavioral regression failed with the initial concatenated-position assumption and passes with the corrected mapping. Nullable, nested-record and enum projection expressions retain their source-field dependencies; their conversions remain after the join.

For two versions of the same row, visibility of the older version cannot authorize the newer one: row ID, transaction time and transaction node remain the match keys within the already selected source/branch. Routed publication retains its binding fields, preventing narrowing from mixing authorization contexts.

## Measurements

Optimized native synthetic cold load: 27,518 visible rows / 39 subscriptions, seeded Core and empty relay/client. Clean candidate repeats were 11.358s and 11.512s, versus a final base repeat of 11.612s (an earlier base repeat was a 13.275s outlier). An initial independent pair was 11.698s base / 11.390s candidate. The defensible conclusion is roughly 1–3% lower total latency, not a large improvement; peak RSS stayed around 3.9GB.

The 1,500-row todo batch phase sums were 230/235ms on base and 238/246ms on candidate: a slight regression in this sample. These native receipts are not browser IndexedDB reopen timings. The measured candidate precedes the final unnamed-slot mapping correction; no timing gain is attributed to that correctness correction.

The symbolized, phase-instrumented comparison shows settle-only exclusive IVM time falling from 2.307s to 1.769s (about 23%), with join output records falling from 848,427 to 691,929. Cumulative projection output bytes increase slightly, from 783.9MB to 799.0MB, because key projections are added. This metric excludes join output bytes. The driver executes Core, relay and client sequentially; both timelines have nine nonoverlapping node ticks. IVM saves 538ms, other exclusive phases add a net 148ms, and total node-tick time falls 390ms (13.488s to 13.098s). Thus overlap is not hiding the gain in this benchmark. Witness decoding/publication counts remain unchanged, and total latency improves only modestly. This is not evidence that join width accounts for most of the remaining cold-load cost.

## Validation and status

2,009 Jazz library tests and 755 Groove library tests passed (2 ignored in each); all three incremental-delivery canaries, the two-seed differential oracle at churn depths 10/1000, Clippy and formatting passed. New checks cover compiled input width/shared-consumer preservation, duplicate-key retractions and same-tick replacement, and witness encoding after exact-version selection. Temporarily clamping maintained multiplicities to presence/absence makes the new behavioral regression fail; the mutation was restored.

An existing microsecond wall-clock subscription-install test failed during concurrent compilation, then passed alone and in the quiet full Groove suite. Its query has no join/projection, and no threshold was changed.

Draft: no independent review or full canonical acceptance is claimed. The coordinator performed this work without subagents as requested. Related to #2746 and #2747; follows #2878.

### DIFF
--- a/crates/groove/SPEC/3_queries_operators.md
+++ b/crates/groove/SPEC/3_queries_operators.md
@@ -147,6 +147,22 @@ _Further invariants._ `INV-QUERY-5` — `MapProject` copies only configured fiel
 and preserves the input weight. `INV-QUERY-6` — `UnwrapNullable` preserves the
 original delta weight.
 
+### Consumer-local column pruning
+
+After resolving field references against the original descriptors, compilation
+may narrow a semi/anti join's right input to its matching fields. Projection
+preserves weights; it is not DISTINCT. Multiple rows sharing a key retain their
+combined multiplicity, including simultaneous insertion and retraction.
+
+For a projection over an inner join, each join input may retain only fields
+required by the output expressions and matching keys. Expressions are rebound
+to the narrowed descriptor by resolved position; logical field labels must not
+be reinterpreted after dropping columns. Comparison modes and exact value types
+remain unchanged. These are consumer-local graphs: independently used sources,
+projections and arrangements remain intact. Pruning does not cross upstream
+filters, winner/window selection, recursive boundaries or fallible expressions;
+it must not suppress errors or change which rows qualify or their ordering.
+
 ### 3.4 Joins
 
 Joins combine or suppress rows by key. groove executes the **inner equi-join**
--- a/crates/groove/src/db/tests/graphs/joins.rs
+++ b/crates/groove/src/db/tests/graphs/joins.rs
@@ -1431,3 +1431,109 @@ async fn anti_join_resubscribe_hydrates_from_storage_after_unretained_changes() 
         .unwrap();
     assert!(subscription.recv().unwrap().is_empty());
 }
+
+#[futures_test::test]
+async fn existence_join_narrowing_keeps_duplicate_key_counts_across_retractions() {
+    for anti in [false, true] {
+        let storage = MemoryStorage::new(&["albums", "artists"]).unwrap();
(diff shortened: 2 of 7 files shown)


aeplay:

aeplay:

PR #2883, opened 2026-09-12
bench: separate terminal and merge-head phase costs

aeplay:
🤖 Split the remaining cold-load bookkeeping into measurable phases before choosing another optimization. This adds opt-in benchmark spans only; normal builds, query results, storage, and wire behavior are unchanged.

The terminal pipeline now separates net-event accumulation from applying net events to retained state. Bulk ingest separately measures initial merge-head construction and the post-persistence rebuild. Existing nested phase accounting continues to subtract child spans; the new spans are absent without `cold-settle-attribution`.

## Measured bounds

Optimized native synthetic workload: seeded Core, empty relay/client, 27,518 visible rows and 39 subscriptions. The run completed all row assertions in 12.798s with phase instrumentation. This is diagnostic timing, not a clean before/after performance result.

- Initial merge-head construction across relay/client: 99ms.
- Post-persistence rebuild: 501ms inclusive, of which 340ms is exclusive rebuild work and 161ms is nested storage work. Do not add those numbers to one another.
- Terminal net accumulation across all node ticks: 584ms exclusive, excluding nested witness decoding.
- Applying net terminal events to retained state: 477ms.

These bound potential improvements; they do not establish that the work is removable. In particular, the first head pass consults incoming/staged ancestry and existing heads, whereas the second derives heads from all persisted row history and transaction fate. Eliminating either requires proving equivalence for partial history, out-of-order ancestry and previously resident versions.

## Validation

The optimized benchmark completed with 27,518 rows, no dropped timeline intervals, and exact exclusive-phase partition assertions. All three existing phase-attribution tests passed, including nested time accounting and pending-future scope restoration. The preceding implementation in #2882 passed the full Jazz/Groove library suites, canaries and bounded differential oracle; those results are not presented as a full canonical gate for this diagnostic follow-up.

Follows #2882; investigation remains tracked by #2746 and #2747. No new optimization or independent review is claimed.

## Experiment memory and current CI

The follow-up documentation records previously rejected experiments in `dev/benchmarks/rejected-experiments.md`, with an AGENTS preflight requiring historical checks before another trial. This prevents redoing universal record sharing or compact-event experiments without a changed premise. The duplicate #2884 trial was closed and is not part of this branch.

The original TypeScript CI job and one rerun both hit the existing #2655 BandChat guest-membership stall (owner memberships=2, guest=1). Both recurrences are recorded on that issue. Other correctness jobs passed; no harmless-timeout claim or assertion/timeout change is made.

PR #2884, opened 2026-09-12
perf: keep maintained event maps compact

aeplay:
🤖 Cold-load terminal accumulation used map keys of 968 bytes and event values of 1,080 bytes, even for ordinary immutable-version events. The size came from much larger result/proof variants sharing the same private enums. This keeps those larger variants behind owning pointers, reducing identity/value sizes to 96/160 bytes on the measured native target.

Only private `EventIdentity` and `NetEvent` representations change. Version/replacement events remain inline. Result/aggregate/proof payloads retain ownership behind `Box`; comparisons still compare values and preserve variant order. The decode-before-state-mutation boundary, weight coalescing, exact raw-version identity, and transition ordering remain unchanged. No public API, protocol type, storage encoding, or wire encoding changes.

For example, a version witness no longer reserves room for a full aggregate payload in every tree slot. A result event still owns its complete member/payload, and an equal result key still combines weights regardless of pointer address. A late decoding error still drops the complete temporary accumulator before retained state changes.

## Measurements

Same optimized native 27,518-row / 39-subscription workload. Allocation instrumentation measures cumulative requests, not live memory:

- Terminal accumulation requested bytes: 1.278GB → 0.452GB (about 65% less).
- Whole measured allocation scope: 27.533GB → 26.707GB (about 3% less).
- Allocation count: 108.597M → 108.805M; large variants add allocations.
- Peak RSS remains around 3.9–4.0GB. No peak-memory improvement is claimed.

Clean ABBA cold-settle runs were base13.745s, candidate11.802s, candidate11.161s, base11.474s. The first base was slow; excluding that outlier, the candidate average is essentially equal to the remaining base. **There is no clear overall cold-load speedup in this sample.** The 1,500-row native todo batch phase sums were base239.7/247.9ms and candidate230.1/231.4ms; this small sample suggests improvement but is not a confidence interval.

## Validation

All 2,010 Jazz library tests passed (2 ignored), including the new internal footprint budget. The pre-change measured sizes exceed that budget; the new layout fits it. Existing behavioral checks exercise coalescing, replacement, exact witnesses, aggregates and failure atomicity. All three incremental-delivery scaling canaries and the two-seed differential oracle at churn depths 10/1000 passed.

Draft experiment, with a measured allocation reduction rather than a claimed large latency win. No independent review is claimed; root-only work as requested. Follows #2883; related to #2746/#2747.

### DIFF
--- a/crates/jazz/src/node/maintained_subscription_view.rs
+++ b/crates/jazz/src/node/maintained_subscription_view.rs
@@ -363,27 +363,31 @@ enum MaintainedTerminalKind {
     AggregateAppRows(AppRowSchema),
 }
 
+// Keep large result/proof carriers out of every version-event map slot. Box
+// comparison remains value-based; indirection must not change identity/order.
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 enum EventIdentity {
-    Result(ResultMemberEntry),
+    Result(Box<ResultMemberEntry>),
     Version(ProgramSourceId, VersionIdentity),
     Replacement(ProgramSourceId, ReplacementKey, VersionIdentity),
-    ProgramFact(ProgramFactEntry),
+    ProgramFact(Box<ProgramFactEntry>),
     StructuredAppRow(RowUuid, Vec<u8>),
 }
 
 #[derive(Clone, Debug)]
 enum NetEvent {
-    Result(ResultMemberEntry, ResultMemberPayloadEntry),
+    Result(Box<(ResultMemberEntry, ResultMemberPayloadEntry)>),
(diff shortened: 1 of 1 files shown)


aeplay:
🤖 Closing this duplicate trial. Historical review found #2869 had already tested the same representation change; it and companion #2872 were discarded because their combined end-to-end gain was insufficient. This repeat confirms the smaller layouts and lower requested allocation bytes but shows no clear cold-load speedup. It does not justify reversing the prior decision. Commit `7a101218bf` and measurements remain preserved; this change is excluded from further work. Follow-up experiments will check the recorded negative results first.

PR #2890, opened 2026-09-12
bench: compare memory-only cold-load topologies

aeplay:
🤖 This benchmark compares the same native permissioned cold load with RocksDB everywhere, memory-only receivers, and memory on every node. It changes no production runtime or durability behavior.

Three rotated runs per configuration give median full-load times of **12.652 s**, **11.509 s**, and **11.213 s**, respectively. Removing persistent receiver storage saves about 9%; removing every persistent backend saves about 11%. Most of the cost remains in the existing in-memory engine and ingestion path; this does not yet test an alternative bulk cache representation.

The workload still materializes 27,518 rows. Core starts from identical physical seed state, verified key/value-for-key/value. An untimed all-memory wire capture matches the prior RocksDB capture's complete bundle multisets, including encoded transaction/version bytes and duplicate counts. All nine runs passed the benchmark's existing result and permission assertions.

The memory modes explicitly support cold load only. They do not simulate durable restart. MemoryStorage's ordered-map implementation is not an ideal RAM cache, so these numbers are not an exact disk-I/O attribution or a browser prediction.

Validation: optimized benchmark build; nine complete measured runs; exact capture comparison; formatting and commit gates. Details and reproduction: `dev/benchmarks/sql-proxy/MEMORY.md`.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -12,6 +12,9 @@ use jazz::db::{
     SubscriptionEvent, SubscriptionStream, Transport,
 };
 use jazz::groove::records::Value;
+use jazz::groove::storage::{
+    BoxedStorage, MemoryStorage, OrderedKvStorage, OwnedWriteOperation, ScanRequest, collect_scan,
+};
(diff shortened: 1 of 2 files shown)


aeplay:

aeplay:

PR #2891, opened 2026-09-12
bench: compare engine work with slim encoded-memory reference

aeplay:
🤖 Adds two diagnostic tools to the native permissioned cold-load benchmark: storage work counters and a slim reference that retains the actual encoded versions, indexes row references, evaluates the fixture's permission graph and materializes every output field.

The final clean reference takes ~271 ms for Core queries plus actual captured payload decode/install and queries at Edge and Client. The unchanged all-memory Jazz comparison was 11.213 s. This is deliberately a workload-specific reference, not a protocol-equivalent replacement: it replays captured delivery sets and omits maintained subscriptions, transport encoding/compression, general history and runtime coordination. Full output IDs and fields match the independently exported fixture at all three nodes. An identity without membership produces no protected outputs.

The engine work budget is the main finding:
- Client's 27,518 rows cause 220,148 point reads, 110,270 scan requests and 334,739 KV set entries.
- Projection operators visit 2.09 million records across the topology and produce 799 MB of new encoded buffers.
- A separate run with allocation counting only (no tracing or storage wrapper) requests 109 million allocations/reallocations and 27.7 GB cumulatively. The reference requests 2.53 million and 363 MB. These are requested allocation volumes, not peak memory or memcpy totals.

The code walk connects this to multiple physical row forms/indexes, merge-head staging followed by persisted-history rebuilding, and repeated intermediate query/publication records. It does not yet establish which production simplification safely eliminates that work.

Production runtime behavior is unchanged. Full method, limits, tables and reproduction commands are in `dev/benchmarks/sql-proxy/SLIM_MEMORY.md`. Validation includes optimized reference runs, exact result/field assertions, permission sensitivity, complete all-memory diagnostic runs, benchmark compile smoke and formatting. No production optimization is proposed in this PR.

### DIFF
--- a/crates/jazz-sim/benches/customer_cold_start.rs
+++ b/crates/jazz-sim/benches/customer_cold_start.rs
@@ -1,3 +1,7 @@
+#[path = "customer_cold_start/slim_memory.rs"]
+mod slim_memory;
+#[path = "customer_cold_start/work_budget.rs"]
+mod work_budget;
 use std::cell::{Cell, RefCell};
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::fs;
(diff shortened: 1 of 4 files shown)


aeplay:

aeplay:

Issue #2892, opened 2026-09-12
Preserve pending terminal edits across evaluation continuations


aeplay:
🤖 An isolated cold-ingestion experiment found lost live query output when another query causes an evaluation continuation.

A synthetic receiver starts empty, opens permissioned subscriptions, then installs accepted row versions. With the first seven fixture queries open, the `group` subscription receives all 38 rows. Adding the eighth query (`res_a`) makes its result remain empty, although supporting-row deltas are produced. The same stored rows are readable with a subsequent one-shot query.

`IncrementalEvaluation::poll` moves `terminal_deltas` and `root_ordering_windows` into its evaluator. Two early exits (an operator yielding and immediately ready hydration triggering recursive polling) drop that evaluator without moving the state back. Other continuation paths already preserve it.

A four-line candidate preserving both maps on those paths makes the full 39-query experiment pass for both captured synthetic datasets (46,740 and 27,518 unique transactions). The original executable fails the eight-query case; the candidate passes all 39. This is currently a local candidate on `bench/isolate-history-current-cost`, not a merged fix.

Follow-up: add a focused public-API regression with sensitivity verification, audit all continuation exits for owned state, and publish the fix separately from performance benchmarks. Storage bytes remain present in this reproduction; the observed failure is missing subscription output. General production reachability and affected release range still need confirmation.

aeplay:
🤖 Post-release triage (2026-09-12): Merged #2903 includes continuation-state preservation and its delivery regression (a8a7f49716 / a26a596297). Closing the pending-terminal-edit loss defect with the integrated checkpoint.

PR #2893, opened 2026-09-12
Preserve live result edits across evaluation continuations

aeplay:
🤖 An incremental evaluation can finish one query’s result edits and then pause to hydrate data for another query. Two continuation paths moved those pending edits into a temporary evaluator and dropped them before resuming. This preserves terminal edits and root ordering windows on both paths.

For example, an empty database with several live queries ingests accepted rows in one batch. The group query should deliver its 38 new rows. In the synthetic performance fixture, seven subscriptions did so, but opening an eighth permissioned query made the group result remain empty even though its supporting rows arrived. With this change, all 39 subscriptions return their exact expected results for both 46,740- and 27,518-transaction captures.

This changes evaluation continuation bookkeeping, not storage or wire formats. The row bytes were present in the failing case; subsequent one-shot reads could find them.

**Draft:** the before/after failure is captured by the companion experimental harness. Three repeated full runs pass with this change. A focused public-API regression and broader verification are still required before this is ready; no independent review is claimed. See #2892. Based on #2891’s performance stack.

### DIFF
--- a/crates/groove/src/ivm/runtime/runtime_tick.rs
+++ b/crates/groove/src/ivm/runtime/runtime_tick.rs
@@ -880,6 +880,9 @@ impl IncrementalEvaluation<'_> {
                 }
                 Poll::Pending => {
                     self.work_queue.requeue_yielded(node);
+                    self.terminal_deltas = std::mem::take(&mut evaluator.terminal_deltas);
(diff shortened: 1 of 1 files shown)


aeplay:

aeplay:

Issue #2895, opened 2026-09-12
Isolate and reduce temporary query-binding cleanup cost in snapshot reads


aeplay:
🤖 The isolated history/read experiment in #2894 found that repeated unretained read passes over the same installed synthetic data can get slower: the 46,740-row control measured a first pass around 1.43s and a second around 3.29s, while retained active queries made subsequent passes around 0.71s (after paying setup and maintenance costs).

A sampled profile and code walk identify `flush_pending_binding_retractions` on snapshot-read stacks. Disposing a temporary routed query binding can schedule a full incremental retraction that a later snapshot must flush. This is a concrete candidate, not yet an isolated explanation of the entire difference; stack truncation prevents complete wall-clock attribution.

Please isolate cleanup cost from query evaluation, then investigate operation-scoped temporary bindings or snapshot execution state that can be discarded without retracting through the whole graph. Preserve live bindings/subscriptions that share graph nodes, authorization isolation, cancellation, and exact supporting-row outputs. Do not simply skip the flush: it currently prevents snapshots from observing bindings already queued for removal.

The research synthesis and effort estimates are in `dev/benchmarks/sql-proxy/RESEARCH_SYNTHESIS.md` on #2894. Related cold-load tracking: #2746.

PR #2896, opened 2026-09-12
Experiment: separate snapshot reads from temporary live bindings

aeplay:
🤖 This research draft separates temporary query lifecycle, initial incremental-state construction, current-layer merging, and explicit history bookkeeping. All behavior experiments are opt-in; this is not ready for production review or merge.

### Why these controls exist

An application reads 39 tables, then reads them again without modifying data. The ordinary path binds each query to the incremental runtime and releases it after receiving a snapshot. Later queries process pending binding removals. Keeping the original lowered graph but substituting exact resolved binding values reduces the larger fixture's first pass from 1,289 to 668 ms and its second from 3,082 to 612 ms (three fresh processes per mode).

Operator traces explain the second-pass amplification: ordinary execution emits 5.19 million intermediate operator records, versus 1.26 million for concrete snapshots. A tracked row appears with negative weights while later queries on other tables flush old bindings. These counts are neither distinct database rows nor allocation bytes.

Crucially, a full Core→Edge→Client checkpoint stays around 11.5 seconds. The subscription-driven cold load does not inherit the one-shot speedup.

### Experiments and boundaries

- `JAZZ_DIRECT_SNAPSHOT_EXPERIMENT`: original permissions, historical winners and output fields; exact concrete parameter leaves replace temporary live binding registration. No settled authority-covered input override. Recursive graph substitution remains a research implementation.
- `JAZZ_FROZEN_SUBSCRIPTION_EXPERIMENT`: retains live incremental evaluation with concrete bindings and the existing routed-terminal normalization. Full fixture oracles pass. Setup + ingest + delivery improves approximately 10%, but subsequent ordinary reads lose prepared-query reuse and regress. This is not a recommendation to replace all prepared subscriptions.
- `JAZZ_HISTORY_LATE_SUBSCRIBE`: registers the same live queries after identical bulk installation. Total improves approximately 9%; most saved ingestion work moves into hydration. This does not implement safe production propagation deferral.
- `JAZZ_HISTORY_GLOBAL_READ`: public Global/deferred-local read options on the all-globally-accepted fixture. Concrete read time falls approximately another 20%. Pending local writes are outside this control's equivalence claim.
- `GROOVE_DIRECT_INITIAL_WINNERS`, behind `performance-experiments`: positive initial winner selection uses a group-to-borrowed-row index while retaining the same arrangement for subsequent changes. Read effects are small. Eighteen existing winner/tie/retraction/prepared-query tests pass with the control enabled.
- `JAZZ_AUTH_JOIN_ORDER_EXPERIMENT`: changes `(wide rows × bindings) join authorized proofs` into `wide rows join (authorized proofs join bindings)`. Every route equality remains required; compact proof matching happens before attaching payloads. Exact output/field checks, live subscriptions and denied-identity checks pass on both datasets. Repeated full-topology timing and targeted routing verification are still underway.

The separate richer history reference retains five versions per row: base, competing siblings, merge and a pending child finalized as accepted or rejected. Descendants arrive before ancestors, and duplicates are delivered. Exact independently generated winner transaction IDs and application/permission outputs are verified. It processes 327,180 deliveries in about 977 ms, versus 47,437 shallow deliveries in 145 ms. Deliberately treating rejected versions as accepted makes the winner oracle fail. This remains a single-row mergeable/default-branch reference, not Jazz admission, cross-row transactions, deletion registers or crash recovery.

### Evidence and status

Optimized native controls, synthetic fixtures, no wire/storage format changes. Detailed measurements, reproduction flags and limits are in `dev/benchmarks/sql-proxy/SNAPSHOT_EXECUTION.md` and `RICH_HISTORY.md`. No browser speedup, independent review, or release-readiness claim is made. Coordinator-owned work; no subagents, as requested.

Continues #2894; investigates #2895, #2746 and #2747. The next decision depends on end-to-end measurements rather than another shortened isolated phase.

### Shared witness trial

The cold topology emits 104125 version witnesses and 104128 replacement
witnesses; all 104125 paired payloads are exactly equal within the same source
and multisink delivery. Their independent role lifetimes still matter.

The optional `JAZZ_SHARED_WITNESS=both` trial normalizes internal payload tags
to share graph work, and reuses decoding across terminal batches. It keeps
separate signed role state. Its temporary cache matches full schemas, actual
record descriptors, and raw bytes; it is cleared after each delivery.

After moving cache preparation out of the row loop and consuming entries on
hits, three alternating fresh-process pairs gave median native cold settle
**11603 → 11064 ms (4.6%)**, with witness decoding **655 → 324 ms**. Baseline
runs: 11420/11603/11621 ms; trial: 11131/11064/10940 ms. All 39 subscriptions
and 27518 rows passed, as did the focused incremental permission-grant/revoke
regression with the flag enabled. Comparisons use one optimized native binary
with phase attribution; these are not browser measurements. Graph sharing
alone did not improve overall time. The decoding saving is firmly explained;
the additional wall-time saving needs replication.

This remains an opt-in experiment, with broader review and peak-memory/browser
checks outstanding before considering default enablement.


aeplay:
🤖 Closing at Anselm’s request and returning to parent `ce21ea7ad5f33e897248e67b674f0023e8eeacb0` (#2894). The experiment and measurements remain preserved at `e75a1f8629`. The shared-witness trial reduced native cold settle about 4.6% but added cache machinery; it is not being included. Next measurement target: one million permissioned resources loaded in under ten seconds.

PR #2900, opened 2026-09-12
bench: replay join arrangements against compact sorted batches

aeplay:
🤖 Research only: compare Groove's current join buckets with compact sorted batches using operations captured from the existing anonymized cold-sync workload. This does not change production query execution and is not part of the release checkpoint.

The current representation maintains nested maps keyed by join key and full encoded record. The experiment stores sorted join keys, offsets, record references and signed weights. Capture is opt-in and limited to the cold-load phase; its timings are discarded.

Replay checks exact weighted contents and probe outputs, preserves prior snapshots, and includes a signed-weight cancellation case. For example, weights +2 and -1 leave +1; a later -1 removes that record while an earlier snapshot remains unchanged. Negative weights remain representable.

The initial capture contains 3,649 cases, 2,923 update calls and 2,606,965 probe keys. It passes equivalence checks. Five quiet allocator-matched runs show median replacement work falling from 585 ms to 72 ms, accumulation from 85 ms to 18 ms, and probes changing from 25 ms to 27 ms. This is approximately 581 ms of replay update savings, not a measured end-to-end improvement. Most update batches have singleton buckets, suggesting full-record hashing and per-bucket allocation deserve particular attention.

Limits: this is a container replay, not an end-to-end performance result. It reconstructs input states, forces a retained snapshot during updates, and excludes key extraction and output encoding. It does not capture every semijoin/antijoin lookup. The initial sorted update rebuilds its batch, deliberately exposing that cost. No wire, storage, permission or query semantics change.

See `dev/benchmarks/sorted-arrangement-replay.md` for the changed premise relative to rejected #2864 and #2869/#2872.

The replay now has a dedicated opt-in benchmark entry point and passes all captured equivalence checks there. The earlier CI lint failure from registering it as an ignored test is addressed. Earlier CI also reported a BandChat guest-invitation failure; that has not been attributed or repaired in this research branch. This draft is not presented as green for release.

### DIFF
--- a/crates/groove/Cargo.toml
+++ b/crates/groove/Cargo.toml
@@ -46,3 +46,8 @@ harness = false
 [[bench]]
 name = "record_validation"
 harness = false
+
+[[bench]]
+name = "arrangement_replay"
+harness = false
+required-features = ["arrangement-replay"]
--- a/crates/groove/Cargo.toml
+++ b/crates/groove/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2024"
 [features]
 default = []
 test = []
+arrangement-replay = []
 cold-settle-attribution = ["dep:tracing"]
 
 [dependencies]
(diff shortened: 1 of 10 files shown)


aeplay:
🤖 Closing by Anselm's decision. The experiment and its findings remain available in this PR and its branch for posterity; the branch is being retained. This is not part of the current integration stack.

PR #2901, opened 2026-09-12
perf: trial compact immutable join buckets

aeplay:
🤖 Experimental follow-up to #2900, separate from the release checkpoint.

The capture/replay found that almost all join-key groups contain one record, yet the existing bucket representation hashes full encoded records. This trial stores each immutable bucket as sorted `(record, weight)` entries. The outer join-key index and copy-on-write update overlays remain in place, so a small update does not rebuild the entire arrangement.

Fresh construction sorts and consolidates equal records, dropping zero-weight entries while preserving negative weights. For example, +2 and -2 for one record cancel; a -1 weight for another record remains. Updating a cloned bucket leaves the earlier snapshot unchanged. These are representation changes only; query, authorization, storage and wire semantics remain unchanged.

Validation so far: full Groove suite passed (757 passed, 2 ignored). A focused signed-weight and snapshot check passed. The cancellation mutation fails and restored source passes. Three quiet alternating cold-load pairs measured 11.364s parent median versus 11.205s trial median (~1.4%); all 27,518 outputs matched. This small result does not justify adopting the trial yet. This is a trial, not release-ready.

The stack is based on the retained research checkpoint, with the disabled-by-default capture experiment below this implementation; it excludes the failed prepared-accessor experiment. #2900 preserves the diagnostic replay separately; its 7.5x faster update result does not establish an end-to-end speedup.

### DIFF
--- a/crates/groove/src/ivm/runtime/join.rs
+++ b/crates/groove/src/ivm/runtime/join.rs
@@ -30,16 +30,43 @@ pub(crate) mod replay;
 pub(super) type JoinKey = SmallVec<[u8; 64]>;
 #[derive(Clone, Debug, Default)]
 struct JoinBucket {
-    base: Rc<HashMap<Bytes, i64>>,
+    base: Rc<Vec<(Bytes, i64)>>,
     overlay: Rc<HashMap<Bytes, Option<i64>>>,
 }
 
 impl JoinBucket {
     #[cfg(test)]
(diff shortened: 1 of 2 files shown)


aeplay:
🤖 Closing by Anselm's decision. The experiment and its findings remain available in this PR and its branch for posterity; the branch is being retained. This is not part of the current integration stack.

PR #2903, opened 2026-09-12
Prepare stable performance checkpoint and repair enum catalogue adoption

aeplay:
🤖 Prepare the stable performance checkpoint for alpha.55. This integrates the performance stack through #2894 with main; the later accessor and arrangement experiments are excluded.

### What this checkpoint adds

It includes the continuation-delivery regression, browser batching receipts, restored WASM local-row export, and subscription completeness checks against the query's validated schema view. Acceptance additionally exposed the catalogue and reconnect problems below.

### Preserve enum meanings when adopting the authority catalogue

A node may persist rows before receiving the authoritative catalogue. Previously, adoption replaced its provisional identity manifest but left enum registries referring to provisional UUIDs. Reopening then rejected the manifest/registry pair as inconsistent.

For example, an offline row whose status is physical tag 1 must still mean the same status after connecting and reopening. The repair replaces semantic UUID references while retaining physical tags, row bytes and introduction coordinates. It includes UUID-qualified nested registry paths and rejects rebinding collisions. Public regressions cover initial adoption and a descendant-shaped local anchor. An internal regression covers payload-owned paths that public schema builders cannot currently express.

The live catalogue may also hold the application's draft schema before its migration is published. That draft is not a second authority genesis. Validation checks the complete authoritative snapshot graph without requiring the local draft to have an authority publication already. The existing integration test verifies that a draft query stays pending and resumes when the migration arrives.

This introduces no storage encoding change or recovery for already-incoherent caches.

### Reuse persisted payloads without claiming fresh query authority

After restart, an Edge retains its durable synchronization cursor but correctly loses live authority confirmation. Declaration generation previously required live confirmation before advertising the cursor, forcing a fresh download despite the persisted rows.

The cursor may now advertise unevicted payload possession before live confirmation. For example, an Edge reopening an unchanged one-row query can receive zero row bodies, but still receives Core's complete supporting-row manifest. Only applying that fresh response restores live authority. Missing bodies still require repair; eviction invalidates the cursor. Client-link authorization-progress checks remain intact. An exact-set declaration without a durable cursor still requires a live receipt.

The warm benchmark keeps its existing declaration requirement and row checks. Tests separate cursor advertisement from live settlement instead of treating them as equivalent.

### Validation and release status

Hosted CI and the complete local CI-equivalent gate now pass on amended revision 6c91eab6e058a35a0556f5c6f724dbf6bc28f252. The local gate completed with exit 0, including Node/browser checks and browser storage compatibility. User review of the prerequisite stack and final merge preflight remain outstanding.

The catalogue integration suite passes all 27 tests. Enum adoption and nested-path tests have mutation sensitivity. The reconnect roundtrip proves cached-body suppression with a fresh supporting set and no premature live authority; restoring the old declaration gate makes it fail. The Jazz library suite passes 2,013 tests with 2 existing ignores on the amended runtime. Full-size native acceptance passes all 27,518 expected rows in both cold and warm phases. Cold readiness was 12.407s; reopened-Edge readiness was 25.502s with all 39 known-state declarations. These are single acceptance runs, not a measured speedup; the warm path remains slow. The original benchmark assertions are unchanged. All hosted CI jobs passed on that revision, including Rust, differential, storage compatibility, TypeScript, React Native and lint.

Optimized browser receipts on fe8: reopening 1,500 persisted rows took 291–304ms. Explicitly batching 1,350 updates took 60ms staging plus 633–639ms commit/local completion. Ordinary sequential persistent updates took 43.423s; these are different operations and must not be presented as interchangeable timings. These performance receipts predate the final catalogue/reconnect amendments; final-tip browser correctness is covered by the completed local gate, but these timing numbers have not been remeasured on the final tip.


aeplay:

PR #2910, opened 2026-09-12
perf: select local deletion witnesses directly from current indexes

aeplay:
🤖 Local-tier supporting-row delivery currently constructs and executes a union/arg-max query graph for each row whose deletion/restore witness must be recovered from storage. This PR selects the same winner directly from the existing current-register indexes.

For example, when a cached row has a globally accepted deletion and a newer local restore, delivery still includes the restore transaction so the receiving node can reveal the row. A row with no deletion register still needs no witness. The difference is how we find that witness: one global point lookup and one lookup for the last ahead entry under the branch/row prefix, rather than compiling and running a query graph.

The ahead index orders entries by transaction time and stored node alias. The comparison with the global candidate preserves that ordering and the graph’s ascending encoded-record-byte tie break. These reads observe applied resident batches. Edge-tier selection, which must filter for its accepted frontier before choosing a winner, is unchanged. This adds no cache, invalidation state, storage format, or wire changes.

This PR contains the direct-lookup change on #2906; #2912 adds the independent publication optimization above it. Independent correctness/security review passed. Matched CodSpeed wall-time results confirm the reopen improvement.

Validation: focused subscription, deletion and restore suites passed; complete Rust library suite passed (2,013 passed, 2 ignored); clippy and formatting checks passed. Matched optimized native reopen measurements on x86_64 used the same two-package Cargo selection, perf profile and mimalloc, with separate preserved baseline/candidate binaries. Alternating seven-sample runs produced baseline medians of 171.2 and 169.1 ms versus candidate medians of 136.8 and 135.5 ms (about 20% less wall time). These are native 1,500-row measurements, not browser timings or a comparison with the ARM64 CodSpeed runner. Measurement revision: 1dc9db32ec; subsequent changes preserve malformed-record diagnostic context and fix inherited manifest formatting. CodSpeed comparison cd3586a → b545cfc reports reopen 876.769 → 669.564 ms minimum (23.6% lower latency), batch updates 1.307 → 1.126 s, and no classified sequential/cold-load improvement. The runtime change has since been restacked onto current main with the repaired foundation; final stacked CI is running.

Independent review verified stored alias/time ordering, encoded-byte ties, branch/schema isolation and applied resident visibility. Discarding ahead candidates made the unchanged restore test fail; the restored candidate passed Local/Edge, ordinary deletion/restore and frozen-branch restore checks. The final follow-up retains the prior malformed-record diagnostic context.


aeplay:
🤖 Takeover review: retain this candidate. The direct lookup preserves the stored time/node-alias ordering, encoded-byte tie break, applied resident visibility and the unchanged Edge frontier path. The patch introduces no cache or durable encoding change.

The completed CodSpeed comparison now supports the native result. Comparing foundation `cd3586a` (run `6aa5eb889dd1ddd9e2189802`) with `b545cfc` (run `6aa5ecb7915c84ce3c2026f4`):

- Reopen 1,500 tasks: minimum 876.769 ms → 669.564 ms, 23.6% lower latency. Medians 976.358 → 677.574 ms; the baseline had considerable spread, so the minimum comparison and prior alternating native runs are the stronger evidence.
- Batch update 1,350 tasks: approximately 1.307 s → 1.126 s, 13.9% lower latency; CodSpeed classifies this as an improvement.
- Sequential updates and permissioned cold load were not classified as significant improvements.

The affected wall-time cases had no detected environment difference. These are ARM64 CodSpeed results; the separate optimized x86 native reopen receipt remains 169–171 ms → 135–137 ms.

CI correctness jobs passed. The failing W1 benchmark is inherited from the foundation and is being investigated separately. I am reconciling the foundation with current main before landing this change. No new external testing request yet.

[Full comparison](https://app.codspeed.io/garden-co/jazz/runs/compare/6aa5eb889dd1ddd9e2189802..6aa5ecb7915c84ce3c2026f4)

PR #2921, opened 2026-09-13
perf(groove): build root-position maps only for consuming outputs

aeplay:
🤖 Trial for #2913: build generic root-position maps only when an affected plain output consumes them.

## Before / after

Every incremental TopBy evaluation previously encoded every selected row's root key into before/after BTreeMaps. Publication never reads those maps for structured CollectBy/Root outputs: their collectors already produce occurrence-aware positional edits. The tick planner now unions demand across affected output consumers and pre-registers only required ordering nodes. TopBy builds maps only for those nodes.

## Invariants and examples

- A structured-only ordered subscription keeps its own inserts, updates, removals and moves, without a second generic position map.
- A plain ordered output sharing that exact TopBy with a structured output still requests maps. Inserting a middle row retains index 1; updating a sort field still emits a move to index 0.
- Removing the plain consumer disables the extra work next tick; attaching it again preserves the ordered initial snapshot and restores map demand. Initial snapshots use hydrated ordering records, not these incremental maps.
- Existing root-ordering metadata, evaluation dependencies and pending/resumed tick ownership remain intact. Demand and partially built maps travel in the same existing continuation state.
- Finite windows, offsets, multiplicities and generic ordering algorithms are unchanged. No storage/wire layout or authoritative encoding changes.

Two additive TickMetrics counters distinguish selected records used to construct maps from records skipped for lack of consumers. No per-row timers or environment checks were added.

## Validation / prediction

Public Groove multisink suite: 8 passed, including the new structured-only → shared plain/structured → detach → reattach regression with output and mechanism assertions. Broader correctness CI and local native timing are pending. Working prediction: 4–7% lower sequential insert/update latency; no material permissioned cold-load gain is predicted. A wider change-proportional membership-delta trial follows separately, only after this checkpoint is measured.


aeplay:
🤖 #2921 local native receipt (source `f76631b3fe1ecb45ac769b81880be4cc3ee24454`) versus #2920 (`13e96f8bf982bfde77e9c7c46edf13dcef78985b`, tree-equivalent to merged `fdc66d04b7`).

Quiet ABBA, three samples per workload per block, six samples/revision. All original result assertions retained. Same RocksDB WALNoSync fixtures, mimalloc, native Rust 1.93.1 perf profile; setup outside timing, final reads inside, verification after. No local builds/tests during timing.

| Workload | Base A / B medians | Candidate A / B medians |
| --- | ---: | ---: |
| sequential insert 1350 | 2.532 / 2.558 s | 2.297 / 2.314 s |
| sequential update 1350 | 4.607 / 4.605 s | 4.153 / 4.137 s |
| batch update 1350 | 181.9 / 187.4 ms | 181.1 / 182.2 ms |
| reopen 1500 | 120.7 / 121.4 ms | 120.6 / 121.3 ms |
| permissioned cold first sync 27518 | 13.25 / 13.15 s | 13.12 / 13.05 s |

Insert latency is 9.3–9.5% lower, updates 9.9–10.2% lower in matched A/B blocks. This exceeds the conditional 4–7% estimate; hosted confirmation remains pending, and removed profile cost is not an exact wall-time prediction. Reopen is flat. Batch has baseline variation; cold differs by only ~1%, so no meaningful cold improvement claimed.

Whole-process max RSS (includes setup/teardown): todo base 206108–206564 KiB, candidate 204428–207984 KiB; cold base 3598356–3600320 KiB, candidate 3589428–3604376 KiB. No meaningful peak-memory claim.

Candidate preserved binary SHA256: todo `cf499b7fb3063e5fa704a036bafd33be13babb04785d7ab40c94dd0829934b97`; cold `b5db3ba36db01b2feb1009b9036e0c90fbe5b55c2f62c593ecc0b61bfefbcdfd`. Parent binaries and hashes are in the #2920 receipt. Packages built separately; compiler artifact paths verified, then executables copied out of shared target before timing. Raw logs/build JSON/script and binaries: `/tmp/jazz-root-position-demand-thJ9nT/`.

Correctness: public multisink suite 8/8; Groove library 757 passed, 2 existing ignores; all three incremental-delivery canaries passed. Correctness CI is in progress; Rust workspace, differential, storage compatibility and lint have passed. This is not a local CI-equivalent claim. Starting the separately measured wider unbounded membership trial now.

Tooling friction: full GQL profiles are available without authentication for this public repository, but the documentation assumed a token and the run list is ascending. The native timing binaries remain preserved so the second trial needs no parent rebuild.

PR #2922, opened 2026-09-13
perf(groove): derive unbounded membership from touched records

aeplay:
🤖 Second ordered-maintenance trial for #2913, building on merged #2921. This PR now targets main and contains only the wider trial.

## Before / after

After skipping unconsumed root-position maps, TopBy still enumerated complete before/after windows and rebuilt a BTreeMap difference for every single-row write. For an unbounded zero-offset window without a plain generic-ordering consumer, this trial calculates membership changes using only touched records: `max(after_weight, 0) - max(before_weight, 0)`.

Existing ordered state remains authoritative for later hydration and consumer attachment. This is not a second ranking index. Update work for this membership calculation is proportional to touched records and tree lookup costs, not retained window length. Other downstream collectors and publication work can still scale with result size; this is not an end-to-end constant-time write claim.

- Inserting one new row visits its changed record, rather than both full windows. Updating a payload or sort key retracts the old encoded record and adds the new one, in the same full-key order as the existing window diff.
- Repeated deltas for one encoded record compare first/final weights. +1 followed by -1 emits nothing. Internal weight -2 → -1 emits nothing, whereas -1 → +1 emits +1; forwarding raw input deltas would be incorrect.
- Removals precede additions within each group; full encoded records retain tie-breaking, signed weights remain in state, and COW snapshots remain isolated.
- If a plain ordered consumer shares the TopBy, the generic window/position path remains active. Detaching it enables the delta path; attaching it later preserves the full initial snapshot and restores generic position demand.
- Finite windows (including finite u64::MAX), nonzero offsets, and Replace/full-snapshot evaluation keep the previous algorithm. Pending tick state, publication ownership, wire and storage encodings do not change.

## Validation

Public shared-consumer ordering regression passes. Internal signed-bag differential oracle compares exact emitted deltas and retained weights against the unchanged full-window algorithm, including tie-breaking, cancellation and preserved COW snapshots; this level is intentional because negative intermediate bag weights are not directly constructible through public table writes. Full Groove library: 758 passed, 2 existing ignores. Public multisink suite: 9 passed. The 10/1000-row scale test checks three changed input records for update+delete and replays terminal edits against fresh hydration. All three incremental-delivery canaries pass. Full correctness CI [34735860671](https://github.com/garden-co/jazz/actions/runs/34735860671) passed on exact head `1e737976ce05afc4e72224973cd73e0ebee2ed00`.

The previous trial's new mechanism-counter assertions now require zero window visits and one delta visit in the structured-only case; public result/order assertions are retained. No benchmark workload, measurement boundary or result assertion changes.

The working prediction was an additional ~8–12% reduction in sequential insert/update latency. Quiet native ABBA against the exact #2921 binaries measured 13.2–14.0% lower insert latency (2.315/2.313s → 1.990/2.007s) and 14.2–15.8% lower update latency (4.229/4.146s → 3.562/3.558s). Batch and reopen are flat; cold varies across blocks with no consistent change. Six samples per revision/workload; fixtures and validation unchanged. Full provenance, all controls and whole-process RSS are in the [receipt comment](https://github.com/garden-co/jazz/pull/2922#issuecomment-5650925878). CodSpeed confirmation remains pending. No independent adversarial review claimed.


aeplay:
🤖 #2922 measured checkpoint, source `1e737976ce05afc4e72224973cd73e0ebee2ed00`, versus merged #2921 source `f76631b3fe1ecb45ac769b81880be4cc3ee24454`.

Quiet native ABBA, 3 samples per workload/block, 6/revision. Same allocator, build profile, fixture, final-read boundary and post-timing assertions as the preceding trial. No local builds/tests during measurement. Binary copies, compiler JSON, raw logs and compare.sh retained under `/tmp/jazz-touched-unbounded-QJK0nO/`.

| Workload | Base A / B medians | Candidate A / B medians |
| --- | ---: | ---: |
| sequential insert 1350 | 2.315 / 2.313 s | 1.990 / 2.007 s |
| sequential update 1350 | 4.229 / 4.146 s | 3.562 / 3.558 s |
| batch update 1350 | 181.3 / 181.9 ms | 181.8 / 181.0 ms |
| reopen 1500 | 121.2 / 120.4 ms | 121.2 / 121.2 ms |
| permissioned cold first sync 27518 | 13.68 / 13.10 s | 13.28 / 13.21 s |

Incremental insert latency is 13.2–14.0% lower; updates 14.2–15.8% lower in matched blocks. This exceeds the working 8–12% estimate, which covered the dominant diff but was not an exact prediction of window enumeration/allocation effects. The two ordered-maintenance steps together move the original ~2.54s insert / ~4.61s update baseline to ~2.00s / ~3.56s here (roughly 22–23% lower latency; comparisons span two runs).

Batch and reopen are flat. Cold baseline A was noisier/slower (samples 13.56–14.49s, versus base B 13.06–13.37s); candidate A 13.02–14.04s, B 13.15–13.47s. Direction changes across blocks, so no cold win or regression claimed. Whole-process RSS (setup/teardown included) is overlapping: todo base 206128–218244 KiB vs candidate 206052–219184; cold base 3559552–3579200 vs candidate 3597788–3598560 KiB, about 1% higher and not isolated to the timed phase.

Binary SHA256: todo `37fd1fe9f89f23a576fcb39dd5d45eed4f9580c15182e790f294104188c13c7e`; cold `65ce07e1a520bb7cfb5bea65f34af4350d80ba9b60e9e9c156adbe0c75e9051b`. Each package built separately with native Rust 1.93.1 perf profile and mimalloc, preserving the prior feature unions. RocksDB WALNoSync throughout.

Correctness: Groove library 758 passed (2 existing ignores), public multisink 9/9 including 10/1000-row delta-count and terminal-replay oracle, all three incremental-delivery canaries passed. CI 34735860671 on exact head: Rust workspace, differential, React Native and lint passed; remaining partitions still running at this receipt. No local CI-equivalent or independent adversarial review claimed. No benchmark assertion, protocol or storage encoding changed. PR is not approved or merged; parent #2921 is already in main.

Tooling friction: preserving exact native binaries made both parent comparisons rebuild-free. Hosted native jobs still spend most turnaround compiling, so local ABBA provides the immediate decision receipt while CodSpeed catches up.

PR #2925, opened 2026-09-13
perf: share publication buffers across resident reads and persistence

aeplay:
🤖 Implements the publication-buffer thesis in #2913, on top of merged #2924 (a68fcd455d).

## Before / after

Previously resident visibility, overlay rebuilding and persistence cloned the staged operation payloads. StagedWriteState now owns append-only shared blocks: a unique tail can grow; once shared, subsequent writes start a new block. Resident overlays and persistence snapshots retain block references. Their latest-write indexes remain independent and address exact operation ordinals with logarithmic block lookup, not a linear scan of publications. Persistence releases its RefCell borrow before awaiting the backend.

An additive borrowed atomic-write method retains the ordinary owned API as its conservative portable fallback. The RocksDB implementation builds the same native WriteBatch from borrowed bytes. Physical layout conversion still owns physical keys, but does not copy value bodies. Adapter forwarding is being verified; no end-to-end gain is claimed yet.

## Invariants and worked cases

- Alice's 1,000-operation publication shares its payload with the resident overlay and a persistence snapshot. Bob's later append starts another block; Alice's snapshot still contains exactly its original operations. No per-operation reference-counted allocation is added.
- If successive publications overwrite the same key, the resident latest-write index still resolves the last operation. Dropping a contiguous durable prefix rebuilds references to the remaining publications, preserving publication order and last-write-wins.
- A persistence attempt takes its snapshot at the same point as before. Cancellation before backend admission remains retryable; uncertain commit failures remain fatal. No staged payload is taken away from an outstanding publication.
- Large-value observers may append metadata before durability or persist follow-on metadata after durability. The existing lifecycle flag, recovery markers and handoff remain authoritative; buffers are not prematurely frozen at initial assignment.
- A batch deleting a valid key and then referring to a missing column family is rejected as Uncommitted before either operation reaches RocksDB. Native write/flush errors retain PossiblyCommitted classification.
- Storage/wire codecs, CF/key mapping, versioned byte contracts, batching order, WAL mode, durability frontier and user-facing result assertions are unchanged. The new sharing representation is ephemeral, not serialized. Existing owned backend implementations remain supported by the default adapter.

## Preflight / prediction

[...]

PR #2926, opened 2026-09-13
perf: reuse join arrangements for identical immutable snapshots

aeplay:
🤖 Trial for #2913, stacked on #2925. Do not merge: Anselm will approve the overnight stack in the morning.

## Before / after

Full-snapshot join consumers rebuilt their shared arrangement even when another consumer had just indexed the exact same immutable input. Logical timestamps alone could not establish snapshot equality, so every Replace call rebuilt both sides.

This trial carries the existing Arc<RecordDeltas> identity to inner/semi/anti joins. A weak reference on the arrangement proves which complete snapshot it contains, without retaining the input row vector. A Replace with the same exact snapshot object AND logical sub-tick reuses the arrangement. Its right side also skips join-key extraction, since replacement output probes only need the left keys. Ordinary incremental deltas keep their existing path.

## Evidence and prediction

Opt-in capture of the current native 27,518-row cold workload at 93ebe46d8e found 4,824 join-side replacement requests / 2,334,418 input records / 482,947,108 bytes. Repeated fingerprints within runtime+arrangement+tick account for 1,151,314 records; 1,100,122 records also reused the same input buffer address. One narrow shared input accounts for most repeated records. Four observations had changed contents at the same logical timestamp. Hashes/addresses are diagnostic hints only, not the production proof.

The complete #2925 GQL cold profile still spends 3.417 s inclusive in join-index building. The parent profile's union of index construction/maintenance scopes was 6.1% overall, not the sum of its overlapping call totals. Original prediction: 2–4% lower cold time, low-to-medium confidence. Quiet native ABBA now measures 6.3–8.5% lower cold medians (12.15/12.00 s → 11.12/11.24 s), with possible 0.2–2.3% sequential-write slowdown. All controls, exact provenance and caveats are in the [receipt](https://github.com/garden-co/jazz/issues/2913#issuecomment-5651502433). CodSpeed confirmation remains pending; the larger native result is not a revised prediction.

Prior-art review is on #2913: #2864 only optimized fresh map construction; #2901 changed inner buckets. This trial removes equivalent builds using actual existing input identity, not another map layout. Typed arrangement scope/lifecycle and hydration memos already existed and remain authoritative.

- Two consumers of the same complete snapshot at one logical sub-tick reuse one index; each still produces its own result and maintains its own semi/anti publication state.
- A new snapshot with different contents at the same tick MUST rebuild. Even byte-equal but separately allocated snapshots conservatively rebuild. Neither a hash nor timestamp alone authorizes reuse.
- A changed descriptor, key/comparison mode or scope is still resolved through the existing typed Arrange key; no cross-binding/permission reuse is introduced.
- Signed duplicates still consolidate as before: +2 and -1 leave +1; a negative record remains represented. Self-join partial clones cannot claim complete-snapshot identity.
- Incremental changes, key/bucket replacement and clear invalidate the identity proof. A later old snapshot must rebuild instead of retaining modified contents. Snapshot cloning keeps old state unchanged when a staged update changes the new state.
- The weak reference prevents allocation-address reuse while the proof exists, but does not keep the RecordDeltas payload alive. Arc mutation/dissociation cannot retain the old identity.
- Cancellation/rollback and recursive sub-tick handling remain on the existing staged state lifecycle. No wire/storage serializer or durable encoding changes, no authorization/result assertion changes.

## Validation

New internal tests cover index pointer reuse, changed/equal-but-distinct snapshots, signed contents, incremental invalidation, partial replacement/clear, weak lifetime and retained old snapshots. Internal tests are needed because output equality cannot prove allocation reuse or lack of retained input vectors. Existing public query/recursive/multiset tests remain unchanged except mechanical adaptation to the private join-input wrapper.

All 761 Groove library tests passed (2 existing ignored), as did all three incremental-delivery canaries. Exact-head CI and CodSpeed remain running. An earlier diagnostic-only CI job failed before compilation because its S3 compiler-cache connection was reset; that was an infrastructure failure, not a test failure. Opt-in capture remains behind both cold-settle-attribution and GROOVE_TRACE_ARRANGEMENT_SNAPSHOTS; capture timing is discarded. No adopter testing requested yet. No independent review or full local CI-equivalent claim.


aeplay:
🤖 #2926 native ABBA receipt: cold improvement is materially larger than the 2–4% prediction. Retain this trial in the unmerged overnight stack pending CI and CodSpeed.

Quiet matched ABBA, three samples per block / six per revision, no local builds/tests during timing. Individual block medians, not pooled statistics:

| Workload | Parent A / B | Candidate A / B | Paired elapsed change |
| --- | --- | --- | --- |
| Permissioned first sync, 27,518 rows | 12.15 / 12.00 s | 11.12 / 11.24 s | −8.5% / −6.3% |
| Sequential insert, 1,350 | 2.003 / 1.989 s | 2.008 / 2.027 s | +0.2% / +1.9% |
| Sequential update, 1,350 | 3.577 / 3.563 s | 3.607 / 3.644 s | +0.8% / +2.3% |
| Batch update, 1,350 | 172.2 / 174.6 ms | 172.7 / 169.5 ms | +0.3% / −2.9% |
| Reopen, 1,500 | 117.4 / 116.5 ms | 115.6 / 115.0 ms | −1.5% / −1.3% |

Cold ranges: parent 11.79–12.18 s, candidate 10.87–11.36 s (no overlap in these six samples). Parent cold CPU user time across three samples is 32.76/32.56 s; candidate 30.55/30.55 s. Candidate cold peak process RSS 3,323,164/3,334,312 KiB versus 3,293,784/3,327,380 parent: no established peak-memory saving. Process RSS includes fixture setup/allocator effects. Sequential writes may be slightly slower; do not call all controls improved. Hosted confirmation is important for those small differences.

All 27,518 visible rows and existing todo assertions are unchanged. Native Linux x86_64, profile.perf opt-level 2/no LTO/256 codegen units, mimalloc, RocksDB WAL-no-fsync. Candidate f1a699ff7910472310b7fb11ef47b2ba5237a273; baseline eccd5b19be3580744d23e7ae3dea5c985170078f (#2925). Capture-only commits between those checkpoints do not compile into the ordinary benchmark feature set.

SHA256:
- parent todo ae8f86072c0c6ba664482608043bb0dfd3e0a01b433f9e324472a41d171c0431
- parent cold 9ad85da1dfd2f7f8ec9f8dc84ebe3c4ea96100a2634f67e6b3ee09071c024bc5
- candidate todo 6bba076b6046c8d005eca7438f555f6eea8d86a55f3195dad6e0a8b941809d20
- candidate cold 8f3575f89d06d89bcbb6e07e9b548042d717fc802ce48274d3036a85847ede7c

Receipts/scripts/build JSON/copied binaries/eight logs: /tmp/jazz-arrangement-snapshots-qICGan. Baseline copies: /tmp/jazz-publication-ownership-PxOwdw. Tests: 761 Groove lib passed, 2 existing ignored; all three incremental-delivery canaries passed. CI/CodSpeed for the exact runtime head are running. No full local CI-equivalent or independent review claim.

The larger local effect does not retroactively change the prediction. The 6.1% cost ceiling came from an ARM hosted parent profile, while native timing here is x86_64; the trial also skips repeated right-side key extraction, outside that index-only union. A candidate capture is being collected to confirm actual identity hits, and the new full GQL graph will distinguish those costs. Retain the result as a local 6–8.5% cold win, not a promised hosted multiplier.

No merges before Anselm's morning approval; no adopter testing requested yet. Next thesis will target the remaining complete-manifest normalization overhead without changing the wire protocol.

Tooling friction: two capture-boundary corrections were needed because the walltime harness skips the diagnostic summary path; exact readiness markers are now in the shared path, and future captures can reuse them.

aeplay:
🤖 #2926 hosted outcome and full-flamegraph follow-up. No adopter testing requested yet; the stack remains draft and unmerged for morning approval.

CodSpeed run `6aa63d0aa8e22370f5dcacb7` resolves through GQL to actual head `3176a335158b9e771ad7f1d8b582a8ca1c8d6993`. Parent #2925 is run `6aa630efb4c66725d41d9acf`, actual `eccd5b19be3580744d23e7ae3dea5c985170078f`. Same existing wall-time benchmark URIs and fixture boundaries.

| Benchmark | Parent minimum | Head minimum | Elapsed change | Parent → head median |
|---|---:|---:|---:|---:|
| permissioned cold 27,518 rows | 58.4305 s | 56.1445 s | −3.91% | 58.8106 → 56.9026 s (−3.24%) |
| sequential inserts 1,350 | 12.4861 s | 12.4486 s | −0.30% | 12.4868 → 12.4917 s |
| sequential updates 1,350 | 21.6135 s | 21.6983 s | +0.39% | 21.6993 → 21.7160 s |
| batch updates 1,350 | 1.09914 s | 1.11528 s | +1.47% | 1.10307 → 1.12008 s |
| reopen 1,500 | 0.657025 s | 0.662416 s | +0.82% | 0.689602 → 0.679882 s |

Cold and sequential workloads have three measured rounds; batch/reopen have five. Both reopen distributions contain an IQR outlier. These are distributions, not a significance claim. Hosted cold improvement fits the original 2–4% prediction, although it is smaller than the paired native x86_64 6–9% result. Todo remains broadly flat, with a small hosted batch-update cost worth keeping visible.

Downloaded complete GQL wall-time graphs for cold (`6aa63fd2ad9a6239bf6b37aa`) and updates (`6aa63fd2ad9a6239bf6b37a8`). The union of cold index build, arrangement update and bucket commit scopes falls from **3.5473 s to 2.1869 s**, a 1.3604 s reduction; nested scopes are counted once. `advance_arrangement` separately falls from 3.2197 s to 1.8596 s, but overlaps that union and must not be added. Whole sampled CPU self time falls 58.1999 → 56.1819 s. This supports the intended rebuild-elision mechanism; opt-in native capture independently counted 1,098,384 skipped input records.

Remaining cold leaders include maintained terminal application (7.03 s inclusive, containing decoding and fact bookkeeping), index construction (2.13 s), and collector work. Full update profiles still show avoidable result-sized root-collector copying/position construction. The latter needs an exact ordering proof: encoded row bytes are part of the tie-break, so unchanged declared sort fields alone are insufficient.

Correctness CI is green at this head. No merge performed.

Tooling friction: MCP run discovery can initially show a synthetic merge hash; resolve the completed run through GQL to bind result IDs and source SHA before recording a comparison.

PR #2927, opened 2026-09-13
perf: normalize supporting manifests in identity order

aeplay:
🤖 Trial for #2913, stacked on #2926. No merge before Anselm's morning approval.

## Before / after

Publication built a temporary physical-row BTreeSet only to return an ordered vector. Reception cloned every physical row into another tree for duplicate validation, expanded rows across source roles, then sorted the larger fact vector.

Publication now sorts/deduplicates the physical-row vector directly. Reception sorts physical rows, rejects adjacent duplicates and records contiguous physical-table ranges. It then emits coverage and row facts in compiler source order, using the matching sorted range for each source. The output is already strictly ordered and unique, so no physical identity tree or expanded-fact sort is needed. Exact expansion capacity is computed with checked arithmetic.

## Invariants and examples

- Two aliases of one table still produce two local source facts per row but only one physical wire row. Different physical-table order and authored native-table labels preserve the existing fact ordering.
- Arbitrarily reordered incoming rows remain accepted; equal rows originally far apart still reject after sorting. Empty snapshots retain coverage for every compiled source, including tables with no rows.
- Content/deletion layers, exact transactions, schema/batch/branch/digest metadata stay intact. Rows outside the compiled physical dataset and malformed native references reject before fact expansion. Later body validation and whole-batch atomic rejection are unchanged.
- Initial/reset, pending markers, later empty snapshots, replay and adjacent predecessors in one receive batch keep their existing behavior. Receiver-local non-peer facts cannot become removals.
- No persistent cache, delta wire protocol, authorization change, benchmark workload or existing result assertion changes. Existing wire/storage serializers and final manifest order are unchanged; this trial changes construction order only.

## Evidence and prediction

The full #2925 sequential-update GQL graph attributes 4.117 s to receiver normalization and a separate 1.941 s to sender physical-manifest construction out of 21.576 s. Nested comparison/difference costs are not added again. Predict 8–15% lower sequential insert/update latency; this is not a receipt. Cold has only ~0.53 s in these scopes out of ~58.2 s, so no material cold win is promised. Keep cold, batch and reopen as unchanged controls.

Preflight: #2917 explicitly retained the sender physical-row tree; #2920 retained the receiver identity tree and row-first expansion followed by a wide-fact sort. This removes those remaining stages using their ordering relationship, not a repeat of the rejected weighted-event representation trials. [Thesis and profile evidence](https://github.com/garden-co/jazz/issues/2913#issuecomment-5651517858).

## Validation / status

[...]

PR #2928, opened 2026-09-13
perf: fuse shared witness source-presence updates

aeplay:
🤖 Continues #2913. Draft linear successor to #2927; do not merge before morning approval.

## Before / after

A compiler-proven shared version/replacement witness previously cloned and looked up the same source-presence key twice, recording the same publication-journal touch on each call. It now applies the two equal signed role deltas in one map lookup and one journal touch. Ordinary unshared terminals still update exactly one role.

- The three signed origin weights remain independent; visible closure presence is `any(weight > 0)`, not their sum. This is not a collapsed refcount.
- Starting at `[0, 0, 0]`, shared `+1` yields `[1, 1, 0]` and one addition. Removing only replacement still leaves version live; removing version later publishes the final removal.
- Starting at `[1, -1, 0]`, shared `-1` yields `[0, -2, 0]` and one removal, retaining the negative maintenance state. A program-fact origin with positive weight would keep the fact visible instead.
- Same-sign paired increments have monotonic union presence, so the old two steps could emit at most one transition. A single before/after comparison preserves that sequence.
- All-zero entries are removed. Journal acknowledgment, retries, independently selected deletion witnesses and multisink coalescing remain unchanged.
- No wire/storage format, authority boundary, terminal netting or payload-sharing proof changes. No durable encoder changes to review.

## Validation and measurement

A new internal signed-state oracle covers 625 combinations of initial role weights and paired deltas. It compares exact transitions with both the old two-call sequence and an independent arithmetic model, and checks journal state, negative retention and closure visibility. Internal testing is needed because public row results do not expose individual signed weights or journal touches. Existing role lifecycle and public correctness assertions are unchanged.

Focused test, broader correctness CI and native timing are pending at this first coherent checkpoint. Prediction: 0.2–0.6% lower permissioned cold time, small/flat todo effects. The separate transition-copy cleanup and larger root-collector ordering work are non-goals.

Tooling friction: source-bound native binaries will provide a direct ABBA comparison against #2927 without rebuilding the parent.


aeplay:
🤖 #2928 native result: modest cold improvement in both orders; retain pending CodSpeed confirmation. No adopter testing requested yet. No merges.

Source `9dc6cfa92f496318696321b4cfeb63665dfb9902` versus parent #2927 `f6897b12fedba7bb5d98a5658cc4b738b1f3316d`. Identical profile.perf/mimalloc/RocksDB WAL-no-fsync standalone fixtures, quiet ABBA, three samples/block and six per revision. Source/binary hashes and all logs live under `/tmp/jazz-fused-source-presence-smUsA3`; parent binaries under `/tmp/jazz-ordered-manifests-uOy4B0`.

| Case | Base A / B medians | Candidate A / B | Paired elapsed change |
|---|---:|---:|---:|
| permissioned cold | 11.36 / 11.24 s | 11.13 / 11.09 s | −2.0% / −1.3% |
| sequential inserts 1,350 | 1.643 / 1.639 s | 1.651 / 1.650 s | +0.5% / +0.7% |
| sequential updates 1,350 | 2.907 / 2.892 s | 2.904 / 2.905 s | −0.1% / +0.4% |
| batch updates 1,350 | 172.9 / 169.4 ms | 174.1 / 172.6 ms | +0.7% / +1.9% |
| reopen 1,500 | 117.2 / 114.9 ms | 114.6 / 115.8 ms | −2.2% / +0.8% |

Cold process user CPU for each three-sample block: base 30.716/30.609 s, candidate 30.370/30.176 s. This supports a small real work reduction, but does not establish a precise 1–2% portable gain. The original prediction was 0.2–0.6%; hosted confirmation matters, especially with small todo/batch costs. No peak-memory claim.

Full local Jazz library: 2,031 passed, 2 ignored; all three incremental-delivery canaries pass. New 625-case signed-role oracle passes. Pre-commit Clippy/format/sensitive-data gates pass. Hosted correctness and benchmark workflow run in parallel; not claiming a full local CI-equivalent run.

Proceeding to the separately logged known-alias point-miss thesis above this checkpoint, while CodSpeed resolves this small effect.

Tooling friction: repeat pairs and CPU receipts help distinguish small cold improvements from incidental first-block wall-time drift.

aeplay:
🤖 #2928 first hosted result **does not confirm an end-to-end win**. Change recommendation is now provisional/unproven, not a counted improvement. No adopter testing requested; do not merge.

Head run `6aa643d8a8e22370f5dcacf6`, actual GQL SHA `9dc6cfa92f496318696321b4cfeb63665dfb9902`, versus parent #2927 run `6aa6403f066a4cec8e339fd3` / `f6897b12fedba7bb5d98a5658cc4b738b1f3316d`.

| Case | Parent → head minimum | Elapsed change | Parent → head median |
|---|---:|---:|---:|
| permissioned cold | 55.8602 → 56.3073 s | +0.80% | 56.3658 → 56.6076 s (+0.43%) |
| sequential inserts 1,350 | 10.4939 → 10.5359 s | +0.40% | 10.7246 → 10.6721 s |
| sequential updates 1,350 | 17.8821 → 18.0116 s | +0.72% | 17.8921 → 18.0540 s (+0.91%) |
| batch updates 1,350 | 1.10174 → 1.10054 s | −0.11% | 1.10305 → 1.10695 s |
| reopen 1,500 | 0.652429 → 0.644069 s | −1.28% | 0.670621 → 0.669471 s |

Complete GQL cold/update graphs downloaded (results `6aa64683ad9a6239bf6b5411` / `6aa64683ad9a6239bf6b540f`). Intended cold presence-bookkeeping scope does shrink: **0.56770 → 0.38047 s** (~0.187 s saved, around 0.34% of total). This fits the original 0.2–0.6% mechanism budget, but does **not** override the slower end-to-end receipt. Local 1.3–2.0% cold improvement was too optimistic as a portable estimate.

Next measurement: repeat only the permissioned cold job at both unchanged parent and head, not the entire suite, to distinguish a real small regression from run-to-run variation. If the cold gain remains unconfirmed, prefer dropping/reverting this lever instead of carrying complexity for a phase-only saving. Todo confirmation can follow if the cold repeat justifies retaining it. Other stack entries remain separately measurable and unmerged.

Correctness CI and all current benchmark jobs pass. This is a performance evidence disagreement, not a correctness failure.

Tooling friction: a sub-percent predicted gain needs unchanged-source parent/head repeats; lower local allocation or scoped CPU work alone is not acceptance.

aeplay:
🤖 #2928 repeat: no convincing end-to-end win. No adopter testing requested; no merge authorized.

The targeted unchanged-source cold jobs both completed successfully. Their repeat minimums are parent #2927 **55.983146962 s** and candidate #2928 **55.912865071 s** (−0.13%); medians 56.283950318 and 56.189620302 s (−0.17%). Three rounds each, no flagged outliers. First pair was +0.80% worse by minimum. Native pairs had looked 1.3–2.0% better, but two hosted pairs do not corroborate a useful end-to-end benefit.

Recommendation: exclude this trial from the eventual landing set unless a later combined receipt changes the conclusion. Its fused presence scope does improve by about 0.187 sampled CPU seconds, but that is not sufficient evidence to claim an application win. Stop spending runner rounds trying to establish a tiny effect. The current unmerged child stack still includes it to preserve exact parent comparisons; remove it in a deliberate clean restack/consolidation before asking for morning landing approval, or explicitly identify it as unresolved. No silent attribution of its phase gain to total wall time.

Important tooling discovery: rerunning a single GitHub job updates the existing CodSpeed run rather than creating a new run ID. GQL result IDs changed: parent cold now `6aa64c62ad9a6239bf6b64ef`, head cold `6aa64c27ad9a6239bf6b5eee`; original result IDs and distributions remain in the earlier receipt. Run ID alone is therefore insufficient to pin a repeated result. Exact source hashes are unchanged.

Tooling-friction: receipts must retain both run ID and result ID plus the distribution, since a rerun can replace the current result behind a stable run ID.

PR #2929, opened 2026-09-13
perf: keep known-alias transaction misses point-bounded

aeplay:
🤖 Advances #2913, now stacked on #2927 after excluding #2928. Original measurements below refer to the original parent; the pruned combined receipt and exact old/new SHA map are on #2913. Draft overnight checkpoint; do not merge before morning approval.

## Before / after

Looking up a transaction under an already-known node UUID first tried its exact `(time, alias)` key. If absent, it scanned all `jazz_nodes` rows and tried that same alias again. Now a known alias returns the exact point lookup's result directly, including absence. Unknown aliases keep the existing durable discovery path.

## Governing invariant

Node UUID → durable alias is unique. Recovery rejects conflicting durable aliases; `ensure_node_alias`, the sole production alias-table writer, installs the resident mapping only after successful alias persistence. Other resident-map writers resolve durable records. The node's database slot is private, not an application mutation API. A known alias therefore cannot hide an absent transaction under a second valid alias.

## Worked cases

- Known UUID, absent transaction: one exact point lookup returns `None`; no catalogue scan or duplicate lookup.
- That exact transaction arrives later: the next read sees it. No negative transaction cache is introduced.
- Unknown resident alias with an existing durable transaction: existing discovery scans and restores the alias, then returns the transaction.
- Truly unknown UUID: discovery still returns absence without inventing a resident mapping.
- Reopen: recovery restores unique aliases, after which misses remain point-bounded. Conflicting persisted aliases still fail recovery; failed alias persistence still cannot install a resident mapping.
- Malformed transaction framing/status/payload errors still cross the same decoder boundary. Status remains a projection; full transaction reads retain payload validation. Pending persistence still clamps durability independently.

## Non-goals / validation

No transaction cache, alias eviction/rebinding, schema/index addition, wire/storage encoding change, authorization bypass, or mutation of existing test assertions. Known-alias lookups no longer incidentally scan unrelated catalogue rows; storage corruption remains governed by recovery and the requested record's validation.

New focused coverage checks repeated status/full-record misses via storage counters, arrival after absence, unknown discovery, and reopen. Existing alias-collision, failed-persistence and status-projection regressions remain unchanged. Broader CI, focused execution and native timing are pending at this first coherent checkpoint. Prediction: 2–3% lower permissioned cold time, 0.5–1% lower sequential update time, likely a small insert improvement.

Also refreshes the existing CodSpeed GQL recipe: public queries need no token, exact result/source resolution matters, and full wall-time subtree accounting must not double-count nested scopes.

Tooling friction: the complete GQL profile identified the real fallback scan after a partial-root summary misleadingly suggested async-frame copies.


aeplay:
🤖 #2929 native result: small todo improvement; **no demonstrated cold-load improvement**. Keep the small simplification pending CodSpeed, not the original cold prediction. No adopter testing requested yet; no merges.

Head `02de7c5e63` versus #2928 `9dc6cfa92f496318696321b4cfeb63665dfb9902`. Identical standalone native profile.perf/mimalloc/RocksDB WAL-no-fsync fixtures, ABBA, three samples/block, no concurrent local builds/tests. Source and binary SHA-256 receipts are preserved under `/tmp/jazz-known-alias-misses-qwnrb3`.

| Case | Base A / B medians | Candidate A / B | Paired elapsed change |
|---|---:|---:|---:|
| permissioned cold | 11.53 / 11.16 s | 11.41 / 11.28 s | −1.0% / +1.1% |
| sequential inserts 1,350 | 1.640 / 1.653 s | 1.621 / 1.635 s | −1.2% / −1.1% |
| sequential updates 1,350 | 2.910 / 2.917 s | 2.859 / 2.907 s | −1.8% / −0.3% |
| batch updates 1,350 | 171.2 / 174.4 ms | 170.0 / 173.3 ms | −0.7% / −0.6% |
| reopen 1,500 | 115.0 / 125.7 ms | 115.3 / 115.8 ms | +0.3% / −7.9% |

The last base reopen block drifted; do not claim a reopen win. Cold CPU is effectively unchanged: base 30.554/30.306 s versus candidate 30.543/30.407 s per three samples. Todo suite user CPU is modestly lower (13.828/13.813 → 13.651/13.744 s).

The original 2–3% cold estimate was too optimistic for this narrow path. The profile's ~1.84 s fallback scan scope does not distinguish known-alias misses from unknown-alias discovery; only the former was removed. A plausible remaining contributor is repeated discovery of an unknown originating UUID before its transaction/alias is installed. Do not infer that it dominates without counters or the next full GQL graph. A bounded negative **alias** lookup memo (not a negative transaction cache), or restoring a discovered alias even when its transaction is absent, would need a separate lifecycle proof and trial before implementation.

New regression verifies 32 known-alias status/full-transaction misses issue exactly 32 transaction reads and no catalogue reads/scans; arrival-after-miss and reopen work. Full Jazz library: 2,032 passed, 2 ignored; three incremental canaries pass. Existing collision/persistence-failure/status assertions are unchanged. CI and CodSpeed pending.

Next, take the larger separately logged whole-member-reconciliation opportunity instead of expanding this uncertain scan premise immediately.

Tooling friction: a scope-level profile measures total fallback cost, not the removable branch's invocation mix; record that distinction in predictions.

aeplay:
🤖 #2929 hosted confirmation, with a narrower cold-load conclusion than initially predicted. No adopter testing requested yet; overnight drafts remain unmerged pending morning approval.

Exact GQL source: `02de7c5e6314dc59a009da9e0ea5290833735bca`, run `6aa64627a8e22370f5dcad09`, compared with direct parent #2928 run `6aa643d8a8e22370f5dcacf6`.

| WallTime case | Parent minimum → candidate minimum | Change | Candidate median |
|---|---:|---:|---:|
| Permissioned cold load | 56.3073 → 55.9768 s | −0.59% | 56.1036 s |
| Sequential inserts, 1,350 | 10.5359 → 10.2712 s | −2.51% | 10.4092 s |
| Sequential updates, 1,350 | 18.0116 → 17.7336 s | −1.54% | 17.7383 s |
| Batch update | 1.10054 → 1.09956 s | −0.09% | 1.10244 s |
| Reopen | 0.64407 → 0.64965 s | +0.87% | 0.67041 s |

Three rounds for cold/inserts/updates and five for batch/reopen, no flagged outliers. The native ABBA was cold-flat and approximately 1% better on todo. Taken together, this supports a modest todo improvement, not the originally predicted 2–3% cold win. Reopen is essentially unchanged by median and slightly worse by minimum.

Downloaded both complete cold/update wall-time graphs via GQL, not just MCP summaries. Cold result `6aa648d3ad9a6239bf6b592b`, update `6aa648d2ad9a6239bf6b5927`. The cold graph's `query_transaction_fields` subtree still totals 2.858 of 55.670 sampled CPU seconds; 1.930 s is under the remaining catalogue-scan path. #2929 only removes catalogue scans for already-known node aliases. Unknown-node lookups can still repeat discovery, and finding a durable alias currently restores the resident map only if the particular transaction exists. Those are a possible next bounded-discovery thesis, not work already eliminated by this PR. Any negative alias memo must be invalidated before alias publication can start, including cancellation/error paths; transaction absence itself must never be cached.

Separately, #2930's previously reported timing-canary CI failure passed its unchanged-source workspace rerun; the full correctness run is now green. No code or assertion was changed to obtain that result. Original failure remains linked in its receipt.

Tooling-friction: MCP initially exposes a synthetic merge SHA; direct GQL run-to-source verification remains necessary for precise receipts.

PR #2931, opened 2026-09-13
perf: keep root ordering state shared across payload-only updates

aeplay:
🤖 Advances #2913. Linear overnight draft on #2930; do not merge before morning approval.

## Before / after

Root collectors copied their entire public order/key indexes on the first prepared-state mutation and rebuilt all public positions even for a payload-only edit. The latest post-manifest update profile assigns about 2.84 s of 17.85 sampled CPU seconds to this collector, including 1.01 s in payload copying.

The order/key indexes now have independent copy-on-write ownership. When every touched group retains its exact public ordering key, only changed payloads are rendered; no full sequence/position bookkeeping is needed.

## Ordering proof and invariants

The existing key includes declared sort values plus complete record bytes. Simply comparing declared values would be wrong for ties. The public index may now omit the record-byte suffix only when **all** group identity fields occur in the effective sort, with non-null UUID/U64 types. Distinct groups then necessarily have distinct sort prefixes. The per-group signed-weight index and representative selection always retain complete bytes.

- Updating the title of ID 32 in an ID-ordered list emits exactly its Update and shares the old public indexes.
- Updating a real ordering field still takes the general positional algorithm. Two crossing moves plus an insertion use sequential edit positions, not independently applied final ranks.
- A group's positive count crossing zero removes its occurrence. Negative maintenance state remains retained; a later zero crossing emits no phantom insert, and positive revival inserts at the correct rank.
- Missing occurrence/route tie fields, unsupported identities, nullable values and floats keep the existing full-byte public key. Floats are deliberately excluded because sort keys canonicalize signed zero. Only sort fields actually consumed by the zipped sort/direction builder count toward the proof.
- Prepared ticks cannot mutate their predecessors: a payload edit shares untouched indexes; real order/membership mutation detaches them. Reset replaces the private indexes without cloning their old contents.

## Scope and tests

Private runtime state only. No storage or wire format, public schema, permission, winner selection, finite-window or non-root collector behavior changes. This does not promise change-proportional inserts/moves; real public index changes still clone their retained trees as before. Existing assertions are unchanged.

Full Groove suite passed before the final additive fallback check (763 library tests passed, 2 ignored; multisink integration 10 passed; other integration/doc suites green). New tests cover unchanged-order updates at two retained scales, predecessor isolation, signed removal/revival, proof refusal and full-byte fallback, and public payload values plus mixed positional edits. CI and native ABBA will run next. Prediction: 8–14% lower sequential-update latency, smaller/flat insert and cold-load effect; not yet a timing claim.


aeplay:
🤖 #2931 local result: payload-only root updates remove another large sequential-update cost. No adopter testing requested yet; do not merge before morning approval.

Exact source `9f26c8ea3f082617e6b3c88608f454369cde247c`, parent #2930. Native ABBA, identical RocksDB WAL-no-fsync fixture, profile.perf + mimalloc, three samples per arm; source/binary hashes and logs in `/tmp/jazz-root-order-stability-Qe1xmz`.

|---|---:|---:|---:|
| 1,350 sequential updates | 2.458 / 2.452 s | 2.005 / 2.014 s | −18.43% / −17.86% |
| 1,350 sequential inserts | 1.400 / 1.405 s | 1.401 / 1.401 s | +0.07% / −0.28% |
| Batch update | 172.2 / 172.9 ms | 171.0 / 170.1 ms | −0.70% / −1.62% |
| Reopen | 117.4 / 116.1 ms | 115.4 / 116.6 ms | −1.70% / +0.43% |
| Permissioned cold load | 11.54 / 11.14 s | 11.23 / 10.93 s | −2.69% / −1.89% |

Sequential updates exceed the 8–14% prediction locally; insert/reopen are flat. Cold is a possible small gain, but its first parent had higher system CPU and one 12.61 s sample. Do not promote that cold result to a stable claim before CodSpeed. Todo process user CPU falls 11.95/12.01 to 10.76/10.88 s; cold 30.68/30.44 to 30.46/30.19 s. No peak-memory claim.

Full Groove suite passed (763 library tests, 2 ignored; multisink 10 passed, remaining integration/doc suites green), final additive full-byte fallback test passed exactly, all three Jazz incremental canaries passed. CI is running in parallel with no known failing job so far. No existing assertions modified; private runtime state only, no authoritative storage/wire encoding changes.

Tooling-friction: an automatic ABBA table renderer with paired drift warnings would save manual receipt transcription; current raw logs and hashes remain authoritative.

aeplay:
🤖 #2931 hosted confirmation: **23.0% faster sequential updates**, with a code-grounded explanation for exceeding the original estimate. No adopter testing requested yet; correctness CI is green, but the stack stays unmerged pending morning approval.

Exact source `9f26c8ea3f082617e6b3c88608f454369cde247c`, run `6aa64e8db43688c29e2a45a7`, direct parent #2930 `f3590b40b8996007911ec72be0cfaf83cd9f739a` / run `6aa649a8a8e22370f5dcad2e`.

| WallTime case | Parent minimum → candidate minimum | Change | Parent median → candidate median |
|---|---:|---:|---:|
| Sequential updates, 1,350 | 15.4843 → 11.9190 s | −23.03% | 15.5277 → 11.9653 s |
| Sequential inserts, 1,350 | 9.00455 → 8.79518 s | −2.33% | 9.01309 → 9.17782 s |
| Permissioned cold load | 56.0740 → 55.6767 s | −0.71% | 56.3667 → 56.0972 s |
| Batch update | 1.11394 → 1.08648 s | −2.47% | 1.11750 → 1.09603 s |
| Reopen | 0.65712 → 0.63797 s | −2.92% | 0.67590 → 0.64759 s |

Three rounds on sequential/cold, five on batch/reopen; candidate has no flagged outliers. Inserts are inconclusive: minimum improves but median regresses, and local pairs were flat. Cold supports a small gain, not the ~2% native headline. The large update gain is consistent across minimum/median and native pairs (−17.9–18.4%).

Complete update GQL graph: root-collector scope falls **2.798729446 → 0.045804078 sampled CPU seconds**. A separate, non-overlapping `IncrementalEvaluation::install` scope falls **0.625543280 → 0.085924438 s**: sharing unchanged indexes also avoids disposing the old full index trees during state installation. Together these account for about 3.293 sampled CPU seconds, explaining most of the 3.565 s wall-time improvement. The prediction budgeted the collector's internal work, not this additional downstream disposal saving. Do not add its nested payload-clone cost again.

The newly dominant update scopes are complete supporting-snapshot normalization (~1.690 s), retained-view footprint refresh (~0.727 s), manifest construction (~0.709 s), and wire decoding (~0.788 s), with overlap within their listed descendants. These remain follow-up candidates; no delta-sized wire protocol is claimed.

Full cold/update artifacts downloaded and analyzed: result IDs `6aa6513aad9a6239bf6b709b` and `6aa6513aad9a6239bf6b70a2`. Native/hosted receipts, exact source/binary hashes and full graphs are in `/tmp/jazz-root-order-stability-Qe1xmz`.

Tooling-friction: include destruction/state-installation scopes when budgeting copy-on-write changes; the saving need not stay within the mutating function.

PR #2932, opened 2026-09-13
perf: bound repeated unknown node alias discovery

aeplay:
🤖 Advances #2913, linear overnight draft on #2931. Do not merge before morning approval.

## Before / after

Unknown node UUIDs caused a complete `jazz_nodes` catalogue scan for every transaction lookup. Known aliases were already point-bounded by #2929. A completed scan that finds no alias now retains exactly one absent UUID. Repeated queries for that UUID avoid both the catalogue scan and impossible transaction reads. A different absent UUID replaces the entry, so arbitrary peer churn cannot grow the cache.

- Repeated timestamps from an unknown UUID return absence without storage reads after the first completed scan. This is an alias-absence proof, **not** a missing-transaction cache.
- A later incoming transaction creates its node alias through the existing sole writer. That writer clears the absence proof **before its first await**, not just after successful persistence. Cancellation or failure cannot retain stale negative evidence after resident/durable publication may have started.
- Once an alias is resident, it takes precedence over the negative slot and every transaction remains an exact point read. A uniquely discovered catalogue alias is restored even when the particular transaction is absent; later transactions are not hidden. Ambiguous discovery retains the old loop and recovery still rejects conflicting durable aliases.
- A cached miss still checks the existing table/database health gate. It cannot mask an unrelated failed write that poisoned the database.
- Recovery starts without a negative entry; normal alias recovery, status projection, transaction payload validation and pending-persistence durability clamps are unchanged. No new authority, schema/index, public API, storage/wire encoding, or transaction-level lifetime cache is introduced.

## Validation / prediction

Three new exact focused tests pass: bounded repeated absence plus later public row visibility and UUID churn; cancelled discovery invalidating before suspension; and cached absence refusing a poisoned database. These internal counters/fault controls are necessary to prove work/ownership boundaries not observable through row equality alone. Existing assertions are unchanged. Broad Jazz/canary checks, CI and native timing follow this checkpoint.

Working prediction: 1–3% lower permissioned cold-load time, flat/small todo effect. The latest complete cold profile has ~1.93 of 55.67 sampled CPU seconds in the remaining transaction-lookup catalogue-scan path; only repeated absence is avoidable, so this is a testable opportunity rather than an assumed full-scope saving.


aeplay:
🤖 #2932 native result: a modest cold-load improvement, broadly matching the 1–3% prediction. No adopter testing requested yet; no merge before morning approval.

Exact source `12703b9b610bbb5e89fc72f73d9cc24c989f9c9f`, parent #2931. Native ABBA, identical RocksDB WAL-no-fsync fixture, profile.perf + mimalloc, three samples per arm. Raw logs and source/binary hashes: `/tmp/jazz-unknown-alias-discovery-4sx0oa`.

|---|---:|---:|---:|
| Permissioned cold load | 11.59 / 11.13 s | 11.02 / 10.90 s | −4.92% / −2.07% |
| Sequential inserts, 1,350 | 1.396 / 1.394 s | 1.404 / 1.395 s | +0.57% / +0.07% |
| Sequential updates, 1,350 | 2.006 / 2.003 s | 2.017 / 2.018 s | +0.55% / +0.75% |
| Batch update | 169.0 / 170.0 ms | 171.2 / 170.5 ms | +1.30% / +0.29% |
| Reopen | 116.4 / 115.8 ms | 116.4 / 116.1 ms | 0.00% / +0.26% |

The first cold baseline again had unusually high system CPU and a 13 s sample. Use the reverse pair's ~2% as the conservative local estimate, not the headline ~5%. Cold process user CPU fell 30.90/30.31 to 29.73/29.77 s, corroborating less work. Todo is flat/slightly slower, not a claimed win. Hosted confirmation is pending.

Validation: 2,036 Jazz library tests passed, 2 ignored; all three incremental canaries passed. New counter tests show repeated unknown-UUID lookups do no storage reads, while existing aliases still read every transaction. Cancellation clears the absence proof before suspension; cached absence still reports a poisoned database. No existing assertions changed.

CI note: the first optional shadow workflow was canceled before its shard receipts existed; its aggregate therefore failed with `expected 2 shard receipts, found 0`. This is not a test failure. Replacement workflow `34745219744` and authoritative correctness CI are still running. Original aggregate log is retained locally; no gate or assertion was relaxed.

Tooling-friction: first-arm cold baseline drift warrants paired/reverse comparisons and CPU receipts; a single favorable pair would overstate this improvement.

aeplay:
🤖 #2932 hosted confirmation: bounded unknown-alias discovery improves permissioned cold load by 2.78%, within the original 1–3% prediction. No adopter testing requested yet; no merge before morning approval.

Exact source `12703b9b610bbb5e89fc72f73d9cc24c989f9c9f`, parent #2931 `9f26c8ea3f082617e6b3c88608f454369cde247c`. CodSpeed run `6aa65236e44d81bc0c5a03f2`, resolved to the actual source through GQL rather than the synthetic merge hash.

| Workload | Parent → candidate minimum | Change |
|---|---:|---:|
| Permissioned cold load | 55.6767 → 54.1307 s | −2.78% |
| Sequential inserts | 8.7952 → 8.8827 s | +1.00% |
| Sequential updates | 11.9190 → 12.0355 s | +0.98% |
| Batch update | 1.0865 → 1.1024 s | +1.46% |
| Reopen | 0.6380 → 0.6519 s | +2.19% |

Cold medians agree: 56.0972 → 54.2550 s (−3.28%). Three rounds for cold/sequential cases and five for batch/reopen; no IQR outliers. Todo is a small regression/tradeoff, not a win; the native comparison also had slightly slower todo results. The cold benefit is real enough to recommend retaining this trial for morning review, while preserving that tradeoff.

Full GQL flamegraphs were downloaded, including cold result `6aa654e1ad9a6239bf6b8125` and update result `6aa654e1ad9a6239bf6b812a`. The union of `query_transaction_fields` cold stack scopes now accounts for 0.861 s sampled CPU, compared with 2.858 s in #2929 before this trial; the repeated unknown-node catalogue scan was 1.930 s of that previous scope. This is mechanism evidence, not an extra additive gain or a walltime measurement. Raw graphs, full distributions and native/source/binary receipts remain in `/tmp/jazz-unknown-alias-discovery-4sx0oa`.

Tooling-friction: pin GQL result IDs as well as run IDs; rerunning one CI benchmark job can replace a result within an existing CodSpeed run.

PR #2933, opened 2026-09-13
perf: share large temporary event identities with accumulator keys

aeplay:
🤖 Advances #2913, draft on #2932. Deliberate changed-premise retry of #2869/#2884; do not merge before morning approval.

## Before / after

The temporary event accumulator reserves 968 bytes per key and 1,080 per event because rare large result/proof variants inflate the enums. It also deep-clones a large result/fact identity into both its key and event.

Large identities now have one temporary thread-local shared owner, referenced by the map key and event. Large result payload alternatives are out of line; ordinary version/replacement entries stay inline. The current executable layout test measures **96-byte keys / 160-byte events**. After consuming an accumulator entry, its key is explicitly dropped and the owned identity is unwrapped for application. This avoids a second deep identity copy without exposing shared owners beyond the temporary accumulator.

- Distinct allocations containing equal identities still compare equal and combine weights. Pointer identity never defines equality/order; the derived enum/value ordering is unchanged.
- Duplicate equal events retain the first event's payload, as before. Positive/negative weights coalesce before application; net-zero entries make no retained-state changes.
- The complete input still decodes into the same BTreeMap before any retained state changes. A late decode failure drops the temporary map and cannot publish a valid prefix.
- Consuming a key releases its owner before applying its event, permitting an owned move. `unwrap_or_clone` retains correct ownership even if a future internal caller legitimately keeps another reference.
- These types are private and thread-local. No protocol structs, authoritative storage/wire encodings, public API, source semantics, alternate vector accumulator, or per-row arena is changed.

## Changed premise / validation

#2869 and #2884 boxed large alternatives but still independently owned/cloned their keys and values; #2872 also changed the accumulator algorithm. Their local speedups were small/inconclusive and do not establish a win here. This trial keeps the original accumulator and shares identity ownership instead. It will be judged on current native ABBA and CodSpeed—not reduced layout/allocation alone—and discarded if the end-to-end result does not justify retention.

New layout test fails on the parent representation (968/1,080) and passes on this candidate (96/160). New ownership tests prove shared allocation, value-based equality, duplicate coalescing and ownership recovery after key drop. Both focused tests pass. Existing behavioral assertions are unchanged; full Jazz/canaries/CI and native timing follow this checkpoint. Prediction: 1–3% lower cold load, small/flat todo, no peak-memory claim.


aeplay:
🤖 #2933 compact shared identities: correctness CI is green; timings show a small benefit, below the original 1–3% cold prediction. Treat as provisional for morning review, not a large allocation-derived win. No adopter testing requested yet; no merge before morning approval.

Exact source `85d772b2173d04c5023ca63de8b67ba2a5cb449d`, parent #2932 `12703b9b610bbb5e89fc72f73d9cc24c989f9c9f`. Optimized native ABBA, RocksDB WAL-no-fsync and mimalloc, three samples per arm:

| Case | Parent A / B medians | Candidate A / B medians |
|---|---:|---:|
| Cold load | 11.30 / 10.87 s | 10.92 / 10.78 s |
| Sequential inserts | 1.381 / 1.416 s | 1.401 / 1.397 s |
| Sequential updates | 1.997 / 2.040 s | 2.011 / 2.011 s |
| Batch update | 168.9 / 171.6 ms | 167.7 / 168.1 ms |
| Reopen | 117.3 / 116.2 ms | 115.1 / 114.4 ms |

Cold improves 3.36% in the first pair but only 0.83% in reverse; first-baseline system CPU is again high. Sequential cases straddle no change. Cold process user CPU is 29.94/29.72 → 29.57/29.46 s. No reliable peak-memory claim.

CodSpeed run `6aa6560c5c9742a4cd5a290b`, exact source verified through GQL:

| Case | Parent → candidate minimum | Change |
|---|---:|---:|
| Cold load | 54.1307 → 53.8737 s | −0.47% |
| Sequential inserts | 8.8827 → 8.8348 s | −0.54% |
| Sequential updates | 12.0355 → 11.9312 s | −0.87% |
| Batch update | 1.1024 → 1.0724 s | −2.72% |
| Reopen | 0.6519 → 0.6308 s | −3.24% |

Cold median 54.2550 → 53.8757 s (−0.70%); update median 12.0797 → 11.9673 s (−0.93%). Three rounds for cold/sequential, five for batch/reopen, no outliers. This corroborates a modest benefit, not the original hoped-for cold gain. Full GQL cold/update graphs are saved as results `6aa658d5ad9a6239bf6b89d9` / `6aa658d5ad9a6239bf6b89d5`; further scope attribution is being checked.

Validation: 2,038 Jazz library tests passed, 2 ignored; all three incremental canaries passed; authoritative hosted correctness CI is green. The new layout test fails on the unchanged parent (key/event 968/1080 bytes) and passes at 96/160 bytes, while ownership/coalescing tests establish value equality and unique consumption. Those sizes do not prove end-to-end performance. The canceled first optional shadow run lacked receipts; its unchanged replacement passed both shards and the aggregate. Existing behavior assertions and codecs are unchanged.

Source/hash files, all eight raw native logs and full hosted distributions are in `/tmp/jazz-shared-net-identities-xOJKF6`.

Tooling-friction: reversed native order matters; the first cold baseline alone would have overstated this small win by roughly fourfold.

PR #2923, opened 2026-09-13
perf(sync): bulk-copy immutable version row wire bytes

aeplay:
🤖 Bulk-copy immutable version row bytes without changing the wire contract.

## Thesis and scope

Advances #2913. The latest complete permissioned cold-load flamegraph (source `13e96f8b`, result `6aa61a88ad9a6239bf6ad0d7`) attributes 2.986s to per-byte row serialization and 0.657s to the intermediate per-byte receive vector out of 65.743s total. Their combined ~5.5% direct ceiling motivates a 3–5% end-to-end cold-load reduction prediction, not a measured claim. Inserts/updates exercise the same codec; their effect will be measured independently.

Preflight: inspected the rejected-experiment ledger, preserved history, and open/closed PR descriptions. The earlier canonical lens byte codec is a precedent, not this row-codec experiment. Rejected witness metadata encoding (#2830) targets different work; no repeated rejected mechanism is proposed here.

## Before / after

Previously a JVRR v1 envelope was serialized as a `Vec<u8>` sequence, pushing each byte through Serde/Postcard separately. Receiving likewise built an intermediate vector one byte at a time before decoding and copying the owned row payload.

Now serialization uses `serialize_bytes` to bulk-extend the output. Deserialization reads the available byte slice directly through the unchanged JVRR decoder. The final owned record still owns its payload. JSON retains its array representation through a sequence visitor without trusting allocation hints.

## Invariants, examples, and non-goals

- Canonical Postcard unsigned-varint byte length followed by the exact JVRR v1 envelope is unchanged. No protocol version bump, descriptor format change, schema translation change, or storage migration.
- An old sender's length-prefixed row followed by another field decodes identically; the new sender emits byte-for-byte the old sequence encoding. Tests cover envelope sizes 127/128 and 16383/16384, plus a following integer sentinel.
- JSON arrays round-trip unchanged; truncated envelopes, unknown JVRR versions, oversized descriptor lengths, and out-of-range JSON bytes still fail.
- Raw row contents remain the responsibility of receipt admission. The representation codec intentionally preserves malformed row payloads just as before; all canonical-message re-encoding, descriptor proof validation, logical-message bounds, and receipt/schema validation remain intact.
- This does not eliminate envelope construction, share arbitrary record ownership, narrow immutable version payloads, change policy evaluation, or optimize join/window maintenance.

[...]

PR #2934, opened 2026-09-13
perf: retain versions directly in transaction order

aeplay:
🤖 Trial for #2913, stacked on #2933. Draft: do not merge before morning approval. No adopter testing requested yet.

## Before / after

Retained versions lived in a global identity map plus a transaction → sort key → identity-set index. The sort key contains the complete identity, so each innermost set is a singleton. Consumers looked up the transaction, traversed those sets and searched the global map again.

Use one transaction → ordered key → weighted shared payload index. Count retained entries and their existing conservative footprint model incrementally, so reading that diagnostic does not walk this index. Membership existence checks borrow its rows rather than cloning the complete transaction. Other callers keep their owned result vectors; selected deletion witnesses remain part of the candidate union.

- Two different rows in one transaction remain separately retained and ordered by table, row UUID, layer and full raw bytes. Distinct versions of the same row/transaction/layer remain distinct because the entire raw record is in the key.
- A +2 then −1 keeps weight 1. A later −2 removes the version and prunes an empty transaction bucket; a subsequent +1 starts from zero. This preserves the existing index's positive-only retention, not the separate source-fact role map's signed-state rules.
- Positive-to-positive updates retain the latest prepared payload and update the exact diagnostic contribution. Keys still compare values, not Arc addresses or hashes.
- Shared version/replacement payloads retain independent role lifetimes. Removing one role cannot mutate or remove the other. Cloning the index gives an independent container snapshot with shared immutable payloads.
- Transaction and row identity derive from canonical immutable version bytes and the existing stable node-alias mapping; no new user-provided freshness or ordering assertion is trusted.
- The borrowed membership predicate includes selected deletion witnesses exactly as before; an existence test does not require deduplication or owned rows. Its replacement fallback is unchanged.

No storage/wire encoding, public type, permission semantics, completeness checks, transaction-wide wire manifests or durable serializer contract changes. The footprint remains the existing conservative referenced-byte model, not a deduplicated allocator measurement. Other footprint components are unchanged.

## Evidence / prediction

Latest full #2933 GQL cold profile attributes 0.848 s sampled CPU to the union of WeightedVersionIndex scopes (~1.6% total); todo has no sampled work in that index. Predict 0.5–1.2% lower cold latency, flat todo. This deliberately does not claim the entire todo footprint budget. Preflight read rejected-experiments.md and preserved branch/PR descriptions: #2874 shared key bytes but retained both indexes; #2923 shared role payloads. This removes containers and lookups while preserving both ownership improvements.

## WIP validation

New internal differential test compares 600 signed mutations with the prior identity-map semantics, checking exact transaction order, retained weights, Arc payload identity, empty-bucket pruning and a full recomputation of cached diagnostics after every step. Internal coverage is necessary for ownership and accounting, which public row equality cannot establish. Existing private ownership-test map-access paths are mechanically adapted; every value/lifetime expectation is preserved. Existing public/integration suites remain unchanged. Focused checks and native/hosted timings are in progress; no independent review or full local CI-equivalent receipt claimed.


aeplay:
🤖 #2934 native checkpoint: cold-load improvement, with a small insert slowdown. No adopter testing requested yet; no merge before morning approval.

Exact head `39d0b85d7651b9a712f6583c77ff7a705061965b`, parent #2933 `85d772b2173d04c5023ca63de8b67ba2a5cb449d`. Quiet ABBA, same RocksDB WAL-no-fsync/perf/mimalloc native fixtures, three samples per arm. All source hashes, binary hashes, raw logs and parsed summary are in `/tmp/jazz-single-version-index-TmVMKB`.

|---|---:|---:|---:|
| Permissioned cold load | 11.12 / 10.97 s | 10.61 / 10.74 s | −4.59% / −2.10% |
| Sequential inserts | 1.386 / 1.393 s | 1.402 / 1.401 s | +1.15% / +0.57% |
| Sequential updates | 2.002 / 2.011 s | 2.007 / 2.013 s | +0.25% / +0.10% |
| Batch update | 168.1 / 167.3 ms | 167.7 / 170.0 ms | −0.24% / +1.61% |
| Reopen | 115.3 / 114.6 ms | 115.5 / 114.9 ms | +0.17% / +0.26% |

Cold user CPU: 29.82/29.59 → 29.01/29.07 s. First baseline again has higher system CPU; use ~2% as the conservative local observation, not ~4.6%. This exceeds the 0.5–1.2% prediction locally, but the profile's sampled index budget was only ~1.6%, so do not promote the larger result to a confirmed portable claim. Hosted comparison/full GQL will test that. No peak-memory claim.

All 2,039 Jazz library tests and all three incremental canaries pass, including the new 600-mutation differential/footprint oracle and preserved shared-role ownership assertions. The first compile caught a new test fixture's integer-width mismatch; fixed before this exact commit and final passing tests. Existing behavior assertions were not weakened.

The first optional shadow workflow was canceled before its shard receipts existed and its aggregate failed with expected 2 receipts, found 0. The unchanged replacement is running; authoritative CI has no failure observed. No local CI-equivalent or independent review claim.

Next trial is the earlier pipeline stage: borrow physical supporting manifests through delta comparison rather than constructing unchanged owned facts. CodSpeed confirmation of this index trial continues in parallel.

Tooling-friction: keep resource counters and reverse ordering; cold system-time drift can make a small structural win look twice as large.

aeplay:
🤖 #2913 / #2934 hosted outcome: retain the single transaction-ordered version index. No adopter testing requested yet; no merge before morning approval.

Exact parent #2933 `85d772b2173d04c5023ca63de8b67ba2a5cb449d`, run `6aa6560c5c9742a4cd5a290b`; head `39d0b85d7651b9a712f6583c77ff7a705061965b`, run `6aa65ce75c9742a4cd5a2998`. GQL resolves the actual head SHA, not MCP's temporary synthetic merge hash. Same maintained RocksDB WAL-no-fsync fixtures, hosted wall time.

| Benchmark | Parent minimum | Head minimum | Change |
|---|---:|---:|---:|
| Permissioned cold load, 27,518 rows | 53.873682 s | 52.806793 s | -1.98% |
| Sequential 1,350 inserts | 8.834771 s | 8.781234 s | -0.61% |
| Sequential 1,350 updates | 11.931208 s | 11.915152 s | -0.13% |
| Batch 1,350 updates | 1.072372 s | 1.082319 s | +0.93% |
| Reopen 1,500 rows | 0.630799 s | 0.633500 s | +0.43% |

Cold median 53.875728 → 53.078954 s (-1.48%), three rounds, no flagged outliers. Updates are flat (head median 11.971075 s). Insert minimum improves but the head's 9.278326 s median is noticeably slower; do not claim an insert win. Batch/reopen show small adverse movement, including a broad reopen range 0.633500–0.681399 s across five rounds. Every result reports a graph and no instrument issue.

Full GQL graphs downloaded: cold result `6aa65f94ad9a6239bf6ba0f8`, update `6aa65f92ad9a6239bf6ba0d6`. The union of WeightedVersionIndex scopes falls from 0.847543 to 0.188231 sampled CPU seconds; `versions_by_tx` falls 0.439987 → 0.098629 s, `apply_delta` 0.387161 → 0.074223 s (these scopes must not be added to their union). This directly supports the intended mechanism: no second global index/singleton sets, direct borrowed transaction membership, and cached accounting. It explains a substantial portion, not all, of the 1.067 s wall-minimum change; no claim that reduced allocation requests alone prove the rest.

The hosted cold gain exceeds the original 0.5–1.2% prediction and aligns with the conservative ~2% local reverse-pair receipt. Sequential operations remain effectively outside the targeted mechanism. Correctness CI is green; local Jazz 2,039 passed/2 ignored, all three canaries passed, signed index/accounting oracle passed. The initially canceled optional shadow invocation has a successful unchanged replacement, as previously recorded.

Next-work confirmation: this newest update graph still spends 0.714142 s in maintained-view footprint refresh, including 0.205617 s in ReplacementIndex accounting. The active #2913 footprint trial targets that separate retained-result/replacement scanning cost, not another version-index rewrite.

Receipts: `/tmp/jazz-single-version-index-TmVMKB/hosted-metrics.json`, `gql-runs.json`, raw `2934-{cold,update}.json.gz`, scoped analysis, source/binary hashes and native logs. Tooling-friction: pin result IDs as well as run IDs; single-job reruns can replace results in place.

PR #2935, opened 2026-09-13
perf: borrow complete manifests through receiver delta comparison

aeplay:
🤖 Trial for #2913, stacked on #2934. Draft: no merge before morning approval; no adopter testing requested yet.

## Before / after

Every complete supporting manifest was sorted/validated and then expanded into an owned Vec<ProgramFactEntry>, even when a successor changed only one row. Comparison borrowed the predecessor but still rebuilt all current facts. Initial resets cloned that expanded vector again for additions.

Retain the validated physical rows and ordered compiler-source/range mappings for this receive batch. Two borrowed fact views represent source coverage and covered inputs in exactly the existing order. Compare those views against the preceding compact snapshot or borrowed settled state; materialize only additions/removals. An initial reset materializes one owned fact list. Full manifest parsing/sorting and O(scope) comparison remain; this is not a delta wire protocol.

## Invariants and worked cases

- Unchanged 1,500-row snapshot: no owned current facts are constructed by normalization. One changed physical version in a two-role self-join: two exact old inputs are removed and two new inputs added, with no unrelated role-path/version clones.
- The order remains coverage before covered inputs, then the same full source, completeness, authored-table, row and version fields as ProgramFactEntry. A retained complete=false coverage entry differs from incoming true coverage; no hash/prefix/physical-table-only equality is substituted.
- Multiple snapshots for one authority usage compare against the immediately preceding snapshot in the same receive frame. Replacement → empty → refill → replay retains the same final facts as before. A claimed empty predecessor remains distinct from initial/pending state.
- Unsorted valid rows are accepted. Duplicate, invalid or out-of-dataset physical rows still reject before any normalization delta is published. Source mappings remain compiler/catalogue-owned, not trusted from the sender. All later body/completeness/atomic-batch validation is unchanged.
- Receiver-local non-peer facts are excluded from removals. Selected deletion rows, exact version discriminators and repeated source roles remain represented. The compact predecessor is temporary batch-local state; failed admission cannot install it.

No public types, durable storage/wire encodings, serializers, permissions, fixture workloads or wire completeness guarantees change. The general ProgramFactEntry enum and its serialization remain untouched.

## Prediction and preflight

[...]

PR #2936, opened 2026-09-13
perf: maintain hot retained-view footprint totals incrementally

aeplay:
🤖 Trial for #2913, stacked on #2935. Draft: no merge before morning approval; no adopter testing requested yet.

## Before / after

Every eager maintained-view footprint refresh scanned retained result weights, current/published payloads, published membership, the touched-member journal, and replacement versions. A one-row update therefore revisited many unchanged rows just to report the current diagnostic.

Private map/set wrappers now update their exact referenced-byte totals alongside insertion/removal/replacement. Result weights maintain their positive-entry count. ReplacementIndex retains the same nested maps and positive-only semantics but tracks entry/key counts and bytes alongside mutations. Footprint refresh still runs at the same points and returns current values, without these full-map scans. Read-only Deref exposes queries but not unchecked mutation.

## Invariants and examples

- Updating one payload in a 1,500-row view subtracts the replaced value's modeled size and adds the new value's size; the retained equal key is counted once. Published/current copies remain separately counted as before, even if their underlying data shares ownership.
- A weight crossing +1 → -1 decreases the positive-row count while retaining the negative weight entry. Removing the entry subtracts its key/value bytes. Zero-weight behavior at call sites is unchanged.
- Replacement weights retain only positive entries, exactly as before. Removing the last version prunes the outer row key and subtracts its bytes; a later positive delta revives from zero, not a discarded negative. Equal-identity updates retain the latest payload as before.
- Taking/resetting a touched-member journal installs fresh counters. The witness-dependent full reconciliation fallback may still rebuild its complete set and computes that set's total once. Cloned views own independent containers/counters.
- Preserve the existing conservative referenced-byte model exactly, including the published-set's old map-overhead convention. This is not a deduplicated allocator/RSS estimate. Structured records, selected deletion witnesses, source journals and witness-table accounting are unchanged.

No public types, row results, permission/admission behavior, wire/storage encoding, serializer or benchmark fixtures/assertions change. Diagnostics are neither delayed nor sampled.

## Thesis and validation

The #2933 full GQL update profile puts 0.675 s / 11.882 s (~5.7%) under eager footprint refresh; only part is removable here. Prediction: 3–5% lower sequential updates, 2–4% lower sequential inserts, roughly flat cold load with possible small mutation-bookkeeping tradeoff. Read rejected-experiments.md and searched branches/PRs: retain #2874's conservative accounting model; extend the separate #2934 index-accounting idea to the result/replacement components actually hot on this workload.

New internal 600-mutation oracles cover signed result weights, payload resizing, membership/journal reset, replacement pruning/revival, value semantics and clone isolation. Internal tests are needed because query rows cannot expose exact private accounting. Every test-mode view footprint refresh also recomputes the old formulas and checks equality. Both focused tests pass; full Jazz library/canaries and CI are in progress. Existing assertions are unchanged. Native ABBA and CodSpeed will decide the outcome; no full local CI-equivalent or independent-review claim.

Tooling-friction: retaining the previous full-scan model behind cfg(test) makes counter drift directly detectable.


aeplay:
🤖 #2913 / #2936 first complete native outcome: incremental footprint accounting improves sequential operations about 2–2.6%; hosted confirmation pending. No adopter testing requested yet; no merge before morning approval.

Parent #2935 `8faeb3fbb713f906fdf73a7da574c9868707a71c`; candidate `2fff17257c8d9275b703436a8c1e8e98b6d039e5`. Native profile.perf + mimalloc, identical maintained RocksDB WAL-no-fsync workload/binaries, sequential ABBA with three samples per case per arm; no overlapping local builds/tests/profile parsing. Values below are the harness's printed medians.

| Case | Parent A / B | Head A / B | Paired change |
|---|---:|---:|---:|
| 1,350 sequential inserts | 1.353 / 1.345 s | 1.323 / 1.310 s | -2.22% / -2.60% |
| 1,350 sequential updates | 1.913 / 1.892 s | 1.867 / 1.852 s | -2.40% / -2.11% |
| 27,518-row permissioned cold load | 11.00 / 10.70 s | 10.80 / 10.87 s | -1.82% / +1.59% |
| 1,350 batch updates | 169.8 / 170.2 ms | 168.6 / 168.0 ms | -0.71% / -1.29% |
| 1,500-row reopen | 115.0 / 117.1 ms | 115.4 / 115.6 ms | +0.35% / -1.28% |

Inserts match the 2–4% prediction's lower end; updates are below the 3–5% prediction. Cold/reopen are inconclusive, not improvements. Todo process user CPU falls 10.555/10.402 → 10.313/10.160 s; cold remains near 29.1–29.6 s. No RSS claim from these arms. The newest pre-trial hosted graph confirms footprint remains ~0.714 s of update CPU, but only the full hosted outcome decides whether that saving survives the extra mutation bookkeeping.

Both new 600-mutation accounting oracles passed. Full local Jazz library: 2,042 passed, 2 ignored; all three incremental-delivery canaries passed. Clippy, formatting and sensitive-data guard passed at commit. Correctness CI is progressing without a reported failed check; native timing is not a full local CI-equivalent result. All existing correctness assertions remain unchanged; test-mode footprint checks validate the original full-scan accounting formulas.

Source/binary hashes, eight raw logs/resource footers, `native-summary.json`, focused tests, broad tests and canaries are in `/tmp/jazz-incremental-footprint-u2pZND`. Tooling-friction: the automated receipt summarizer rejects incomplete arms and avoids hand-copying timing units.

aeplay:
🤖 #2936 first hosted receipt: stronger write win, unresolved cold-load tradeoff

Pinned source `2fff17257c8d9275b703436a8c1e8e98b6d039e5`, CodSpeed run `6aa665aa5c9742a4cd5a2a20`, compared with exact parent #2935 `8faeb3fbb713f906fdf73a7da574c9868707a71c`.

| Workload | Parent → head minimum | Change |
|---|---:|---:|
| Sequential inserts | 8.398206 → 7.935695 s | −5.51% |
| Sequential updates | 11.141076 → 10.387811 s | −6.76% |
| Permissioned cold load | 52.796217 → 54.670029 s | **+3.55%** |
| Batch updates | 1.073705 → 1.069946 s | −0.35% |
| Reopen | 0.631520 → 0.661513 s | +4.75% |

Cold median also worsened, 52.903219 → 54.977958 s (+3.92%). Reopen's distribution is broad (maximum 1.548837 s); no outlier was flagged. These are not enough grounds to dismiss either adverse result as noise. All correctness and benchmark CI checks passed.

Downloaded the complete GQL walltime graphs before reruns: cold result `6aa6685fad9a6239bf6bd4ed`, update result `6aa6685fad9a6239bf6bd4f4`. Update footprint accounting drops from approximately 0.714 s to 0.170 s sampled CPU, supporting the intended mechanism and a substantial portion of the 0.753 s walltime reduction. Sampled CPU scopes are not interchangeable with minimum wall time.

The local ABBA had smaller write wins and inconclusive cold movement. I am requesting unchanged cold reruns of both parent and candidate to distinguish a real mutation-bookkeeping tradeoff from run variance. GitHub's first job-rerun requests returned HTTP 502/500; the originals and full graphs are preserved and requests will be retried. Until confirmation this is a **write-positive, cold-unresolved trial**, not an unconditional recommendation.

No adopter testing requested yet; no merge before morning approval.

Tooling-friction: single-job reruns replace result IDs inside an existing CodSpeed run; preserve exact IDs and full graphs before requesting them.

aeplay:
🤖 #2936 unchanged-source confirmation: the cold-load regression is real enough to treat as a tradeoff, not dismiss as noise. No adopter testing requested yet; no merge before morning approval.

The original GH job rerun ended in an inconsistent queued/completed API state, so I dispatched the same benchmark workflow on the unchanged branch. Fresh GH run 34750165412 / CodSpeed `6aa672c35c9742a4cd5a2aa5` is bound by GQL to `2fff17257c8d9275b703436a8c1e8e98b6d039e5`. Full cold graph result `6aa6757ead9a6239bf6bfffa` was downloaded and preserved.

| Unchanged cold confirmation | Minimum | Median | Maximum |
|---|---:|---:|---:|
| Parent #2935, source 8faeb3fbb7 | 52.672406 s | 53.222013 s | 53.290209 s |
| #2936, source 2fff17257c | 54.603915 s | 54.795748 s | 54.852060 s |

Three rounds each, no flagged outliers. This is +3.67% minimum / +2.96% median cold time, consistent with the original +3.55% / +3.92%. #2935's rerun baseline itself stayed close to its original (minimum -0.23%, median +0.60%). The initial #2936 write wins remain useful, and the fresh all-job dispatch also reports insert 8.014423 s and update 10.469480 s, still below #2935's original 8.398206 / 11.141076 s. But this is not a free cross-workload improvement.

Interpretation: maintaining exact footprint totals pays mutation-time bookkeeping during bulk cold hydration, while avoiding repeated retained-state scans benefits sequential writes. This is a plausible mechanism, not yet a causal breakdown of the whole regression. Morning decision: keep as a conscious write/cold tradeoff, improve the bulk path, or exclude it. The stack currently retains it so each subsequent trial has a precise parent.

This confirmation also removes the suspected slow-parent caveat from #2937's empty-history-head proof: #2937's 48.126185 s is 11.86% faster than this repeated #2936 minimum, close to its original 11.97% comparison. The strong #2937 cold win is not explained by a one-off parent run.

Receipts: `/tmp/jazz-incremental-footprint-u2pZND/{confirmation-metrics.json,2936-cold-confirmation.json.gz}`; parent `/tmp/jazz-borrowed-manifest-delta-O6saG2/{cold-confirmation-metrics.json,gql-runs-confirmation.json,2935-cold-confirmation.json.gz}`. Original receipts remain intact.

Tooling-friction: fresh dispatch recovered an orphaned job-rerun state; immutable copies of original metrics/graphs were essential because successful job reruns replace result IDs within a CodSpeed run.

PR #2938, opened 2026-09-13
perf: scan pending constraints once per complete parent batch

aeplay:
🤖 Draft performance trial for #2913, stacked on #2937. No merge before morning approval; no adopter testing requested yet.

## Before / after

Bulk complete-parent preflight scanned the complete staged pending-edge table once per parent. It now scans once, groups owned raw constraint records by the admitted parent, then invokes the existing child-fate and exact-coordinate validation in the original parent order. Empty tables return immediately. The ordinary single-parent entry point still performs one scan and uses the same validator.

- A cold batch of 1,024 parents with no pending constraints makes one storage range read, rather than 1,024. An empty parent batch performs no scan.
- Two parents with Accepted partial children: matching first-parent constraints are staged for deletion; if the second parent's complete body contradicts its child's exact row/branch/layer constraint, the entire caller batch fails before any parent persists. The earlier deletes are not published.
- A constraint deleted earlier in the same open batch is absent from the discovery snapshot. Unrelated parent constraints remain untouched. Pending children keep their constraints for post-persistence fate settlement.
- A repeated parent cannot acquire new constraints inside this read/delete-only boundary. Its first pass already removed matching Accepted constraints; Pending-child validation makes no mutation. Subsequent occurrences therefore need no second group evaluation.
- Raw record ownership moves into the groups without copying encoded payloads. Parent aliases are validated during discovery, as all rows were inspected by the old initial scan. Child and coordinate decoding stays deferred until that parent's turn. Corrupt storage remains a typed error; no claim is made that multiple simultaneous corruptions have a stable diagnostic precedence.
- There is no durable schema, storage byte encoding, wire protocol, permission, conflict resolution or result change. In particular, the existing pending-edge key/record encoder is unchanged. No serializer defaults are introduced.

## Evidence / checks

The full parent #2935 cold graph attributes ~0.712 s sampled CPU to preflight. Prediction: 0.5–1.3% cold improvement, approximately flat sequential todo cases. Its single-parent update scan has a separate ~1.271 s cost that this trial does not claim to remove.

Read the rejected-experiment ledger and searched preserved parent/preflight branches and PR descriptions. Prior parent-settlement work optimized the post-commit phase; #2878 removed bundle clones. This is the pre-persistence staged scan.

New tests pass for the 0/1/1,024-parent work bound and accepted-child atomicity, duplicate parents and staged deletes. These tests intentionally use the internal receiver seam because public result rows do not expose uncommitted constraint state or scan counts. Existing receiver-level atomicity and settlement tests remain unchanged. Runtime compilation passed; broad Jazz, three incremental-delivery canaries, actual correctness CI and native/hosted timing are next. No full local CI-equivalent or independent review is claimed.

Tooling-friction: separating preflight from settlement in full profiles prevents repeating an already completed optimization.


aeplay:
🤖 #2938 local receipt — single-snapshot bulk preflight

PR creation recovered after transient GitHub server failures: #2938. Exact parent `7f472c782daae36ad33c1e9ae543b02d387fd66e`, candidate `c39c2ee25a` (full source/hash receipt preserved). Quiet native ABBA, profile.perf + mimalloc, three samples per arm, unchanged RocksDB WAL-no-fsync workloads.

| Workload | Parent medians A / B | Candidate medians A / B | Paired change |
|---|---:|---:|---:|
| Permissioned cold | 10.63 / 10.17 s | 10.22 / 9.978 s | −3.86% / −1.89% |
| Sequential inserts | 1.297 / 1.316 s | 1.312 / 1.330 s | +1.16% / +1.06% |
| Sequential updates | 1.847 / 1.869 s | 1.875 / 1.863 s | +1.52% / −0.32% |
| Batch updates | 166.1 / 167.8 ms | 167.5 / 168.8 ms | +0.84% / +0.60% |
| Reopen | 114.7 / 115.4 ms | 115.6 / 114.8 ms | +0.78% / −0.52% |

Conservative local cold effect ~1.9%; the first baseline is slower and does not justify a 3.9% claim. This exceeds the 0.5–1.3% prediction locally, but hosted confirmation is required. Inserts have a consistent ~1.1% adverse local movement; other todo cases are not proven improvements.

Jazz 2,046 passed / 2 ignored, all three incremental-delivery canaries passed, new focused tests passed, runtime compilation and pre-commit Clippy/format/guard passed. Initial new-test compilation used an encoded key where the public batch API requires a typed key; the fixture now uses the existing `pending_edge_primary_key` codec. No existing assertion changed.

New research from the full #2937 update graph: **preflight 1.257105 s + post-ingest pending-child settlement 1.149783 s** are separate scopes, ~23% of the 10.46 s result. A durable by-parent index would need old-store backfill and stale-entry cleanup. Also, `index_scan_raw_in_batch` currently stages primary writes without synthesizing secondary-index changes; it has no production callers, but must not be adopted as a read-your-writes index shortcut without a correctness test/fix. Keep that API follow-up open.

A simpler potential next proof: track a conservative upper bound on every referenced pending-parent TxTime, including Accepted partial children. A newly completed parent strictly above that bound cannot have a stored constraint. Advancing the bound before staging, never lowering it on delete/failure, and recovering it from all durable edges could skip both scans without a new index or storage format. This needs explicit recovery/unknown-state and staged-write coverage before becoming a trial.

No adopter testing requested yet; no merge before morning approval.

Tooling-friction: saved native binaries allowed quiet ABBA despite GitHub PR-creation failures; no timing was taken while local builds or graph parsing were active.

aeplay:
🤖 #2938 hosted receipt — cold gain confirmed, small todo tradeoff visible

Pinned GQL source `c39c2ee25accb79b6c68d88b04f9d17b643cca98`, run `6aa670a8e6a06de818a061e8`, exact parent #2937 run `6aa669fde6a06de818a06174`.

| Workload | Parent → head minimum | Change |
|---|---:|---:|
| Permissioned cold | 48.126185 → **47.642391 s** | **−1.01%** |
| Sequential inserts | 7.984448 → 8.072462 s | +1.10% |
| Sequential updates | 10.461503 → 10.608037 s | +1.40% |
| Batch updates | 1.075388 → 1.075777 s | +0.04% |
| Reopen | 0.640853 → 0.662904 s | +3.44% |

Cold median 48.254964 → 47.699227 s (−1.15%), three rounds, no outliers. This matches the 0.5–1.3% prediction, below the conservative ~1.9% local pair. Insert/update slowdowns are visible, not dismissed. Reopen has a broad distribution (maximum 1.557221 s); no dependable conclusion from this small case alone.

Full walltime GQL graphs downloaded: cold `6aa67355ad9a6239bf6bf4ea`, update `6aa67355ad9a6239bf6bf4ed`. The cold graph has no samples under the named `preflight_complete_parent` scope after removing the per-parent loop. The deterministic 1/1,024-parent test independently proves the one-scan bound; absence of samples is not a claim of zero runtime.

All correctness checks are green. Keep as a cold-positive tradeoff pending stack review; #2939 separately avoids the sequential path's repeated preflight/settlement work and has a large local update win, but its hosted result is not yet confirmed. No merges.

New cold profile attribution for further research: `build_join_delta_index` 2.153795 s sampled CPU, including `JoinBucket::commit_overlay` 1.138415 s and `add_weight` 0.538281 s (nested, do not add them to the parent). This is still meaningful after the previous exact-snapshot reuse optimization.

No adopter testing requested yet; no merge before morning approval.

Tooling-friction: full graph downloads and pinned SHA/result IDs let small tradeoffs remain visible while the next PR runs.

PR #2941, opened 2026-09-13
perf: reuse singleton root hydration seeds for snapshots

aeplay:
🤖 Draft performance trial for #2913, stacked on #2940. No adopter testing requested yet; no merge before morning approval.

## Before / after

Root collector hydration rendered terminal Insert seeds, then built a second join arrangement and rendered the same groups again for its relational snapshot. Now the relational snapshot can reuse the already-rendered seeds when each freshly hydrated group has one distinct positive input record. The retained collector state and terminal output are still installed normally; the redundant collector arrangement is not needed by later Tick-mode terminal updates.

## Invariants and examples

- Rows with IDs 2, 3 and 8 hydrate once. Terminal Insert indexes still address the evolving public sequence (for descending order, successive insert-at-zero operations are valid). The relational snapshot remains in encoded group-key order, independently of terminal rank order.
- Two positive copies of the same input record remain one positive root. Two distinct records in one group fail the reuse proof: the old hash-arrangement and ordered-collector paths can choose different representatives, so they must not be silently equated.
- The shortcut requires Hydrate, Root mode, strictly positive input deltas, singleton retained groups and complete root Insert operations. Signed/zero inputs, non-root Collect/Expand, ambiguous groups and incremental Tick updates retain the old paths. Existing graph-input arity validation still runs before returning the reused snapshot.
- Hydrate removes any old collector operator before building these seeds, including shared-subscription reattachment. No incremental subset is mistaken for a complete snapshot. Retained state still supports later Updates/Removes; discarded speculative work does not mutate a predecessor.
- Record values/descriptors are unchanged. The snapshot still owns its bytes separately from terminal seeds; this removes duplicate grouping and rendering, not all copying. No cross-binding cache, schema, permission, public API, storage/wire encoding or serializer change.

## Thesis / preflight

The full #2939 cold graph shows 2.623537 s in update_collect_by, including 1.547309 s in root terminal construction. Prediction: ~1–2% end-to-end cold improvement, possibly ~3% including temporary-index savings; todo flat/small benefit. Samples are a budget, not an end-to-end guarantee.

Rejected ledger read; preserved branches and open/closed PR descriptions searched. #2849 removed the duplicate render inside the terminal Insert/Update passes, not this separate relational hydration. Its negligible latency outcome is noted. This is not #2851's rejected encoded-field projection machinery. #1262 reuses source snapshots across equivalent terminals, a different boundary.

## Validation

[...]

PR #2940, opened 2026-09-13
perf: inline singleton records in shared join buckets

aeplay:
🤖 Draft trial for #2913, stacked on #2939. No merge before morning approval; no adopter testing requested yet.

## Before / after

Each join bucket used a shared hash map for its committed base and another for staged overrides. A singleton still allocated and hashed a collection in both phases. Now each existing shared allocation contains Empty, One(record, weight), or Many(hash map). A second distinct record promotes to the ordinary hash map; multi-record buckets retain the single-probe Entry update path. No sorting or automatic demotion is introduced.

## Invariants and worked cases

- Insert record A with weight +2: the override and committed base each remain inline singletons. Rc snapshots still share the base, and mutation still uses copy-on-write.
- Delete A with -2: the override stores an explicit tombstone, not a missing override. A later -1 starts from zero, not from the old base weight; negative multiset weights remain valid.
- Insert distinct B: promotion preserves A's signed weight and the exact encoded record keys. Previously cloned snapshots remain unchanged. Replacing A keeps its existing key just as HashMap does.
- Cancellation/drop discards the staged snapshot as before. Existing arrangement reuse and snapshot pointer-identity proofs remain unchanged.

This is a private in-memory representation change only: no schema, durable/wire encoding, permission, public API, or benchmark assertion changes. Rc allocations and the outer join-key map remain; this does not claim allocation-free joins. No generic replacement of other maps or universal record sharing.

## Thesis and evidence

The full #2938 cold graph attributes 2.153795 s to join-index construction, including 1.138415 s in commit_overlay and 0.538281 s in add_weight (nested, not additive). Prediction: 1–3% end-to-end cold improvement and flat/small todo effects. Existing singleton-heavy workload evidence motivates the representation; actual comparable native/hosted timings decide retention.

Preflight: read the rejected-experiment ledger and searched preserved branches plus open/closed PR descriptions. #2901 used a heap-allocated sorted Vec for the immutable base and left the hash-map overlay intact; its ~1.4% local cold change was inconclusive. This trial changes that premise by removing singleton collection allocation and hashing in both base and overlay while preserving the existing sharing boundary. It does not repeat #2864's direct fresh-base build.

Validation so far: runtime check and all 767 Groove tests pass (2 ignored), including two new private representation/signed-map oracles across promotion, tombstones, negative weights and retained snapshots. Existing assertions are unchanged. Broad Jazz/canaries and actual CI follow; native ABBA and CodSpeed pending. No full local CI-equivalent or independent-review claim.

[...]

PR #2939, opened 2026-09-13
perf: prove newer parents have no waiting constraints

aeplay:
🤖 Draft performance trial for #2913, stacked on #2938. No merge before morning approval; no adopter testing requested yet.

## Before / after

Every complete transaction scanned pending-parent constraints before ingest, then scanned them again during later fate settlement. Those constraints often refer only to older transactions. Now a local conservative timestamp ceiling can prove that a strictly newer completed parent has no waiting constraint, avoiding both scans.

`RejectionTracking` holds Unknown, Empty, or Through(max referenced parent TxTime). Startup remains Unknown until recovery has validated every durable pending-edge record. A shared staging helper advances the ceiling before every local or receiver constraint insert/update, including constraints owned by already-Accepted partial children. Deletes and abandoned/failed batches never lower it. Database-slot rebuilds retain it; full reopen reconstructs it from durable rows.

## Invariants / worked cases

- Stored constraints reference parents at times 10 and 20. Completed parent time 21 cannot match either, so both single-parent preflight and settlement check database health and return without scanning. This does not infer freshness from a peer claim or require monotone arrival/clocks.
- Parent time 20, even from a different node, does not satisfy the strict inequality. Parent time 19 likewise falls back. The old exact row/branch/layer validation, immutable Accepted-child conflict and Pending-child settlement rules remain unchanged.
- An Accepted partial child references a missing parent at time 100. The ceiling includes 100 even though that child is absent from the Pending-only rejection graph. A contradictory complete parent at 100 is still rejected before it persists.
- Staging an edge to time `u64::MAX` advances the bound before any later read or suspension. A preflight in that same open batch sees it. If the batch is abandoned, the high bound merely disables skipping; it cannot create a false absence. Reopen tightens the bound back to the actual durable maximum. No increment/overflow arithmetic is needed.
- Unknown startup/recovery state never skips. Recovery continues validating every persisted alias and coordinate, including corrupt-state rejection. Poisoned storage still returns `DatabasePoisoned` through the shortcut.
- For post-ingest bulk settlement, all requested parents must be provably newer/absent before skipping the single scan. Mixed sets preserve the existing path. #2938's separate bulk preflight remains unchanged.

The ceiling is constant-size and local-only: no storage schema/index, durable encoding, serializer, wire protocol or permission change. The existing pending-edge codec remains the sole record/key encoding. No public API or benchmark assertion changes. The single-parent method becomes parent-module-visible for internal regression coverage, not publicly exported.

## Evidence / validation

[...]

PR #2924, opened 2026-09-13
perf: share proven witness payloads across maintained roles

aeplay:
🤖 Advances #2913: share compiler-proven witness payloads across independent maintained roles.

## Before / after

Version and replacement terminals can have identical payload graphs except for their event-role literal. Previously they executed and decoded independently, each reconstructing a row body and full-row identity, then retaining its own row and derived metadata.

The compiler now records a one-to-one sharing proof when the complete witness schemas (including source, branch fields and routing) match and the complete graphs differ only in the expected top-level role literal. Logical terminal contracts remain unchanged. Runtime opening executes only the version payload producer for each proven pair; its decoder applies both role consumers. Nonmatching terminals retain the ordinary execution path.

The two retained indexes share an immutable prepared payload and the exact full-row identity allocation. Their weights, lookup indexes, transaction membership and replacement-winner selection remain separate. This does not change all OwnedRecord instances or add a raw-byte decode cache.

## Invariants and examples

- A row present in both roles is decoded and prepared once, then retained by both indexes. Removing replacement presence must not remove version presence, and vice versa; the payload lives until its last consumer releases it.
- Paired +1/−1 deltas cancel within the shared event identity. Different roles on unshared paths retain independent signed weights; they are never summed into one permission fact.
- A replacement-only terminal is not dropped. Different source aliases, branch interpretation, routing fields, payload schema, filters or upstream graphs do not qualify merely because their row IDs match.
- Deletion witnesses preserve authoritative membership reconciliation; permission revocation does not become a tombstone.
- Wire/storage formats, full immutable payload requirements, canonical bytes, policies and public result assertions are unchanged. The sharing map and prepared payload are ephemeral runtime metadata, not durable identities.

Logical output metadata is deliberately preserved for compiler consumers; physical sink selection and maintained decoding use the same sharing proof. No general Groove API or serializer change is introduced.

## Thesis / preflight

The newest complete #2922 cold GQL graph has 2.370s witness decoding inside 6.747s decoded-event handling / 65.409s total. The full-design prediction is 3–6% lower cold load, not a measured result. Decoding alone has only ~1.8% direct headroom; the remaining prediction depends on eliminating duplicate construction and retention.

Read the rejection ledger and preserved #2896 implementation/description. Changed premise: remove duplicate payload production and retained ownership at their boundaries, not merely cache a second decode. The user's updated direction explicitly permits revisiting opportunities more broadly. #2848's universal record sharing and #2830's metadata-only encoder are not repeated here.

## Status

[...]

PR #2943, opened 2026-09-13
perf: borrow supporting-row table names during decode

aeplay:
🤖 Performance trial on #2942; thesis/outcome log #2913. Draft, no merge before morning approval. No adopter testing requested yet.

## Before / after

Before: SupportingRow deserialization builds an owned String and enters the global intern pool for every row, including repeated table names in a sorted manifest. After: its one field requests a borrowed string; an exact-equality, one-entry thread-local cache reuses the last permanent intern handle. On misses, borrowed input uses the existing `Intern::from_ref`; owned input reuses its allocation through `Intern::new`.

Prediction: approximately 1–3% sequential inserts/updates; cold potentially 0.2–1%. This is a measured trial, not an allocation-count performance claim. Local timings and CodSpeed confirmation will follow.

## Invariants and encoding review

- Serialization, field order, wire version and all durable bytes are unchanged. Existing postcard string framing remains unsigned length varint + UTF-8, with a pinned string corpus and old/new SupportingRow byte/decode comparisons.
- Exactly the old String visitor's UTF-8/byte validation is retained. Empty strings still decode, then fail the unchanged SupportingRow wire-validity check.
- Cached identity is exact string content, not a table/schema/authority decision. Physical table identity remains authoritative; all receiver manifest/permission/body validation is unchanged.
- Intern handles refer to permanent existing pool entries; the cache never retains an input-buffer reference and is bounded to one handle per thread. No schema lifetime invalidation is needed for immutable strings.
- No new dependency, public type, general interner replacement, wire compression, delta manifest or permission cache.

## Worked cases

- A 1,000-row single-table manifest consults the pool once, then compares the same name and reuses its handle without allocating owned strings during postcard decode.
- Alternating A/B/A names miss the single cache entry each time and use the canonical global pool; no local unbounded map grows.
- The same spelling in different schema/authorization scopes shares only an immutable string handle, exactly as the old global interner did.
- Unicode labels round-trip unchanged; invalid UTF-8 fails; dropping the wire buffer leaves all decoded handles valid. A different thread has its own one-entry cache and returns the same canonical handle.
- Truncated and trailing-byte behavior is compared with the previous field decoder; no stricter or more permissive framing is silently introduced.

## Validation

New internal wire corpus and cache-work tests are appropriate because byte compatibility and intern lookup counts are not observable through row queries. Existing public sync suites remain the user-visible correctness gate. Existing assertions are unchanged. Focused wire corpus passes; full Jazz suite, canaries, clippy/format/guard and actual CI are running/pending. No local CI-equivalent or independent-review claim.

Preflight read rejected experiments and searched preserved branches plus PR descriptions. #2101 stored descriptor caching is a different mechanism; this trial changes only repeated wire-label decoding.

Tooling-friction: borrowed lookup already exists in the current intern crate, avoiding a dependency or protocol change.


aeplay:
🤖 #2943 native outcome: updates improve 1.2–1.9%; other scenarios are inconclusive. No adopter testing requested yet; no merge before morning approval.

Exact head `91f8f5af305133c878229efc9f6702cf6f153878`, parent #2942 `0005a0a99a9d9b0f514b68815167ced9235221d3`. Quiet ABBA, three samples per arm/case, unchanged profile.perf + mimalloc + RocksDB WAL-no-fsync fixture and assertions.

| Scenario | Parent A / B medians | Head A / B medians | Paired change |
|---|---:|---:|---:|
| Permissioned cold | 9.684 / 9.315 s | 9.499 / 9.379 s | -1.91% / +0.69% |
| Sequential inserts | 1.260 / 1.277 s | 1.267 / 1.263 s | +0.56% / -1.10% |
| Sequential updates | 1.334 / 1.336 s | 1.309 / 1.320 s | -1.87% / -1.20% |
| Batch update | 162.9 / 166.0 ms | 164.0 / 164.6 ms | +0.68% / -0.84% |
| Reopen | 109.9 / 111.0 ms | 110.9 / 110.9 ms | +0.91% / -0.09% |

The 1–3% update prediction is supported locally; no cold or insert benefit claimed. The wire corpus confirms unchanged old/new postcard and JSON bytes/decoding, UTF-8 rejection, empty-label admission behavior, all truncated boundaries and trailing bytes. A 1,000-name decode makes one pool lookup and no owned-string visitor calls, while alternating names miss the bounded cache and cross-thread handles remain canonical. These work counts are mechanism receipts, not substitutes for timing.

All 2,055 Jazz library tests pass (2 ignored), all three scaling canaries pass, fmt/clippy/guard pass. Actual CI and CodSpeed continue. No existing assertions changed; no full local CI-equivalent or independent review claimed. Receipts `/tmp/jazz-borrowed-supporting-names-0c5863/{source.txt,binary-hashes.txt,native-summary.json,*-todo.log,*-cold.log,wire-test.log,cache-test.log,jazz-tests.log,canaries.log}`.

While hosted confirmation runs, a separate phase-scoped native cold profile is being prepared. The latest GQL graph has implausible broad inline ancestry (e.g. query execution beneath a derived claims clone), so those inclusive values are not reliable removable-time budgets. Preserve raw graphs and corroborate attribution with code plus non-inline/phase-scoped samples before choosing a large next lever.

Tooling-friction: byte compatibility was checked against the previous field decoder rather than changing the wire protocol to improve interning.

aeplay:
🤖 #2943 hosted confirmation: sequential updates improve 4.15%, inserts 2.45% by minimum; cold is 1.04% slower. No adopter testing requested yet; no merge before morning approval.

Exact head `91f8f5af305133c878229efc9f6702cf6f153878`, GQL-verified run `6aa68b2c5d98b71dc6a81bb5`; parent #2942 run `6aa685e66ad2c92994448ec0`. All actual correctness CI is green.

| Scenario | Parent minimum / median | Head minimum / median |
|---|---:|---:|
| Permissioned cold | 45.623192 / 45.856896 s | 46.096825 / 46.138589 s |
| Sequential inserts | 7.619886 / 7.622823 s | 7.433235 / 7.554566 s |
| Sequential updates | 7.682278 / 7.683915 s | 7.363233 / 7.368268 s |
| Batch update | 1.044439 / 1.054016 s | 1.049489 / 1.055807 s |
| Reopen | 0.614794 / 0.631513 s | 0.608937 / 0.630216 s |

Update rounds are tight (7.363–7.372 s) and improve 4.11% by median, exceeding the 1–3% prediction and the 1.2–1.9% local result. Insert rounds remain broad, median improvement only 0.90%; do not advertise the minimum alone as a stable 2.45% throughput gain. Cold is +1.04% minimum/+0.61% median; local cold was inconclusive. Batch +0.48% minimum/+0.17% median is near flat. Reopen remains noisy with one IQR outlier and a 1.515 s maximum. No other listed issue/outlier flags.

Downloaded full cold/update walltime GQL graphs (`6aa68dd7ad9a6239bf6c5f6f`, `6aa68dd7ad9a6239bf6c5f74`). Exact SupportingRow **Deserialize** inclusive time falls 0.675359→0.406553 s (0.269 s removed), consistent with most of the 0.319 s whole-update minimum difference. The new field decoder costs 0.085590 s inclusive. Correction to the earlier loose SupportingRow scope: that substring also included Serialize, so its 0.775→0.488 s union is not exclusively decode time; use the exact Deserialize frame for this thesis. No double-counting the nested visit_seq frame.

Recommendation: retain provisionally as a simple write-focused optimization, with the small cold adverse result explicit. This is not a cross-workload win. Final combined-stack timing will keep both cold and write targets visible. Raw distributions, byte-compatibility tests, exact hashes and complete profiles are preserved in `/tmp/jazz-borrowed-supporting-names-0c5863/`.

Tooling-friction: exact frame selection matters—broad substring filters can silently combine serialization and deserialization.

aeplay:
🤖 Overnight consolidation checkpoint, 05:18 PDT. No merge into main and no adopter testing requested yet.

The retained stack is now linear: #2925 → #2926 → #2927 → #2929 → #2930 → #2931 → #2932 → #2933 → #2934 → #2935 → #2936 → #2937 → #2938 → #2939 → #2940 → #2941 → #2942 → #2943. Current main remains `a68fcd455d7e479794cf889ada655b95fe56a830`; combined tip is `0487a56a67ffadc8197f4fe862ce3e0980ecdf8b`.

Closed/excluded #2928 (inconclusive after hosted repeat) and #2944 (updates slower, no established cold gain). #2945 is still a separate pending experiment against its original parent while its hosted receipt resolves; it is not yet in the retained stack. Every original PR tip is preserved locally under `preserve/overnight-20260913-prNNNN`; the original complete stack is also pushed as `preserve/overnight-20260913-pr2945`. Nothing was deleted or merged.

Clean-worktree preflight ran before restack/ref propagation. Removing #2928 caused one adjacent-test conflict in #2933: retained the event-identity tests and removed only the rejected fused-transition test. No existing baseline correctness assertion was weakened. The final runtime diff against original #2943 is only the removal of #2928. Two documentation changes preserve negative trials and the full-GQL attribution caveats.

Original per-thesis measurements remain linked above; they describe their original source/parent, not the new SHAs. Lower rewritten PRs rerun correctness CI, but their benchmark labels were removed to avoid a redundant full macro run at every level. The combined tip retains the benchmark label and will get a fresh complete receipt. Native combined binaries are building; endpoint/source hashes and mapping are in `/tmp/jazz-overnight-consolidated-20260913`.

Explicit retained tradeoffs remain visible: #2936's timestamp-independent footprint bookkeeping improves writes but has a twice-observed ~3–4% cold penalty whose exact cause is not yet established; #2942/#2943 improve writes with small cold penalties. These are not silently called universal wins. The morning report will compare actual combined endpoints rather than adding per-PR percentages.

Tooling-friction: preserved immutable binaries and exact old/new ref mappings make pruning possible without losing the evidence chain; only combined-tip macro reruns are needed for this consolidation.

aeplay:
🤖 Pruned 18-PR checkpoint confirmed. No merges; no adopter testing requested yet. #2945 is now restacked above this checkpoint and #2946 is the next isolated trial, both awaiting their own fresh confirmation.

Candidate #2943 `0487a56a67ffadc8197f4fe862ce3e0980ecdf8b`, run `6aa696e3499cfbed9c30796f`. Baseline run `6aa62c81bc6abc0f9f4e911e` actually names #2924 commit `b699aa1a4a2456af1bda3c19f565e4a8dac60c2f`, not main's rebased SHA. Verified both it and current main `a68fcd455d7e479794cf889ada655b95fe56a830` have the identical complete tree `fc87f3200b80e96abe55b3259c85080c8a0de400`; this is a tree-equivalent main baseline, not an assumed synthetic-hash equivalence.

| Workload | Baseline minimum / median (s) | Pruned minimum / median (s) | Minimum latency change |
|---|---:|---:|---:|
| Permissioned cold, 27,518 rows | 58.861929 / 59.076509 | 46.233142 / 46.317243 | −21.45% |
| Sequential inserts, 1,350 | 12.406636 / 12.434892 | 7.339768 / 7.431289 | −40.84% |
| Sequential updates, 1,350 | 21.623139 / 21.648056 | 7.274297 / 7.349811 | −66.36% |
| Batch updates, 1,350 | 1.111481 / 1.115392 | 1.047067 / 1.052197 | −5.80% |
| Reopen, 1,500 | 0.654776 / 0.673252 | 0.608030 / 0.633112 | −7.14%, but slow candidate rounds reach 1.471s |

Todo sequential throughput from these minima is ~109 → 184 insert roundtrips/s and ~62 → 186 update roundtrips/s. These native workload roundtrips include foreground authoring, wire encode/decode, RocksDB worker ingest/persistence and delivery back to the maintained in-memory foreground; JS staging and per-write fsync are excluded. They are not raw SQL storage-engine insert rates or adopter-machine measurements.

The pruned checkpoint is essentially unchanged versus original #2943 for cold (minimum +0.30%, median +0.39%), and slightly better on write minima; removing #2928 did not establish a material regression on the hosted endpoint. Local pruning ABBA was mixed: cold first parent drifted 10.190→9.235s while candidate blocks were 9.448/9.410s; the second candidate write block slowed ~2–3%. Do not use that drift to claim a pruning speedup. All actual correctness CI at the pruned tip is green, and the local Jazz library passes 2,054 tests (2 existing ignored). Original per-PR receipts remain historical comparisons, not measurements at rewritten SHAs.

## Broad-suite guard: one unresolved regression flag

The complete CodSpeed comparison reports 19 improvements, 55 unchanged and 1 regression across its common benchmarks, with no reported environment mismatch. Some W1 subscription deltas/resumes improve ~2.5–4.4x; these are secondary benefits, not the workload that selected the trials.

`update_activity_indexed_predicate_no_subscription_rocksdb` flags a slower minimum: 736.850µs → 852.832µs (+15.74% latency, −13.60% throughput). Its median is 894.795µs → 893.752µs (flat); each run has only ten single-iteration samples and broad tails. Do not dismiss it as noise, and do not call the whole performance check green.

Fetched this exact benchmark from every original overnight run. The floor first rises at #2925 (860.758µs; median 1,022.976µs); most later minima cluster around 840–880µs. #2934/#2938/#2943 have still slower minima in individual runs. This narrows a potential mechanism to publication/shared-write ownership or a baseline/setup/cadence effect, not the latest source-transition change. The benchmark toggles an indexed predicate and waits for local durability on a persistent 24k-row fixture with no maintained subscription. Next safe checks: full GQL profile/code audit and a longer unchanged native sample comparison at the exact baseline versus #2925. No benchmark assertion or durability policy will be weakened to make it pass.

Receipts, full distributions, source tree identities, original-to-pruned ref map and all-run W1 history live in `/tmp/jazz-overnight-consolidated-20260913`; full graph IDs are captured there. GQL baseline hash correction above supersedes any shorthand identifying that run directly with main's SHA.

Tooling-friction: full-suite comparison caught a control-path warning that the five target cases alone would miss; retain exact distributions and inspect both minimum and median before assigning cause.

PR #2945, opened 2026-09-13
perf: prepare RocksDB atomic write batches once

aeplay:
🤖 Part of #2913. Now stacked on #2943 after excluding #2928/#2944; original measurements describe the original parent and fresh restacked receipts are in the comments/#2913. No adopter testing requested yet. No merge before morning approval.

## Before / after

Owned and borrowed outcome-aware writes first validated every family, then resolved each family again while constructing the same native WriteBatch. Both implementations duplicated preparation and submission logic. Prepare the batch once, borrowing the input operations, and reuse the previous non-default family handle within that call. All entry points share preparation and submission.

## Invariants and worked cases

- A batch setting `records/a`, overwriting it, and deleting `other/b` still submits once in its original order. Default-family operations remain default operations; they do not invalidate the one-entry non-default handle cache.
- A final operation naming a missing or reserved internal family discards only the local native buffer. Earlier sets/deletes never reach RocksDB; outcome-aware APIs return `Uncommitted`. Namespace checks are unchanged, including the internal-family guard.
- After a native write is attempted, either the write or a required WAL flush can fail with uncertain durability. These still return `PossiblyCommitted`, retain flush debt, and must not be treated as safe-to-retry uncommitted failures.
- Submission stays under the mutation gate and advances the existing cadence once, including empty batches. There is no persistent handle cache, schema-cache invalidation change, or cross-call borrowing.

## Scope / encoding review

No storage keys, values, physical families, manifests, wire encoding, serialization, or WAL policy changes. Existing byte corpus and durability tests remain unmodified. This changes only construction of the same native atomic batch.

## Evidence / prediction

Native phase-scoped profiling puts cf_handle near 0.82% of all sampled cycles. Predict 0.3–1% permissioned cold load, 0–1% todo; retain only with end-to-end evidence or a compelling simplification with no regression. New public storage mutation/readback test covers mixed families, repeated handles, overwrite/delete order, late invalid/reserved-family rejection and reopen. Its private counter observes only otherwise-invisible handle lookup work. Full backend test suite passes (27 tests); actual correctness CI and local/native/CodSpeed results will follow. No local CI-equivalent claim.

Tooling-friction: the prepared native batch itself is a successful validation receipt; a second family-name pass is unnecessary.

PR #2946, opened 2026-09-13
perf: accumulate multisink source transitions directly

aeplay:
🤖 Continues #2913 on rebased #2945. Draft; no adopter testing requested yet; no merge before morning approval.

## Before / after

Single-terminal source changes were copied into an ordered presence vector and into program-fact add/remove vectors. The multisink caller combined the presence vector into its existing drain-local BTreeMap, then discarded the closure facts in the add/remove vectors. Route those changes directly into that same accumulator. Final output still uses the existing first/final presence folding and sorted list generation.

Standalone helper output remains complete; non-closure facts keep their existing lists. There is no new retained cache or different fact representation. Independent signed origin weights, immutable version identity, membership/witness reconciliation and publication acknowledgment are unchanged. The rejected #2928 role-weight fusion is not reintroduced.

## Worked cases / invariants

- Two terminal roles witness one physical version: the first appearance contributes one addition; removing one role retains it; removing the final role contributes one removal.
- One drain removes and re-adds the same fact: changes reach the same accumulator in the same terminal/event order, and its initial/final presence comparison emits no spurious disappearance.
- A closure fact moves into the accumulator once instead of being retained in intermediate vectors that are immediately discarded. Ordinary relation/non-closure facts remain in their existing add/remove lists.
- All rows of a terminal must decode before its retained-state application begins. A late decode error leaves the accumulator and that terminal's state unchanged. A later application error can still follow earlier retained-state work, as before; the multisink caller returns the error and discards its drain-local accumulator, never publishing a partial drain. This does not add transactional rollback to terminal application.
- Standalone helper callers still receive complete fact lists and presence transitions. One existing late-decode test receives the mechanically required `None` argument; no assertion changes.

No public external API, wire/storage encoding, codec, schema, permission or benchmark assertion changes. Exact final facts/order remain authoritative; no serializer layout is introduced.

## Evidence / prediction / preflight

The current non-inline native cold profile, captured only during connect/subscribe/readiness and without dense phase tracing, confirms copying/comparison beneath multisink and decoded-output processing. Full GQL profiles place record_source_fact_transition copying near 0.15–0.19s. Broad caller scopes include other work and are not wholly removable. Predict 0.5–2% cold, flat to 1% todo, low-to-medium confidence.

Read the full rejected ledger and searched preserved branches plus open/closed PR descriptions. This removes an audited throwaway consumer list, unlike #2928's rejected role-weight fusion, #2917's final manifest borrowing, #2933's temporary event representation, or rejected vector/sort netting. Original sources/binaries remain preserved.

Two new focused tests pass: direct-vs-standalone equality over signed/shared role transitions with an 8KB witness and no temporary lists; late-decode failure before any accumulator/state mutation. These internal tests are necessary to observe intermediate ownership and inject a decoder failure. Existing public permission/subscription tests and final delivery canaries remain unchanged. Production check, Clippy and formatting pass. Broad Jazz/canaries, native ABBA and hosted exact-head CI/CodSpeed follow; no full local CI-equivalent or independent-review claim.

Tooling-friction: tracing-disabled native profiles avoid attributing instrumentation clock overhead to production work; zero opt-in counters in that diagnostic mean disabled, not zero actual work.


aeplay:
🤖 #2946 outcome: exclude this trial from the overnight stack. No adopter testing requested; no merge before morning approval.

Exact parent #2945 `7c6ef07258eeec56c3ea7bf80575422d9399f04b`, run `6aa6998fbbf79fe271c9347c`; candidate `21121eace403526e3ee91411f711e6fa65d483cd`, run `6aa69b8baddcedc409371bc7`. GQL confirms both exact source hashes. Full walltime cold/update graphs downloaded for both, including result IDs `6aa69e36ad9a6239bf6cb1be` and `6aa69e36ad9a6239bf6cb1c5` for the candidate.

| Workload | Parent minimum / median (s) | Candidate minimum / median (s) |
|---|---:|---:|
| Permissioned cold | 45.669011 / 45.918665 | 45.575061 / 45.872700 |
| Sequential inserts | 7.484891 / 7.988058 | 7.472747 / 7.754041 |
| Sequential updates | 7.299445 / 7.367348 | 7.362472 / 7.367541 |
| Batch updates | 1.042799 / 1.051711 | 1.039305 / 1.047690 |
| Reopen | 0.607649 / 0.634410 | 0.605636 / 0.622226 |

Cold is only −0.21% minimum / −0.10% median versus the predicted 0.5–2%; updates are +0.86% minimum / flat median. Insert medians have broad parent variation, so the faster candidate median does not establish a throughput win. Local ABBA was similarly unpersuasive: warmed cold pair +0.36%; insert pairs +1.20/+0.24%, updates +1.38/+0.15%, batch +1.35/+0.86%. The first parent cold block drifted from 10.120s to 9.338s, while head blocks were 9.462/9.372s; its apparent first-block speedup is not evidence.

Moving facts directly into the multisink accumulator does eliminate audited throwaway intermediate lists, but avoiding those copies did not yield a useful end-to-end gain. Keep the simpler existing call interface; independent signed role weights remain unchanged either way. The full graph still places multisink at ~5.76s and decoded application at ~4.23s inclusive, overlapping scopes that contain much more than those removed copies.

Correctness CI is green, as are 2,056 local Jazz library tests (2 existing ignored) and all three incremental-delivery canaries. This is a performance rejection, not a correctness failure. Branch, immutable binaries, complete distributions and graphs remain preserved at `/tmp/jazz-direct-source-transitions-VcwaFl` and the pushed PR branch.

Tooling-friction: direct ownership simplifications still need end-to-end confirmation; neither fewer clones nor a smaller profiled phase is a sufficient retention criterion.

PR #2947, opened 2026-09-13
perf: allocate shared join bucket maps only when populated

aeplay:
🤖 Draft #2913 trial on #2945. No adopter testing requested yet; no merge before morning approval.

## Before / after

Join buckets already inline singleton entries (#2940), but their default base and overlay each allocate an empty Rc. Committing the overlay creates another empty allocation. A private SharedJoinBucketMap now represents the empty state with None, allocating an Rc only on first mutation. Populated snapshots retain the same Rc copy-on-write boundary; committing consumes the overlay and returns it to allocation-free absence. An empty map reached by deleting populated entries is not automatically shrunk or demoted.

- A new bucket and its clones contain no allocated base/overlay. Reading or committing the empty bucket does not allocate either map.
- Add A with weight −2: allocate only the overlay and retain the negative weight. Committing creates the populated base, consumes the overlay, and leaves any pre-commit snapshot unchanged.
- Add +2 to that A: the overlay contains an explicit tombstone, not absence. A later +3 starts from zero. Snapshots before and after the tombstone remain independent after commit and future mutation.
- Two distinct records still promote to the existing hash map; its one-probe update path, exact encoded keys and iteration semantics are unchanged. Arrangement snapshots and same-input reuse proofs are unchanged.
- This changes no storage/wire representation, serializer, schema, permissions, public API, or benchmark assertion. The wrapper is private in-memory ownership state only. No global cache or unsafe code.

## Evidence / prediction / preflight

Current mimalloc allocation-site capture estimates ~1.88M requests in JoinBucket::default and ~0.91M in commit_overlay during the permissioned cold interval (83.63M total requests). These are sampled cumulative requests, not live/peak memory. The complete current GQL cold graph attributes ~1.056s inclusive to join-index construction. Predict 0.5–2% cold, approximately flat todo; allocation savings can be offset by extra branches.

Read the full rejected ledger and testing guidance; searched preserved branches and open/closed PR descriptions, including #2901 and #2864. Those changed bucket contents/build algorithms; this only eliminates empty Rc ownership. #2946's rejected accumulator trial is recorded in the ledger, not included in this branch.

All 771 Groove library tests pass (2 existing ignored), including a new internal allocation-absence/snapshot/tombstone proof and the unchanged signed-map oracle. Two existing pointer-equality assertions mechanically unwrap the now-optional populated pointers, preserving their exact Rc identity requirement; no assertion is weakened. Internal tests are necessary because public query results cannot establish allocation absence or snapshot pointer sharing. Broader Jazz/canaries and actual CI follow. Native timing/CodSpeed pending; no full local CI-equivalent or independent-review claim.

Tooling-friction: caller-attributed allocation profiles revealed the remaining empty ownership objects after the earlier singleton-map improvement; uninstrumented end-to-end timing decides retention.


aeplay:
🤖 #2947 first native receipt; hosted confirmation pending. No adopter testing requested yet; no merge before morning approval.

Exact parent #2945 `7c6ef07258eeec56c3ea7bf80575422d9399f04b` versus head `04a581b237` (full source and binary hashes in `/tmp/jazz-lazy-empty-join-buckets-1T7lEi`). Same profile.perf/mimalloc/RocksDB workloads, both immutable binaries warmed once per workload before ABBA, three measured samples per block. This preconditioning is explicit; older un-warmed first-block receipts are not silently replaced.

| Case | Parent block medians | Head block medians | Paired latency change |
|---|---:|---:|---:|
| Permissioned cold | 9.465 / 9.388s | 9.314 / 9.345s | −1.60 / −0.46% |
| Sequential inserts | 1.257 / 1.267s | 1.256 / 1.260s | −0.08 / −0.55% |
| Sequential updates | 1.315 / 1.317s | 1.299 / 1.302s | −1.22 / −1.14% |
| Batch updates | 163.7 / 165.1ms | 164.6 / 163.7ms | +0.55 / −0.85% |
| Reopen | 110.2 / 110.9ms | 111.4 / 112.3ms | +1.09 / +1.26% |

Cold is consistent with the 0.5–2% thesis but still overlapping; updates are a plausible small secondary gain, reopen a small contrary movement. All 771 Groove / 2,054 Jazz library tests and three incremental canaries pass, along with Clippy and formatting. Actual CI is still running; do not call this fully CI-equivalent locally. Proceeding to the next isolated ownership thesis while hosted confirmation runs.

Tooling-friction: warming both binaries made the remaining cold variability more interpretable; complete distributions and per-process resource receipts remain retained.

aeplay:
🤖 #2947 hosted decision: reject as unproven; preserve the experiment and remove it from the linear stack after the active trial's comparable timing finishes.

Exact GQL source 04a581b23796afe78e9ffdde977aaafe58d04930, run 6aa6a2b4addcedc409371c8e. Cold result 6aa6a57cad9a6239bf6cd77e and update result 6aa6a57cad9a6239bf6cd785; full graphs downloaded. Parent #2945 source 7c6ef07258eeec56c3ea7bf80575422d9399f04b; its original distributions and later todo-only rerun are saved separately.

Cold minimum 45.669011 → 45.788861s (+0.26%); median 45.918665 → 46.415937s (+1.08%). This does not confirm the predicted 0.5–2% win or the small warmed local gain. Updates 7.299445 → 7.355265s minimum (+0.76%), 7.367348 → 7.365882s median (flat). Against the parent's later todo repeat, update differences are likewise small. Insert minima 7.484891 → 7.460667s (−0.32%), with broad medians in both runs. Batch/reopen remain broadly flat.

Avoiding millions of estimated empty-Rc allocation requests is not sufficient reason to keep a new wrapper and its branches without a demonstrated end-to-end gain. #2940's inline singleton buckets are a separate, confirmed change and remain. No correctness failure: 771 Groove and 2054 Jazz library tests plus canaries passed; exact correctness CI is green. No assertions were weakened.

No adopter testing requested yet; no merge before morning approval. Tooling-friction: allocation counts made this look promising, but the endpoint receipt is the acceptance criterion.

PR #2948, opened 2026-09-13
perf: preserve immutable buffers through hydration readiness

aeplay:
🤖 Draft #2913 retained trial, restacked directly on #2945. No adopter testing requested yet; no merge before morning approval.

## Restack status

#2947 was rejected and removed from this ancestry. Current head is `1ea2ea0b66969562d8a56313eb3b1175a5b75ed4`. Fresh exact-head correctness CI and CodSpeed both passed. Restacked cold is 44.910388s minimum / 45.159534s median, −1.66% / −1.65% versus #2945; sequential write minima are effectively flat against its preserved repeat. This confirms the gain without the rejected wrapper. The original-parent −2.68% / −3.19% result remains a separate historical receipt. Full GQL graphs and source/result IDs are in [the restacked outcome](https://github.com/garden-co/jazz/issues/2913#issuecomment-5654040462). The whole stack and explicit remaining tradeoffs are in [the morning summary](https://github.com/garden-co/jazz/issues/2913#issuecomment-5654069210).

## Before / after

Hydration cloned the entire RecordDeltas vector returned by every evaluated node, even though interior-node clones were immediately discarded. Roots were then materialized into another owned vector; already-inline records still copied their raw bytes. Keep the existing Arc<RecordDeltas> through evaluation readiness and paused root requests. Only root output construction allocates the required output vector. A borrowing materializer returns the original inline slice after the same encoded probe, allowing the caller to retain the existing Bytes storage; changed indirect values keep their normal owned encoding.

The old owned materializer API remains a wrapper for its other callers. No universal record sharing or new retained cache is introduced.

## Invariants / worked cases

- Interior node completes: its existing memo retains the immutable result; hydration marks the node complete without allocating a throwaway row-vector clone.
- Inline root row: run the same encoded indirect-value probe, then clone the existing Bytes handle. Descriptor, weight, order and bytes are unchanged; output remains owned after its input and database are dropped.
- A root contains inline A, indirect B, inline C: B still registers all required chunk requests. Nothing publishes until all rows materialize. While paused, pending_outputs owns the input Arc. Retry uses those unchanged inputs and produces all three rows exactly once.
- Chunk failure/cancellation follows the existing scoped failure and dropped-session rules. No memo, operator-state or arrangement state installs earlier. CPU-yield budget, work queue order, temporal predecessor barriers and waker registration are unchanged.
[...]

PR #2949, opened 2026-09-13
perf: move staged operation vectors across ownership boundaries

aeplay:
🤖 Draft #2913 trial, stacked on #2948. No adopter testing requested yet; no merge before morning approval.

## Before / after

StagedWriteState::from(Vec<OwnedWriteOperation>) pushed an already-owned vector into a new vector one operation at a time. Consuming a single uniquely-owned staging block likewise flattened it into a fresh vector. Wrap the incoming nonempty vector directly; on consumption recover a unique single block directly. Empty input remains the allocation-free empty state. Shared or multi-block inputs retain the existing clone/flatten behavior.

- A unique 100-operation vector retains its allocation and capacity through wrapping, lazy-index lookup, a same-key overwrite and consumption. Last-write order remains ordinal 100 after appending the overwrite.
- A published single-block snapshot remains immutable: consuming a shared staging state clones its operations, and mutating the returned payload cannot affect the snapshot.
- Multiple blocks still flatten in append order. Empty input discards unused capacity as before. Last-write lookup, publication lifetime, cancellation, storage/WAL cadence, durable encodings and public APIs are unchanged.

## Evidence / validation

Prediction: 0.3–1.5% permissioned cold load, 0–1% todo, possibly 1–5% W1 updates; low confidence. The W1 floor regression begins near #2925, but the incoming-vector reconstruction predates that PR. This trial tests a plausible constant overhead, not a proven explanation for the hosted regression.

Full rejected-experiment ledger and testing guidance read; preserved branches and PR descriptions searched, including #2925, #2809 and perf/index-staged-scan-snapshots. This does not repeat value-codec staging or add an index/cache. 774 Groove library tests pass, with two existing ignored; production Clippy and diff checks pass. Two new allocation/aliasing tests preserve all existing assertions. Actual CI, Jazz checks and timings follow; no full local CI-equivalent or independent-review claim.

Also documents a measurement discovery: rerunning one CodSpeed CI job replaces that job's result IDs inside the same run ID. Saved distributions, result IDs and CI attempt are necessary immutable receipts.

Tooling-friction: ownership-boundary allocation tests expose copies that broad async flamegraph frames cannot reliably attribute.


aeplay:
🤖 #2949 first local outcome: modest cold/W1 gain with slightly slower todo; await hosted confirmation, do not claim a universal win.

Exact parent b4c78ccc3b16a0bab426bee4358d4b0a9cff5d94 vs head 0058d2ad9ca9cedb2444e79a8471aec27cddb896. Immutable binaries and hashes preserved. Warmed each arm once; profile.perf/mimalloc todo and cold, then ABBA with three samples per arm. No builds/tests/heavy graph parsing overlapped timings.

- Cold medians 9.252/9.275 → 9.190/9.153s: −0.67% / −1.32%, within the 0.3–1.5% prediction.
- Inserts 1.257/1.258 → 1.261/1.272s: +0.32% / +1.11%.
- Updates 1.300/1.303 → 1.314/1.312s: +1.08% / +0.69%.
- Batch 163.3/162.0 → 163.0/163.8ms; reopen 108.9/108.4 → 109.4/110.6ms. These are small movements, not independently established regressions.

Additional W1 System-allocator control: five fresh-fixture ABBA cohorts, 20 processes, exactly ten updates per process (no changed history depth), both arms warmed once. Median of per-process fastest samples 140.65 → 138.90µs (−1.24%); median of per-process medians 152.80 → 149.85µs (−1.93%). Broad mean/tails nearly flat. This is a small improvement consistent with constant overhead, not an explanation or fix for the whole hosted W1 floor regression. Original main/#2925 receipt remains separately preserved.

774 Groove and 2054 Jazz library tests pass (two existing ignores each), all three canaries and the two-seed maintained-vs-one-shot oracle pass. Actual correctness CI is nearly complete; do not call partial local checks CI-equivalent. No adopter testing requested yet; no merge before morning approval.

Tooling-friction: keep the W1 ten-update fresh-fixture protocol; lengthening one fixture changes compaction/history behavior and invalidates the comparison.

aeplay:
🤖 #2949 decision: reject as unproven and remove from the proposed linear stack. Preserve branch, tests, exact binaries and receipts.

Exact GQL source 0058d2ad9ca9cedb2444e79a8471aec27cddb896, run 6aa6aa6abbf79fe271c9356d. Cold result 6aa6ad14ad9a6239bf6cf264, updates 6aa6ad14ad9a6239bf6cf262, W1 update 6aa6ad15ad9a6239bf6cf27d. Full graphs downloaded. Parent is #2948 b4c78ccc3b16a0bab426bee4358d4b0a9cff5d94.

Cold minimum 44.563656 → 44.775260s (+0.47%); median 44.933264 → 44.785752s (−0.33%). Inserts minimum 7.465922 → 7.467608s (flat), with broad medians. Updates minimum 7.281792 → 7.313845s (+0.44%), median 7.346597 → 7.324488s (−0.30%). Batch flat; reopen has a 1.469s tail outlier. CodSpeed classifies all 75 benchmarks unchanged.

W1 update minimum 871.820 → 863.153µs (−0.99%) but median 953.355 → 1030.590µs (+8.1%) across just ten samples with broad tails. Local fresh-fixture cohorts were ~1–2% faster, but this does not establish a hosted improvement or resolve the main/#2925 floor regression. Local todo also moved ~1% slower. Taken together, this does not justify retaining a perf change solely because its internal allocation assertion passes, even though the runtime change is small.

No correctness failure: 774 Groove and 2054 Jazz library tests, all three canaries and the two-seed oracle passed. Full correctness CI result is tracked separately. No existing assertion weakened. The reusable CodSpeed rerun-receipt documentation added here will be retained independently when pruning the runtime trial.

No adopter testing requested yet; no merge before morning approval. Tooling-friction: a mechanically better ownership transfer can still be end-to-end neutral; preserve the negative result rather than extrapolate from pointer identity.

Issue #2985, opened 2026-09-14
perf: filtered top-N and pagination read the full scan set (policy discards the index, and no ordered access path exists for the sort)


tobiaslins:
## Summary

Owner-scoped reads are **bounded by the owner's whole row scope, not by the requested page.** On a table of 1,000,000 documents with 10,000 per owner, asking for **one** row costs exactly the same as asking for the owner's entire scope: 10,000 index reads plus 10,000 current-row reads, ~45 ms. A `LIMIT 1` on a primary-key-equality-ish owner index should be ~1 read.

The mechanism is a **policy gate**. On the *same* data, *same* table definition, *same* query, only the policy differs:

| table has SELECT policy | LIMIT | index reads | row reads | time |
|---|---:|---:|---:|---:|
| no | 1 | **1** | **1** | **0.76 ms** |
| no | 50 | 50 | 50 | 0.62 ms |
| no | 500 | 500 | 500 | 3.07 ms |
| **yes** | 1 | **10,000** | **10,000** | **51.2 ms** |
| **yes** | 50 | 10,000 | 10,000 | 50.7 ms |
| **yes** | 500 | 10,000 | 10,000 | 50.6 ms |

With a policy, the page size has **zero** effect on the work. That is not #2079's ORDER BY problem: the policy case is scope-bounded even with **no ORDER BY and no LIMIT interaction at all**, and the no-policy case is page-bounded with the identical query shape.

This matters for every hosted/SaaS workload, because row-level security is the normal case there, not the exception.

## Reproduction

Native, `[profile.perf]`, RocksDB `WalNoSync`, mimalloc, source `a2cd41ba1` (clean), local arm64.

Schema in the fast/slow pair is byte-identical apart from `.policies(...)`:

```rust
documents (owner_id TEXT, done BOOL, updated_at TIMESTAMP, title TEXT)
  .index_only(["owner_id", "updated_at"])
  .policies(select: owner_id == @user.identity.subject)   // policy arm only
```

Seed: one owner with 10,000 documents (1M rows across 100 owners in the larger run). Query:

```rust
Query::from("documents")
    .filter(eq(col("owner_id"), param("owner")))
    .limit(N)
```

Also measured on the 1M-row table (100 owners x 10k, 5,000-row seed batches, 77.6 s seed):

| shape | result rows | index reads | row reads | total reads | time |
|---|---:|---:|---:|---:|---:|
[...]

Issue #2079, opened 2026-08-26
Perf thesis: ordered LIMIT reads should scale with the page, not all filter matches


aeplay:
## Thesis

For an indexed equality filter followed by `ORDER BY ... LIMIT n`, cold one-shot work should be bounded by the requested page plus a small visibility/policy factor when a compatible ordered access path exists. It should not hydrate and decode every equality match merely to retain `n` rows.

## Current receipt

The existing public-core `selective_global_hydration` benchmark already models this shape on persistent RocksDB:

```text
documents.where(team = ? and active = true)
  .orderBy(updated_at DESC, id DESC)
  .limit(50)
```

At checkpoint `de1663cdd`, table size fixed at 10k:

| matching `team` rows | result rows | index reads | current-row reads | logical reads | query time |
|---:|---:|---:|---:|---:|---:|
| 100 | 50 | 100 | 100 | 200 | 3.532 ms |
| 10,000 | 50 | 10,000 | 10,000 | 20,000 | 162.948 ms |

A 100x increase in matching rows with a fixed 50-row result produced ~46x wall time and exactly 100x physical candidate reads. The declared `team` index successfully avoids unrelated table rows, but Jazz hydrates the entire equality bucket before Groove's finite `TopBy` retains the page.

## Code mechanics

- `one_shot_access_paths` selects equality indices but explicitly refuses a source cap when `order_by` is present (`crates/jazz/src/node/query_eval/read_sources.rs`).
- Public `index_only` currently exposes independent single-column indices; it cannot describe a compound `(team, updated_at, id)` order.
- The query lowers ordering/limit to Groove `TopBy`, which correctly bounds retained state but only after all upstream candidates have been read and materialized.

## Candidate improvements

1. Add schema/API support for compound ordered indices and encode their key direction/order.
2. Teach one-shot planning to recognize an order-compatible prefix, scan it in the requested direction, and stop after enough *visible, predicate-passing* rows.
3. Until compound indices exist, consider an order-only scan plus residual equality filtering when estimated cheaper, but do not cap before visibility/residual predicates prove enough results.
4. Preserve the existing `TopBy` path for subscriptions, unsupported orders, joins, policies, and fallback correctness.

Expected improvement for the measured 10k-match/50-row case: physical reads should approach O(50) rather than 20k, with a plausible 20–50x cold latency reduction on this machine. Sparse deletions/policy denials may require bounded over-fetch/retry rather than an unconditional 50-row source cap.

## Proof obligations

[...]

PR #2996, opened 2026-09-14
bench(jazz): add a policy-scoped SaaS read benchmark

tobiaslins:
🤖 Adds a benchmark for the read shape a hosted multi-tenant product actually issues: a page of rows the requesting user owns, or rows shared through the organisation that owns them.

## Why a second bench rather than extending the existing one

`selective_global_hydration` is well-formed but models a configuration that does not occur in a hosted product. It uses a `team` equality with **no read policy** and `AuthorSubject::SYSTEM`. Both of those are exactly the conditions under which the planner's declared-index path works, so it measures index selection while holding row-level security out of scope. Its receipt shows `index_reads` tracking the match count one-for-one, which is the healthy case.

`owner_filter_diagnostic` covers adjacent ground but is **born-red on unmodified `main`**:

```
thread 'main' panicked at crates/jazz/benches/owner_filter_diagnostic.rs:91:
assertion `left == right` failed: policy_only row count changed
  left: 0
  right: 500
```

It is also not referenced by CI. So it is not a usable control today.

| dimension | `selective_global_hydration` | `saas_policy_reads` |
| --- | --- | --- |
| read policy | none | `owner == @user.account`, or `owner == @user.account OR INHERITS via org` |
| identity | `AuthorSubject::SYSTEM` | the requesting user |
| equality bucket | 100 rows | `table_rows / 100` |
| cases | one equality, two orderings | owner page, owner top-10, org page |
| page sizes | fixed at 50 | 1 / 10 / 50 |
| relations | one table | documents plus org membership |

**Both are kept deliberately.** The pair is the signal: the policy-free arm selects the declared index while a policy-scoped arm does not. A trend on either one alone hides that difference.

## What it reports

Per case: `index_reads`, `current_row_reads`, `history_row_reads`, `total_reads`, `query_us`, and `reads_per_result_row`. The last is the boundedness check — a page-bounded read is roughly constant in it as the table grows, a scope-bounded read grows with the owner's share.

Local release run, arm64 (M1 Max), RocksDB `WalNoSync`, single runs:

| case | median |
| --- | ---: |
| `saas_owner_page_10k` | 69.62 ms |
| `saas_owner_page_100k` | 629.2 ms |
| `saas_owner_top10_100k` | 649.0 ms |

100k costs ~9x the 10k case, while top-10 and page-50 at 100k cost the same. That is scope-bounded behaviour, and it is what this bench keeps visible.

Receipt at 20k rows, per-owner share 200:

| case | limit | index_reads | current_row_reads | total_reads |
| `owner_page` | 1 | 0 | 20000 | 20000 |
| `owner_page` | 10 | 0 | 20000 | 20000 |
| `owner_page` | 50 | 0 | 20000 | 20000 |
| `owner_recent` | 10 | 0 | 20000 | 20000 |

## Two measurement details worth review attention

- **Every case reopens the database before timing.** A second query on the same instance is served from the retained maintained view and reports zero reads, which measures the cache rather than query planning. My first version did not reopen and reported zero reads for every case after the first.
- **Receipt JSONL is emitted only under `JAZZ_SAAS_RECEIPT`.** Divan owns stdout for the wall-time lane and CodSpeed parses that stream; an early version leaked seed lines into the benchmark output.

## Two structural facts the bench documents rather than assumes

- `index_only(["owner_id", "updated_at"])` declares **independent single-column** global-current indexes, because `TableSchema::global_current_indexed_columns()` is a `BTreeSet` (`crates/jazz/src/schema.rs:1323`) and each entry becomes its own index. There is no ordered `(owner_id, updated_at)` access path, so an `ORDER BY updated_at DESC` page cannot be capped before the sort and the equality bucket must be materialised.
- The org arm uses `INHERITS via org_id`, which the planner currently keeps on the full-scan path by design, per `policy_access_path_planner_falls_back_for_or_and_non_equality` and the reasoning in `read_sources.rs`.

Both are recorded in [SAAS_POLICY_READS.md](crates/jazz/benches/SAAS_POLICY_READS.md) so the next reader does not re-derive them.

## Scope

Measurement only. **No runtime behaviour changes.** The shared `schema_fixture` change is additive — one import plus two policy helpers (`owner_or_org_access`, `select_only`) totalling 22 lines.

## Validation

- `cargo fmt -p jazz -- --check` clean.
- `cargo clippy -p jazz --features testing --bench saas_policy_reads` clean, apart from a pre-existing unused-import warning in the shared fixture that this change does not touch.
- `cargo check -p jazz --features testing --benches` passes for every bench, confirming the shared-fixture addition breaks nothing.
- New bench compiles and runs in both Divan and receipt modes.
- YAML lint passes on the modified workflow.
- Pre-commit hooks (clippy, oxfmt, rustfmt, sensitive-data) pass.

## CI

Adds a `saas-policy-reads-walltime` CodSpeed job alongside the existing maintained-selective-hydration job, gated on the `benchmark` label like its neighbours so it does not run on every PR. The benchmark label is set.

## Caveats

- Local arm64 figures are a **separate series** from CodSpeed ARM64; do not mix them.
- Single runs, not repeated medians. Seeding dominates the receipt: 1M rows takes minutes and reseeding per rung is the largest cost in a full ladder run. Noted as tooling friction in the bench doc.
- The bench asserts returned row counts; it does not digest the ordered result. Ordering is checked by construction of the `updated_at` values.

Related: #2079 (ordered LIMIT reads should scale with the page), #2985 (policy-scoped reads are scope-bounded, not page-bounded).


aeplay:
Superseded by #3072 at Anselm’s request. Thank you @tobiaslins for identifying this gap and contributing the workload; the replacement credits your work in its documentation and co-authored commit. It adopts the fixture as an example-owned CodSpeed suite with matched policy controls, stable organization grouping, exact authorization/order tests, and a consistent query-only timer. We will record a fresh baseline there before touching the read planner. Your branch is preserved.

PR #3093, opened 2026-09-17
Validate raw record keys against primary keys

joeinnes:
🤖 Reject raw batch writes whose supplied storage key disagrees with the primary key encoded in the record.

For example, storing a record whose ID is 7 under key 9 would make keyed lookup disagree with the record contents. Raw insert, fresh insert, update and batch-overlay inspection now reject that mismatch with InvalidDirectRecordStoreKey before applying the operation. Matching-key writes continue to succeed.

The check reads only primary-key fields at raw-write admission. It does not re-encode records, validate unrelated fields or add read-time validation. Existing storage encodings and public logical write semantics are unchanged. Includes a jazz-tools PATCH changeset.

Independent review initially found that the proposed test used a variant-tagged envelope where the API expects payload bytes; unrelated errors concealed missing validation. Corrected fixtures assert the exact error and include successful matching writes. Disabling each of six owned/borrowed validation paths now fails its regression while the positive control stays green.

Full Groove verification passed after the fixture correction (793 library tests passed, 2 existing ignored; integration and documentation targets green). Coordinator and independent reviewer each reran the three focused tests on current-main integration 8d2703e26fa2c2b00d861f04da83eb211f53e916, all passed. Independent final review has no remaining findings. Fresh CI and user approval remain pending.


aeplay:
🤖 Independent review found the original regression could pass for the wrong reason: variant-envelope bytes were supplied to an API expecting raw payload bytes. Removing the insert/fresh/update key checks still left that test green.

Pushed 87e87eb95c to correct the fixture, require the intended InvalidDirectRecordStoreKey error, verify matching-key writes succeed, and exercise borrowed batch-overlay validation. Runtime behavior is unchanged by this follow-up, and a PATCH changeset is included. Focused tests pass; per-path mutation sensitivity and full Groove verification are running before readiness.

PR #3094, opened 2026-09-17
Validate complete JSON documents for string pointers

joeinnes:
## Summary
Validate the complete JSON document before resolving a String large-value JSON pointer.

## Before
String large-value pointer reads could accept a valid prefix while ignoring malformed trailing text, and could diverge from serde JSON pointer semantics for leading-zero array indices.

## After
The full UTF-8 document is validated before pointer lookup, with pointer resolution matching the inline/serde behaviour for malformed suffixes and array indices.

## Test plan
- [x] `cargo test -p jazz --no-default-features --features testing,transport-compression-zstd --lib large_string_json_pointer`


aeplay:
🤖 Review decision from Anselm: validate JSON values on insert/update, then rely on that invariant for efficient pointer reads. The public TypeScript selection API already rejects non-JSON columns. We will align the lower-level helper with that contract instead of adding full-document parsing for ordinary String columns.

There is a related useful regression to retain: the fast parser's nested array traversal accepts a leading-zero component such as /0/01 through usize parsing, although the top-level array component already checks canonical spelling. The current String-specific serde path avoids that error only for strings; the JSON fast path still needs correction.

Preparing a replacement implementation: require JSON columns, fix nested array pointer semantics while preserving object keys such as "01", and retain partial-fetch/chunk-count coverage. No JSON storage encoding or descriptor change is intended. This PR remains held; the current implementation is not approved for merge.

aeplay:
🤖 Closing per Anselm's review decision. Thank you, Joe, for identifying the inline/chunked pointer inconsistency and supplying the regressions.

The intended contract is to validate JSON values on insert/update, then rely on that guarantee for efficient pointer reads. The public TypeScript API already requires JSON columns for pointer selections. We will align the lower-level Rust helper with that rule rather than add full-document parsing for arbitrary String columns, so this PR's String-specific implementation will not land.

There is a valid related bug worth retaining: nested array traversal in the fast JSON parser accepts noncanonical indices such as "01". The replacement work will fix that for actual JSON values, preserve object keys named "01", reject pointer reads on non-JSON columns, and retain chunk-count tests proving reads stay partial. No storage encoding or descriptor change is intended.

Replacement implementation is underway separately; its link will be added here when the first coherent commit is published. This branch is preserved for the original investigation and regression ideas.

aeplay:
🤖 Replacement work is now tracked in #3158: JSON-only pointer reads, canonical nested array indices, and partial-fetch regressions. This PR remains closed; its original branch is preserved.

PR #3096, opened 2026-09-17
Surface permanent queued-send failures

joeinnes:
🤖 Queued transport failures now reach the caller, and reconnecting subscriptions cannot lose a fast confirmation while draining messages from the previous connection.

## Behavior and examples

A temporary send blockage keeps valid messages in order and schedules another attempt. A permanent send failure ends the operation with the transport error instead of silently continuing. Updates already consumed before a temporary blockage still finish their normal processing.

During a connection handoff, messages left over from the previous authority cannot confirm a query on the new connection. We now finish draining that old backlog before reissuing subscriptions. Previously, a fast reply to the new subscription could arrive while the old backlog was still being discarded; that valid reply was then discarded too, leaving the query waiting indefinitely.

For example, after reconnecting, an old queued snapshot remains ineligible to settle the query. Once that queue drains, Jazz sends the new subscription and accepts its immediate fresh snapshot. If draining encounters backpressure, Jazz schedules another attempt before opening the new subscription; a permanent failure is surfaced. The fix does not relax authorization or allow an old receipt to become current.

## Validation

- Independent correctness review passed after fixing the fast-reply ordering gap found during review.
- Five reversible mutations prove sensitivity to drain ordering, transport backlog quarantine, pre-staged receipt invalidation, consumed-update finalization, and permanent flush failures.
- Nine focused regressions passed. Final integrated revision passed the Rust-workspace partition: 4,310 tests, 12 skipped; exhaustive workspace target compilation also passed.
- PATCH changeset included. No storage or wire encoding change. GitHub CI is completing separately; merge remains subject to user approval and green gates.


aeplay:
🤖 Independent review found a handoff liveness blocker at f4a127ce36f42fef7663b35b004b15de31323aca.

After handoff polling encounters backpressure, pending Subscribe is sent before the quarantined transport backlog has reached a clean idle poll. A fast authority can return the fresh confirming snapshot in that interval. The receiver marks every reply ineligible until Ok(None), discards this fresh response together with the old backlog, then clears quarantine without reissuing the waiting subscription. The read never settles.

A reversible timing-only change to the existing handoff regression ticks the server immediately after the client sends its fresh Subscribe. The unchanged assertion then fails: `a fresh confirming view must settle` (0 passed / 1 failed). The original serialized interleaving passed, as did the 12 focused baseline checks.

Please preserve the rule that old staged/transport-resident views cannot establish fresh coverage, while ensuring the fresh subscription response is not permanently lost. Candidate approaches: drain quarantine before issuing the fresh opens, or reliably reissue still-waiting opens when the drain completes. We are preparing the fix directly on this PR; not ready for merge yet.

Issue #2964, opened 2026-09-14
[RN] Native writes block JS and defer local subscription echoes


joeinnes:
## Regression

After moving from `2.0.0-alpha.53` to `2.0.0-alpha.54`, Xelson reports a React Native message-composer regression:

- while typing, the last character briefly disappears and reappears later;
- fast typing is delivered in bursts instead of continuously;
- the previous release did not show this behaviour.

## Expected behaviour

A local composer write should not block the JS/UI thread, and the write's own local result should become visible without waiting for a periodic relay poll. Keystrokes should remain responsive and subscription delivery should not introduce visible batching.

## Reported diagnosis (to verify)

The write path blocks JavaScript but does not return/apply its own delta. The visible echo arrives only through the 50 ms poll, whose `tick()` performs a full relay pump. Each keystroke also re-evaluates all subscriptions, including subscriptions that join `jazz_transactions`. Simulator profiling reportedly measured about 300 ms of blocked JS per letter.

These are the reporter's measurements and code reading; the causal path and exact attribution still need a focused reproduction.

## Investigation / acceptance

- Reproduce on the alpha.54 React Native runtime with a minimal local write and a maintained subscription.
- Measure JS-thread blocking, poll/tick duration, subscription evaluations, and time from write completion to the local echo.
- Ensure the foreground write result is visible without waiting for the 50 ms poll.
- Make subscription invalidation proportional to affected data rather than re-evaluating unrelated graphs on every keystroke.
- Preserve relay ordering, authorization, retries, and remote-delta semantics.
- Add an installed Android/iOS or equivalent native-runtime regression receipt for single-character and burst typing.

Related: #2698 covers the broader native synchronous core-tick / JS-thread stall regression. This issue tracks the composer-visible lost/deferred local echo and subscription invalidation behaviour specifically.

PR #3103, opened 2026-09-17
Keep React Native foreground mutations responsive

joeinnes:
## Summary
Keep foreground React Native mutations responsive by separating native relay admission from the queued owner work.

## Before
Direct mutation commands could synchronously drive local publication and related relay work before returning to JavaScript. Reads admitted immediately afterwards could also observe the wrong ordering boundary.

## After
Mutation commands admit writes without polling the queued owner inline. The relay preserves FIFO ordering, fences subsequent reads behind admitted writes, and retains transaction-local ordering and settlement behaviour.

## Test plan
- [ ] direct_mutations_admit_before_local_publish_and_fence_reads
- [ ] standalone_mutation_queues_behind_a_held_owner
- [ ] ordinary_multi_large_scalar_insert_settles_through_native_relay
- [ ] mutation_command_rejects_unsupported_target_options_before_admission
- [ ] No React Native iOS Simulator run was performed in this validation pass.


aeplay:
🤖 Integration handoff (@aeplay's decision, alpha.57 review). @joeinnes, a note for when you're back.

This PR is being superseded by a from-scratch implementation tracked in #3273. We're leaving it open for reference until that PR lands. The review found:
- **The work stays on the JS thread.** The queued write still runs there through the platform wake → `pump_foreground`, and `all()` still calls `tick()` before reads. So #2964's blocking gets split across turns rather than removed.
- **Errors surface asynchronously, but only on RN.** `db.update` of a missing row no longer throws on RN, while NAPI and WASM still throw.
- **The direct-mutation queue has no bound.** `direct_foreground_mutation` never calls `ensure_mutation_operation_capacity`.
- **Subscriptions aren't fenced.** Only `all()` waits on the barrier, so an insert followed by a same-turn subscribe can flicker.

#3273 records the decided contract. Failures that admission can determine cheaply throw synchronously, for parity with web and NAPI. Everything else goes through the write handle. The queue is capped with backpressure, reads and subscriptions are both fenced, and the heavy work runs off the JS thread. That PR also requires device acceptance. Your admission and barrier hooks here are the reference for it. Moving this PR out of alpha.57; #2964 stays tracked.

---

PR #3104, opened 2026-09-17
Allow migration app IDs from the environment

joeinnes:
## Summary
Allow `migrations push` to resolve the app ID from supported environment variables when the positional app ID is omitted.

## Before
The command required a positional app ID even when the CLI environment already provided one.

## After
The two-hash form resolves the app ID from the supported environment variables, while the explicit three-positional form and option forwarding remain unchanged.

## Test plan
- [x] `pnpm --filter jazz-tools run build:runtime`
- [x] `pnpm --filter jazz-tools exec vitest run src/cli.test.ts -t "env app ID"`


aeplay:
🤖 Closing as superseded by the approved, merged deployment API in #3120. This PR improves environment app-ID resolution specifically for `migrations push`; that separate command has now been removed in favor of `deploy`, which publishes the schema, explicit permissions, and migration path together. Reintroducing the retired command would undo that decision. The branch remains preserved.

Thank you for the fix and regression coverage—the underlying ergonomics goal remains relevant to the unified deploy flow.

PR #3109, opened 2026-09-17
Compare exclusive predicate results without unrelated conflicts

joeinnes:
## Stack and scope
Depends on #3107 (retained authored/physical schema identity); this diff contains only predicate conflict detection and its regressions. Review and merge #3107 first. No encryption features, new wait-tier API, or storage/wire encoding changes.

## Before / after
Local exclusive validation could reject a filtered read merely because a newer version existed elsewhere in the same table. For supported single-table predicates, compare the selected `(RowUuid, TxId)` set against a comparison snapshot incorporating newly visible non-rejected transactions. Keep the transaction's original evidence snapshot unchanged.

## Worked cases and invariants
- Read `bucket = destination`, insert an unrelated `bucket = other` row, then commit the intended write: the selected row/version set is unchanged, so this now succeeds.
- Insert a matching phantom on a disconnected local client: the exclusive transaction still rejects locally and its write does not leak. The online/global matching-phantom case also still rejects.
- Rewriting an aggregate input with identical values continues to conflict locally even when aggregate output is identical; authority-side unchanged-payload semantics are unchanged.
- Relational, aggregate, and degenerate whole-table local checks remain conservative. This is not a general relaxation of exclusive serializability.

## Verification
- Red proof on parent #3107 with fresh source-bound artifacts: unrelated filtered-row insertion failed with `transaction_conflict`; the other three public cases passed, including both phantom controls.
- After the fix, nine exact Rust controls passed: irrelevant filtered changes, relevant phantoms, other-table changes, ingested remote phantoms, binding sensitivity, inline-shape validation, and three aggregate-frontier/payload cases.
- Rust formatting, `cargo clippy -p jazz -- -D warnings`, TypeScript fixture formatting/lint/typecheck, and invariant registry passed.
- Standards and Spec reviews reported no findings.
- Fresh final-source native/WASM producer and full jazz-tools build passed. Predicate plus migration integration: 13 tests across 3 files passed. All remote CI checks passed (12 passed, 9 conditional skips).

Separate regression and implementation commits. Full local CI-equivalent verification is not claimed.


aeplay:
🤖 **Verification receipt** (integration review, PR head `e90238f3` merged locally with main `849eaf9d`)

**Commands:** `dev/t --exact <name>` for each Rust test below; `pnpm build:correctness-artifacts`; `JAZZ_NODE_TEST_COMMAND="pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/db.exclusive-predicate.test.ts" JAZZ_BROWSER_TEST_COMMAND=true pnpm test:typescript-consumers`

**Baseline (merged tree):** 5/5 pass: `node::tests::harness::{exclusive_filtered_shape_phantom_conflict_rejects, exclusive_filtered_shape_ignores_irrelevant_changes, filterless_shape_and_degenerate_predicate_validation_agree, local_exclusive_predicate_rejects_remote_phantom_ingested_after_begin}`, `node::query_eval::tests::execution::grouped_aggregate_frontiers_preserve_identity_and_validate_payloads`.
Note: the INV-QUERY-14 registry cites 3 of these as `jazz::node::tests::<name>`. `dev/t` can't resolve those paths; the real paths are `node::tests::harness::<name>`.

**Mutation** (`shape_predicate_outputs_differ(...).await? && false` in `open_tx.rs` local check; rebuild confirmed, then reverted):
- Caught it: **none**
- Didn't catch it: all 5 above. `local_exclusive_predicate_rejects_remote_phantom_ingested_after_begin` reads with `tx_current_rows`, which takes the degenerate whole-table early `return Ok(false)` before the mutated compare. `exclusive_filtered_shape_phantom_conflict_rejects` is rejected by the core authority, not locally.
- Not run: the TS suite under the mutation. Its existing "rejects a matching phantom at local durability" case probably reaches this branch.

**TS scratch tests** (not committed): 6/6 pass in the file (4 existing + 2 scratch).
- (a) Offline unrelated insert: `commit().txId` resolves offline with the write visible locally, and `wait()` resolves after `reconnect()`. No historical-cut error. Exclusive `wait()` takes no tier, so the offline check awaits `txId`.
- (b) Matching row deleted before commit (delete awaited at `{ tier: "local" }` while **online**): rejected with `transaction_conflict`. Because the client was online, this doesn't prove the rejection was local.

<details><summary>Scratch test code</summary>

```ts
it("[scratch] offline: does not reject a filtered read when an unrelated row is inserted locally", async () => {
  const app = s.defineApp({ notes: s.table({ bucket: s.string(), title: s.string() }, {}) });
  const permissions = definePermissions(app, ({ policy }) => {
    policy.notes.allowRead.always();
    policy.notes.allowInsert.always();
  });
  const server = await startLocalJazzServer({ allowLocalFirstAuth: true, inMemory: true });
  let db: Awaited<ReturnType<typeof createDb>> | undefined;
  try {
    await deploy({ serverUrl: server.url, appId: server.appId, adminSecret: server.adminSecret, schema: app, permissions });
    db = await createDb(await localAccountConfig(server.appId, server.url));
    await db.all(app.notes, { tier: "global" });
    await db.disconnect();
    const tx = db.beginExclusiveTransaction();
    expect(await tx.all(app.notes.where({ bucket: "destination" }), { tier: "local" })).toEqual([]);
    await db.insert(app.notes, { bucket: "unrelated", title: "Other data" }).wait({ tier: "local" });
    const note = tx.insert(app.notes, { bucket: "destination", title: "Intended write" });
    const handle = tx.commit();
    await handle.txId; // local publication, offline
    expect(await db.all(app.notes.where({ bucket: "destination" }), { tier: "local" })).toEqual([
      { id: note.id, bucket: "destination", title: "Intended write" },
    ]);
    await db.reconnect();
    await handle.wait(); // authority fate after reconnect
  } finally {
    await db?.shutdown();
    await server.stop();
  }
}, 60_000);

it("[scratch] rejects when a matching row is deleted locally before commit", async () => {
  const app = s.defineApp({ notes: s.table({ bucket: s.string(), title: s.string() }, {}) });
  const permissions = definePermissions(app, ({ policy }) => {
    policy.notes.allowRead.always();
    policy.notes.allowInsert.always();
    policy.notes.allowDelete.always();
  });
  const server = await startLocalJazzServer({ allowLocalFirstAuth: true, inMemory: true });
  let db: Awaited<ReturnType<typeof createDb>> | undefined;
  try {
    await deploy({ serverUrl: server.url, appId: server.appId, adminSecret: server.adminSecret, schema: app, permissions });
    db = await createDb(await localAccountConfig(server.appId, server.url));
    const existing = await db.insert(app.notes, { bucket: "destination", title: "Existing" }).wait({ tier: "global" });
    const tx = db.beginExclusiveTransaction();
    expect(await tx.all(app.notes.where({ bucket: "destination" }))).toHaveLength(1);
    await db.delete(app.notes, existing.id).wait({ tier: "local" });
    tx.insert(app.notes, { bucket: "destination", title: "Must not publish" });
    await expect(tx.commit().wait()).rejects.toThrow(/transaction_conflict/);
  } finally {
    await db?.shutdown();
    await server.stop();
  }
}, 60_000);
```
</details>

**Verdict:** The local relaxation is **not adequately pinned in Rust**: no cited Rust test fails when the local filtered-shape compare is disabled. Missing tests:
- a Rust local filtered-shape phantom test (`tx_query` with a filter, then an ingested or pending matching insert, expecting `Err(TransactionConflict)` locally)
- an offline local-delete-of-matching-row case

Also fix the 3 registry test paths.

---

aeplay:
🤖 **Integration tests pushed** (for @joeinnes, approved by @aeplay)

Follow-up to the verification receipt above. The branch now pins the local filtered-shape comparison in `open_exclusive_is_locally_serializable`.

**Commits** (plain commits only, no rebase or force-push):
- `067b59d6` Merge `origin/main` into the branch. No conflicts.
- `254953de` test: pin local filtered exclusive predicate comparison
- `08ff60b0` docs: resolve INV-QUERY-14 harness test citations

**New tests** (in `crates/jazz/src/node/tests/exclusive_transactions.rs`, `node::tests::harness`, same NodeState style as the existing tests):
- `local_exclusive_filtered_predicate_rejects_remote_matching_phantom` (a): filtered `tx_query`, then a matching row from another node is ingested (unit + fate). The commit returns `Err(TransactionConflict)` locally.
- `local_exclusive_filtered_predicate_rejects_pending_local_matching_insert` (a): filtered `tx_query`, then a matching mergeable insert on the client with **no authority node**, so the write stays pending. The commit returns `Err(TransactionConflict)`.
- `local_exclusive_filtered_predicate_ignores_non_matching_inserts` (b): filtered `tx_query`, then a non-matching remote row is ingested and a non-matching local write is left pending. `commit_exclusive` returns `Ok` and publishes the unit, and the core then accepts it. This is the relaxation this PR introduces.
- `local_exclusive_filtered_predicate_rejects_pending_local_removal` (c): a matching row is globally settled before begin, and `tx_query` returns it. The authority is then offline, and a local pending write either **updates the row out of the filter** or **deletes it**. Both cases return `Err(TransactionConflict)`.

Each test writes to a fresh row (`row(9)`), so the only conflict source is the predicate read. `tx_query` records only a `PredicateRead`, not row reads. A comment explains why these are node-level tests rather than `JazzClient` tests, as `TESTING_GUIDELINES.md` asks: the local check runs before any authority sees the unit, and it depends on exactly which versions were ingested between begin and commit. The public client API can't pin that interleaving.

**Mutation results** (`open_tx.rs`: `shape_predicate_outputs_differ(...).await? && false`, reverted with `git checkout`):

| Test | Head | Mutated |
|---|---|---|
| `local_exclusive_filtered_predicate_rejects_remote_matching_phantom` | pass | **FAIL** |
| `local_exclusive_filtered_predicate_rejects_pending_local_matching_insert` | pass | **FAIL** |
| `local_exclusive_filtered_predicate_rejects_pending_local_removal` | pass | **FAIL** |
| `local_exclusive_filtered_predicate_ignores_non_matching_inserts` | pass | pass |
| the 5 previously cited tests | pass | pass (reproduces the receipt) |

Extra check, not requested: I also ran the inverse mutation (`... .await? || true`, so every newly visible version conflicts). Only `local_exclusive_filtered_predicate_ignores_non_matching_inserts` fails, so (b) pins the relaxation from the other direction. Reverted.

**Registry fix (INV-QUERY-14, still one sorted physical line):**
- Changed `jazz::node::tests::{exclusive_filtered_shape_phantom_conflict_rejects, exclusive_filtered_shape_ignores_irrelevant_changes, filterless_shape_and_degenerate_predicate_validation_agree}` to `jazz::node::tests::harness::<name>`.
- Added the 4 new tests, which fit the invariant's "Local single-table source-row validation MUST include newly visible non-rejected transactions, including pending local writes, and MUST ignore changes outside the predicate output".
- All 9 Rust citations now resolve with `dev/t --exact <path minus jazz::>`.

**Why the gate accepted the old paths:** `dev/gates/invariant-registry.sh` does not check the libtest module path. `citation_matches_test_source` indexes test functions by **leaf name**. It then only checks that the citation's non-skipped middle components appear, in order, in the physical source path. It deliberately skips `tests`, `harness` and `*_tests` components ("`harness` is the registry's stable logical mount"). So `jazz::node::tests::X` reduces to "leaf `X`, under a path containing `node`", and `crates/jazz/src/node/tests/exclusive_transactions.rs` satisfies that. The file is `include!`d into `node::tests::harness`, and the gate never checks that. Five more jazz rows have the same stale `jazz::node::tests::<name>` form: INV-QUERY-13 (`exclusive_shape_predicate_is_binding_sensitive`, `exclusive_shape_predicate_validation_uses_inline_shape_without_registration`) and three `db_facade_*`/`view_update_result_set_*` citations. I left those alone because they are outside this PR. Making the gate resolve citations against `cargo test -- --list` would catch this.

**Commands run (on `08ff60b0`):**
- `dev/t --exact node::tests::harness::<each new test>`: 4/4 pass. The 5 previously cited tests also pass.
- The mutation and inverse-mutation runs above, using the same `dev/t --exact` commands.
- `dev/gates/invariant-registry.sh`: pass (484 rows, 0 missing citations, 0 covered rows without a test, 18 now + untested, which does not fail the gate).
- `cargo clippy -p jazz -- -D warnings`: pass (also run by the lefthook pre-commit).
- `cargo fmt -p jazz --check`: pass.
- `dev/t --test incremental_delivery_canary` for `maintained_relation_include_single_row_changes_are_scale_independent`, `complete_supporting_snapshot_receiver_cost_is_at_most_linear` and `mergeable_transaction_write_cost_is_scale_independent`: 3/3 pass.
- Not run: the jazz-private sensitive-data guard (the hook reported it missing in this container), the TS suite, the full canonical gate set and the oracle soak.

---

aeplay:
🤖 Integration note (not blocking; @aeplay asked for this to be recorded).

A review comment worried that the local exclusive-predicate check "only scans `predicate.table`, so it could miss writes a migration lens routes to another physical table". I traced this on `main` by reading the code, and it isn't reachable:
- **Lenses don't span physical tables.** A lens maps exactly one source table to one target, and the lineage partition validator rejects splits and merges (`node/ingest/catalogue.rs` ~1141-1186). A rename inherits the source's physical `table_id`.
- **The local scan covers every schema version.** `query_table_versions(name)` takes the union of `mapping.tables.get(name)` over every schema's physical mapping (`currency.rs` ~750-759). A renamed table is still scanned. A dropped-then-reused name scans both tables, which can cause unnecessary local rejections but never misses a write.
- **Core validates independently.** For predicate reads it finds the authoring schema by `shape_id` and compares output sets through that schema's lensed view (`node/ingest/fates.rs` ~400-455). If it can't resolve the schema, it treats the read as changed and fails closed.
- **The rename case is already tested:** `catalogue_lenses/projected_reads.rs::global_changes_span_table_renames_for_history_and_conflict_detection`.

One case has no test: a predicate read authored under an **old** schema, with the concurrent matching write arriving under a **newer** schema the client hasn't ingested yet. The authority's `shape_id` re-validation should handle it, but it isn't pinned. It would be a good addition to the catalogue-lens suite at some point. It doesn't block this PR.

---

PR #3129, opened 2026-09-18
Isolate branch preparation from root update poll frames

joeinnes:
## Before / after

Ordinary root-table updates shared an async poll frame with branch-view-only preparation. Root writes can refresh resident subscriptions beneath that frame. Move the branch preparation into its own nested async body so branch-only temporaries do not occupy the root update poll frame.

This is the exact generic correction extracted from the repaired E2EE source, not an encryption-specific path. It targets `main` independently and will be included in the temporary prerequisite base until landing.

## Unchanged behavior

- Root updates continue to publish ordinary mergeable writes and refresh current subscriptions.
- Branch updates still require read visibility, reject an empty authored patch, preserve inherited-cell provenance and record the same authored columns.
- No schema, storage/wire encoding, authorization, transaction-kind or public API change.

## Verification boundary

A temporary public root-update/resident-subscription scenario passed on a 2MiB thread **before** this change. It is therefore a smoke scenario, not a demonstrated regression, and is not added as a misleading permanent regression test. The same path and existing branch read-policy/relay regressions are exercised after this change; exact receipts follow. No measured stack-size or performance improvement is claimed from source inspection alone.

Formatting and pre-commit checks passed. This draft is not a full CI-equivalent or landing-gate receipt.

### Additional executed evidence

At `95804a225`, the existing public branch read-policy regression and branch update/upsert relay-to-authority regression each passed exactly once. The throwaway 2MiB root-update/resident-subscription scenario also passed after isolation and was removed. It passed before too: this is explicitly smoke coverage, not a reproduced stack-overflow regression or a measured performance claim.


aeplay:
🤖 Integration review for alpha.57: this has moved out of the release milestone. @aeplay's decision is that the PR stays open pending the changes below.

**Finding:** the unboxed nested `async move { … }.await` around the `WriteTarget::BranchView` arm of `update` (`crates/jazz/src/db/mutations.rs` ~2243-2330) probably doesn't isolate anything:
- An unboxed inner future is stored by value inside the outer `update` future.
- The two match arms already share storage, so the outer future's size is still about max(root-arm state, branch-arm state), the same as before.
- Any poll-stack saving would depend on LLVM choosing to inline the inner poll function. Nothing guarantees that.
- This file already uses `Box::pin(async move { … })` to isolate frames, e.g. around line 519.

The upsert (~2402) and delete (~2530) `BranchView` arms aren't isolated either. Logic, lock order and cancellation behaviour are unchanged, so there is no correctness risk.

**Requested:**
1. Switch the branch arms of update, upsert and delete to `Box::pin(async move { … }).await`. This costs one heap allocation, and only on branch writes.
2. Show a measured effect. Either compare `size_of_val` of the `update` future before and after, or run the 2 MiB-stack root-update/subscription-refresh scenario and show it overflowing before and passing after.
3. Reword the changeset to match what was measured. If nothing measurable remains, drop the changeset, or close the PR.

---

PR #3141, opened 2026-09-18
Fix standalone Inspector admin query authority

### LINKED ISSUE #3140: Standalone Inspector admin queries can show empty rows after alpha.54
## Summary

The standalone Inspector can load the deployed schema and list every table, while every table renders with zero rows and no Jazz error. This reproduces on the protocol-v2 Inspector with `jazz-tools` 2.0.0-alpha.54 and alpha.55.

The failure depends on the application's read policies: permissive/public applications may appear to work, while applications using ownership, organization membership, claims, or relation-based read policies return empty results.

## Reproduction

1. Deploy an application to a hosted Edge/Core topology with at least one table whose rows require a non-public read policy.
2. Insert rows and confirm they are synchronized to the hosted server.
3. Open the standalone v2 Inspector with the app ID, server URL, and valid admin secret.
4. Select the table.

### Actual

The schema and table list load, but the table shows no rows and no error.

### Expected

A valid admin Inspector connection should query with the server-admitted SYSTEM authority and display synchronized rows regardless of application read policy.

## Confirmed diagnosis

The standalone Inspector creates an admin-backed memory runtime through `createInspectorAdminClient`.

`runtimeAuthorFromConfig()` gives that runtime a syntactically valid placeholder author (`https://jazz.invalid` / `backend-open`) before the native backend ABI internally derives SYSTEM. However, `NativeRuntimeAdapter.peerIdentity` retains the placeholder.

`peerIdentityForWebSocketAuth()` explicitly maps credential-only `backend_secret` connections to the canonical SYSTEM author, but an `admin_secret` connection falls through and keeps the placeholder transport identity.

The server's alpha.54-era `ws_link_admission()` gate grants `AuthorityQueryDelegate` only when all of these are true:

- credential is Admin
- identity is SYSTEM
- trust is TrustedAuthority

The Inspector therefore authenticates successfully but misses the authority-query capability. Policy-scoped queries can settle as empty rather than surfacing an authentication error.

This was confirmed against a hosted application without exposing customer data:

- unmodified alpha.55 client: all 29 table queries returned zero rows;
- same client with only the admin WebSocket identity changed to canonical SYSTEM: populated tables immediately returned their existing remote rows.

This also reproduces outside the browser UI, so unrelated injected `reportAllChanges`/Performance API errors are not causal.

## Suggested fix

Make `peerIdentityForWebSocketAuth()` return the canonical Jazz SYSTEM author for an `admin_secret` credential, analogous to the existing credential-only backend path. Keep the server-side triple gate intact; possession of an asserted SYSTEM identity alone must never grant this capability.

Update the existing WebSocket identity test that currently expects admin credentials to retain the fallback identity.

Add a black-box standalone Inspector regression covering:

1. hosted Edge connected to Core;
2. admin credential;
3. a restrictive application read policy that rejects the placeholder identity;
4. synchronized remote rows appearing in the Inspector;
5. a non-admin SYSTEM assertion still failing to acquire `AuthorityQueryDelegate`.

## Regression window

The server-side `AuthorityQueryDelegate` identity requirement landed immediately before 2.0.0-alpha.54. The client admin transport identity was not updated to match, which explains why the symptom begins around alpha.54 and remains in alpha.55.

tobiaslins:
## Summary

Fix standalone Inspector admin connections so their WebSocket prelude uses the canonical Jazz `SYSTEM` identity expected by the server's `AuthorityQueryDelegate` admission gate.

Closes #3140.

## Before / after

Before, `createInspectorAdminClient()` opened its native database with backend authority, but `peerIdentityForWebSocketAuth()` sent the adapter's placeholder author over the WebSocket. The admin secret authenticated successfully, schema catalogue requests worked, and public rows could appear, but policy-protected tables could settle as empty without an error.

After, an `admin_secret` always maps to the canonical `SYSTEM` wire identity. The server still requires the independently validated Admin credential, `SYSTEM` identity, and `TrustedAuthority` trust before granting query delegation.

## Governing invariants

- Admin admission is sessionless and takes precedence over incidental JWT, backend-secret, or backend-session fields.
- Claiming `SYSTEM` without a valid admin credential does not grant admin query authority; the server-side triple gate is unchanged.
- Credential-only backend connections retain their existing `SYSTEM` behavior.
- Backend sessions retain their explicit end-user identity when no admin credential is present.

## Worked examples

- Normal path: standalone Inspector + valid admin secret + restrictive table read policy -> the synchronized rows are visible.
- Permissive application: public rows remain visible; they no longer mask an incorrectly scoped admin connection in the browser fixture.
- Mixed credential payload: admin secret plus incidental bearer/backend session -> wire identity remains sessionless `SYSTEM`, matching server credential precedence.
- Non-admin path: a bearer or backend session continues to use its canonical session author; an asserted `SYSTEM` identity alone gains no server capability.

## Tests

- `pnpm --filter jazz-tools exec vitest run --config vitest.config.ts src/runtime/native-runtime/websocket.test.ts` (35 passed)
- `pnpm --filter jazz-tools check` (passed)
- `pnpm --filter inspector exec playwright test --config playwright.config.ts standalone-inspector.spec.ts` (12 passed)
- The standalone Inspector fixture now uses `allowRead.never()`, so row loading and edits exercise admin authority instead of public access.
- Branch-bound NAPI and fast-WASM compilation succeeded locally. The repository artifact-store sealer hit a local macOS filesystem `EACCES` while renaming its already-sealed staging directory; the focused browser suite therefore used the freshly built mutable artifacts directly. CI will run the canonical sealed producer/consumer path.


joeinnes (approved):
Thanks for the PR!

Issue #3156, opened 2026-09-19
Add short-lived Inspector sessions with Cloud login and self-hosted authorization


aeplay:
🤖 Standalone Inspector correctly no longer persists root admin secrets in localStorage, but repeatedly pasting the secret makes normal use awkward. Add short-lived Inspector sessions, with convenient renewal backed by administrative authentication. Anselm approved this design for alpha.56.

## Shared server primitive

Provide an authenticated server-side exchange from administrative authority to a short-lived Inspector credential (initial proposal: 15 minutes). Bind it to the exact tenant/app, issuing authority, operator identity, expiry and permitted Inspector operations. Validate it for both relevant HTTP endpoints and WebSocket admission.

Use a distinct Inspector credential/admission path. Do not reuse the raw admin-secret admission path in a way that grants trusted-node/replication powers. Distinguish protected-data read, data edit, and schema/permission administration capabilities and enforce them server-side. Define the initial default capability set explicitly before implementation.

Expiry must also end or renew authorization on an already-open WebSocket; validating expiry only at initial connection is insufficient. Define renewal, logout and revocation semantics, including the maximum remaining access window after dashboard access is removed. Retain only connection metadata in localStorage; short-lived browser access tokens stay in memory.

## Jazz Cloud flow

1. Dashboard offers **Open Inspector**; standalone Inspector offers **Connect with Jazz Cloud**.
2. Tenant manager/dashboard uses the existing WorkOS login and checks the operator's administrative permission for the specific app.
3. Initially, tenant manager can obtain an Inspector token through a server-to-server exchange using its existing administrative authority. Jazz core does not need to integrate directly with WorkOS.
4. Use a one-time authorization-code handoff with PKCE, bound to the initiating browser and approved redirect destination. Do not put admin secrets or bearer access tokens in URLs.
5. Renew through the authenticated dashboard session, rechecking app authorization. Reloading Inspector can restore access through that session without repasting a root secret or storing one in Inspector.

## Self-hosted flow

- Support exchanging a supplied admin secret for an Inspector session, then discarding the root secret from browser memory.
- Offer a CLI flow such as `jazz-tools inspector`, using locally configured credentials and a protected one-time browser handoff.
- A server-hosted Inspector may use a secure HttpOnly session cookie with appropriate CSRF/origin protection. A remotely hosted standalone Inspector needs an explicit authorization/renewal flow; do not assume cross-site cookies are available.

## Acceptance

[...]

PR #3157, opened 2026-09-19
Add short-lived Inspector sessions and Cloud authorization handoff

aeplay:
🤖 Add short-lived Inspector sessions for #3156 (alpha.56). Implementation, independent review and local acceptance are complete; fresh CI is running. No Cloud deployment is implied.

## What changes for an operator

The standalone Inspector can open protected application data without retaining the root admin secret. A session lasts at most 15 minutes and defaults to read-only. Explicit edit capability enables data changes; separate admin capability permits catalogue publication. These credentials cannot act as a replication peer, bootstrap a catalogue, impersonate an application account or perform account administration.

For self-hosted development, the CLI exchanges its configured admin secret for a short-lived session and opens Inspector through a loopback handoff. The launch URL contains a random, one-use authorization code, never the root secret or access token. That code and a browser-generated proof bind redemption to the intended launch; the fragment is immediately removed from browser history.

For Cloud, Inspector opens the configured dashboard login flow. Tenant manager checks the operator's access to the exact app, redeems a single-use PKCE authorization code, and returns a scoped session plus the canonical server URL. Before expiry, Inspector renews through the authenticated dashboard cookie session and CSRF-protected endpoint. Browser memory holds tokens; persisted metadata contains only the connection choice and signed-out preference. Reload can restore access through the dashboard session; explicit logout suppresses automatic restoration.

## Failure and authority boundaries

A read-only session can inspect protected rows but cannot edit them; both UI controls and server admission enforce that distinction. Edit sessions use a fixed SYSTEM author on Core and dynamic Edges. An Edge can forward its own protected query to Core without giving the browser trusted-node privileges.

Expiry ends the live connection, including a connection waiting for its initial protocol message or queued runtime admission. An already accepted transaction may finish; expiry is not rollback. Denied renewal clears the client rather than falling back to root authority. Stateless credentials have no immediate individual revocation: a copied token can remain valid until its original expiry, at most 15 minutes.

Blocked cookies or expired dashboard login produce an explicit sign-in action. Automatic renewal does not repeatedly open popups. CLI handoff expires and is single-use. Server and adapter failures must not expose credentials in errors or logs.

## Cloud integration and rollout

The exact proposed tenant-manager contract is in `docs/engineering/inspector-sessions.md`, with client details in `docs/engineering/inspector-session-client.md`. It specifies authorization, code redemption, cookie/CSRF session renewal, exact origins/redirects, app permission rechecks, and response shapes. Those endpoints are assumptions for this PR; tenant manager still needs to implement them.

Current main uses wire v3; deployed Cloud uses wire v2. Deploy matching server/client artifacts before enabling this flow. This PR does not upgrade the Cloud fleet or authorize production deployment.

## Verification

Independent server/core and UI/CLI reviews passed, including final connection-token forwarding and optional-field serialization. Review-driven tests cover dynamic-Edge author binding, expiry during admission, cancellation cleanup, one-use CLI launch ownership, read-only controls, and logout even when browser metadata storage fails.

Source-bound native/WASM acceptance at `7d0663a4cf` passed 78 Tools tests, 169 Inspector tests, six real-server browser cases and three Chromium contract cases. The browser cases cover protected reads on Core and Edge, explicit edits, renewal/access removal, reload restoration, logout suppression and hard expiry. The final `0e02f708b3` delta preserves ordinary JSON preludes while retaining complete positional binary fields: 19 native transport tests, JSON/postcard round trips, and real Edge Inspector read/edit passed. Independent final delta checks and the coordinator's frozen-prelude rerun passed; mutation checks detect missing auth forwarding and malformed binary omission.

Fresh integrated CI is running on `0e02f708b3`. Tenant-manager behavior is tested with a synthetic adapter; this is not evidence that Cloud endpoints are deployed. The exact handoff contract above is ready for infrastructure implementation. Follow-up #3163 tracks making connection admission failures visibly distinct from empty query results.


aeplay:
🤖 Release scope update: Anselm narrowed alpha.56 to current main, the multiplexed transport work, the combined performance stack, and the explicitly approved correctness additions. This Inspector-session work moves to alpha.57. That supersedes the earlier alpha.56 scheduling note; it does not reject the design or imply that Cloud endpoint integration is complete.

PR #3185, opened 2026-09-19
Diagnose ARM64 W1 channel-credit misclassification

aeplay:
🤖 Investigate and validate a narrow correction for the alpha.56 ARM64 channel-credit failure. This remains a diagnostic branch; the release fix will be separated from its probe workflow and trace instrumentation.

## What failed

A receiver returned consumed bytes under the wrong channel class, so the sender correctly rejected the grant as exceeding that class’s outstanding balance. For example, an early W1 run returned 721,647 bytes as Control credit while those bytes were outstanding in Writes. This occurs during initial fixture loading, before reconnect.

A probe using the actual Jazz credit implementation now reproduces a wrong Auxiliary → Control grant before opening any database, in an ordinary optimized ARM64 Cargo executable. It does not need W1, RocksDB or CodSpeed. The trace shows the wrong class before grant construction or encoding. The preserved binary loads the selected stack byte before initializing it; for physical bucket 4, the load precedes the store of Auxiliary’s value 5.

## Candidate correction

Commit `475d7b0713e431d65de032d2788ee6820182c8da` expresses the two existing constant-array mappings as scalar matches. Physical and decoded-resource bucket assignments are identical, including LargeValue sharing Writes’ physical window and bulk Control’s separate decoded-resource bucket. No credit limits, accounting, wire encoding or storage encoding change.

Focused tests pass locally. Mutating the physical and decoded mappings to Control makes the two corresponding regressions fail; restoring the mappings passes. The coordinator independently reran the decoded-resource test and inspected the failing assembly. Independent review and ARM64 candidate verification are in progress; **no verified fix SHA is claimed yet**.

## Evidence and boundaries

- Unchanged main control failure: https://github.com/garden-co/jazz/actions/runs/35477561996
- Earlier initial-load trace: https://github.com/garden-co/jazz/actions/runs/35478559254
- Actual production credit probe, ordinary ARM64 Cargo release, fails before database setup: https://github.com/garden-co/jazz/actions/runs/35480816849
- Full diagnostic binary also fails its pre-database self-check: https://github.com/garden-co/jazz/actions/runs/35480223781. Its execution steps use continue-on-error; their green step conclusions do not mean successful executions. Both processes exit 101 and the final gate fails. No workload samples completed.
- Candidate ARM64 probe: https://github.com/garden-co/jazz/actions/runs/35481424179

The workflow is read-only and does not publish benchmark results. Preserve the diagnostic branch and failing binaries for compiler investigation, but exclude the diagnostic workflow and instrumentation from the release. The precise upstream compiler defect is not yet identified.


aeplay:
🤖 Superseded by the clean runtime fix #3192, now merged into main at 2c14d7e4cf9779f6c702661d5950dd1a0716160f. The clean fix passed correctness and all benchmark jobs, including W1. This diagnostic branch and its assembly/run evidence remain preserved; they are not release code. Further compiler reduction is tracked in #3193.

PR #3230, opened 2026-09-21
fix(server): enable TCP_NODELAY for small remote-read replies

### LINKED ISSUE #3229: alpha.56: remote reads stall ~40ms on server TCP buffering; enable TCP_NODELAY
## Summary

On Linux loopback, a repeated one-row remote read regresses from a warmed median of about **8ms in 2.0.0-alpha.55 to 47ms in alpha.56**. Local reads stay around 1–2ms. Enabling `TCP_NODELAY` on **accepted server sockets** removes the extra delay.

This reproduces with one table, one row, explicit allow policies, fresh in-memory storage, a backend session, and sequential reads. No application framework, authentication middleware, persistent dataset, or concurrent workload is required.

## Why the versions differ

Both releases pass a bare Tokio TCP listener into Axum, leaving Nagle enabled. Alpha.56 introduced the multiplexed v3 transport and changed the small-write pattern.

A header-only packet trace of the anonymous reproduction shows alpha.55 normally replying with a single ~734-byte payload. Alpha.56 repeatedly sends an initial 27-byte packet and then stalls before its ~794-byte follow-up:

| Relative time | Direction | TCP event |
| --- | --- | --- |
| 0ms | server → client | 27 bytes, sequence 9703:9730 |
| 43.153ms | client → server | ACK 9730, no data |
| 43.182ms | server → client | 794 bytes, sequence 9730:10524 |

The follow-up leaves **29 microseconds after the delayed ACK**. This is consistent with the first small write remaining unacknowledged while Nagle holds the next short write. This observation plus the server-only TCP_NODELAY experiment isolates the extra latency to the accepted-socket policy. I have not decoded the semantic identity of those wire frames, and am not claiming every query has this packet shape or that alpha.55 can never encounter Nagle stalls.

Control experiments with the same tiny fixture:

| Version / socket policy | Warm median remote read |
| --- | ---: |
| alpha.55, published defaults | 8.0ms |
| alpha.56, published defaults | 47.1ms |
| alpha.56, NODELAY only on outgoing connections | 48.2ms |
| alpha.56, NODELAY only on accepted connections | 7.2ms |

The accepted-socket experiment initially used a Linux diagnostic interposer; it is not proposed as a deployment workaround. The actual source fix uses Axum's `ListenerExt::tap_io` and `TcpStream::set_nodelay(true)` before serving a connection.

## Reproduction

Use Node 26.8.2 on Linux x64 (Debian 13 in this case). Install each version in a separate directory, save the following as `repro.mjs` beside that directory's package.json, and run it in a fresh process. The SDK installs its matching native payload.

```sh
npm install jazz-tools@2.0.0-alpha.55
node repro.mjs
# Repeat in another directory with jazz-tools@2.0.0-alpha.56.
```

```js
import assert from 'node:assert/strict';
import { schema as s } from 'jazz-tools';
import { createJazzSession } from 'jazz-tools/backend';
import { deploy, startLocalJazzServer } from 'jazz-tools/testing';

const app = s.defineApp({ item: s.table({ name: s.string() }, {}) });
const permissions = s.definePermissions(app, ({ policy }) => {
  policy.item.allowRead.always();
  policy.item.allowInsert.always();
});
const server = await startLocalJazzServer({ inMemory: true, enableLogs: false });
let session;
try {
  await deploy({
    appId: server.appId, serverUrl: server.url, adminSecret: server.adminSecret,
    schema: app, permissions,
  });
  session = await createJazzSession({
    app, permissions, appId: server.appId, serverUrl: server.url,
    driver: { type: 'memory' }, env: 'test',
    initial: { backendSecret: server.backendSecret },
  });
  const { db } = session.getSnapshot().client;
  await db.insert(app.item, { name: 'example' }).wait({ tier: 'global' });
  for (const tier of ['local', 'remote']) {
    const samples = [];
    for (let i = 0; i < 15; i++) {
      const rows = await db.all(app.item.where({ name: 'example' }), { tier });
      assert.equal(rows.length, 1);
      assert.equal(rows[0].name, 'example');
      samples.push(performance.now() - started);
    }
    const warm = samples.slice(3).sort((a, b) => a - b);
    console.log(tier, { samples, medianMs: (warm[5] + warm[6]) / 2 });
  }
} finally {
  await session?.close();
  await server.stop();
}
```

## Suggested change

Enable TCP_NODELAY once for each accepted socket in both `crates/jazz-server/src/lib.rs` (standalone server) and `crates/jazz-server/src/server/testing.rs` (embedded/NAPI server). Axum documents this exact socket-configuration hook: https://docs.rs/axum/0.8.9/axum/serve/trait.ListenerExt.html#method.tap_io

A deterministic regression should inspect the flag on real accepted sockets instead of imposing a flaky latency threshold. Separate end-to-end receipts should verify the performance change. This does not change TCP ordering/reliability, Jazz consistency tiers, or wire/storage formats; the tradeoff is potentially more small packets.

## Source-fix verification

A standalone server built from alpha.56 commit `8af685e184c8691a244da1171a51c745a7208e3b` with only the accepted-socket change, using Rust 1.93.1, reduces the same tiny fixture from **48.0ms (published CLI) to 7.6ms (patched CLI)**. Both use the unchanged published alpha.56 client. The accepted-socket test passes. A separate process-level test also confirms the patched server reopens storage written by the published alpha.56 server and accepts globally confirmed updates.

Proposed source fix: https://github.com/garden-co/jazz/pull/3230

husseinraoouf:
## Problem and behavior

Fixes #3229. With the alpha.56 multiplexed transport, a small server write can remain unacknowledged while Nagle holds the follow-up write. On Linux loopback the one-row remote-read reproduction increases from roughly 8ms on alpha.55 to 47ms on alpha.56. Captured traffic shows a 27-byte packet, a delayed ACK 43ms later, and the following 794-byte payload sent 29 microseconds after that ACK.

Configure `TCP_NODELAY` once on every accepted standalone and embedded/NAPI server connection, using Axum's standard `ListenerExt::tap_io`. The follow-up write can leave immediately instead of waiting for the first packet's ACK. Share the listener policy across both entry points so the test server and production server do not diverge. Failure to set the socket option is logged and retains the existing connection behavior.

TCP remains reliable and ordered. Jazz authorization, consistency, transaction ordering, durability, storage and wire formats are unchanged. This does not change client socket options, message batching, or application query semantics. The tradeoff is potentially more small TCP packets.

## Validation

- Added a deterministic regression checking `nodelay()` on two real accepted sockets. An internal socket test is intentional: correct row results cannot distinguish the socket policy, and a wall-clock threshold would be flaky under load.
- The exact source change is built and tested as an alpha.56 backport at `8af685e184c8691a244da1171a51c745a7208e3b`, with Rust 1.93.1. `cargo test --locked --release -p jazz-server --lib tcp::tests::accepted_connections_disable_nagle -- --exact` passes.
- Unchanged published alpha.56 client + published CLI: 48.0ms warm median remote read. Same client + source-patched CLI: 7.6ms. Local reads remain about 1.2ms. Details and the anonymous standalone reproduction are in #3229.
- A process-level qualification writes with the published alpha.56 server, restarts the patched executable on that storage, then reads and globally confirms an update successfully.

This PR targets current main; the local executable/test receipts above are explicitly for the alpha.56 backport, not a full current-main CI-equivalent run. Upstream CI qualification remains pending. No existing tests are rewritten or skipped.


aeplay:
🤖 Thank you @husseinraoouf. The diagnosis, packet capture and fix were excellent, and they pinned down a real alpha.56 regression.

We've moved this into #3270 on a garden-co branch so it can run trusted CI, which doesn't run for fork PRs. Your commit is kept as-is with co-author credit. The one addition is the same `TCP_NODELAY` setting on the `jazz-server` binary's loopback accept loop (`LoopbackWebSocketServer::accept_loop`), which doesn't go through `axum::serve`. The client-side `disable_nagle = false` you noted as a non-goal is now tracked in #3268.

Closing this in favour of #3270. Thanks again!

---

